PR #12, opened 2021-07-09
Use/integrate with strict types

jdeepee:


src/core/db.ts
```diff
@@ -26,7 +27,7 @@ export class PerspectivismDb {
         return `${pUUID}-to_target-${target}`
     }
 
-    storeLink(pUUID: string, link: object, linkName: string) {
+    storeLink(pUUID: string, link: object, linkName: string): void {
```
lucksus, line 30: Isn't this redundant?

lucksus (approved):
This is great work!

PR #13, opened 2021-07-13
Neighbourhood Integration Test

lucksus:
- [x] make GraphQL and Holochain ports configurable so we can start two instances on the same machine
- [x] add test setup that runs two ad4m-executor instances
- [x] make IPFS repo path configurable
- [x] debug all port collisions
- [x] add test that creates a neighbourhood and expects links to be shared


jdeepee (approved):
All tests working except neighbourhood link tests which is flaky due to holochain gossip not always arriving correctly

PR #16, opened 2021-07-28
Integrate with neighbourhood on PerspectiveHandle

jdeepee:
Depends on https://github.com/perspect3vism/ad4m/pull/13 being merged and published for proper release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@perspect3vism/ad4m-executor",
-  "version": "0.0.18",
+  "version": "0.1.0",
   "description": "Node.js package that allows the running/interfacing of AD4M Languages & Perspectives.",
   "main": "lib/main.js",
   "files": [
@@ -60,7 +60,7 @@
   },
   "dependencies": {
     "@holochain/conductor-api": "0.0.2",
-    "@perspect3vism/ad4m": "file:../ad4m",
+    "@perspect3vism/ad4m": "0.1.5",
     "@rollup/plugin-commonjs": "^17.1.0",
     "@rollup/plugin-json": "^4.1.0",
     "@rollup/plugin-node-resolve": "^11.2.0",
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
   },
   "dependencies": {
     "@holochain/conductor-api": "0.0.2",
-    "@perspect3vism/ad4m": "0.1.4",
+    "@perspect3vism/ad4m": "file:../ad4m",
     "@rollup/plugin-commonjs": "^17.1.0",
     "@rollup/plugin-json": "^4.1.0",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -60,7 +60,7 @@
   },
   "dependencies": {
     "@holochain/conductor-api": "0.0.2",
-    "@perspect3vism/ad4m": "0.1.4",
+    "@perspect3vism/ad4m": "file:../ad4m",
```
lucksus, line 63: Local path needs changing before merge

src/core/PerspectivesController.ts
```diff
@@ -24,28 +24,14 @@ export default class PerspectivesController {
...
-                this.#perspectiveHandles.set(k, fileObject[k])
-                if (neighbourhoodsFileObject) {
-                    //@ts-ignore
-                    let neighbourhood = neighbourhoodsFileObject[k];
-                    if (neighbourhood != {}) {
-                        let neighbourhoodTyped = neighbourhood as Neighbourhood;
-                        console.debug(`Found neighbourhood for perspective "${k}":`, neighbourhoodTyped)
-                        this.#perspectiveInstances.set(k, new Perspective(fileObject[k], this.#context, neighbourhoodTyped))
-                    }
-                }
```
lucksus, line 48: Is this correct? You're removing the code that was recently added..

jdeepee: Yep having neighbourhood on the PerspectiveHandle means we dont need another .json file to store neighbourhood data as it is done on the PerspectiveHandle objects already

lucksus: ah, perfect

src/tests/neighbourhood.ts
```diff
@@ -33,8 +34,12 @@ export default function neighbourhoodTests(testContext: TestContext) {
                 //Check that we got an ad4m url back
                 expect(publishPerspective.split("://").length).toBe(2);
 
+                const perspective = await ad4mClient.perspective.byUUID(create.uuid);
+                expect(perspective?.neighbourhood).toBeDefined();
+
                 const join = await ad4mClient.neighbourhood.joinFromUrl(publishPerspective );
                 expect(join.sharedUrl).toBe(publishPerspective);
+                expect(join.neighbourhood).toBeDefined();
```
lucksus, line 42: Should we also check the neighbourhood's fields (`linkLanguage` and `meta`)?

lucksus (approved):
Maybe needs one or two more assertions in the test as mentioned above, but looks good to merge.

src/tests/neighbourhood.ts
jdeepee: Sure adding and then merging

lucksus (approved):

PR #22, opened 2021-08-28
WIP: Entanglement proof

jdeepee:


jdeepee:
@lucksus this is basically ready for review and merge, its just blocked by the sign crypto function not being exposed by the conductor, ive been recommended on the holochain discord to just expose this function using a holochain DNA... maybe this is okay for now

jdeepee:
# Notes on implementation details / changes

- If languages share a DNA, when the DNA emits a signal both languages will receive it. This was a by-product of using the DNA hash as the key in the `signalCallbacks` map on `HolochainService`
- There is a case where a DNA can be installed before an agent init'd, this can only happen on system level DNA's on ad4m-executor first starting up, for now I have just omitted the creation of a did x holochain EP proof if DID the is not present on the agent service. Let me know if you think we need this EP proof generated, if so we may need to change the way we init languages. 
- The `signing.dna` is currently loaded from the `public` directory, this is working fine in dev but I am pretty sure its not gonna work in a production build (I am yet to try however). Do you have any ideas about how we want to package/load these static DNA's / resources that the executor contains? All the current loading of languages/DNA's happens by passing paths into the executor so this is the first time we have had to deal with packaging our own DNA.
- UI for the dapp server is very basic text, but should be fine for now
- I have upgraded the `listActiveApps()` call on `HolochainService.ensureInstallDNAforLanguage`, to the new `listApps()` method with status filter: `AppStatusFilter.Enabled`, I am not 100% sure if we are getting a difference in behaviour with these two methods, all tests are passing and junto app working okay but worth keeping an eye on.

# Todo

- [ ] Once DID x Holochain EP proof has been generated, save to file to avoid regenerating on each DNA install 

# Further questions

- Do we want to associate any type information with EP proof's denoting what type of keys in the proof are?
-  Are we happy with using hex encoding for signature/key data coming out of holochain?

src/core/agent/AgentService.ts
```diff
@@ -29,6 +30,7 @@ export default class AgentService {
     constructor(rootConfigPath: string) {
         this.#file = path.join(rootConfigPath, "agent.json")
         this.#fileProfile = path.join(rootConfigPath, "agentProfile.json")
+        this.#entanglementProofs = path.join(rootConfigPath, "entanglementProofs.json")
```
lucksus, line 33:
```suggestion
        this.#fileEntanglementProofs = path.join(rootConfigPath, "entanglementProofs.json")
```

src/core/storage-services/Holochain/HolochainService.ts
```diff
@@ -123,6 +134,37 @@ export default class HolochainService {
...
                 console.debug("HolochainService: Holochain app interface connected on port", this.#appPort)
             };
+
+            //Install signing service DNA
+            const activeApps = await this.#adminWebsocket!.listApps({status_filter: AppStatusFilter.Enabled});
+            if (!activeApps.map(value => value.installed_app_id).includes("signing_service")) {
+                const pubKey = await this.pubKeyForLanguage("main");
+
+                const hash = await this.#adminWebsocket!.registerDna({
+                    //Pretty sure this is not gonna work in production
```
lucksus, line 144: Let's just publish the DNA as release in its repo and download it the same way we download the bootstrap languages..

src/core/storage-services/Holochain/HolochainService.ts
```diff
@@ -230,6 +307,32 @@ export default class HolochainService {
...
+                    const hashHex = hash.toString("hex");
+                    let callbacks = this.signalCallbacks.get(hashHex);
+                    let newCallbacks = [];
+                    if (callbacks) {
+                        callbacks.push(callback);
+                        newCallbacks = callbacks;
+                    } else {
+                        newCallbacks = [callback] as AppSignalCb[]
+                    }
+                    this.signalCallbacks.set(hashHex, newCallbacks);
```
lucksus, line 333: The signal callback registration can be extracted out of the two cases to avoid c/p code. See my PR: https://github.com/perspect3vism/ad4m-executor/pull/29/files#diff-7cb52e8044275b999765d1fd40e5b6810eb9d1a0fbf0781dc523dc77810bb994

src/tests/agent.ts
```diff
@@ -81,6 +81,31 @@ export default function agentTests(testContext: TestContext) {
...
+                expect(getProofs[0]).toStrictEqual(preFlight);
+
+                //Check can delete entanglement proofs
+                const deleteProofs = await ad4mClient.agent.deleteEntanglementProofs([preFlight as EntanglementProofInput]);
+                expect(deleteProofs.length).toBe(0);
+
+                //Check entanglement proof is deleted on get
+                const getProofsPostDelete = await ad4mClient.agent.getEntanglementProofs();
+                expect(getProofsPostDelete.length).toBe(0);
+            })
```
lucksus, line 108: Hm, to really test what is important, we should check the created signature (from `entanglementProofPreFlight()`) to be valid with regards to the agent's DID...

lucksus: I would suggest to move the new signing code from `AgentService.ts` to `Signature.ts` and have a `verify()` function there as well that does not add the Expression cludder/wrapper so that we could call that here to verify the signature..

lucksus (commented):
Basically great stuff! Just a few remarks..

src/core/storage-services/Holochain/HolochainService.ts
jdeepee: hah yeah right...

src/core/storage-services/Holochain/HolochainService.ts
jdeepee: Ill just merge your code here

jdeepee:
This PR is now working correctly on my machine. All tests passing, Flux app still functioning with this branch & dapp ui working and correctly submitting entanglement proofs. There are still a few things that should be added but can likely be done in more PR's, let me know what you think @lucksus: 

- DApp should scan for the ad4m-executor graphql server searching all ports on the machine. (This requires some interface to be implemented which will return a "status" response if the correct port is hit and perhaps can be added in another PR)
- Add ability to create an agent via the dapp if one is not already present (currently can only unlock)
- Add support for wallet connect
- Improve UI for Dapp
- Dapp should show graphql errors within notification system
- Some mechanism within the executor which allow you to declare which EP's you allow shared with given languages; or is this something that we want to do explicitly via a new interface on languages?

dapp/App.svelte
```diff
@@ -0,0 +1,146 @@
+<script>
+	import Paper, {Title, Subtitle, Content} from '@smui/paper';
+	import Button, {Label} from '@smui/button';
+	import Textfield from '@smui/textfield'
+    import Icon from '@smui/textfield/icon/index';
```
lucksus, line 5: indentation is flaky

executor/src/core/storage-services/Holochain/HolochainService.ts
```diff
@@ -168,7 +178,62 @@ export default class HolochainService {
...
                 console.debug("HolochainService: Holochain app interface connected on port", this.#appPort)
             };
+
             resolveReady!()
+
+            //Install signing service DNA
+            const activeApps = await this.#adminWebsocket!.listApps({});
+            if (!activeApps.map(value => value.installed_app_id).includes("signing_service")) {
+                const dest = path.join(this.#dataPath, "signing.dna");
+                const res = await fetch(`https://github.com/perspect3vism/signing-service/releases/download/${signingServiceVersion}/signing.dna`);
```
lucksus, line 188: Uh, can we somehow package this with the executor?

lucksus (changes requested):
Signing-DNA downloading mid-initialization feels a bit funky, otherwise happy to merge this .. finally!

Also, changelog entry needed.

PR #26, opened 2021-09-20
Added limit param support for local link limit

fayeed:


jdeepee:
Looks good to me but not sure why the tests are failing

fayeed:
@jdeepee Fixed the other error, it's still failing because it is not able to find the limit property on LinkQuery, merging the PR for ad4m should fix this.

lucksus (approved):
👌

PR #29, opened 2021-10-02
Direct Message Language integration

lucksus:


jdeepee (dismissed):
Looks good, only thing id like to see is the agent direct message stuff be optional. Right now I can see the language is always created on `agentGenerate` Did we reach consensus when we last spoke about if we wanted to make this optional or not?

lucksus:
In that conversation I explained that it is basically opt-out, since the agent/user can set the language to anything, also null. I said that I prefer having a working default and you seemed to agree, so I thought it's ok to have it this way..

Also, since the agent (their keys and the agent expression) is part of ad4m which means outside the scope of an app using ad4m. Even if you don't use the DM features in the Junto UI, the same agent is used in Perspect3ve and other apps. So not using the feature should not (per default) mean that it should be deactivated.

Of course, if somebody gets followed by a lot of agents and needs something different (maybe even not be reachable via DM) they can always change the DM language or set it to null.

PR #32, opened 2021-10-13
update version to 0.1.21

kaichaosun:
Ad4m cli need this to fix installation.


lucksus:
I see the latest published version is already 0.1.23, so I guess we don't need/want to merge this anymore, right?

PR #34, opened 2021-11-02
Added a way to find open port for holochainAppPort & holochainAdminPort

fayeed:


fayeed:
Currently, we are only storing the `holochainAdminPor`t port in the config so it's persistent but `holochainAppPort` is not and can have a different port if the default one is not found, should this also be persisted in the config?

lucksus (approved):
This looks good. I don't think we need to persist `hcPortAdmin` since, as you say, it is already persisted in the conductor config - and that is normative.

fayeed:
@lucksus I mean `holochainAppPort` is not saved in the config.

jdeepee:
@lucksus still good to merge?

PR #35, opened 2021-11-02
Prolog queries on Perspective data / links

lucksus:
Needs https://github.com/perspect3vism/ad4m/pull/27 merged and next ad4m version released
- [x] Add a Prolog engine (uses `swipl-stdio` which calls and wraps `swipl` binary pulled in from Nix)
- [x] Add a `query` function to Perspectives that runs a Prolog query over an engine instance initialized with all the links in the Perspective as facts.


jdeepee (approved):
Looks awesome! Will test deeper with our social-context's once I understand prolog querying some more.

jdeepee (approved):

PR #43, opened 2022-01-11
Fixes an issue where if a property for bootstrapFixtures was not defined it would still run

fayeed:
Fixes an issue where if a property for `bootstrapFixtures` was not defined it would still run causing it to return `undefined`.


lucksus (approved):
Looks good, though tests are failing. Seems like CI got flaky with Holochain stalling on `app_info`, which is not related to these changes.

PR #44, opened 2022-01-12
Added resolver for perspective from url method

fayeed:


lucksus:
Basically ok, though I've solved this in Perspect3ve inside the UI only by iterating over all perspectives and searching for the one with the given sharedURL, which I would suggest to do here on this level as well as it would be less error prone, since no need to update any map cache...

lucksus:
... which BTW is what is missing in this PR. The map/cache update happens in perspectiveSnapshot() and not in add() which can make the map get stale..

fayeed:
> Basically ok, though I've solved this in Perspect3ve inside the UI only by iterating over all perspectives and searching for the one with the given sharedURL, which I would suggest to do here on this level as well as it would be less error prone, since no need to update any map cache...

Right now we are doing exactly this getting all the perspectives using the `all` method and looping over it. The reason I thought this method might be needed is as the users join more neighbourhoods the number of perspectives will increase and looping every time would not be great.

lucksus:
Well, I don't think looping over Perspectives will really become a problem.. Even less so if the looping would happen here in the ad4m-executor (because it would save GQL interface round-trips).

And I also don't oppose adding/merging this, but then I would like to see proper tests :)

PR #45, opened 2022-01-13
Allow init ad4m service by connecting to existing holochain

kaichaosun:


src/core/storage-services/Holochain/HolochainService.ts
```diff
@@ -199,18 +219,6 @@ export default class HolochainService {
...
-        if(alreadyExisting) {
-            const pubKey = Buffer.from(alreadyExisting.pubKey)
-            console.debug("Found existing pubKey", pubKey.toString("base64"), "for language:", lang)
-            return pubKey
-        } else {
-            const pubKey = await this.#adminWebsocket!.generateAgentPubKey()
-            this.#db.get('pubKeys').push({lang, pubKey}).write()
-            console.debug("Created new pubKey", pubKey.toString("base64"), "for language", lang)
-            return pubKey
-        }
```
lucksus, line 213: FYI: left this in because using the same key for all DNAs should only be a temporary thing...

lucksus (approved):
Great!

src/core/storage-services/Holochain/HolochainService.ts
kaichaosun: ah I see.

PR #50, opened 2022-01-22
Hdk 119

jdeepee:
this PR uses the new languages built with hdk version 119 to be compatible with holochain version 0.0.123.

Tests are never completing for me and generally very flaky so cannot asses the stability of PR. Prolog tests are also consistently failing.


default.nix
```diff
@@ -1,16 +1,14 @@
...
+  holonix = import (holonixPath) { };
   nixpkgs = holonix.pkgs;
 in nixpkgs.mkShell {
   inputsFrom = [ holonix.main ];
-  packages = [
-  ];
   buildInputs = with nixpkgs; [
     binaryen
     nodejs-16_x
-    swiProlog
```
lucksus, line 14: This is what breaks the prolog tests. We need that line.

lucksus (changes requested):
All good except the removal of swiProlog

PR #54, opened 2022-01-28
Stable tests

jdeepee:
uses local storage bootstrap languages for more stability


lucksus (approved):
So it's rather IPFS than Holochain that made our tests flakey...?

PR #61, opened 2022-03-09
add runtimeInfo query endpoint

jdeepee:
Added this endpoint so that UI running in browsers can scan for the ad4m-executor server like so:

```
for port in 0..65535 {
    http://locahost:port?query={runtimeInfo{ad4mExecutorVersion}}
}
```

and then once it gets a response it can connect.

AD4M interface implemented here: https://github.com/perspect3vism/ad4m/pull/35


src/core/graphQL-interface/GraphQL.ts
```diff
@@ -177,6 +179,13 @@ function createResolvers(core: PerspectivismCore) {
                 console.log("runtimeMessageOutbox")
                 const { filter } = args
                 return core.runtimeService.getMessagesOutbox(filter)
+            },
+            //@ts-ignore
+            runtimeInfo: (parent, args) => {
+                return {
+                    ad4mExecutorVersion: ad4mExecutorVersion,
```
kaichaosun, line 186: Should we instead return this, since GraphQL/client users don't need to care the version of ad4m, it's encapsulated in ad4m-executor
```json
{
"name": "ad4m",
"version": "ad4m-executor-version"
}
```

src/core/Config.ts
```diff
@@ -2,6 +2,8 @@ import * as path from 'path';
 import * as fs from 'fs';
 import { Expression } from '@perspect3vism/ad4m';
 
+export let ad4mExecutorVersion = "0.1.32";
```
kaichaosun, line 5: The current package version can be get by
```
const pkg = require('./package.json');
console.log(pkg.version);
```

src/core/graphQL-interface/GraphQL.ts
```diff
@@ -5,6 +5,8 @@ import { typeDefsString } from '@perspect3vism/ad4m/lib/src/typeDefs'
 import type PerspectivismCore from '../PerspectivismCore'
 import * as PubSub from './PubSub'
 import { GraphQLScalarType } from "graphql";
+import { config } from 'process'
```
kaichaosun, line 8: It seems not used.

src/core/Config.ts
jdeepee: awesome I was looking for something like this

src/core/graphQL-interface/GraphQL.ts
jdeepee: Yeah I think you are right actually, chances are the person using the ad4m-client will also know which version of ad4m they are using since they would have installed it

kaichaosun (dismissed):
👍

jdeepee:
(tests will now pass once new version of ad4m is merged and published)

lucksus (dismissed):

PR #64, opened 2022-03-18
Improve error handling with subscription

kaichaosun:
TODO:
- [x] bump ad4m https://github.com/perspect3vism/ad4m/pull/36

### DIFF
--- a/src/core/LanguageController.ts
+++ b/src/core/LanguageController.ts
@@ -219,7 +219,9 @@ export default class LanguageController {
         const languagePath = path.join(Config.languagesPath, address);
         const sourcePath = path.join(languagePath, 'bundle.js')
         const metaPath = path.join(languagePath, 'meta.json')
-        fs.mkdirSync(languagePath)
+        if(!fs.existsSync(languagePath)) {
+            fs.mkdirSync(languagePath)
+        }
         fs.writeFileSync(sourcePath, source)
         fs.writeFileSync(metaPath, JSON.stringify(languageMeta))
         // console.log(new Date(), "LanguageController: installed language");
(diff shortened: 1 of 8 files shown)


src/core/LanguageController.ts
```diff
@@ -211,13 +219,7 @@ export default class LanguageController {
         const languagePath = path.join(Config.languagesPath, address);
         const sourcePath = path.join(languagePath, 'bundle.js')
         const metaPath = path.join(languagePath, 'meta.json')
-        try {
-            fs.mkdirSync(languagePath)
-        } catch(e) {
-            console.error("Error trying to create directory", languagePath)
-            console.error("Will proceed with installing language anyway...")
-        }
-
```
jdeepee, line 220: I think this should likely still catch the error, and then just make a console.warn()

jdeepee (approved):
Nice idea! Looks good to me except leaving the try/catch block above

src/core/LanguageController.ts
kaichaosun: In below `fs.writeFileSync` will throw exception if directory is not exist. try catch here seems not much helpful.

jdeepee: I think the above catch is useful in the case where the language directory already exists, but something went wrong later in the install process and you want to re-install. If the `fs.mkdirSync()` actually does fail for some reason other than the directory already existing then `fs.writeFileSync()` should definitely fail (since the path to write the file is invalid) and then we can throw a gql pub sub error here.

kaichaosun: Improved with checking dir exist or not, pub sub can be added later.

lucksus (dismissed):

PR #71, opened 2022-04-19
Session management

kaichaosun:
TODOs,

- [x] decide if use store based session manager or JWT, resources [1](https://hackmd.io/@2dHpJQYgQrCqfxqXMtJTFQ/ByzkcO6Xc), [2](https://hackmd.io/Epa06kVQQ5mkfB1i_Nqcqw?view)
- [x] remove duplicate logs
- [x] extract permission logic
- [x] bump ad4m
- [ ] protect all the interface

Related changes,
https://github.com/perspect3vism/ad4m/pull/39
https://github.com/perspect3vism/ad4m-host/pull/26
https://github.com/perspect3vism/ad4min/pull/12


src/core/agent/AgentService.ts
```diff
@@ -21,18 +26,24 @@ export default class AgentService {
     #agent?: Agent
     #agentLanguage?: Language
     #pubsub: PubSub
+    #tokens: Map<string, AuthInfo>
```
lucksus, line 29: Should the name of this map be `#requests`? It is not storing the actual capability tokens, or is it?

src/core/agent/AgentService.ts
```diff
@@ -21,18 +26,24 @@ export default class AgentService {
...
+    #tokenValidPeriod: number
+    #adminCredential: string
```
lucksus, line 31: I don't really get the idea/reason for why we have this variable `adminCredential`. I remember talking about printing a `*` capability token out to stdout so ad4min can get it from there. Then there should be no need for handling the 'admin case' in a special way within the executor, no?

src/core/agent/AgentService.ts
```diff
@@ -272,10 +283,93 @@ export default class AgentService {
...
+        } as AuthInfoExtended
+        
+        this.#pubsub.publish(
+            PubSubInstance.EXCEPTION_OCCURRED_TOPIC,
+            {
+                title: "Request to authenticate application",
+                message: `${appName} is waiting for authentication, go to ad4min for more information.`,
+                type: ExceptionType.RequestAuth,
+                addon: JSON.stringify(authExtended),
+            } as ExceptionInfo
```
lucksus, line 325: If we generate the random number already here and send it to ad4min through this exception, we only need one more function to implement the [sketched out flow](https://hackmd.io/Epa06kVQQ5mkfB1i_Nqcqw) - that is we can do it without `permitAuth` below..

lucksus: ...which also means, we wouldn't need `AuthInfoExtended` since we don't need to send the requestId to ad4min. We would need to send the random number with the `AuthInfo` though.

src/core/agent/AgentService.ts
```diff
@@ -272,10 +283,93 @@ export default class AgentService {
...
+    }
+
+    permitAuth(adjustedAuth: string, capabilities: Capability[]) {
+        console.log("admin user capabilities: ", capabilities)
+        console.log("auth info: ", adjustedAuth)
+
+        let { requestId, auth }: AuthInfoExtended = JSON.parse(adjustedAuth)
+        let rand = genAuthRand()
+        this.#tokens.set(genAuthKey(requestId, rand), auth)
+
```
lucksus, line 338: So basically we can move this code into `requestAuth()` (which I think should be called `requestCapability()`.

lucksus (changes requested):

src/core/agent/AgentService.ts
kaichaosun: Yeah we talked about grep the output for the token, but after research on tauri side and more depth thinking, it's not a consistent design since only if we already generate an agent, it can further sign an jwt. And if the adminCredential is a jwt, it needs to refreshed periodically, this is not perfect for long running tasks.

In this new design, 
* the adminCredential gets passed in when starting ad4m-host, this is the practical way to run a binary with credentials suggested from [12 factor apps](https://12factor.net/config).
* restart the ad4min app, a new adminCredential generated.

Overall, I think jwt is good in authentication, but not fit for administration of long running tasks.

src/core/agent/AgentService.ts
kaichaosun: This is tricky though, since user can inspect the content of this exception to get the reqestId and number, without actually confirm the request.

src/core/agent/AgentService.ts
kaichaosun: `requests` sounds better, it stores the `requestid-randnumber` -> `auth(capabilities, desc)`.

src/core/agent/AgentService.ts
kaichaosun: Authenticate is a process, capability is part of the resource of such process. I tried to rename `requestAuth` to `requestCapability`, and `permitAuth` to `permitCapability`, but it feels not straight, since we have `requestCapabilities` in the params of `requestCapability`.

lucksus (changes requested):
Good to merge when the new interface functions have tests.

src/core/agent/AgentService.ts
```diff
@@ -21,18 +26,24 @@ export default class AgentService {
...
+    constructor(rootConfigPath: string, reqCredential?: string) {
         this.#file = path.join(rootConfigPath, "agent.json")
         this.#fileProfile = path.join(rootConfigPath, "agentProfile.json")
         this.#pubsub = PubSubInstance.get()
         this.#readyPromise = new Promise(resolve => {
             this.#readyPromiseResolve = resolve
         })
+        this.#requests = new Map()
+        this.#tokenValidPeriod = DefaultTokenValidPeriod
+        this.#adminCredential = reqCredential || ''
```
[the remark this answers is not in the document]
kaichaosun: With empty string, reqCredential is not needed to take into consideration when making local dev/test setup. It has an implicit meaning that if no reqCredential passed in, a empty string token is able to access all the interface.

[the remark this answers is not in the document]
lucksus: Yeah, exactly. Not sure if it is a good idea to have the default (missing argument) bypass all the capability checks.

Hm, I get that it helps for dev/testing purposes.
So, keep this in, maybe it should create a big banner in the log, saying that in any production use-case, reqCredential should be set..?

src/tests/authentication.test.ts
```diff
@@ -0,0 +1,242 @@
...
+
+            await expect(call())
+                .resolves
+                .toMatch(/.+/)
+        })
+
+        it("admin user can permit capability", async () => {
+            let requestId = await unAuthenticatedAppAd4mClient!.agent.requestCapability("demo-app", "demo-desc", "https://demo-link", '[{"with":{"domain":"agent","pointers":["*"]},"can":["QUERY"]}]')
+            const call = async () => {
+                return await adminAd4mClient!.agent.permitCapability(`{"requestId":"${requestId}","auth":{"appName":"demo-app","appDesc":"demo-desc","appUrl":"demo-url","capabilities":[{"with":{"domain":"agent","pointers":["*"]},"can":["QUERY"]}]}}`)
```
lucksus, line 183: Hm, seeing the repetition of the capability request here in permitCapability makes it clear again for me, why I was arguing for the other solution (no permit function but finalising the request in generateJwt (which I think should be called finalise_request) by checking the random number - and instead having a decline function). 

I'm sorry I missed that when I approved the PR in ad4m with your set of functions yesterday, which means this solution is a bit more carved in stone.
But, as I said in our call, I really think this should be different.

Allowing for a cap request to return a capability that is different to what was requested only creates more complexity in the requesting UIs. So with this implementation, to avoid that complexity there, we need to make sure that any UI that calls `permitCapability` never changes the cap subject. Which is what AD4MIN definitely should do.

But also by not having a decline function, and not storing requests before they are permitted, UIs can't find out, if their request was declined.

I don't want to have another big turn-around on these PRs and I don't want to repeat myself again. And I don't like holding up the next steps of adding checks to all functions. So I want to request just one last change on this PR: can you please add comments (I guess best next to the definition of permitCapability) describing how this should be changed and why? Basically what I said in our calls and summarised above. Then I'm happy to merge this for now and let you continue on the rest of the capability story, and I will put up a PR myself in the not-too-distant future to change this around. 

Ok?

lucksus (changes requested):
Summarised from my comments above, last changes I request before merge
1. adding a log warning for when no reqCredential is set, saying that now capabilities are overridden
2. adding comments to permitCapability, saying how it should be changed and why

src/tests/authentication.test.ts
kaichaosun: Comments added, also link to the related issue here,
https://github.com/perspect3vism/ad4m-executor/issues/73

lucksus (approved):

PR #72, opened 2022-05-10
Windows support

fayeed:
- Forked swipl package to be able to work with windows & also added support to use the passed swipl binary rather than using one in the $PATH. https://github.com/perspect3vism/node-swipl-stdio
- Tests are passing in windows.
- Added a new job for windows in CI.

### DIFF
--- a/src/core/PrologInstance.ts
+++ b/src/core/PrologInstance.ts
@@ -1,14 +1,16 @@
 import fs from "fs-extra";
+import path from "path";
 //@ts-ignore
 import swipl from 'swipl-stdio'
 //@ts-ignore
 import tmp from 'tmp'
+import { resourcePath } from "./Config";
 
 export default class PrologInstance {
     #engine
 
     constructor() {
-        this.#engine = new swipl.Engine()
+        this.#engine = new swipl.Engine(path.join(resourcePath, "swipl"))
     }
 
     async query(input: string) {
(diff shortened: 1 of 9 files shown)


jdeepee (dismissed):
Looks good. Shame we cannot use nix when installing holochain for windows tests. I guess will just have to remember to update the test workflow holochain versions whenever we bump our nix file. Can you add a comment in the nix file mentioning this @fayeed ?

fayeed:
@jdeepee Added comment

lucksus (approved):
Nice!

fayeed:
@lucksus can you approve the PR again.

src/core/PrologInstance.ts
```diff
@@ -1,14 +1,16 @@
...
 //@ts-ignore
 import tmp from 'tmp'
+import { resourcePath } from "./Config";
 
 export default class PrologInstance {
     #engine
 
     constructor() {
-        this.#engine = new swipl.Engine()
+        this.#engine = new swipl.Engine(path.join(resourcePath, "swipl"))
```
lucksus, line 13: Why does this work on Linux as well without a change in this PR that copies (or creates a link of) the swipl binary to resourcePath?

lucksus (approved):
Looks good! I have a question but that shouldn't block a merge since tests are obviously green.

src/core/PrologInstance.ts
fayeed: if you check the `swipl-stdio` package it doesn't use the swipl binary that we have, instead it would only use the binary if its in the $PATH

jdeepee:
can we merge this @lucksus

PR #74, opened 2022-05-25
Social-DNA example "TODO" with tests

lucksus:
- [x] TODO example SDNA
- [x] Tests


src/tests/social-dna-flow.ts
```diff
@@ -0,0 +1,106 @@
...
+                expect(flowLinks[0].data.target).toEqual("todo://doing")
+
+                expressionsInTodo = await perspective.expressionsInFlowState('TODO', 0.5)
+                expect(expressionsInTodo.length).toEqual(1)
+                expect(expressionsInTodo[0]).toEqual('test-lang://1234')
+
+                // continue flow
+                flowActions = await perspective.flowActions('TODO', 'test-lang://1234')
+                expect(flowActions.length).toEqual(1)
+                expect(flowActions[0]).toEqual("Finish")
```
kaichaosun, line 89: This is hard to tell with index 0 is `Finish`, can it be make type safe?

src/tests/sdna/todo.pl
```diff
@@ -0,0 +1,8 @@
+register_sdna_flow("TODO", t).
+flowable(_, t).
+start_action(t, '[{action: "addLink", source: "this", predicate: "todo://state", target: "todo://ready"}').
+flow_state(ExprAddr, 0, t) :- triple(ExprAddr, "todo://state", "todo://ready").
```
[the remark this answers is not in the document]
kaichaosun: I find it in the test, but this prolog syntax is hard to read and not friendly to devs, is it the only way to make sdna?

src/tests/social-dna-flow.ts
```diff
@@ -0,0 +1,106 @@
...
+                const sdna = [
+                    // The name of our SDNA flow: "TODO"
+                    'register_sdna_flow("TODO", t).',
+
+                    // What expressions can be used to start this flow? -> all
+                    'flowable(_, t).',
+
+                    // This Flow has 3 states (0=ready, 0.5=doing, 1=done), 
+                    // which are represented by links with predicate 'todo://state'
+                    'flow_state(ExprAddr, 0, t) :- triple(ExprAddr, "todo://state", "todo://ready").',
```
kaichaosun, line 21: It would be good if you can explain this line a bit..

kaichaosun (approved):
I think I understand the ideas behind SDNA, which is reuse the expression and make it flowable with logic written in prolog.. 

But it seems I can do the same thing just with another holochain language to facilitate the todo state flow? so what's the advantages to use prolog? It looks light weight with prolog to easily register such flow though.

And how is the validation works, like only a specific role could change the state?

kaichaosun:
I have been learning a little about prolog, but I would not expect eco ad4m devs being strong with this language 😢 .

PR #79, opened 2022-06-06
lair-keystore update 0.1.3

fayeed:
With the new version `lair-keystore` now expects passphrase with it in the previous version we would save the passphrase in `conductor-config.yaml` and it was `foobar` hardcoded. This also changes the flow of how the executor starts different services. With the new version, `holohain` & `lair-keystore` only start when the agent unlocks.

Previous flow (Had hardcoded passphrase `foobar`):
`Start Perspectivsm Core -> Start IPFS -> Start Lair Keystore -> Start Holohchain -> Start Graphql Server`

New Flow (Uses the agent passphrase):
`Start Perspectivsm Core -> Start IPFS  -> Start Graphql Server  -> User unlocks the agent -> Start Lair Keystore -> Start Holohchain`


jdeepee:
Lets also keep an eye out an ensure that lair-keystore process still gets killed correctly on executor close with the SIGINT command we send

src/core/storage-services/Holochain/HcExecution.ts
```diff
@@ -30,6 +32,7 @@ export interface ConductorConfiguration {
     useProxy: boolean,
     useLocalProxy: boolean;
     useMdns: boolean;
+    connectionUrl: string;
```
lucksus, line 35:
```suggestion
    lairConnectionUrl: string;
```
?

src/tests/integration.test.ts
```diff
@@ -97,6 +97,14 @@ export class TestContext {
...
     }
 
+    get aliceCore() : PerspectivismCore {
+      return this.#aliceCore!
+    }
+
+    get bobCore(): PerspectivismCore {
+      return this.#bobCore! 
+    }
+
```
lucksus, line 108: I don't see these being used anywhere...

lucksus (commented):
I have two minor remarks. Looks good otherwise!

fayeed:
@lucksus Fixed the comments.

lucksus (approved):

PR #82, opened 2022-07-01
Holochain 146 bump & lair-keystore 0.2.0

jdeepee:
- [x] Holochain property templating to use integrity zome definition
- [x] Bump holochain & lair versions
- [x] Use `//@ad4m-template-variable` to denote the location of template values in a language
- [x] Update language to new holochain version 
- [x] Update languages to use new template variable declaration
- [x] Link signals are emitted on first pulls
- [x] Polling loop for LinkLanguages


src/core/PerspectivesController.ts
```diff
@@ -35,8 +35,8 @@ export default class PerspectivesController {
             })
         }
 
-        this.#context.languageController!.addLinkObserver((diff: PerspectiveDiff, lang: LanguageRef) => {
-            let perspective = Array.from(this.#perspectiveInstances.values()).find((perspective: Perspective) => perspective.neighbourhood?.linkLanguage == lang.address);
+        this.#context.languageController!.addLinkObserver((diff: PerspectiveDiff, lang: string) => {
```
lucksus, line 38: Saw this above already. Why this change to a less expressive type when we have `LanguageRef`?

src/core/LanguageController.ts
```diff
@@ -523,25 +525,36 @@ export default class LanguageController {
...
+        if (templateData["dna"]) {
+            let dnaIndex = 0;
+            for(let i = 0; i < sourceLanguageLines.length; i++) {
+                if (sourceLanguageLines[i].includes(ad4mTemplatePattern)) {
+                    dnaIndex = i;
+                }
+            }
+            sourceLanguageLines.splice(dnaIndex, 1);
+        }
+
```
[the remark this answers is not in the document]
jdeepee: Just resolved this with a much more robust system for templating that replaces the variables. Hopefully this will also remove the templating has issues we are seeing.

src/core/PerspectivesController.ts
jdeepee: When I was console logging the lang value received in the callback it was for some reason only receiving the lang hash and not the name, so I decided just to remove the name and only send the string since the name is not used anyhow

lucksus: Hm, but even if the name is not set, I think using `LanguageRef` (which was intended for variables like this) is better already because this code is clearer. With a string you could wonder if it is the name or address... Don't think this change is an improvement.

jdeepee: Agreed, ill revert, I think I made this change whilst diagnosing the language hash bug we had

jdeepee:
apollo-server here is now updated to use v3, leveraging the new ws-graphql protocol. Hopefully this fixes the websocket CI errors we had before. Tests should be green once: https://github.com/perspect3vism/ad4m/pull/46 is merged.

@lucksus

src/core/graphQL-interface/GraphQL.ts
```diff
@@ -37,6 +45,7 @@ function createResolvers(core: PerspectivismCore, config: any) {
             },
             //@ts-ignore
             agentStatus: (parent, args, context, info) => {
+                console.warn('gggg', context);
```
lucksus, line 48:
```suggestion
```

src/core/graphQL-interface/GraphQL.ts
```diff
@@ -797,33 +806,83 @@ function createResolvers(core: PerspectivismCore, config: any) {
...
+                        async drainServer() {
+                            await serverCleanup.dispose();
+                        },
+                    };
+                },
+            },
+        ]
+    });
+
+    console.log('server', server)
```
lucksus, line 854: This is debugging leftover as well, right?
```suggestion
```

src/tests/authentication.test.ts
```diff
@@ -205,12 +229,15 @@ describe("Authentication integration tests", () => {
             let rand = await adminAd4mClient!.agent.permitCapability(`{"requestId":"${requestId}","auth":{"appName":"demo-app","appDesc":"demo-desc","appUrl":"demo-url","capabilities":[{"with":{"domain":"agent","pointers":["*"]},"can":["READ"]}]}}`)
             let jwt = await adminAd4mClient!.agent.generateJwt(requestId, rand)
 
+            console.warn("Init ad4m client1");
```
lucksus, line 232:
```suggestion
```

src/tests/authentication.test.ts
```diff
@@ -205,12 +229,15 @@ describe("Authentication integration tests", () => {
...
+            // @ts-ignore
             let authenticatedAppAd4mClient = new Ad4mClient(apolloClient(gqlPort, jwt))
-
+            console.warn("Finish init 1");
```
lucksus, line 235:
```suggestion
```

src/tests/authentication.test.ts
```diff
@@ -227,7 +254,10 @@ describe("Authentication integration tests", () => {
             let rand = await adminAd4mClient!.agent.permitCapability(`{"requestId":"${requestId}","auth":{"appName":"demo-app","appDesc":"demo-desc","appUrl":"demo-url","capabilities":[{"with":{"domain":"agent","pointers":["*"]},"can":["CREATE"]}]}}`)
             let jwt = await adminAd4mClient!.agent.generateJwt(requestId, rand)
 
+            console.warn("Init client 2");
```
lucksus, line 257:
```suggestion
```

src/tests/authentication.test.ts
```diff
@@ -227,7 +254,10 @@ describe("Authentication integration tests", () => {
...
+            // @ts-ignore
             let authenticatedAppAd4mClient = new Ad4mClient(apolloClient(gqlPort, jwt))
+            console.warn("Finish init2");
```
lucksus, line 260:
```suggestion
```

lucksus (commented):
Found several console.logs/warns, otherwise good!

lucksus (approved):

PR #87, opened 2022-07-15
Extend the expiration date.

kaichaosun:
Extends the token expiration date to be 30 days. 
We make want users to set this expiration date when making capability request.

Make it no expiration date still hold a lot of risk to data when jwt leaks.

### DIFF
--- a/src/core/agent/Auth.ts
+++ b/src/core/agent/Auth.ts
@@ -359,7 +359,7 @@ export const checkCapability = (capabilities: Capabilities, expected: Capability
     }
 }
 
-export const DefaultTokenValidPeriod = 7 * 24 * 60 * 60; // 7 days in seconds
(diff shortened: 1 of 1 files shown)


src/core/agent/Auth.ts
```diff
@@ -359,7 +359,7 @@ export const checkCapability = (capabilities: Capabilities, expected: Capability
     }
 }
 
-export const DefaultTokenValidPeriod = 7 * 24 * 60 * 60; // 7 days in seconds
+export const DefaultTokenValidPeriod = 30 * 24 * 60 * 60; // 30 days in seconds
```
lucksus, line 362:
```suggestion
export const DefaultTokenValidPeriod = 180 * 24 * 60 * 60; // 180 days in seconds
```

lucksus (commented):
What about 6 months for now?

lucksus (approved):

PR #92, opened 2022-08-10
Unlock agent with wrong password should return quickly

kaichaosun:
Related to https://github.com/perspect3vism/ad4min/issues/36


jdeepee:
Should this also throw an error, or do we catch the error field in the response from the ad4min side?

lucksus (dismissed):

kaichaosun:
> Should this also throw an error, or do we catch the error field in the response from the ad4min side?

We can catch the error field in ad4min side, and show notification once error field is present.

PR #95, opened 2022-08-16
disable cors on apollo server

jdeepee:


jdeepee:
Closing since its off by default....

jdeepee:
It was actually brave blocking requests with its shield feature

PR #106, opened 2022-09-22
remove pull from getLinks()

jdeepee:


src/core/Perspective.ts
```diff
@@ -420,12 +420,6 @@ export default class Perspective {
     }
 
     async getLinks(query: LinkQuery): Promise<LinkExpression[]> {
-        const remoteLinks = await this.callLinksAdapter('pull')
```
kaichaosun, line 423: Instead of removing it, is it better to trigger a `pull and save` call in background, but without waiting the result?

PR #110, opened 2022-10-07
Intelligent link language pull

jdeepee:
This PR ensures that after joining a neighbourhood, we do not commit any entries to a link language until we have seen something ourselves, so long as the neighbourhood as one we joined and not created. 

With the holochain implementation of the link language, this has the benefit of enabling pull to use snapshots to fetch vs pulling on the raw graph. This is because the snapshot fetching code will only get triggered when currentRevision is null.

Additionally the first commit made to a link language will be batched into one diff mutation, reducing the number of commit entries made into the link language.

### DIFF
--- a/src/core/graphQL-interface/GraphQL.ts
+++ b/src/core/graphQL-interface/GraphQL.ts
@@ -33,15 +33,19 @@ function createResolvers(core: PerspectivismCore, config: OuterConfig) {
             agentByDID: async (parent, args, context, info) => {
                 checkCapability(context.capabilities, Auth.AGENT_READ_CAPABILITY)
                 const { did } = args;
-                const agentLanguage = core.languageController.getAgentLanguage().expressionAdapter;
-                if (!agentLanguage) {
-                    throw Error("Agent language does not have an expression adapter")
-                }
-                const expr = await agentLanguage.get(did);
-                if (expr != null) {
-                    return expr.data;
+                if (did != core.agentService.did) {
+                    const agentLanguage = core.languageController.getAgentLanguage().expressionAdapter;
+                    if (!agentLanguage) {
+                        throw Error("Agent language does not have an expression adapter")
+                    }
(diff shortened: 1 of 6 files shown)


lucksus (commented):
Hm, I was expecting to see a flag added to links in the DB signaling if it was shared, which would help with the future story of having local-only links. But I think i get why that is not that easy: link removals. Is that the reason why you're storing the pending diffs?

src/core/graphQL-interface/GraphQL.ts
```diff
@@ -33,15 +33,19 @@ function createResolvers(core: PerspectivismCore, config: OuterConfig) {
...
+                    }
+                    const expr = await agentLanguage.get(did);
+                    if (expr != null) {
+                        return expr.data;
+                    } else {
+                        return null
+                    }
                 } else {
-                    return null
+                    return core.agentService.agent
```
lucksus, line 48: Is this needed for this feature for some reason, or just optimization on the side? I was wondering about scenarios where `core.agentService.agent` could get outdated compared to the value in the language, but I think it should be ok... ?

jdeepee:
> Hm, I was expecting to see a flag added to links in the DB signaling if it was shared, which would help with the future story of having local-only links. But I think i get why that is not that easy: link removals. Is that the reason why you're storing the pending diffs?

Yes exactly. I started with this approach, but once I saw removals I realised how it doesn't work and then went forward with this approach.

src/core/graphQL-interface/GraphQL.ts
jdeepee: Just an optimization. I don't think `core.agentService.agent` in its current form can become out of sync with the language since the agent service gets updated whenever we mutate the users profile via `agentUpdatePublicPerspective` & `agentUpdateDirectMessageLanguage`. But definitely something to keep an eye on if we change these functions at all.

lucksus (approved):

PR #121, opened 2022-11-04
AD4M cli

lucksus:
* Rust port of `Ad4mClient`. Published to https://crates.io/crates/ad4m-client
* New Rust based AD4M cli app using above Ad4mClient to implement a full generic and scriptable AD4M client binary. Published to https://crates.io/crates/ad4m.

Plus new integration tests that spawn an executor through ad4m-host binary and using ad4m cli to exercise it. (breaking since they reproduce a real bug in current main branch)


cli/src/agent.gql
```diff
@@ -0,0 +1,71 @@
...
+        isUnlocked
+        did
+        didDocument
+        error
+    }
+}
+
+mutation RequestCapability($appName: String!, $appDesc: String!, $appUrl: String!, $capabilities: String!) {
+    agentRequestCapability(appName: $appName, appDesc: $appDesc, appUrl: $appUrl, capabilities: $capabilities)
+}
```
jdeepee: are we able to have these queries generated / imported from core in anyway?

cli/src/main.rs
```diff
@@ -0,0 +1,698 @@
...
+                let agent = agent::run_me(cap_token).await?;
+                print_agent(agent.into());
+            }
+            AgentFunctions::Status => {
+                let status = agent::run_status(cap_token).await?;
+                println!(
+                    "\x1b[36mDID: \x1b[97m{}",
+                    status.did.unwrap_or_else(|| "<undefined>".to_string())
+                );
+                println!("\x1b[36mis_initiliazed: \x1b[97m{}", status.is_initialized);
```
jdeepee: Shall we put this whole match block inside its own function so we can easily call it from tests?

cli/src/main.rs
```diff
@@ -0,0 +1,698 @@
...
+                    print_prolog_results(results)?;
+                }
+                PerspectiveFunctions::Watch { id } => {
+                    perspectives::run_watch(cap_token, id).await?;
+                }
+                PerspectiveFunctions::Snapshot { id } => {
+                    let result = perspectives::run_snapshot(cap_token, id).await?;
+                    println!("{:#?}", result);
+                }
+                PerspectiveFunctions::Repl { id } => {
```
jdeepee: WOAH! This is so useful!!!

cli/src/main.rs
```diff
@@ -0,0 +1,698 @@
...
+                    neighbourhoods::run_publish(cap_token, link_language, None, perspective_id)
+                        .await?;
+                println!("Neighbourhood shared as: {}", neighbourhood);
+            }
+            NeighbourhoodFunctions::Join { url } => {
+                let neighbourhood = neighbourhoods::run_join(cap_token, url).await?;
+                println!("Neighbourhod joined!\n{:#?}", neighbourhood);
+            }
+        },
+        Domain::Runtime { command } => match command {
```
jdeepee: another match block which should be its own function

jdeepee (approved):
Looks amazing! Just a few small ideas

jdeepee (dismissed):
Also just realised a few other things: where can I set the bootstrap seed file path? And also should we handle ad4m init like we did in the old cli tool. I believe ad4min uses this to fetch binaries & bootstrap seeds, guess we would need something similar for ad4min to still work. If not we should also update ad4min in this PR to work or without that function.

jdeepee:
Also when publishing a language it would be great to have the option to pass language language only (and this getting put through to executor config) for cases where we need to publish languages for new holochain version, of which we do not have any existing compatible languages

cli/src/agent.rs
```diff
@@ -0,0 +1,144 @@
...
+#[graphql(
+    schema_path = "schema.gql",
+    query_path = "src/agent.gql",
+    response_derives = "Debug"
+)]
+pub struct RequestCapability;
+
+pub async fn run_request_capability() -> Result<String> {
+    let query = RequestCapability::build_query(request_capability::Variables {
+        app_name: "AD4M cli".to_string(),
```
kaichaosun: Is this mock data expected?

cli/src/formatting.rs
```diff
@@ -0,0 +1,128 @@
...
+
+use crate::types::{Agent, LinkExpression, PerspectiveExpression, SentPerspectiveMessage};
+
+pub fn print_prolog_results(results: Value) -> Result<()> {
+    match results {
+        Value::Bool(true) => println!("true ✅"),
+        Value::Bool(false) => println!("false ❌"),
+        Value::String(string) => println!("{}", string),
+        Value::Array(array) => {
+            println!("\x1b[90m{} results:", array.len());
```
kaichaosun, line 12: The log format seems confused to me, can you elaborate a bit?

cli/src/formatting.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+        _ => bail!("Unexpected non-obhect value in prolog result: {:?}", result),
+    }
+    Ok(())
+}
+
+fn maybe_decode_literal(uri: String) -> String {
+    if uri.starts_with("literal://") {
+        let literal = uri.replace("literal://", "");
+        if literal.starts_with("string:") {
```
kaichaosun, line 71: Use pattern match usually will make the logic more readable.

cli/src/formatting.rs
```diff
@@ -0,0 +1,128 @@
...
+    } else {
+        uri
+    }
+}
+
+pub fn print_link(link: LinkExpression) {
+    let source = maybe_decode_literal(link.data.source);
+    let target = maybe_decode_literal(link.data.target);
+
+    if let Some(pred) = link.data.predicate {
```
kaichaosun: It's better use `match` here.

lucksus:
> Also just realised a few other things: where can I set the bootstrap seed file path? And also should we handle ad4m init like we did in the old cli tool. I believe ad4min uses this to fetch binaries & bootstrap seeds, guess we would need something similar for ad4min to still work. If not we should also update ad4min in this PR to work or without that function.

This is NOT a replacement to ad4m-host. This is just a UI... a command-line UI... So bootstrap seed handling is still being managed by the old code.. (for now)

lucksus:
> Also when publishing a language it would be great to have the option to pass language language only (and this getting put through to executor config) for cases where we need to publish languages for new holochain version, of which we do not have any existing compatible languages

Yeah, this is also an executor parameter. This does not include the executor.. But the whole bootstrap seed creation is something we should look into including here..

I would say, next step should be having another Rust binary that replaces the current JS/vercel ad4m-host binary and do those things there..

cli/src/agent.rs
lucksus: That's not mock data, but the name of this app.

cli/src/formatting.rs
lucksus: That's ANSI escape codes for color: https://en.wikipedia.org/wiki/ANSI_escape_code#Colors

cli/src/agent.gql
lucksus: They are currently inline in the JS client code. And they are a bit different here (the name) since it has to match with the Rust struct name in order to use clap's derivation style... so I would keep it this way which is how GraphQL is meant: the client is in charge of the query...

cli/src/main.rs
lucksus: I have extracted all functions into modules now.

rust-client/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+All functions require a capability token passed as parameter (plus potential other parameters)
+and return a future.
+
+Note that the executor URL needs to be set globally first by calling `set_executor_url(url String)`:
+
+```Rust
+use ad4m_client::{set_executor_url, agent};
+
+set_executor_url("https://localhost:3000/graphql".to_string());
```
kaichaosun: It would be better if we make API like,
```
let client = Ad4mClient::new(executor_url);
let status = client.agent.status().await?;
```

lucksus: done and published as v0.0.2: https://crates.io/crates/ad4m-client

lucksus:
The failing (new) integration test makes visible the problem we currently have with some macOS systems. That tests passes on my local machine but fails in the same way we saw on @leifriksheim's device. I suggest we merge this with this new failing test since it is real feedback that also applies to what is in master.

kaichaosun (approved):

PR #122, opened 2022-11-09
chore: add correct download links

leifriksheim:


connect/src/components/Start.ts
```diff
@@ -1,5 +1,45 @@
 import { html } from "lit";
 
+function downloadAd4m() {
+  const windowsLink =
+    "https://github.com/perspect3vism/ad4m/releases/download/v0.0.11/AD4Min_0.0.11_x64_en-US.msi";
```
lucksus: this version is wrong. needs to be 0.2.x.

lucksus (changes requested):

PR #128, opened 2022-11-30
WIP: Runtime binary fetch

jdeepee:
As reported by Leif, with our latest attempt to fix the holochain binaries. When the binaries that AD4MIN runs come from the AD4MIN bundle, he receives a websocket error / crash. But if he manually copies the signed binaries to his .ad4m/binary directory directory... everything works fine. This led me to believe there is something damaging our binaries in the build process, so instead of including holochain at build time... I instead fetch at runtime. 

This PR also fixes the issue of ad4min log output being 300mb by removing the UI window log in tauri.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -48,5 +48,5 @@ jobs:
       run: ./tests/bats/bin/bats tests/binaries.bats
 
     - name: Run integration test script
-      run: ./tests/bats/bin/bats tests/integration.bats
+      run: ./tests/bats/bin/bats tests/integration.bats || cat ~/Library/Logs/DiagnosticReports/*
     
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,7 +55,7 @@ dependencies = [
 
 [[package]]
 name = "ad4min"
-version = "0.0.10"
+version = "0.2.1"
 dependencies = [
  "directories",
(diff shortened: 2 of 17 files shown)


jdeepee:
If we wish to move forward with this approach, I suggest we setup a feedback process between the ad4m init command and the splash screen, so the user knows that the initial setup is taking place...

lucksus:
@jdeepee, check the log of the integration test in this PR. It is still failing but not with "websocket closed" anymore. It failed because of the central language language. Seems to fix it at least for CI.

jdeepee:
@lucksus interesting, I do however see a timeout when installing the agent-language DNA. Maybe the languages store is returning bad results as well as being flaky on post, seems unlikely though

jdeepee:
Its also possible that the 504 just stops us getting to the point where we have the socket is not opened error. See this run (https://github.com/perspect3vism/ad4m/actions/runs/3586440356/jobs/6035632540). The `Socket is not open` error comes after agent & direct message language templating, which the test run for this PR never did.

lucksus:
Hm, ok. I will restart the test here anyway to see if it gets any further now. Central Language language was working for me just now.

jdeepee:
closing since I believe this will not solve the socket issue, but is instead something else more systemic

PR #130, opened 2022-11-30
add client & executor methods for perspective add local link

jdeepee:


lucksus:
Hm, I thought we could introduce some kind of link-status such that we:
1. can get more info on links when we query them, to know which link is shared and which ist not
2. can use the existing link-add functions plus a new optional parameter for the link-status

jdeepee:
Yeah those sounds like nice additions, will implement those asap

fayeed:
@jdeepee @lucksus I have added a new property for links `status` which can be for now `shared` or `local` to indicate link status.

Right when running the test locally one tests fail with the proof set to `invalid = true`

jdeepee:
@fayeed the reason the agent proof is failing is because the snapshot contains link share information, but then the holochain DNA which stores the agent expression does not capture the information. This means the signature is generating to contain the shared status, but the stored expression does not actually contain this. I will update the agent language so that it just deletes the shared status.

jdeepee:
@fayeed lets just add one more test case to test if the links actually get shared or not. This can be done inside the neighbourhood tests by adding a local link and then checking the other agent does not receive it. Then also adding a shared link and checking that the other agent does receive it.

jdeepee:
@lucksus please also add a review, id mark this complete once the above two points are implemented

jdeepee:
Since the latest agent-language uses the new holochain version, I will first get this holochain version working, merge into dev, then we can merge those changes into here with the fixed agent-language which deletes agent-status

fayeed:
@jdeepee Added new tests.

executor/src/tests/neighbourhood.ts
```diff
@@ -81,6 +94,36 @@ export default function neighbourhoodTests(testContext: TestContext) {
                 expect(bobLinks.length).to.be.equal(1)
             })
 
+            it('local link created by Alice received by Bob', async () => {
```
lucksus:
```suggestion
            it('local link created by Alice NOT received by Bob', async () => {
```

lucksus (dismissed):
Looks good. Just spotted a misleading test name. Suggested change in comments.

PR #132, opened 2022-12-03
WIP: CLI command for publishing and creating a new bootstrap

jdeepee:
WIP because the code needs to be cleaned up, and the logs improved. Otherwise it is functional and has been used to succesfully generate a new bootstrap seed.

Note: this spawns its own ad4m-host instance, and moves agent data around. This is added for convience since when updating the language language, if you did not automate this process you would have to manually create the temporary publishing bootstrap seed, copy agent and then run ad4m-host init + serve.  With this structure that is all handled via the cli.

Would like to hear you first thoughts @lucksus


cli/src/languages.rs
```diff
@@ -147,6 +159,114 @@ pub async fn run(ad4m_client: Ad4mClient, command: Option<LanguageFunctions>) ->
...
+            seed_proto,
+        } => {
+            println!(
+                "Attempting to generate a new bootstrap seed using ad4m-host path: {:?} and agent path: {:?}\n",
+                ad4m_host_path,
+                agent_path
+            );
+
+            //Warn the user about deleting ad4m agent during publishing
+            println!("WARNING... THIS WILL DELETE YOUR EXISTING AD4M AGENT AND REPLACE WITH SUPPLIED PUBLISHING AGENT, PLEASE BACKUP BEFORE PROCEEDING\n\n");
```
lucksus: I don't think this needs to be the case, no? We don't need to use `executor_data_path()` below, but could just use a temp directory.

cli/seed_proto.json
```diff
@@ -0,0 +1,59 @@
+{
+    "languageLanguageRef": "/Users/josh/dev/language-persistence/build/bundle.js",
+    "perspectiveDiffSync": {
```
lucksus: `perspectiveDiffSync` is different to the other languages in that it is the name of a specific link language implementation, while the other languages need to have exactly one - and `agentLanguage` for instance describes the role of that language.

We could have this be an array with the name "linkLanguages" and then we could really fill the `knownLinkLanguages` array with multiple.

lucksus (commented):
This is awesome, and I'd be happy to merge after some cleanup. I have two small suggestions though, that I wouldn't consider blocking the merge..

cli/src/languages.rs
jdeepee: Ah yes I think that is possible and would avoid having to backup and delete existing agents.

cli/seed_proto.json
jdeepee: yeah right; ill rename the key to linkLanguages and make it an array

lucksus (approved):
perfect

PR #149, opened 2022-12-22
AD4Min -> AD4M: also change name of files in ad4m-connects download links

lucksus:
The Tauri action in our publish workflow picks up the filenames from the project, right?
Is there anything more to do here?


jdeepee (approved):
Looks good to me, @fayeed have a double check and let us know if there are any other changes needed. Will merge now anyhow since I expect we will release again today.

PR #157, opened 2023-01-05
WIP: Fix release workflow

lucksus:
After renaming AD4Min to AD4M, the artifact names have changed as well. The Tauri publish action did turn "ad4min" into "ad4-min" only on Linux. Now it creates files called "ad4m.deb.." etc. but still expects "ad4-min.deb..". Just had a quick look and not sure how to fix this. Just leaving this WIP here for later..


jdeepee:
closing since this is fixed by: https://github.com/perspect3vism/ad4m/pull/158

PR #158, opened 2023-01-05
Ad4m launcher binaries are uploaded manually

fayeed:


jdeepee:
squashed and merged into dev manually, closing here

PR #177, opened 2023-01-09
fix joined neighbourhood not setting isJoined value

jdeepee:
This PR ensure that the executor will wait to receive changes before it publishes any commits! 1 line fix, yay!


lucksus (approved):
Would be nice to have a test dir this :)

PR #182, opened 2023-01-09
Wait for repo.lock to disappear before running ad4m-host startup

### LINKED ISSUE #167: If AD4M launcher detects a `repo.lock` file in the .ad4m directory, it should delay starting until it is gone
This usually occurs when trying to start ad4m launcher in rapid succession and the old IPFS engine is still closing

jdeepee:
Closes: #167. 

Right now if you open AD4M launcher, quit and then quickly re-open another window, IPFS init will error due since the old IPFS node is still shutting down. It seems that the previous implementation of deleting the file did not work... here we will just wait until the lock is release and then open the launcher.

### DIFF
--- a/ui/src-tauri/src/main.rs
+++ b/ui/src-tauri/src/main.rs
@@ -64,8 +64,9 @@ fn main() {
         let _ = remove_dir_all(data_path());
     }
 
-    if data_path().join("ad4m").join("ipfs").join("repo.lock").exists() {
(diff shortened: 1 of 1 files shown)


lucksus (commented):
Basically good, but couldn't this also result in the Launcher just not showing up if the repo.lock won't be freed. Naive users might not understand what's going on.

So maybe we should have a user prompt in the GUI after some time..?

lucksus (approved):

PR #185, opened 2023-01-10
Show Connection error screen if localhost blocked - closes #69

bnhovde:
After some debugging I believe GraphQLWsLink -> on.error is the place to hook into these connection errors.

This needs to be tested, but locally I get the following results:

- localhost unblocked: Normal flow
- localhost blocked: New screen
- network disconnected: No change

@jdeepee - The below used to specify `localhost:3000` as the URL to be unblocked, but I think `ws://localhost:12000/graphql` is the actual URL requested. Do we specify this or leave it as 'localhost'?

image

### DIFF
--- a/connect/src/components/CouldNotMakeRequest.ts
+++ b/connect/src/components/CouldNotMakeRequest.ts
@@ -6,15 +6,19 @@ export default function CouldNotMakeRequest() {
       <div class="text-center">
         <h1 class="heading">Request to AD4M blocked</h1>
         <div class="body">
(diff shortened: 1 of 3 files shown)


bnhovde:
FYI: To recreate locally using brave:

Screenshot 2023-01-10 at 15 25 56
Screenshot 2023-01-10 at 15 26 24

jdeepee (approved):
Amazing! Working great for me, feel free to merge.

PR #186, opened 2023-01-10
Agent deletion/migration logic between AD4M versions

jdeepee:
@lucksus made a simple naive pass at implementing this here. Do we feel it is right to have this in ad4m-host or should it be in the executor?


host/src/commands/prepare.ts
```diff
@@ -0,0 +1,84 @@
...
+        // Path to our config
+        const configPath = path.join(appDataPath, 'ad4m-host-config.json');
+        // Path to bootstrap seed
+        const bootstrapSeedPath = path.join(appDataPath, 'mainnet_seed.json');
+        // Path to holochain data
+        const holochainDataPath = path.join(appDataPath, 'ad4m', 'h');
+        // Path to languages
+        const languagesPath = path.join(appDataPath, 'ad4m', 'languages');
+        // path to perspectives
+        const perspectivesPath = path.join(appDataPath, 'ad4m', 'perspectives.json');
```
lucksus: Can't we keep the perspectives? I guess the problem is the Neighbourhood status, after deleting all Languages, but maybe we can handle that without deleting?

lucksus (approved):

PR #189, opened 2023-01-10
Added a way to revoke tokens & remove apps.

fayeed:


jdeepee (changes requested):
Seems to be working well for me locally, just two things:

1. Lets add some tests into the executor which test the new graphql methods added.
2. Deleted apps should arrive in a "Deleted Apps" list in the ad4m launcher. From here it should then be possible to remove the app from the deleted apps list and give it access once again. How does this addition feel to you @lucksus ?

jdeepee:
Or even if there is not UI for an app being enabled again, atleast the app itself should be able to request capabilities once again without having that request blocked.

executor/src/core/agent/AgentService.ts
```diff
@@ -314,6 +314,10 @@ export default class AgentService {
 
         return payload.capabilities
     }
+
+    isAd4minCredential(token: string) {
```
lucksus: I'm confused about the name. It should be 
```suggestion
    isAdminCredential(token: string) {
```
right?

executor/src/core/graphQL-interface/GraphQL.ts
```diff
@@ -15,22 +15,27 @@ import * as PubSub from './PubSub'
...
 function createResolvers(core: PerspectivismCore, config: OuterConfig) {
     const pubsub = PubSub.get()
-
+    const rootConfigPath = path.join(config.appDataPath, 'ad4m');
     return {
         Query: {
             //@ts-ignore
             agent: (parent, args, context, info) => {
+                const isAd4minCredential =  core.agentService.isAd4minCredential(context.authToken)
+                checkTokenAuthorized(rootConfigPath, context.authToken, isAd4minCredential)
```
lucksus: I would like to see the name changed (isAd4min to isAdmin), but why not put `checkTokenAuthorized()` into `AgentService` and have one line less in this block that is copied in each of these functions? (Or maybe even merge with / put into `checkCapabality()` to avoid even more copy-pasta)..

lucksus (changes requested):

executor/src/core/graphQL-interface/GraphQL.ts
fayeed: I initially thought to put it within the `checkCapability` but then I thought maybe it's better to have it separate as it does something different than the function.

Yeah I guess it's better to be handled as a middleware in graphql.

fayeed:
@jdeepee maybe having a filter dropdown or something in the apps page to filter out revoked, deleted, active & all apps would work.

fayeed:
> Or even if there is not UI for an app being enabled again, atleast the app itself should be able to request capabilities once again without having that request blocked.

This has been taken care in the ad4m-connect.

jdeepee:
> > Or even if there is not UI for an app being enabled again, atleast the app itself should be able to request capabilities once again without having that request blocked.
> 
> This has been taken care in the ad4m-connect.

How / where was this resolved in ad4m-connect?

jdeepee:
This looks good to merge once we have some tests for this in the executor

fayeed:
I have pushed a fix for this in PR and made a small change in ad4m-connect where it treats it as an invalid token and they can request a new token.

executor/src/tests/app.test.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      // @ts-ignore
+      unAuthenticatedAppAd4mClient = new Ad4mClient(apolloClient(gqlPort))
+  })
+
+  after(async () => {
+      await agentCore!.exit();
+  })
+
+  it("once token issued user can one app", async () => {
```
jdeepee: Fix the name

executor/src/tests/app.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      let authenticatedAppAd4mClient = new Ad4mClient(apolloClient(gqlPort, jwt))
+  
+      const call = async () => {
+          return await authenticatedAppAd4mClient!.agent.getApps();
+      }
+  
+      await expect((await call()).length).to.be.equal(1);
+  });
+  
+  it("can revoke token", async () => {
```
jdeepee, line 120: Lets also add test case here that we can request the capability again here

jdeepee (changes requested):

executor/src/core/agent/Auth.ts
```diff
@@ -351,6 +353,25 @@ export const RUNTIME_EXCEPTION_SUBSCRIBE_CAPABILITY: Capability = {
     can: [SUBSCRIBE]
 }
 
+export const checkTokenAuthorized = (rootConfigPath: string, token: string, isAd4minCredential: boolean) => {
+    let apps: any[] = [];
+    if (!isAd4minCredential) {
+        if (fs.existsSync(path.join(rootConfigPath, "apps.json"))) {
+            apps = JSON.parse(fs.readFileSync(path.join(rootConfigPath, "apps.json")).toString())
```
lucksus: I'm a bit worried that this file access can slow things down since this function gets called in every interface query.  Might make sense to hold the `apps` array in memory as well?

lucksus (dismissed):
🥇

jdeepee (approved):

PR #207, opened 2023-01-23
Link updated listener

### LINKED ISSUE #206: Add update link listener to perspectiveClient

fayeed:
Updating a link triggers two events remove & add, which means apps would have first to remove a link & then add it if they rely on the listeners. This PR adds a new event listener to handle this use case.


executor/src/core/Perspective.ts
```diff
@@ -509,14 +515,11 @@ export default class Perspective {
...
         const perspective = this.plain();
         this.#prologNeedsRebuild = true;
-        this.#pubsub.publish(PubSub.LINK_REMOVED_TOPIC, {
-            perspective: perspective,
-            link: oldLink
-        })
-        this.#pubsub.publish(PubSub.LINK_ADDED_TOPIC, {
-            perspective: perspective,
-            link: newLinkExpression
-        })
```
lucksus, line 519: Could it be that tests are failing because of these removals? Should we not trigger all three signals?

fayeed: Fixed this it was just an error due to an undefined callback array 😅

lucksus (approved):

PR #209, opened 2023-01-25
Fixed IPFS lock issue

### LINKED ISSUE #205: Fix fast restarts of AD4M launcher causes error

fayeed:


leifriksheim (dismissed):

ui/src-tauri/src/main.rs
```diff
@@ -70,9 +70,10 @@ fn main() {
         let _ = remove_dir_all(data_path());
     }
 
-    while data_path().join("ad4m").join("ipfs").join("repo.lock").exists() {
-        std::thread::sleep(std::time::Duration::from_secs(1));
-        println!("IPFS repo.lock exists, waiting...");
+    while data_path().join("ipfs").join("repo.lock").exists() {
```
lucksus, line 74: Looks like removing "ad4m" in this line might be the fix already? I'm a bit weary of always and immediately deleting the lock...

lucksus: Pushed a change.

lucksus (approved):

ui/src-tauri/src/main.rs
fayeed: Actually tried just doing that but it didn't fix it for me, could be just because of windows.

fayeed: Also, I checked the IPFS go implementation there was an issue there too for this and they just deleted the IPFS lock.

lucksus: So do you think we should revert this? Could you try on your windows machine? This is working for me on macOS.

fayeed: This works it's just that for me on windows the lock never releases, it could just be a windows thing.

PR #217, opened 2023-01-30
add check if neighbourhood is already installed

### LINKED ISSUE #201: Deleting a DNA does not cause the holochain database state to be deleted
(Needs to be investigated further, but likely the case)

jdeepee:
I did not need to implement holochain DNA deletion logic for languages since we now install DNA's provided by languages at the holochain app level, with each language having its own app & internal DNA's. 

Since it is not possible to install multiple languages, it should not be possible to install duplicate holochain apps. It may however be possible to have two different languages expose the same underlying holochain DNA, in this case maybe it would be best to have the two languages share one app / dna. We can address this later if it arrises as a common pattern.

closes #201

### DIFF
--- a/executor/src/tests/neighbourhood.ts
+++ b/executor/src/tests/neighbourhood.ts
@@ -37,12 +37,8 @@ export default function neighbourhoodTests(testContext: TestContext) {
 
                 const perspective = await ad4mClient.perspective.byUUID(create.uuid);
                 expect(perspective?.neighbourhood).not.to.be.undefined;
-
-                const join = await ad4mClient.neighbourhood.joinFromUrl(publishPerspective );
-                expect(join.sharedUrl).to.be.equal(publishPerspective);
-                expect(join.neighbourhood).not.to.be.undefined;
-                expect(join.neighbourhood!.linkLanguage).to.be.equal(socialContext.address);
-                expect(join.neighbourhood!.meta.links.length).to.be.equal(1);
+                expect(perspective?.neighbourhood!.linkLanguage).to.be.equal(socialContext.address);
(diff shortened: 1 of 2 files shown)


executor/src/tests/neighbourhood.ts
```diff
@@ -37,12 +37,8 @@ export default function neighbourhoodTests(testContext: TestContext) {
 
                 const perspective = await ad4mClient.perspective.byUUID(create.uuid);
                 expect(perspective?.neighbourhood).not.to.be.undefined;
-
-                const join = await ad4mClient.neighbourhood.joinFromUrl(publishPerspective );
-                expect(join.sharedUrl).to.be.equal(publishPerspective);
-                expect(join.neighbourhood).not.to.be.undefined;
-                expect(join.neighbourhood!.linkLanguage).to.be.equal(socialContext.address);
-                expect(join.neighbourhood!.meta.links.length).to.be.equal(1);
```
lucksus, line 45: Why removing the `joinFromUrl()` here? Now we're not testing the whole Neighbourhood flow anymore. That doesn't seem right.

lucksus: Ah, I now see that the test was exercising only Alice. So it must have triggered the new condition in `installNeighbourhood`. What we need to do instead is having Bob joining the NH and keeping these assertions.

lucksus (changes requested):

executor/src/tests/neighbourhood.ts
jdeepee: That test is already implement, its in the same file, called: 'can be created by Alice and joined by Bob'

lucksus (approved):

PR #218, opened 2023-01-31
Tweaks holochain gossip loops to be much faster

jdeepee:
Seems to be working well, but need to do more testing for CPU load increases & improvements on DHT propogation time


lucksus:
Should this go into today's release?

PR #229, opened 2023-02-06
Subject API updates

fayeed:
Example implementation: https://github.com/fluxsocial/flux/pull/366


core/src/subject/SDNADecorators.ts
```diff
@@ -127,102 +122,107 @@ function makeRandomPrologAtom(length: number): string {
...
-            let propertyCode = `property(${uuid}, "${property}").\n`
-
-            let { through, initial, required, resolve, resolveLanguage } = properties[property]
-
-            if(resolve) {
-                propertyCode += `property_resolve(${uuid}, "${property}").\n`
+ interface SDNAClassOptions {
+    through: string;
+    initial?: string,
+    required?: boolean,
```
lucksus: Not sure how these options make sense on the class level, since they modify a property.
I know this is a WIP, was just eager to see how the class decorator might work. :) Adding `generateSDNA` in there like you do below is great - but I just couldn't get it to work such that TSC knows about that function in the decorated classes...

fayeed: We can rename the properties to something else that might fit their usecase at the class level, As for the `generateSDNA` my understanding was others won't directly use it if we provide an ORM, I see that we are using it in the tests so just adding a `generateSDNA` method should suffice, I did try to implement `ActiveRecord` implementation but there was something weird happening I would always get the community subject class back.

lucksus: Oh yeah, that's right! We don't need the type-safety for `generateSDNA`... great!

core/src/subject/SDNADecorators.ts
```diff
@@ -75,14 +75,20 @@ interface PropertyOptions {
...
     return function <T>(target: T, key: keyof T) {
         target["__properties"] = target["__properties"] || {};
         target["__properties"][key] = target["__properties"][key] || {};
         target["__properties"][key] = { ...target["__properties"][key], ...opts }
+
+        if (opts.resolveLanguage) {
+            const value = key as string
+            target[`set${capitalize(value)}`] = () => {}
+        }
+
```
lucksus, line 90: As per our discussion earlier today and my comments in https://github.com/fluxsocial/flux/pull/366, `resolveLanguage` should not be what decides if we generate a setter. Instead add an explicit option to the decorator: `writable: true/false`.

lucksus:
I've added most comments in https://github.com/fluxsocial/flux/pull/366 since that's where we have example code using this. But we need to make the tests work here anyways, so the examples in https://github.com/perspect3vism/ad4m/blob/dev/tests/js/integration.test.ts need to be updated.

And another change that needs to happen then is the code in the CLI (in perspectives -> repl). But we don't need to add that to this PR. I'm happy to have a go at it once this is merged...

core/src/subject/SDNADecorators.ts
```diff
@@ -74,15 +74,43 @@ interface PropertyOptions {
...
+}
+export function subjectFlag(opts: FlagOptions) {
+    return function <T>(target: T, key: keyof T) {
+        target["__properties"] = target["__properties"] || {};
+        target["__properties"][key] = target["__properties"][key] || {};
+        target["__properties"][key] = { 
+            ...target["__properties"][key],
+            through: opts.through,
+            required: true,
+            initial: opts.value
```
lucksus: Hm, don't we need to store that it is a flag, which should have the consequence that the instance check not only checks for any link with the `through` predicate, but also that the target is `value`?

core/src/subject/SDNADecorators.ts
```diff
@@ -98,143 +126,153 @@ export function subjectCollection(opts: CollectionOptions) {
...
+                    propertyCode += `property_resolve(${uuid}, "${property}").\n`
+                    propertyCode += `property_resolve_language(${uuid}, "${property}", "${resolveLanguage}").\n`
+                }
+                
+                if(through) {
+                    propertyCode += `property_getter(${uuid}, Base, "${property}", Value) :- triple(Base, "${through}", Value).\n`
+    
+                    if(required) {
+                        instanceConditions.push(`triple(Base, "${through}", _)`)
+                    }
```
lucksus, line 198: Yeah, here we are missing the case the this property is a flag in which case the last triple paramter must also be set to the flag value..

lucksus (changes requested):
Amazing work! 👏  
Just one little remark about the flag...

lucksus (dismissed):
Implemented my change requests myself

lucksus (approved):

PR #259, opened 2023-02-20
connect exports are updated

fayeed:


leifriksheim (dismissed):

leifriksheim:
Could we maybe export getAd4mClient from the web, so that we can do `import Ad4mConnect, { getAd4mClient}`?

lucksus:
Looks good, needs changelog

PR #273, opened 2023-02-27
Launcher language fixes

### LINKED ISSUE #250: Launcher's exception pop-up UI is broken
When the executor tries to install a Language but can't verify the author of the code is trusted, the Launcher flashes (the main window pops up) but it doesn't show any information or UI to add this untrusted author to the list of trusted agents.

Even when navigating to the trusted agents overview, there is no button to add an agent.

This is a regression as both was working/possible before.

fayeed:
closes #250 and #221


jdeepee (approved):

ui/src/components/Language.tsx
```diff
@@ -234,13 +234,13 @@ const Language = (props: Props) => {
...
                 variant="link"
-                onClick={() => setInstallLanguageModalOpen(false)}
+                onClick={() => setPublishLanguageModalOpen(false)}
               >
                 Cancel
               </j-button>
               <j-button
                 variant="primary"
-                onClick={installLanguage}
+                onClick={publishLanguage}
```
jdeepee, line 220: if this is for installing a language, it should not be calling publish right?

jdeepee (changes requested):

PR #278, opened 2023-02-28
Launcher UX improvements

jdeepee:
- currently after login / registration the system tray will open in the middle of the screen; this is due to tauri not knowing the system tray position yet (since the user has not clicked the system tray icon). I don't think there is currently any way around this and for now we will have to leave it as it is.
- @fayeed can you confirm on windows if the system tray is opening at the bottom middle; if not we will have to add a custom position for windows in the open_tray command function
- @leifriksheim please make the App.tsx look pretty if the user has no apps installed yet <3


fayeed:
Its aligned top now for windows.

ui/src-tauri/src/commands/app.rs
```diff
@@ -15,6 +20,23 @@ pub fn close_main_window(app_handle: tauri::AppHandle) {
...
 }
 
+#[tauri::command]
+pub fn open_tray(app_handle: tauri::AppHandle) {
+    let window = get_main_window(&app_handle);
+    if let Ok(true) = window.is_visible() {
+        let _ = window.hide();
+    } else {
+        let _ = window.show();
+        let _ = window.move_window(Position::TopRight);
```
fayeed, line 30:
```suggestion
        let _ = window.move_window(Position::TrayRight);
```
This should work

fayeed: Actually this breaks on windows, I guess the tray icon is not created and it tries to get its position to align it self.

lucksus: Yeah, I remember bumping into the problem that we can't get tray coordinates before we have initialized the tray icon...

leifriksheim:
I added appUrl to requestCapabilities, but need some more work to get the graphQL request working. Could you have a look @fayeed ?

Commit: 3b12014

connect/index.html
```diff
@@ -16,8 +16,8 @@ <h1 id="connection"></h1>
       import Ad4mConnectUI from "./src/web.ts";
 
       const ui = Ad4mConnectUI({
-        appName: "ad4m-connect-example",
-        appDesc: "hello",
+        appName: "AD4M Connect",
+        appDesc: "A description of the app",
         appDomain: "http://localhost:3000",
```
lucksus:
```suggestion
        appUrl: "http://localhost:3000",
        appDomain: "dev.ad4m.connect.example",
```

lucksus (dismissed):
👏 👏 👏

lucksus (approved):

PR #296, opened 2023-03-03
Convenience fn: sendSignalU(p: PerspectiveUnsignedInput) -> avoid temp Perspectives for signals

lucksus:
This adds a new type `PerspectiveUnsignedInput` which is a `Perspective` but with `Link`s instead of `LinkExpression`s. It is used as replacement for `Perspective` in new functions:
  * `setOnlineStatusU`
  * `sendSignalU`
  * `sendBroadcastU`

These functions allow sending of signals without prior creation of a temp Perspective for signed Links. Instead, these functions sign all links and the whole perspective just before sending in one round-trip.


jdeepee (dismissed):
Awesome! Just needs a changelog

PR #303, opened 2023-03-06
Login proxy server when unlock or generate agent.

### LINKED ISSUE #285: Proxy is not always stable
Seems like I sometimes struggle to have the Proxy started

kaichaosun:
Fix https://github.com/perspect3vism/ad4m/issues/285


lucksus (changes requested):
Great! Just need a changelog entry.

lucksus (commented):
Ok, I've added a changelog entry, but now I'm wondering if we should actually have it this way. It will create more load on our proxy even for users that are not using this feature. Would be better to remember between startups if the proxy was activated and only restore the previous state.

kaichaosun:
> if we should actually have it this way. It will create more load on our proxy even for users that are not using this feature.

The login process when startup ad4m will not setup the proxy connection, it only login agents did in cloudflare kv db, which be used for credential check when set up the proxy connection later.

Have every user to login is not ideal, we can add the advanced setting when initialize the ad4m, an item in this advanced setting can be `do you want to use proxy (mobile maybe better?)`

lucksus (approved):

PR #308, opened 2023-03-08
Support multiple agents ad4m-test

jdeepee:


test-runner/src/cli.ts
```diff
@@ -115,26 +113,13 @@ export function startServer(relativePath: string, bundle: string, meta: string,
...
 
-    await findAndKillProcess('holochain')
+    //await findAndKillProcess('holochain')
     const seedFile = path.join(__dirname, '../bootstrapSeed.json')
     const agentSeedFile = path.join(__dirname, `../${relativePath}-bootstrapSeed.json`);
 
 
     const tempSeedFile = JSON.parse(fs.readFileSync(seedFile).toString())
 
-    if (!fs.pathExistsSync(`${tempSeedFile.languageLanguageSettings.storagePath}-${relativePath}`)) {
```
fayeed: I had this because the storagePath in the language bootstrap would determine where the language would install and overtime it would pile up if using the same path, also if I remember correctly there was a issue around trusted agent because the language installed by one agent would have same hash so other agent would throw an error.

PR #325, opened 2023-03-16
Perspective sync state subscriptions

jdeepee:


host/package.json
```diff
@@ -29,7 +30,7 @@
       "temp/binary/holochain.exe",
       "mainnet_seed.json",
       "package.json",
-      "oldestSupportedVersion.txt",
+      "oldestSupportedVersion.json",
```
lucksus:
```suggestion
```

lucksus: Fayeed fixed the startup problem this way, and I fixed (simultaneously) by changing this file from json to js and including it in the build process so we don't need vercel's snapshot filesystem at all...

lucksus (dismissed):

PR #347, opened 2023-03-29
version localstorage

bnhovde:
Attempt at versioning localstorage tokens to prevent possible states where a user is stuck with invalid token.


lucksus (dismissed):

ui/src/util.ts
```diff
@@ -1,90 +1,125 @@
...
+      const msg = successful ? "successful" : "unsuccessful";
+      console.log("Copying text command was " + msg);
+    } catch (err) {
+      console.error("Unable to copy", err);
+    }
+  } else {
+    navigator.clipboard.writeText(text);
+  }
+}
+
```
lucksus, line 97: Looks like all changes above are just lints, correct?

And the added functions below are redundant in this PR. Same code was added above in connect. Let's move these functions to core and use them in both connect and ui.

lucksus (changes requested):

lucksus (approved):

lucksus (approved):

PR #354, opened 2023-04-03
Apps changed

fayeed:
Adds a new listener that gets triggered when an app is added/ removed, useful for the launcher to update the apps list.


jdeepee:
@fayeed please add a changelog and then you can merge 👍

PR #356, opened 2023-04-03
SDNA Active record implementation

fayeed:


core/src/subject/SubjectEntity.ts
```diff
@@ -0,0 +1,250 @@
...
+
+export class SubjectEntity {
+  #baseExpression: string;
+  #subjectClass: string;
+  #perspective: PerspectiveProxy
+  author: string;
+  timestamp: string;
+
+  constructor(perspective: PerspectiveProxy, baseExpression?: string) {
+    this.#baseExpression = baseExpression ? baseExpression : `entry://${makeRandomPrologAtom(24)}`;
```
lucksus:
```suggestion
    this.#baseExpression = baseExpression ? baseExpression : Literal.from(makeRandomPrologAtom(24)).toUrl();
```

core/src/subject/SubjectEntity.ts
```diff
@@ -0,0 +1,250 @@
...
+  private async getData(id?: string) {
+    const tempId = id ?? this.#baseExpression;
+    let isInstance = await this.#perspective.isSubjectInstance(tempId, this.#subjectClass)
+    if (!isInstance) {
+      throw `Not a valid subject instance of ${this.#subjectClass} for ${tempId}`
+    }
+
+    const subject = await this.#perspective.getSubjectProxy(tempId, this.#subjectClass);
+
+    if (!subject) {
```
lucksus: Don't really understand what this is for.. ?
`subject` isn't used below. if the base is not an instance of the class (which createSubject() would fix) we already jumped out of this function with the throw above...

core/src/subject/SubjectEntity.ts
```diff
@@ -0,0 +1,250 @@
...
+      await this.#perspective.createSubject(this.#subjectClass, tempId);
+    }
```
lucksus: I think these lines can go
```suggestion
```

core/src/subject/SubjectEntity.ts
```diff
@@ -0,0 +1,250 @@
...
+      const resolveExpressionURI = await this.#perspective.infer(`subject_class("${this.#subjectClass}", C), property_resolve(C, "${p}")`)
+      const getProperty = async () => {
+        let results = await this.#perspective.infer(`subject_class("${this.#subjectClass}", C), property_getter(C, "${tempId}", "${p}", Value)`)
+        if (results && results.length > 0) {
+          let expressionURI = results[0].Value
+
+          if (resolveExpressionURI) {
+            const expression = await this.#perspective.getExpression(expressionURI)
+            this['author'] = expression.author;
+            this['timestamp'] = expression.timestamp;
```
lucksus: That doesn't seem right. The expression we're setting in a property could be created by somebody else an definitely at a different timestamp than the base..

core/src/subject/SubjectEntity.ts
```diff
@@ -0,0 +1,250 @@
...
+
+      this[c] = await getProperty()
+    }
+
+    this.#baseExpression = tempId;
+
+    return this
+  }
+
+  private async setProperties(key: string, value: any) {
```
lucksus: This only sets one property, right?
```suggestion
  private async setProperty(key: string, value: any) {
```

core/src/subject/SubjectEntity.ts
```diff
@@ -0,0 +1,250 @@
...
+    const setters = await this.#perspective.infer(`subject_class("${this.#subjectClass}", C), property_setter(C, Property, Setter)`)
+    if (setters.length > 0) {
+      const setter = setters.find(e => e.Property === key)
+
+      if (setter) {
```
lucksus: Why not let Prolog do the matching and only look at the given property?
```js
await this.#perspective.infer(`subject_class("${this.#subjectClass}", C), property_setter(C, ${key}, Setter)`)
```

core/src/subject/SubjectEntity.ts
```diff
@@ -0,0 +1,250 @@
...
+          await this.#perspective.executeAction(actions, this.#baseExpression, [{ name: "value", value }])
+        }
+      }
+    }
+  }
+
+  async save() {
+    this.#subjectClass = await this.#perspective.stringOrTemplateObjectToSubjectClass(this)
+    
+    const subject = await this.#perspective.createSubject(this, this.#baseExpression);
```
lucksus:
```suggestion
```

core/src/subject/SubjectEntity.ts
```diff
@@ -0,0 +1,250 @@
...
+            break;
+          default:
+            await this.setCollectionSetter(key, value.value)
+            break;
+        }
+      } else {
+        await this.setCollectionSetter(key, value)
+      }
+    }
+  }
```
lucksus: update() and save() are the same, right?

tests/js/integration.test.ts
```diff
@@ -584,6 +586,73 @@ describe("Integration", () => {
...
+                    let root = Literal.from("Active record implementation test").toUrl()
+                    const recipe = new Recipe(perspective!, root)
+
+                    recipe.name = "recipe://test1";
+
+                    const recipes = await recipe.find();
+
+                    expect(recipes.length).to.equal(1)
+                })
+            })
```
lucksus, line 708: Needs tests for collections. Would suggest to clone the collection tests from the subject proxy above.

lucksus (changes requested):

core/src/subject/SubjectEntity.ts
fayeed: The find all doesn't work if I don't create a subject instance for it.

core/src/subject/SubjectEntity.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  async get() {
+    this.#subjectClass = await this.#perspective.stringOrTemplateObjectToSubjectClass(this)
+    
+    return await this.getData()
+  }
+
+  // TODO: implement simple quering like limit, skip etc.
+  async find() {
```
lucksus: Find it weird that this is an instance method, and not a static class method. I understand that we need a PerspectiveProxy to do anything, but we can provide that as argument. Clearer interface than having to create an instance which is not used as an instance of the Subject Class..
```suggestion
  static async all(perspective: PerspectiveProxy) {
```

core/src/subject/SubjectEntity.ts
```diff
@@ -0,0 +1,213 @@
...
+    this.#subjectClass = await this.#perspective.stringOrTemplateObjectToSubjectClass(this)
+    const proxies = await this.#perspective.getAllSubjectProxies(this.#subjectClass)
```
lucksus:
```suggestion
    let subjectClass = await perspective.stringOrTemplateObjectToSubjectClass(this)
    const proxies = await perspective.getAllSubjectProxies(subjectClass)
```

tests/js/integration.test.ts
```diff
@@ -584,6 +586,129 @@ describe("Integration", () => {
...
+
+                    expect(recipe2.name).to.equal("recipe://test1")
+                })
+
+                it("find()", async () => {
+                    let root = Literal.from("Active record implementation test").toUrl()
+                    const recipe = new Recipe(perspective!, root)
+
+                    const recipes = await recipe.find();
+
```
lucksus: With above changes, this becomes much clearer...
```suggestion
                    const recipes = new Recipe.all(perspective!)

```

lucksus (changes requested):
Would like `find()` to be changed to `static all()` as described below... Rest is great!

lucksus (approved):
👍

PR #368, opened 2023-05-01
Fixed Perspective handle's state was not updated when the instance state got updated which caused the issue

fayeed:


lucksus (approved):
@jdeepee please take a look as well

PR #370, opened 2023-05-15
Local SDNA links

fayeed:


core/src/subject/SDNADecorators.ts
```diff
@@ -217,6 +219,7 @@ export function SDNAClass(opts: SDNAClassOptions) {
                             source: "this",
                             predicate: through,
                             target: "value",
+                            local
```
lucksus: We shouldn't mention local here at all if it's false

tests/js/subject.pl
```diff
@@ -6,37 +6,37 @@
 
 property(c, "state").
 property_getter(c, Base, "state", Value) :- triple(Base, "todo://state", Value).
-property_setter(c, "state", '[{action: "setSingleTarget", source: "this", predicate: "todo://state", target: "value"}]').
+property_setter(c, "state", '[{action: "setSingleTarget", source: "this", predicate: "todo://state", target: "value", local: false}]').
```
lucksus: These `local: false` mentioned shouldn't be needed. **Social** DNA is about shared data anyways. Adding local properties to our SDNA tooling is for dev convenience and makes sense, but let's have shared stay the default in the normative SDNA definition (i.e.: here).

lucksus (changes requested):

lucksus (approved):

PR #379, opened 2023-05-30
Add update prompt in nav and settings

bnhovde:
Added update checking to the launcher. 

- Compares local package.json version number to npm/latest
- Shows notification dot in top right
- Shows button on settings page (button says "latest version installed" when no update needed) 

Example of update detected:

Screen Recording 2023-05-30 at 14 27 03

### DIFF
--- a/ui/src/components/Settings.tsx
+++ b/ui/src/components/Settings.tsx
@@ -1,5 +1,6 @@
 import { Agent, Literal } from "@perspect3vism/ad4m";
 import { useContext, useEffect, useState } from "react";
+import { version } from "../../package.json";
 import {
   PREDICATE_FIRSTNAME,
   PREDICATE_LASTNAME,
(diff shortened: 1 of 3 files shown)


ui/src/components/Settings.tsx
```diff
@@ -195,6 +198,28 @@ const Profile = (props: Props) => {
     }
   };
 
+  const checkLatestVersion = async () => {
+    try {
+      const response = await fetch(
+        `https://registry.npmjs.org/@perspect3vism/ad4m/latest`
```
lucksus, line 204: I think we should watch the github releases (https://github.com/perspect3vism/ad4m/releases) for new version instead of npm

lucksus (changes requested):

ui/src/components/Settings.tsx
jdeepee: +1

PR #381, opened 2023-06-06
Decentralized language for language & neighborhood language

fayeed:


bootstrap-languages/language-language/hc-dna/zomes/language_storage/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    Ok(file_chunk_hash)
+}
+
+#[hdk_extern]
+pub fn get_language_expression(
+    file_expression_hash: LanguageAddress,
+) -> ExternResult<Option<LanguageExpression>> {
+    let expression_links = get_latest_link(
```
lucksus: With this and without validation rules it is possible to publish the same language (same code file) multiple times from multiple agents which causes problems with trusted agents (we had that problem some months ago with the centralised language language).

An easy fix could be to not take the latest link but that oldest link here.

A better solution would be to make it impossible to add a second link off of a LanguageAddress anchor.

bootstrap-languages/language-language/hc-dna/zomes/language_storage/src/utils.rs
```diff
@@ -0,0 +1,26 @@
...
+pub(crate) fn get_latest_link(base: EntryHash, tag: Option<LinkTag>) -> ExternResult<Option<Link>> {
+    let language_info = get_links(base, LinkTypes::LanguageLink, tag)?;
+
+    // Find the latest
+    let latest_info =
+        language_info
+            .into_iter()
+            .fold(None, |latest: Option<Link>, link| match latest {
+                Some(latest) => {
+                    if link.timestamp > latest.timestamp {
```
lucksus: Yeah, might be easy to change this to `get_oldest_link` like this.
```suggestion
                    if link.timestamp < latest.timestamp {
```

lucksus: And we should think about enabling multiple LanguageMetas (just keeping the first as the normative one) so that multiple agents actually can sign a language and the language controller iterates through all signatures/metas from these different agents if it can't establish a trusted signature.

bootstrap-languages/language-language/putAdapter.ts
```diff
@@ -1,60 +1,79 @@
...
+        const data_compressed = pako.deflate(data_uncompressed)
+        const blob = new Blob([data_compressed])
+
+        const hashes = await storage.upload(blob);
+
+        const fileMetadata = {
+            name: language.meta.name,
+            description: language.meta.description,
+            address: language.meta.address,
+            checksum: "1234",
```
lucksus: We should get rid of `checksum` if we're not using it really.

lucksus (changes requested):
Looks very good! Just the mentioned thing about overwriting languages. That caused quite some pain with trusted agent changes. Better to avoid that by just changing `get_latest_link` to `get_oldest_link` like mentioned below.

lucksus (dismissed):

PR #401, opened 2023-07-27
Rust integration with launcher

fayeed:
- Adds rust-executor to the launcher.
- Updated rust-executor to work with launcher using tokio.
- Updated logging for JSCore & rust.


lucksus (dismissed):
This looks amazing! But the error (about not being able to create publishing agent) should be fixed. Or is this failing in the rust refactor branch as well?

jdeepee (changes requested):
Looks great, the only missing change I can see is updating the `clear_state` function inside `ui/src/commands/app.rs`

PR #402, opened 2023-08-02
SDNA fixes

fayeed:
Fixes bugs in SDNA from this card: https://github.com/orgs/coasys/projects/3/views/1?pane=issue&itemId=33212832

### DIFF
--- a/core/src/subject/SDNADecorators.ts
+++ b/core/src/subject/SDNADecorators.ts
@@ -235,12 +235,12 @@ export function SDNAClass(opts: SDNAClassOptions) {
                         target: initial,
                     })
 
-                    classRemoverActions = [{
+                    classRemoverActions.push({
                         action: "removeLink",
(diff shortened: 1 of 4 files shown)


core/src/subject/SDNADecorators.ts
```diff
@@ -235,12 +235,12 @@ export function SDNAClass(opts: SDNAClassOptions) {
                         target: initial,
                     })
 
-                    classRemoverActions = [{
+                    classRemoverActions.push({
```
lucksus, line 238: Ah, does this fix the deletion of subject instances?

lucksus (dismissed):

core/src/subject/SDNADecorators.ts
fayeed: yes

lucksus:
Please also add a section to the changelog

PR #410, opened 2023-10-05
Ad4m launcher now works on windows

fayeed:
- Updated all the node std modules to use `node:` specifier to work in deno.
- Removed rlimit from windows build.
- Patched all the packages that used node std lib.
- Updated Deno to latest


jdeepee (commented):
🔥 Lots of work here thanks @fayeed ! 

I have a few questions; I believe I tried before updating to the latest deno version but found that imports with the node prefix like: 'node:os' were not able to be imported due to the module loaders we use not being compatible. Was this somehow fixed in the deno versions, was the loader added by default?

fayeed:
Yeah, it's using the latest deno which was released a week ago, also had to set global variables manually since we dont have `global.ts` which seems to solve the problem.

edit: Also had problem with dynamic imports so even if we used `node:` specifiers so changed it to static import.

PR #411, opened 2023-10-10
New link languages

jdeepee:
- Adds link language which uses socket.io for signaling and holochain as data storage
- Adds link languages which uses socket.io for signaling and centralized cloudflare store for data storage (TODO)
- Fixes bug where JS event loop would stop polling and cause link language to starve


jdeepee:
Closing in favour of: https://github.com/coasys/ad4m/pull/413

PR #417, opened 2023-10-31
SDNA improvements

fayeed:
- Added `query` method to `SubjectEntity` for paginate to work.
- Added override option to `ensureSDNASubjectClass`.

### DIFF
--- a/core/src/subject/SubjectEntity.ts
+++ b/core/src/subject/SubjectEntity.ts
@@ -1,4 +1,5 @@
 import { Literal } from "../Literal";
+import { Link } from "../links/Links";
 import { PerspectiveProxy } from "../perspectives/PerspectiveProxy";
 import { makeRandomPrologAtom } from "./SDNADecorators";
 import { singularToPlural } from "./util";
@@ -11,13 +12,15 @@ export type QueryPartialEntity<T> = {
 export class SubjectEntity {
   #baseExpression: string;
   #subjectClass: string;
+  #source: string;
   #perspective: PerspectiveProxy
   author: string;
   timestamp: string;
 
-  constructor(perspective: PerspectiveProxy, baseExpression?: string) {
+  constructor(perspective: PerspectiveProxy, baseExpression?: string, source?: string) {
     this.#baseExpression = baseExpression ? baseExpression : Literal.from(makeRandomPrologAtom(24)).toUrl();
-    this.#perspective = perspective
+    this.#perspective = perspective;
+    this.#source = source || "ad4m://self";
   }
 
   get baseExpression() {
@@ -95,7 +98,7 @@ export class SubjectEntity {
       if (resolveLanguageResults && resolveLanguageResults.length > 0) {
         resolveLanguage = resolveLanguageResults[0].Language
       }
-      
+
       if (resolveLanguage) {
         value = await this.#perspective.createExpression(value, resolveLanguage)
       }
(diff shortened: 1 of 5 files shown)


jdeepee (dismissed):
Looks good to me! Would also tag in @leifriksheim for a review here since he is close to the SubjectEntity definitions we are using in Flux and making sure we have parity with that

core/src/subject/SubjectEntity.ts
```diff
@@ -157,6 +160,14 @@ export class SubjectEntity {
 
     await this.#perspective.createSubject(this, this.#baseExpression);
 
+    await this.#perspective.add(
+      new Link({
+        source: this.#source,
+        predicate: "rdf://has_child",
```
leifriksheim: What if this is part of the constructor as well @fayeed ? If you don't pass any predicate, this is the default, but I think it is helpful to be able to control this.

leifriksheim: I also wonder if this save() is the right function name, because this creates a new class instance. 
I notice for example here:

https://typeorm.io/#updating-in-the-database

That you still use save() to update the instance, but we have the update() function. 

I think there are two options:

- Rename save() to create()
- save() could be the same as the update() function, just that it checks if there is a class instance of the baseExpression and if not creates a new entry, if it already exist it just updates the values.

core/src/subject/SubjectEntity.ts
```diff
@@ -221,10 +232,62 @@ export class SubjectEntity {
 
     return []
   }
+
+  static async query(perspective: PerspectiveProxy, query?: SubjectEntityQueryParam) {
```
leifriksheim, line 236: I think it would make sense to actually create a separate SubjectRepo class, or something similar that provides more of the query, findAll, find etc methods. It doesn't quite feel right to have a query method on a SubjectEntity, as it refers more to *one* Entity. What do you think @lucksus ?

leifriksheim: Still, it is very nice to not worry about more different types of classes. So I don't have a very strong opinion, and I would let this be up to you backend folks!

leifriksheim (changes requested):

core/src/subject/SubjectEntity.ts
fayeed: Yeah I could do that. Also add a option to change the predicate.

lucksus: Hm, I think `save()` is the right name for this and I'm fine it being both create and update. (it means: making sure that after this call the graph stores the local object).

I have a problem with this `rdf://has_child` predicate. Does that really exist, or are we making something up and calling it RDF?. 

But then, why do have this `source` link here in SubjectEntity at all?

core/src/subject/SubjectEntity.ts
lucksus: This is a `static` function, so it's a class method, not an object method. See the test below. So I think that is fine.

I think the `source` parameter makes sense here in the query.

Only thing I don't feel good about is enforcing the creation of these source links. And if we do, I think the predicate should at least be `ad4m://something`.

Talked about this with Harlan last week and mentioned a predicate for arbitrary "associations" between objects...

lucksus (changes requested):
Minimum change: rename the predicate `rdf://has_child` to either `ad4m://has_child` or `ad4m://association`.

I would really like to take the time to design these things properly as we are laying the ground work for interoperability. So I'm ambivalent between saying "remove source completely for now, and have apps use collections to explicitly say how they associate subject class instances" and "ok merge it but let's have a design session around these patterns next week".

I do like the static query function and having `source` in the param makes sense. I'm only talking about the forced link creation in `save()`.

core/src/subject/SubjectEntity.ts
fayeed: I will remove the link creatiion from here and we could do this in flux/ override the save method to this

lucksus (approved):

leifriksheim (approved):

PR #422, opened 2023-11-22
Simple SDNA governance

fayeed:


core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -320,36 +320,49 @@ export class PerspectiveProxy {
...
-     * This will replace all previous SDNA code elements with the new one.
-     */
-    async setSdna(sdnaCode: string) {
-        await this.setSingleTarget(new Link({
-            source: "ad4m://self",
-            predicate: "ad4m://has_zome",
-            target: Literal.from(sdnaCode).toUrl()
-        }), 'shared')
-    }
-
```
lucksus, line 333: Seeing this removed make me wonder if it doesn't break the Rust client and the CLI since it's used there. But that has it's own Rust implementation of some parts of PerspectiveProxy. So this change doesn't break that, but I think we should have according changes there as well.

core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -580,32 +593,15 @@ export class PerspectiveProxy {
...
-            })
+        if(subjectClass.length > 0) {
+            return
+        }
 
-            await this.remove(link);
+        const { name, sdna } = jsClass.generateSDNA();
 
-            await this.addSdna(jsClass.generateSDNA())
-        }
```
lucksus, line 608: Hm, I think we need this code still. It was an approach for dev-mode where you want to make sure SDNA always gets overwritten. It needs to change but should not be removed.

core/src/neighbourhood/Neighbourhood.ts
```diff
@@ -12,6 +12,9 @@ export class Neighbourhood {
     @Field()
     meta: Perspective
 
+    @Field()
+    author: string;
+
```
lucksus: This is redundant since the Neighbourhood exists as expression in the Neighbourhood Language. But I think I understand why you thought it's needed: Perspective.ts only inlcudes a `Neighbourhood` property and not a `NeighbourhoodExpression`. Two ways to solve this without introducing this redundancy:
1. Change `Perspective.neighbourhood` to be a `NeighbourhoodExpression` including the author (and signature!)
2. We have `Perspective.sharedUrl` which is the URL of the `NeighbourhoodExpression`. We could get it from there. Actually `Perspective.neighbourhood` is just a cached version of that without the expression fields. 

So I think better would be option 1, which means changing the initialisation when publishing and joining Neighbourhoods in `Ad4mCore.neighbourhoodPublishFromPerspective()` and `Ad4mCore.installNeighbourhood()`

lucksus (changes requested):

lucksus (approved):

PR #423, opened 2023-11-28
Added agent expression language

fayeed:
Centralized agent expression endpoints here https://github.com/coasys/centralized-socket-link-language/pull/1


jdeepee:
@fayeed both PR's look good, I can make the mainnet seed for these and get it into a release. Have you tested this language locally?

PR #424, opened 2023-11-28
Update scryer & deno to latest

fayeed:


fayeed:
Prolog tests don't seem to pass, there are multiple issues one is if it's not able to find results it will not return an empty array.

```prolog
subject_class("${this.#subjectClass}", C), property_getter(C, "${this.#baseExpression}", "${p}", Value)
```
 results in (can be solved in js code to ignore this result)

```json
[{"C":" c", "Value":" Value"}]
```

another issue is that when fetching a property for a subject class results in that fact to disappear from the engine, don't know whats happening here.

PR #440, opened 2024-01-23
Launcher fixes

fayeed:
- Updated tauri to the latest
- Fixed system tray message popup not working on windows production.
- Fixed JScore crashing on app restart.
- Added back window decorations for splash screen & auth screens.


.github/workflows/integration_cli.yml
```diff
@@ -28,6 +28,11 @@ jobs:
         override: true
         toolchain: 1.73.0
 
+    - name: Use Python 3.11
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.11'
+
```
lucksus: What is this for? I've deactivated the CLI tests (for now) just wonder if we need to add python the the cirlceci config now..?

lucksus (dismissed):

.github/workflows/integration_cli.yml
fayeed: Ahh yeah this could be the issue you are facing as well, need python version 3.11, I thought this was just for CI.

PR #451, opened 2024-02-19
Fix-up HC AgentInfo retrieval and sharing

lucksus:
This is a start to get properly serialized and base64 encoded AgentInfos so we can grab them through the CLI and share via chat messages in case a node is not connected. For that, we need to be able to import them from the shared string again.

I've based this PR on-top of https://github.com/coasys/ad4m/pull/448 so we can do this all in Rust and avoid serializing Buffer instances as string maps.

### DIFF
--- a/cli/src/runtime.rs
+++ b/cli/src/runtime.rs
@@ -32,7 +32,7 @@ pub enum RuntimeFunctions {
     Friends,
     HcAgentInfos,
     HcAddAgentInfos {
-        infos: String,
+        infos_file: Option<String>,
     },
     VerifySignature {
         did: String,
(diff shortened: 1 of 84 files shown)


cli/src/runtime.rs
```diff
@@ -118,7 +118,21 @@ pub async fn run(ad4m_client: Ad4mClient, command: RuntimeFunctions) -> Result<(
         }
         RuntimeFunctions::HcAgentInfos => {
             let infos = ad4m_client.runtime.hc_agent_infos().await?;
-            println!("{}", infos);
+            println!("RAW\n{}\n\n", infos);
+            let encoded_agent_infos: Vec<String> = serde_json::from_str(&infos)?;
```
lucksus, line 123: This decoding is just me testing the serialization. I think we can just print out the string here (or maybe deserialize it just to print a pretty list with agent addresses), but we need this code in the resolver for `HcAddAgentInfos`.

PR #457, opened 2024-02-26
SDNA improvements

fayeed:
- Fixed value returns as undefined if the property was boolean and set to false.
- Change Decorator names to `PascalCase`
- Add validation for decorators.
- Fixed links in docs.
- Improved Docs for sdna & subject classes.
- Added typedoc for decorators.

### DIFF
--- a/core/src/subject/SDNADecorators.ts
+++ b/core/src/subject/SDNADecorators.ts
@@ -23,10 +23,30 @@ export function hasLink(predicate: string): string {
 }
 
 interface InstanceQueryParams {
+    // An object representing the WHERE clause of the query
     where?: object;
+
+    // A string representing the condition clause of the query
     condition?: string;
 }
 
+/**
+ * Decorator for querying instances of a subject class.
+ * 
+ * @description
+ * NOTE: Only works on methods that return a promise and will throw an error if not used on a method that returns a promise.
+ * This will allow you to query for instances of a subject class with custom clauses within the instance clauses.
+ * 
+ * @example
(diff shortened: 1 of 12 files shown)


docs/pages/social-dna.mdx
```diff
@@ -17,6 +17,12 @@ based on the structure of the graph data. Using [Prolog](https://en.wikipedia.or
 turing complete logical programming language, any kind of alorithm can be run based on the
 state of the graph.
 
+<Callout type="info" emoji="ℹ️">
+  Don't worry if this looks very unfamiliar with [Prolog](https://en.wikipedia.org/wiki/Prolog). You will normally either
```
lucksus:
```suggestion
  Don't worry if you are unfamiliar with [Prolog](https://en.wikipedia.org/wiki/Prolog). You will normally either
```

core/src/subject/SDNADecorators.ts
```diff
@@ -71,17 +91,60 @@ export function instanceQuery(options?: InstanceQueryParams) {
 
 
 interface PropertyOptions {
```
lucksus: This gets ignored by typedoc because the whole interface doesn't have a doc-comment, and it's not exported.

lucksus (commented):
Most of this is adding docs, but am I right that all code changes in here are the checks in the decorator functions that throw if for instance a required property doesn't have an initial value?

Could you add a short list of changes/improvements to the PR description?

lucksus (approved):

PR #459, opened 2024-02-26
Launcher handling multiple different agents/config files (esp. for dev and main)

fayeed:


lucksus (changes requested):
Can we change the UX for this? Clicking the toggle in settings make ADAM restart, which I think I would not expect from activating "dev mode".

I think it should rather be a selection box in the splash / login. And also I would not conceptualize it as "dev" and "not-dev" but rather enable multiple different named instances.

What this should entail in the future (not necessarily now) is picking a bootstrap seed.

I think we should think about this and make it look like picking the network in MetaMask.

..hm, and maybe we should actually introduce bootstrap-seed selection with this already and provide a dev bootsrap-seed so dev agents don't spam the public network.. I think there is a chance with this to communicat/introduce the whole concept of the ADAM bootstrap language network in good way..

lucksus (approved):

PR #464, opened 2024-02-28
Split ad4m-cli into two binaries

fayeed:


fayeed:
Need to update the redme for CLI will do it now

lucksus (approved):

PR #466, opened 2024-03-01
Runtime service rust implementation

fayeed:


executor/src/core/graphQL-interface/GraphQL.ts
```diff
@@ -338,11 +318,6 @@ export function createResolvers(core: Ad4mCore, config: OuterConfig) {
                 return await dmLang.directMessageAdapter!.inbox(filter)
             },
             //@ts-ignore
-            runtimeMessageOutbox: (args, context) => {
-                const { filter } = args
-                return core.runtimeService.getMessagesOutbox(filter)
-            },
-            //@ts-ignore
             runtimeInfo: () => {
```
lucksus: runtimeInfo() should be easy to port too, no?

rust-executor/src/runtime_service/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+
+pub fn add_trusted_agent(new_agents: Vec<String>) {
+    Ad4mDb::global_instance()
+        .lock()
+        .expect("Couldn't get write lock on Ad4mDb")
+        .as_ref()
+        .expect("Ad4mDb not initialized")
+        .add_trusted_agents(new_agents)
+        .map_err(|e| e.to_string())
+        .unwrap_or(());
```
lucksus: Could save a lot of redundant lines in this file using `Ad4mDb::with_global_instance()` like so:
```suggestion
Ad4mDb::with_global_instance(|db| 
   db.add_trusted_agents(&new_agents)
).map_err(|e| e.to_string()))?;
```

lucksus (dismissed):

PR #469, opened 2024-03-07
Agent & entanglement rust implementation

fayeed:


executor/src/core/agent/AgentService.ts
```diff
@@ -155,30 +113,18 @@ export default class AgentService {
...
-    this.#signingKeyId = key.id;
-  }
-
+  // TODO: need to be removed once runtime stuff gets merged
   isInitialized() {
-    return fs.existsSync(this.#file);
+    return AGENT.isInitialized();
   }
 
+  // TODO: need to be removed once runtime stuff gets merged
```
lucksus, line 121: Runtime PR is merged now. Is `isUnlocked` not used anymore?

rust-executor/src/agent/mod.rs
```diff
@@ -86,6 +100,205 @@ impl Into<crate::graphql::graphql_types::AgentSignature> for AgentSignature {
...
+    pub agent: Option<Agent>,
+}
+
+lazy_static! {
+    static ref AGENT_SERVICE: Arc<Mutex<Option<AgentService>>> = Arc::new(Mutex::new(None));
+}
+
+impl AgentService {
+    pub fn new() -> AgentService {
+        let agent_path = format!("{}/ad4m/agent.json", std::env::var("APPS_DATA_PATH").unwrap());
```
lucksus: Hm, I saw that environment variable in one of the other PRs and thought to comment on it but didn't. I also don't really want to block merging this, but I'm a bit concerned about doing this way. If we want to read this config setting from the environment, I think it should happen once in the initialization. But I saw you're setting it there so it can be read here.

Is there reason why we can't pass this config variable as a variable? `instance()` doesn't have to call `new()`. The value inside the `Arc<Mutex<>>` is already an `Option`, so instance could just return `None` and then we could have an initializtion function which gets the path passed in as variable...

This env variable just looks like its asking for trouble..

rust-executor/src/entanglement_service/mod.rs
```diff
@@ -0,0 +1,93 @@
+use std::{collections::HashSet, env, fs::File, io::{BufReader, Write}, path::{self, Path}, sync::Mutex};
+
+use crate::{agent::{sign, sign_string_hex, AgentService}, graphql::graphql_types::EntanglementProof};
+
+pub(crate) mod entanglement_service_extension;
+
+lazy_static! {
+  static ref ENTANGLEMENT_PROOF_FILE: Mutex<String> = Mutex::new(format!("{}/entanglement_proof.json", env::var("APPS_DATA_PATH").unwrap_or_else(|_| ".".to_string())));
```
lucksus: <insert comment on env var from above>

rust-executor/src/graphql/mutation_resolvers.rs
```diff
@@ -104,34 +107,72 @@ impl Mutation {
...
-        let result = js.execute(script).await?;
-        let result: JsResultType<AgentStatus> = serde_json::from_str(&result)?;
-        result.get_graphql_result()
+        js.execute(script).await?;
+
+        let agent = {
+            let mut agent_service = agent_instance.lock().expect("agent lock");
+            let agent_ref: &mut AgentService = agent_service.as_mut().expect("agent instance");
+            agent_ref.dump().clone()
+        };
```
lucksus: Seeing these blocks with boilerplate mutex locking code, I think `AgentService` (like all our singleton objects) would benefit from having a `with_global_instance()` function like [the one in Ad4mDb](https://github.com/coasys/ad4m/blob/d75127f7a31328a9a8f14863a29e6128f04979d0/rust-executor/src/db.rs#L648) so we both have more readable code without the risk of creating deadlocks by keeping locks for too long..

lucksus (changes requested):
I didn't want to block this PR only because of the env var thing, but if we have more changes then that should be improved as well.

But the main change request I have is, not having a good portion of added lines be the same 
```Rust
let agent_instance = AgentService::instance();
let agent_service = agent_instance.lock().expect("agent lock");
let agent_ref: &AgentService = agent_service.as_ref().expect("agent instance");
```
over and over again. Adding something like [Ad4mDb::with_global_instance()](https://github.com/coasys/ad4m/blob/dev/rust-executor/src/db.rs#L648) to `AgentService` (or all our singletons) will reduce noice and deadlock risk.

rust-executor/src/entanglement_service/mod.rs
fayeed: my bad I think we change this to use the SQL database

rust-executor/src/runtime_service/mod.rs
```diff
@@ -25,101 +26,123 @@ use serde::{Deserialize, Serialize};
...
+        }).map_err(|e| e.to_string());
+    }
+
+    pub fn remove_trusted_agent(&self, agents_to_remove: Vec<String>) {
+        let _ = Ad4mDb::with_global_instance(|db| {
+            db.remove_trusted_agents(agents_to_remove)
+        }).map_err(|e| e.to_string());
+    }
+
+    pub fn get_know_link_languages(&self) -> Vec<String> {
```
lucksus, line 92: typo `get_knowN_..`

lucksus (approved):

PR #474, opened 2024-04-04
Host root

fayeed:
Need to change the url from localhost to the  service


connect/src/web.ts
```diff
@@ -496,6 +547,82 @@ export class Ad4mConnectElement extends LitElement {
     this.loadFont();
   }
 
+  private async loginToHosting() {
+    try {
+    const response = await fetch('http://localhost:3000/login', {
```
lucksus: `localhost` in `loginToHosting()` seems wrong, or something only for local dev mode.

fayeed: yeah this needs to be updated with hosted API url

lucksus (dismissed):

lucksus (approved):

PR #483, opened 2024-05-24
Notification Prolog helper predicates and OS notification push

fayeed:
# Added 
### Prolog predicates needed in new Flux mention notification trigger:
 - agent_did/1
 - remove_html_tags/2
 - string_includes/2
 - literal_from_url/3
 - json_property/3
 
### Triggered notifications are handled through operating system notifications

# Fixed
- Prolog engine gets respawned when a query caused an error to clean the machine state
- Catch panics in Scryer and handle as error instead of killing the engine thread which caused `channel closed` error in future queries

### DIFF
--- a/ui/src/components/Notification.tsx
+++ b/ui/src/components/Notification.tsx
@@ -1,0 +1,77 @@
+import { useContext, useEffect, useState } from "react";
+import { Ad4minContext } from "../context/Ad4minContext";
+
+const Notification = () => {
+  const {
+    state: { client, notification },
+    methods: { handleNotification },
+  } = useContext(Ad4minContext);
+
+  const [requestModalOpened, setRequestModalOpened] = useState(true);
+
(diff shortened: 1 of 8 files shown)


ui/src/components/Notification.tsx
```diff
@@ -0,0 +1,78 @@
...
+                <div>
+                  <j-avatar size="xl" src={notification?.appIconPath}></j-avatar>
+                </div>
+                <div>
+                  <j-text variant="heading-sm">{notification?.appName}</j-text>
+                  <j-text nomargin size="500">
+                    {notification?.description}
+                  </j-text>
+                </div>
+              </j-flex>
```
lucksus, line 69: This should also show the perspective/neighbourhood in which the notification is checked.

And even more important: if there is a webhook URL, making clear that this way data can get leaked outside of ADAM!

PR #495, opened 2024-06-20
Notifications Fix

lucksus:
Flux escapes apostrophes "'" with a backslash "\'" which is not a valid JSON escape sequence. We have handcrafted code that turns Prolog results into JSON. With these "\'" it failed to produce valid JSON. This fixes that.


fayeed (dismissed):

jhweir (approved):
ChatGPT's nitpicking suggests it would be would be more "idiomatic and concise" to use `&wrapped_s` and `&fixed_s` instead of `wrapped_s.as_str()` and `fixed_s.as_str()` on lines 16 and 21 of `rust-executor/src/perspectives/utils.rs`, otherwise looks good XD

PR #501, opened 2024-07-22
Update brand colors & styles for launcher

fayeed:
Screenshot 2024-07-24 at 15 17 11


lucksus (changes requested):
The app icon for the bundle is unchanged.
Also, the ADAM logo was changed in the redesign slightly. It's not just a different color but also less spacing. We need a new export from Figma.

lucksus (approved):

PR #507, opened 2024-07-30
Change connection in ad4m connect on mobile & hosting email connection reset

fayeed:


connect/src/components/Settings.ts
```diff
@@ -79,6 +81,17 @@ export default function Settings({
                 Connect
               </button>
             </div>`}
+            ${isMobile ? html`
```
lucksus: I think we could always show this button, no?

connect/src/core.ts
```diff
@@ -393,6 +397,22 @@ export default class Ad4mConnect {
...
+
+      const response = await fetch('https://hosting.ad4m.dev/api/service/checkStatus', {
+        method: 'GET',
+        headers: {
+            'Content-Type': 'application/json',
+            'Authorization': 'Bearer ' + token
+        },
+      });
+
+      if (response.status === 200) {
```
lucksus: Should this be 
```suggestion
      if (response.status != 200) {

```
?

connect/src/web.ts
```diff
@@ -861,8 +891,19 @@ export class Ad4mConnectElement extends LitElement {
 
   render() {
     console.log(this.authState,  this.connectionState, this.uiState, this._isOpen);
-    if (this._isOpen === false) return null;
-    if (this.authState === "authenticated") return null;
+    if (this._isOpen === false) {
+      if (this.authState === "authenticated" && this._isMobile) {
```
lucksus: Also here, I think we could always show this new logo button. Or: we make it an option. But I think it could help us when dogfooding and debugging also in the web/desktop case..

lucksus (approved):

PR #509, opened 2024-08-01
Connect improvements

fayeed:
- Shows the connect icon even when the connect popup is visible
- Now Websocket errors are handled and doesn't break the connect UI.
- Changed the text for connecting to remote executor.


lucksus (approved):
Good changes. But I'm missing the probing of the executor connection via a client function. Have you tried this with a proxy that lost its proxied executor? Does the websocket.onerror trigger in that case? Anyways good to merge :)

PR #521, opened 2024-08-23
Fix problems with Subject Classes

fayeed:
- Improve isSubjectInstance method with retry logic for better reliability
- Fix addSdna() to not omit adding a new SDNA definition if the same name was added by someone else (which breaks use-case of some other non progenitor agent adding a subject class for them, which we won't run because not signed by a trustet agent, but we don't get out of that situation because we can't add it with our signature since the old implementation doesn't regard the author of the SDNA, but just finds it existing already)


lucksus (commented):
Great! Just need a changelog entry and then merge this!

lucksus (approved):

PR #522, opened 2024-09-05
Local AI model inference

lucksus:
This adds AI model inference to ADAM, including downloading and managing of models from Huggingface, registering tasks and prompting models - based on [Kalosm](https://github.com/floneum/floneum/tree/main/interfaces/kalosm). 

As a first step, this PR only focusses on managing and making available a couple of hard-coded models, but introducing an new section "AI" to the Ad4mClient. Other, future PRs will add the ability to have the user select and manage the exact models (https://github.com/coasys/ad4m/pull/504), as well as registering external modals via API, and UI in the ADAM Launcher for all of this.

# New AI interface functions in Ad4mClient
This introduced 3 different kinds of model interactions that ADAM provides to apps via it's interface

## Language processing with LLMs via tasks
```Javascript
const task = await ad4mClient.ai.addTask(
  //Model name, currently irrelevant with hard-coded model,
  "Llama", 
  //System prompt
  "You analyse incoming text for topics and respond with a JSON array including the topic names", 
  // Examples
  [
    {
      input: "Hey guys, how is it going?",
      output: '["greeting"]'
    },
    {
      input: "I really want to try the new Synergy plugin in Flux",
      output: '["testing", "Synergy", "Flux"]'
    },
  ]
);
```
This spawns a session of the given model and configures it with the prompt and the examples, so that subsequently it can be run with:
```Javascript
const answer = await ad4mClient.ai.prompt(task.taskId, "Hello there, yeah I also would like to test that plugin!")
```

## Emedding of text
```Javascript
const vector = await ad4mClient.ai.embed("Bert", "This is a test string");
```

## Transcription of audio
```Javascript
const streamId = await ad4mClient.ai.openTranscriptionStream("Whisper", (text) => {
  console.log(text);
  });
  
// Feed raw sample data (float 32)
await ad4mClient.ai.feedTranscriptionStream(streamId, [0, 10, 20, 30]);
```

# Progress
- [x] ADAM interface/client definitions
- [x] Task CRUD
- [x] Task execution
- [x] Embedding vectors
- [x] Audio transcription

# Deno/V8/Abseil linking issues fixed

This branch also switches Deno dependencies to a new branch in our forks which use our new fork of rusty_v8. V8, and more precisely its dependency Abseil needs to get compiled within a separate namespace since one of Kalosm's C sub-dependcies (libort) also uses Abseil. Since building C dependencies inside Rust crates is not something that cargo is able to track, this problem only becomes apparent at link-time (Linux) or at run-time (macOS and Windows) with a crash.

Understanding and solving this riddle is what took most of the time to complete this PR, even if the work is barely visible here, only with the switch of the branch of the Deno crates.


rust-executor/src/ai_service/mod.rs
```diff
@@ -0,0 +1,713 @@
...
+        }
+
+        rx.await?;
+        Ok(())
+    }
+
+    pub async fn open_transcription_stream(&self, _model_id: String) -> Result<String> {
+        let stream_id = uuid::Uuid::new_v4().to_string();
+        let stream_id_clone = stream_id.clone();
+        let (samples_tx, sampels_rx) = futures_channel::mpsc::unbounded::<Vec<f32>>();
```
lucksus: typo: sampELs

rust-executor/src/ai_service/mod.rs
```diff
@@ -0,0 +1,713 @@
...
+        Ok(true)
+    }
+
+    pub fn get_tasks() -> Result<Vec<AITask>> {
+        let tasks = Ad4mDb::with_global_instance(|db| db.get_tasks())
+            .map_err(|e| AIServiceError::DatabaseError(e.to_string()))?;
+        Ok(tasks)
+    }
+
+    pub async fn spawn_embedding_model(&self, model_id: String) {
```
lucksus: Don't think this function should be public.

jhweir (approved):
Looks good to me. Great work!

PR #575, opened 2025-03-03
Enhanced SubjectEntity Query System

jhweir:
# Enhanced SubjectEntity Query System
This PR introduces a new, more powerful query system for SubjectEntity that enables complex queries with a fluent interface and real-time subscriptions to query results.

# Key Features
1. Unified Query Building: All instance queries now use a single Prolog query that fetches all properties, collections, and metadata in one go
2. Rich Query Constraints: Support for filtering by:
 - Properties and their values
 - Source
 - Collections
 - Author and timestamp metadata
 
3. Result Shaping: Control over what data is returned via:
 - Property selection
 - Collection selection
 - Ordering
 - Pagination (limit/offset)

4. Real-time Query Subscriptions: Subscribe to query results and receive updates when matching instances change
**Huge for a much more performant implementation of our hooks! - coming next**

5. Fluent Interface: Chain-able query building for better readability

# Example usage
```JavaScript
// Basic query with where clause
const recipes = await Recipe.findAll(perspective, { 
  where: { name: "Pancakes" } 
});

// Complex query with multiple constraints
const recipes = await Recipe.findAll(perspective, {
  where: { 
    name: "Pancakes",
    author: currentUser.did
  },
  properties: ["name", "ingredients"],
  collections: ["comments"],
  limit: 10,
  offset: 0
});

// Using the fluent query builder
const query = Recipe.query(perspective)
  .where({ name: "Pancakes" })
  .properties(["name", "ingredients"])
  .collections(["comments"])
  .limit(10);

// Run once
const results = await query.run();

// Or subscribe to updates
query.subscribeAndRun((recipes) => {
  console.log("Query results updated:", recipes);
});
```

# Implementation Details
The new query system works by:
1. Building a comprehensive Prolog query that includes all necessary constraints
2. Executing the query to get all matching instances and their data in a single round-trip
3. Converting the Prolog results into properly typed SubjectEntity instances
4. For subscriptions, watching for changes and re-running the query automatically

This replaces the previous approach of fetching instances and their data separately, resulting in better performance and enabling more complex queries.


core/src/subject/SubjectEntity.ts
```diff
@@ -8,26 +8,62 @@ export type QueryPartialEntity<T> = {
...
+    }
+
+    return this.classNamesByPerspectiveID[perspectiveID];
+  }
+
+  /**
+   * Constructs a new subject.
+   * @param perspective - The perspective that the subject belongs to.
+   * @param baseExpression - The base expression of the subject.
+   * @param soruce - The source of the subject, the expression this instance is linked too.
```
lucksus:
```suggestion
   * @param source - The source of the subject, the expression this instance is linked too.
```

core/src/subject/SubjectEntity.ts
```diff
@@ -49,75 +85,169 @@ export class SubjectEntity {
     return this.#perspective;
   }
 
-  private async getData(id?: string) {
-    const tempId = id ?? this.#baseExpression;
-    let data = await this.#perspective.getSubjectData(this.#subjectClass, tempId)
-    Object.assign(this, data);
-    this.#baseExpression = tempId;
-    return this
+  private static async assignValuesToInstance(
```
lucksus: Why have this be a static method when it takes instance as parameter? Could be `instance.assignFromPrologResults()` or so...

core/src/subject/SubjectEntity.ts
```diff
@@ -8,26 +8,208 @@ export type QueryPartialEntity<T> = {
...
+  // return `
+  //   findall([PropertyName, PropertyValue, Resolve], (
+  //     % Constrain to specified properties if provided
+  //     ${properties ? `member(PropertyName, [${properties.map((name) => `"${name}"`).join(", ")}]),` : ""}
+
+  //     % Get the property name, URI, and resolve boolean
+  //     property(SubjectClass, PropertyName),
+  //     property_getter(SubjectClass, Base, PropertyName, PropertyURI),
+  //     (property_resolve(SubjectClass, PropertyName) -> Resolve = true ; Resolve = false),
+
```
lucksus: something like:
```prolog
Resolve -> expression_data_url(PropertyValue, PropertyURI) ; PropertyValue = PropertyURI
```
with
```prolog
expression_data_url(ExpressionData, Uri) :- 
  language_name(Uri, "literal") -> (
    literal_from_url(Uri, Literal, Scheme),
    (
      (Scheme = "json" -> (json_property(Literal, "data", ExpressionData); ExpressionData = Literal))
      ;
      ExpressionData = Literal
    )
  )
```

core/src/subject/SubjectEntity.ts
```diff
@@ -8,26 +8,208 @@ export type QueryPartialEntity<T> = {
...
+    if (key === "author") isString = true;
+    if (key === "timestamp") isString = false;
+    // Wrap strings in quotes
+    return isString ? `"${value}"` : value;
+  }
+
+  return (Object.entries(where) as [string, WhereCondition][])
+    .map(([key, value]) => {
+      const isSpecial = ["author", "timestamp"].includes(key);
+      const getter = `property_getter(SubjectClass, Base, "${key}", URI), literal_from_url(URI, V, _)`;
```
lucksus: if the URI is not a literal, I guess literal_from_url won't match the property is omitted

lucksus: should be handled generically with property query code above

core/src/subject/SubjectEntity.ts
```diff
@@ -8,26 +8,208 @@ export type QueryPartialEntity<T> = {
...
+          else return `${getter}, V >= ${formatValue(key, between[0])}, V =< ${formatValue(key, between[1])}`;
+        }
+      }
+
+      // Default to direct equality
+      if (isSpecial) return `${field} = ${formatValue(key, value)}`;
+      else return `${getter}, V = ${formatValue(key, value)}`;
+    })
+    .join(", ");
+}
```
lucksus, line 149: nice!

core/src/subject/SubjectEntity.ts
```diff
@@ -49,75 +231,192 @@ export class SubjectEntity {
...
+    perspective: PerspectiveProxy,
+    instance: SubjectEntity,
+    values: ValueTuple[]
+  ) {
+    // Map properties to object
+    const propsObject = Object.fromEntries(
+      await Promise.all(
+        values.map(async ([name, value, resolve]) => {
+          // Resolve the value if necessary
+          const finalValue = resolve ? (await perspective.getExpression(value)).data : value;
```
lucksus: So if we have the literal handling in Prolog as above, we will need to understand here that we don't have to resolve those values. Maybe we should have an internal `ResolveInternal` prolog variable above and only if it's not a literal language URI set `Resolve = true`...

core/src/subject/SubjectEntity.ts
```diff
@@ -49,75 +231,192 @@ export class SubjectEntity {
...
+    if (query?.where) console.log("whereQuery!!!", buildWhereQuery(query?.where));
+
+    const result = await perspective.infer(fullQuery);
+    if (!result?.[0]?.AllInstances) return [];
+
+    // Map results to instances
+    const requestedAttribtes = [...(query?.properties || []), ...(query?.collections || [])];
+    const allInstances = await Promise.all(
+      result[0].AllInstances.map(async ([Base, Properties, Collections, Timestamp, Author]) => {
+        const instance = new this(perspective, Base) as any;
```
lucksus: While this `Base` *should* really be a base of `this` Subject Class, defensive programming would be to wrap all of this in a `try {} catch()`, log errors and filter out the elements that couldn't be instantiated...

core/src/subject/SubjectEntity.ts
```diff
@@ -49,75 +231,192 @@ export class SubjectEntity {
...
+        const values = [...Properties, ...Collections];
+        if (!query?.properties || query.properties.includes("timestamp")) values.push(["timestamp", Timestamp]);
+        if (!query?.properties || query.properties.includes("author")) values.push(["author", Author]);
+        await SubjectEntity.assignValuesToInstance(perspective, instance, values);
+
+        return instance;
+      })
+    );
+
+    return allInstances;
```
lucksus, line 341: 👏 👏 👏 
I guess next step could be to extract the prolog result parsing into its own function, well basically all 3 steps (query creation, infer, result parsing), so that we can have a version that only runs once, and one that subscribes to the query which we want to use in the hooks.

core/src/subject/SubjectEntity.ts
```diff
@@ -49,75 +231,192 @@ export class SubjectEntity {
...
+      const resolveLanguageResults = await this.#perspective.infer(
+        `subject_class("${this.#subjectClassName}", C), property_resolve_language(C, "${key}", Language)`
+      );
+      let resolveLanguage;
       if (resolveLanguageResults && resolveLanguageResults.length > 0) {
-        resolveLanguage = resolveLanguageResults[0].Language
+        resolveLanguage = resolveLanguageResults[0].Language;
       }
 
       if (resolveLanguage) {
```
lucksus, line 358: Hm, weird, so we do have the code that would create the literal URLs from values when resolve language is set. Should investigate if and/or when that is failing...

tests/js/tests/prolog-and-literals.test.ts
```diff
@@ -570,6 +570,22 @@ describe("Prolog + Literals", () => {
                     })
                     booleanTest: boolean = false
 
+                    // used in where query tests
+                    // todo: allow boolean type and auto convert to URI
+                    @SubjectProperty({
+                        through: "recipe://boolean_test",
+                        writable: true,
```
lucksus:
```suggestion
                        writable: true,
                        resolveLanguage: "literal",
```

tests/js/tests/prolog-and-literals.test.ts
```diff
@@ -570,6 +570,22 @@ describe("Prolog + Literals", () => {
...
+                    })
+                    boolean: string = ""
+
+                    // used in where query tests
+                    // todo: allow number type and auto convert to URI
+                    @SubjectProperty({
+                        through: "recipe://number",
+                        writable: true,
```
lucksus, line 584:
```suggestion
                        writable: true,
                        resolveLanguage: "literal",
```

lucksus (commented):
amazing work so far!

tests/js/tests/prolog-and-literals.test.ts
```diff
@@ -818,6 +802,476 @@ describe("Prolog + Literals", () => {
...
+                    const recipesWithAllAttributes = await Recipe.findAll(perspective!);
+                    expect(recipesWithAllAttributes[0].name).to.equal("recipe://test_name")
+                    expect(recipesWithAllAttributes[0].booleanTest).to.equal(true)
+                    expect(recipesWithAllAttributes[0].author).to.equal(me!.did)
+                    
+                    // Test recipes with name only
+                    const recipesWithNameOnly = await Recipe.findAll(perspective!, { properties: ["name"] });
+                    expect(recipesWithNameOnly[0].name).to.equal("recipe://test_name")
+                    expect(recipesWithNameOnly[0].booleanTest).to.be.undefined
+                    expect(recipesWithNameOnly[0].author).to.be.undefined
```
lucksus: I'm not sure about the utility of constraining it to certain properties, but I could fiture a case where we have big properties (especially if at some point we would have automatic loading/instantiation as default).. but even if we would have that, I think author and timestamp should always be there.

lucksus (dismissed):

lucksus (approved):

PR #602, opened 2025-04-28
Fix port collision crash

lucksus:
Tests revealed crashes on some machines where our default ports (1337, 2000) are used by other services. The JS lib get-port in the remaining JS executor implementation doesn't work (anymore?) and instead throws when the preferred port is used. This does free port scanning in Rust.

## Summary by CodeRabbit

- **New Features**
  - Automatic assignment of available ports for server components if not already configured.

- **Bug Fixes**
  - Fixed crashes caused by default ports being in use by other processes.
  - Improved error handling and logging when no free ports are found.

- **Chores**
  - Updated dependencies to improve port selection reliability.
  - Streamlined port selection logic for better maintainability.


executor/src/main.ts
```diff
@@ -83,14 +83,23 @@ export interface SeedFileSchema {
...
-    if(!gqlPort) gqlPort = 4000
-    // Check to see if PORT 2000 & 1337 are available if not returns a random PORT
-    if(!config.hcPortAdmin) config.hcPortAdmin = await getPort({ port: 2000 });
-    if(!config.hcPortApp) config.hcPortApp = await getPort({ port: 1337 });
-    if(!dAppPort) dAppPort = await getPort({port: 4200})
+
+    // Moved the port check to Rust
+    // BUT: we have weird problem with our JS runtime 
+    // (which will be refactored soon, when we move over the last remaining JS code to Rust)
+    // when this function doesn't actually some async I/O operations, the JS event loop doesn't
```
jhweir:
```suggestion
    // when this function doesn't actually do some async I/O operations, the JS event loop doesn't
```

jhweir (approved):

PR #620, opened 2025-08-28
Special character fix in prolog & ad4m model

jhweir:
# Fix UTF-8 Emoji Support in Literal URL Decoding
### Summary
Fixes issue where UTF-8 emojis (👋, 🎉, etc.) and special characters (ç, ð, etc.) in literal URLs were causing property resolution to fail silently, returning undefined instead of the expected emoji content in the UI.

### Root Cause Analysis
The URL decoding pipeline had multiple cascading issues:

1. **Initial Problem**: Prolog URL decoder failed silently on UTF-8 bytes outside ASCII range (128-255), causing resolve_property to return undefined for emoji messages
2. **Infinite Loop Issue**: During fix attempts, removed cuts caused Prolog backtracking loops, maxing out CPU
3. **Character Corruption**: UTF-8 byte sequences were being treated as individual characters instead of being properly reconstructed into Unicode

### Issues Encountered & Solutions
🚨 Issue 1: Silent Failures (Original Problem)
**Problem**: UTF-8 emojis like %F0%9F%91%8B (👋) caused URL decoding to fail silently
**Symptom**: Properties returned undefined instead of emoji content
**Root Cause**: General URL decoder rule only handled bytes 0-127, rejecting UTF-8 bytes 128-255

🔄 Issue 2: Infinite Loops (During Fix)
**Problem**: Removing constraints caused Prolog backtracking loops
**Symptom**: 100% CPU usage, UI freezing when loading communities
**Root Cause**: Missing cuts (!) allowed infinite backtracking in DCG rules

🔤 Issue 3: Character Corruption (During Fix)
**Problem**: UTF-8 bytes decoded as separate characters
**Symptom**: Emoji displayed as ð instead of 👋 
**Root Cause**: No UTF-8 byte sequence reconstruction on JavaScript side

### Key changes in `sdna.rs`:
- Extended byte range from 0-127 to 0-255 to handle UTF-8 bytes
- Added deterministic cuts (!) to all rules to prevent infinite backtracking
- Preserved performance by keeping all existing ASCII-specific rules
- Maintained compatibility with all existing URL decoding functionality

### Key changes in `Ad4mModel.ts`:
- Added UTF-8 reconstruction to properly decode byte sequences from Prolog
- Safe error handling with fallback to original value
- Selective processing only for unresolved string values from URL decoding

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of emoji, Unicode, and special characters to prevent corruption during save, retrieval, and subscriptions.
  * More reliable percent-encoding decoding to correctly interpret UTF-8 sequences and special characters.
  * Safer handling of byte-like strings: attempts recovery decoding but preserves original content on failure.

* **Tests**
  * Added comprehensive emoji/Unicode tests (complex sequences, HTML entities, mixed content) and subscription coverage.
  * Improved test isolation and teardown to avoid cross-test interference.


lucksus (dismissed):
Great! Can we add an integration test for this too?

lucksus (approved):

PR #670, opened 2026-02-14
fix(test-runner): update ad4m-test for v0.10.1+ compatibility

HexaField:
## Summary

Fixes `@coasys/ad4m-test` (test-runner) for compatibility with the AD4M executor v0.10.1+. Verified working in CI on [HexaField/nextgraph-adam-language](https://github.com/HexaField/nextgraph-adam-language) — a real AD4M Language with both unit and e2e tests green.

### CLI Flag Renames (v0.10.1 uses kebab-case)

| Old (camelCase) | New (kebab-case) |
|---|---|
| `--dataPath` | `--data-path` / `--app-data-path` |
| `--reqCredential` | `--admin-credential` |
| `--port` | `--gql-port` |
| `--networkBootstrapSeed` | `--network-bootstrap-seed` |
| `--languageLanguageOnly` | `--language-language-only` |
| `serve` | `run` |
| `--ipfsPort` | _(removed — IPFS removed from executor)_ |
| `--overrideConfig` | _(removed)_ |

### stderr → stdout Re-emit

The executor logs via Rust's `log` crate to **stderr**, but all server/init detection logic in `cli.ts` and `installSystemLanguages.ts` only checked `stdout`. Added `child.stdout.emit('data', data)` in the stderr handler so detection works for both old and new executor versions.

### Alternative Server Detection

v0.10.1 emits `"listening on http://127.0.0.1:PORT"` instead of `"GraphQL server started, Unlock the agent to start holohchain"`. Added fallback detection for both patterns.

### GitHub API Safety (`utils.ts`)

- **Check binary existence before API call** — avoids unnecessary API calls and rate-limit errors
- **Wrap fetch in try/catch** — previously crashed with `Cannot read properties of undefined (reading 'replace')` when rate-limited
- **Safe version extraction** — `data?.name` with fallback to `'unknown'`
- **Safe asset lookup** — `data.assets?.find(...)` with explicit error when platform binary not found
- **Updated org** — `perspect3vism/ad4m` → `coasys/ad4m`
- **Fixed Linux asset pattern** — `"-linux-"` → `"-linux"` (actual filename: `ad4m-cli-executor-linux-0.10.1-x64`)

### Package Publishing Fix

Added `files` field to `package.json` so `build/`, `scripts/`, `bootstrapSeed.json`, and `public/` are included in the npm tarball. Previously only shipped TypeScript source — consumers got no compiled output.

### Storage Path Guards

- Guard `storagePath` operations with truthiness checks (empty string in fresh `bootstrapSeed.json` caused crashes)
- Use `{ recursive: true }` in `mkdirSync` for robustness
- Simplified copy-and-suffix logic

### Cleanup

- Removed `get-port` dependency (only used for `--ipfsPort` which is removed)
- Removed dead if/else branches with identical bodies
- Removed unused `defaultLangPath` variable in `installSystemLanguages.ts`
- `defaultLangPath` now properly controls `--language-language-only` flag in `cli.ts`

## Testing

CI verified on [HexaField/nextgraph-adam-language](https://github.com/HexaField/nextgraph-adam-language/actions):
- ✅ 35/35 unit tests (vitest, Node 22)
- ✅ Single-agent e2e test (real AD4M executor v0.10.1 on ubuntu-latest)
- Language published, wallet created, session started, expression created + retrieved + verified

## Note on System Language Bundles

The language bundles downloaded by `get-builtin-test-langs.js` are CJS format, but the executor's Deno runtime requires ESM. This is a broader issue not addressed here — consumers need to convert bundles at build time or use `--language-language-only true`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during initialization and language installation
  * Enhanced binary download process with better OS-specific asset detection and error messaging
  * More reliable server startup status detection

* **Chores**
  * Refactored internal execution mechanisms
  * Streamlined dependency requirements for leaner installation

### DIFF
--- a/test-runner/package.json
+++ b/test-runner/package.json
@@ -36,7 +36,6 @@
     "express": "4.18.2",
     "find-process": "^1.4.7",
     "fs-extra": "^10.0.1",
-    "get-port": "5.1.1",
     "glob": "^7.2.0",
     "graphql": "15.7.2",
     "graphql-ws": "5.12.0",
--- a/test-runner/package.json
+++ b/test-runner/package.json
@@ -3,6 +3,12 @@
   "description": "Testing library to test ad4m languages",
   "license": "MIT",
   "bin": "./build/cli.js",
+  "files": [
+    "build/",
+    "scripts/",
+    "bootstrapSeed.json",
+    "public/"
+  ],
   "scripts": {
     "build": "tsc",
     "build:util": "browserify build/client.js --standalone Bundle -p esmify > public/client.js",
--- a/test-runner/src/cli.ts
+++ b/test-runner/src/cli.ts
@@ -6,8 +6,7 @@ import path from 'path';
 import { v4 as uuidv4 } from 'uuid';
 import yargs from 'yargs'
 import { hideBin } from 'yargs/helpers';
-import { ChildProcessWithoutNullStreams, spawn } from 'child_process';
-import { execSync } from 'child_process';
+import { ChildProcessWithoutNullStreams, execFileSync, spawn } from 'child_process';
 import kill from 'tree-kill'
 import { resolve as resolvePath} from 'path'
 import { ad4mDataDirectory, cleanOutput, deleteAllAd4mData, findAndKillProcess, getAd4mHostBinary, getTestFiles, logger } from './utils.js';
@@ -129,7 +128,9 @@ export function startServer(relativePath: string, bundle: string, meta: string, 
       if (!fs.pathExistsSync(suffixedPath)) {
         fs.mkdirSync(suffixedPath, { recursive: true });
       }
-      fs.copySync(langStoragePath, suffixedPath, { overwrite: true });
+      if (fs.pathExistsSync(langStoragePath)) {
+        fs.copySync(langStoragePath, suffixedPath, { overwrite: true });
+      }
       tempSeedFile.languageLanguageSettings.storagePath = suffixedPath;
     }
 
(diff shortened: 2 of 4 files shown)


HexaField:
@coderabbitai review

@lucksus @data-coasys — This PR fixes the `@coasys/ad4m-test` test runner for compatibility with the v0.10.1+ executor. All CLI flags, subcommands, and detection logic are updated. Verified working in CI on [HexaField/nextgraph-adam-language](https://github.com/HexaField/nextgraph-adam-language) (unit + e2e tests green).

Key changes:
- CLI flags from camelCase → kebab-case (`--dataPath` → `--data-path`, etc.)
- `serve` → `run` subcommand
- stderr→stdout re-emit (executor logs via Rust `log` crate to stderr)
- GitHub API: `perspect3vism` → `coasys`, safe error handling
- `files` field in package.json so `build/` is actually included in npm
- Storage path guards for empty bootstrapSeed paths
- Removed dead `get-port` dependency and `--ipfsPort` flag

⬡

HexaField:
The `integration-tests-js` CircleCI failure appears unrelated to this PR — our changes only touch `test-runner/`, and the failing job runs `tests/js/` which has no dependency on the test-runner package. The same test passes on the current `dev` branch tip (`6800bbb`), so this is likely a flaky test or transient environment issue.

All other CI checks pass:
- ✅ build-and-test  
- ✅ integration-tests-multi-user-simple  
- ✅ integration-tests-email-verification  
- ✅ CodeRabbit

Happy to re-trigger if needed. ⬡

lucksus:
This looks great, thanks @HexaField! Only, I think coderabbit has a point with replacing all `v`s. We do have versions with `-dev` which would get broken.

data-bot-coasys:
Hey @HexaField! Just a heads up - I'm now using this account (`data-bot-coasys`) going forward. The old `data-coasys` account got locked out (Google OAuth issues 🙃).

Thanks for the great work on this PR - the test runner fixes are exactly what we needed. I reviewed it earlier and recommended merge. The `-dev` version stripping issue Nico mentioned is a good catch - that `.replace('v', '')` would turn `0.10.1-dev` into `0.10.1-de`.

— Data 🤖

HexaField:
Hey @lucksus — good catch, and yeah coderabbit was right on that one! I actually already fixed it in c18b9f4 — it now uses `replace(/^v/, '')` instead of `replace('v', '')`, so `v0.10.1-dev` correctly becomes `0.10.1-dev` rather than `0.10.1-de`.

That same commit also addressed the other CodeRabbit feedback (execFileSync for safe paths, fs.pathExistsSync guard, duplicate log writes, HTTP status check). Should be good to go! ⬡

Issue #671, opened 2026-02-16
@coasys/ad4m-test: multiple compatibility issues preventing language development


HexaField:
## Summary

`@coasys/ad4m-test` (all published versions 0.8.0–0.11.1) has multiple issues that prevent it from working out of the box for developing and testing AD4M languages. These were discovered while building [nextgraph-adam-language](https://github.com/HexaField/nextgraph-adam-language) and are addressed in PR #670.

### 1. No compiled output shipped (all versions)
The npm package ships TypeScript source only — no `build/` directory. The `bin` field points to `build/cli.js` which does not exist. Users must manually compile with `npx tsc --noImplicitAny false`.

**Fix:** Add `build/` to the `files` field and ensure `prepublishOnly` runs `tsc`.

### 2. CLI flag names don't match executor (since v0.10.1)
The test runner passes `--dataPath`, `--languageLanguageOnly` etc. but the executor expects kebab-case: `--data-path`, `--language-language-only`. The `serve` subcommand was also renamed to `run`.

### 3. GitHub API URL uses wrong org
Binary download URL references `perspect3vism/ad4m` but releases are at `coasys/ad4m`. Also no error handling on the API response (crashes on 403/rate-limit).

### 4. Binary download URL pattern mismatch
Code constructs URL with `-linux-` separator but the actual release asset uses `_linux_` (`ad4m-cli-executor-linux_0.11.1_amd64`). Binary download silently fails.

### 5. Init detection only checks stdout
The executor logs "AD4M init complete" via Rust `log::info!` to **stderr**, but the test runner only watches stdout. This causes a 10-minute hang until timeout.

### 6. System language bundles are CJS
The executor uses a Deno JS runtime that requires ESM modules. All downloaded system language bundles (Language Language, agent-expression-store, etc.) are CJS and crash with `ReferenceError: exports is not defined`.

### 7. Language Language bundle references removed IPFS
`context.IPFS.add()` is called during `languagePublish` but IPFS has been completely removed from the executor. This crashes language publishing.

### 8. Missing `bootstrapSeed.json`
The executor requires a bootstrap seed with the Language Language bundle, but the test runner doesn't create one, resulting in "No Language Language installed!" errors.

### 9. SSH-only git dependencies in lockfile
`appdata-path` dependency uses SSH URLs (`git@github.com:perspect3vism/...`) which fail in CI environments without SSH keys configured.

### 10. Node builtin imports need `node:` prefix
[...]

PR #677, opened 2026-02-18
feat(test-runner): auto-transform CJS bundles to ESM for Deno runtime

HexaField:
## Summary

Follow-up to #670 — addresses the remaining items from #671 (issues 6, 7, 10) that could not be fixed in the first PR.

### Problem

System language bundles downloaded by `@coasys/ad4m-test` from the perspect3vism repos are CommonJS, but the executor's Deno JS runtime requires ESM. This causes:

- **`ReferenceError: exports is not defined`** — CJS `exports` object doesn't exist in ESM context
- **`ReferenceError: require is not defined`** — CJS `require()` calls fail in Deno
- **IPFS crash** — Language Language bundle calls `context.IPFS.add()` which was removed from the executor
- **Node builtin resolution** — `require("path")` needs `node:path` prefix for Deno

Previously users needed a [custom patch script](https://github.com/HexaField/nextgraph-adam-language/blob/main/scripts/patch-ad4m-test.cjs) to handle all of this. This PR integrates the transform directly into the test-runner.

### Changes

**New: `test-runner/src/bundleTransform.ts`**
- `convertCjsToEsm()` — converts `require()` → `import`, `exports.x` → `export`, adds `node:` prefix for builtins
- `patchIpfsReferences()` — removes dead IPFS code from Language Language bundle
- `transformBundle()` — orchestrates both transforms with detection and logging

**Modified: `test-runner/src/installSystemLanguages.ts`**
- Language Language bundle is transformed (CJS→ESM + IPFS patch) before embedding in `bootstrapSeed.json`
- All system language bundles are transformed before publishing to the executor

### Testing

This transform logic has been running successfully in CI for [nextgraph-adam-language](https://github.com/HexaField/nextgraph-adam-language) and [harvest AD4M integration](https://github.com/HexaField/harvest) — both have green CI pipelines using the exact same conversion logic.

Closes remaining items from #671.

/cc @lucksus @data-bot-coasys

## Release Notes

* **Improvements**
  * Language bundles are now automatically transformed to ensure compatibility with the runtime environment.
  * Enhanced bundle processing with support for automatic format conversion and optimization.

### DIFF
--- a/test-runner/src/bundleTransform.ts
+++ b/test-runner/src/bundleTransform.ts
@@ -1,0 +1,162 @@
+/**
+ * Bundle Transform Utilities
+ *
+ * Converts CJS system language bundles to ESM for Deno runtime compatibility,
+ * and patches out removed IPFS references from the Language Language bundle.
+ *
+ * Background: System language bundles downloaded from perspect3vism repos are
+ * CommonJS, but the AD4M executor's Deno runtime requires ESM. Additionally,
+ * the Language Language bundle still references `context.IPFS.add()` which was
+ * removed from the executor, causing crashes during language publishing.
+ *
+ * @module bundleTransform
+ */
+
+import { logger } from './utils';
+
+/** Node built-in modules that need 'node:' prefix in Deno */
+const NODE_BUILTINS = new Set([
+  'assert', 'buffer', 'child_process', 'cluster', 'console', 'constants',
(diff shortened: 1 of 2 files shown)


HexaField:
Closing this — the CJS→ESM conversion is better handled at the consumer level (each repo's CI scripts) rather than baked into the test-runner. Both repos that need it (Harvest, nextgraph-adam-language) already have working patch scripts, and the upcoming release will likely make this moot entirely. Thanks for the review feedback. ⬡

PR #654, opened 2026-02-04
feat(shacl): SHACL SDNA Migration - Prolog to SHACL links parser

data-coasys:
# SHACL SDNA Migration - Queryable Schema Links

## Overview

This PR migrates AD4M's schema definition system (SDNA) from opaque Prolog strings to **queryable W3C SHACL RDF links**. This enables schemas to be stored, queried, and reasoned about using the same graph infrastructure as instance data.

### Link Structure (Named Property Shapes - Option 3)

```
┌─────────────────────────────────────────────────────────────────────────┐
│                         SHACL LINK STRUCTURE                            │
├─────────────────────────────────────────────────────────────────────────┤
│                                                                         │
│  ┌──────────────────┐                                                   │
│  │   NodeShape      │                                                   │
│  │ recipe://Recipe  │                                                   │
│  └────────┬─────────┘                                                   │
│           │                                                             │
│           ├──────────────────────────────────────────────────────────┐  │
│           │ sh:targetClass                                           │  │
│           ▼                                                          │  │
│  ┌──────────────────┐                                                │  │
│  │  recipe://Recipe │ (the class being described)                    │  │
│  └──────────────────┘                                                │  │
│           │                                                          │  │
│           ├──────────────────────────────────────────────────────────┤  │
│           │ sh:property (one per property)                           │  │
│           ▼                                                          │  │
│  ┌────────────────────┐   ┌────────────────────┐                     │  │
│  │ recipe://Recipe    │   │ recipe://Recipe    │                     │  │
│  │       .name        │   │   .ingredients     │                     │  │
│  └─────────┬──────────┘   └─────────┬──────────┘                     │  │
│            │                        │                                │  │
│   ┌────────┴────────┐      ┌────────┴────────┐                       │  │
│   │ sh:path: name   │      │ sh:path: ingr.. │                       │  │
│   │ sh:datatype:xsd │      │ sh:class: Ingr  │                       │  │
│   │ sh:minCount: 1  │      │ ad4m:collection │                       │  │
│   │ ad4m:getter     │      │ ad4m:adder      │                       │  │
│   │ ad4m:setter     │      │ ad4m:remover    │                       │  │
│   └─────────────────┘      └─────────────────┘                       │  │
│                                                                      │  │
└──────────────────────────────────────────────────────────────────────┘  │
                                                                          │
```

### Code Architecture

```
┌─────────────────────────────────────────────────────────────────────────┐
│                         CODE RELATIONSHIPS                              │
[...]

PR #685, opened 2026-02-19
refactor: use batching for SHACL/Flow link operations

HexaField:
Addresses review feedback from @lucksus on #654:

### Changes:
1. **Top-level imports** — Moved `SHACLShape` and `SHACLFlow` imports from inline `import()` to top-level static imports
2. **Batch link writes** — `addShacl()` and `addFlow()` now use `addLinks()` to batch all links in a single call instead of individual `add()` calls
3. **Single surreal query for reads** — `getShacl()` and `getFlow()` now use a single `querySurrealDB()` call instead of many individual `get()` calls

Net result: ~74 lines added, ~146 removed. Much fewer round-trips to the executor.

### DIFF
--- a/core/src/perspectives/PerspectiveProxy.ts
+++ b/core/src/perspectives/PerspectiveProxy.ts
@@ -14,6 +14,8 @@ import { PERSPECTIVE_QUERY_SUBSCRIPTION } from "./PerspectiveResolver";
 import { gql } from "@apollo/client/core";
 import { AllInstancesResult } from "../model/Ad4mModel";
 import { escapeSurrealString } from "../utils";
+import { SHACLShape } from "../shacl/SHACLShape";
+import { SHACLFlow } from "../shacl/SHACLFlow";
 
 type QueryCallback = (result: AllInstancesResult) => void;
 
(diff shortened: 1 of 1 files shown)


core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -1078,84 +1080,34 @@ export class PerspectiveProxy {
...
+        const sourceUris = [shapeUri, ...propertyLinks.map(l => l.data.target)];
+        const escapedSources = sourceUris.map(u => `'${escapeSurrealString(u)}'`).join(', ');
         
-        // Get destructor actions
-        const destructorLinks = await this.get(new LinkQuery({
-            source: shapeUri,
-            predicate: "ad4m://destructor"
-        }));
-        shapeLinks.push(...destructorLinks.map(l => l.data));
+        const query = `SELECT in.uri AS source, predicate, out.uri AS target FROM link WHERE in.uri IN [${escapedSources}]`;
```
lucksus, line 1095: soo much cleaner!

lucksus (approved):
Great! Thanks Hex :)

PR #686, opened 2026-02-19
fix(test-runner): remove stale process kills and perspect3vism refs

HexaField:
Cleanup follow-up from [#678](https://github.com/coasys/ad4m/pull/678#pullrequestreview-3827669994) — Nico noted the `findAndKillProcess('holochain')` and `findAndKillProcess('lair-keystore')` calls are stale (pre-Rust refactor). Everything is a single `ad4m-executor` process now.

## Changes

- **Replace** all `findAndKillProcess('holochain')` / `findAndKillProcess('lair-keystore')` with `findAndKillProcess('ad4m')` across `cli.ts` and `installSystemLanguages.ts`
- **Remove** stale `--ipfsPort` flag from executor spawn args (IPFS was removed)
- **Remove** dead code: identical `if/else` branches on `defaultLangPath`, unused `get-port` import
- **Update** binary download URL from `perspect3vism/ad4m` → `coasys/ad4m`
- **Update** README and example `package.json` URLs from `perspect3vism` → `coasys`

Small, safe cleanup — no behaviour changes beyond killing the right process name.

⬡ Hex

## Summary by CodeRabbit

* **Chores**
  * Updated repository references and dependencies from perspect3vism to coasys across configuration files

* **Refactor**
  * Simplified test-runner startup by removing port discovery and conditional logic
  * Standardized process handling with fixed argument sets and streamlined error handling


test-runner/src/cli.ts
```diff
@@ -86,8 +86,7 @@ async function installLanguage(child: any, binaryPath: string, bundle: string, m
           neighbourhood,
           clear: () => {
             kill(child.pid!, async () => {
-              await findAndKillProcess('holochain')
-              await findAndKillProcess('lair-keystore')
+              await findAndKillProcess('ad4m')
```
lucksus: it would be `ad4m-executor`

lucksus (changes requested):
I'm wondering if we need the kill at all even if the binary name would be correct `ad4m-executor`. can you explain what's going on here?

HexaField:
The kill is a safety net — if a previous test run crashed or was interrupted, there may be orphaned `ad4m` processes still holding ports. Without the kill, the next `startServer` spawn would fail to bind.

The binary is downloaded and spawned as `ad4m` (see `utils.ts` line ~68, ~87), so `findAndKillProcess('ad4m')` matches the right process name. The old `holochain`/`lair-keystore` names were stale since those are now embedded in the single executor binary.

Happy to remove the kills entirely if you prefer — the trade-off is that a crashed test run could block subsequent runs until the port is manually freed.

⬡

HexaField:
Good call — removed `findAndKillProcess` entirely along with the `find-process` dependency. It was matching by name which could kill a real AD4M instance, not just a test leftover. The teardown paths already use `tree-kill(child.pid!)` to target the specific process.

⬡

lucksus (dismissed):

PR #704, opened 2026-03-02
feat: State of Affairs (SoA) ontology for knowledge representation

data-bot-coasys:
Fix applied: Node 18-compatible path resolution in esbuild.ts

## New Features

* Added State of Affairs (SoA) ontology module for knowledge representation with six distinct modalities: belief, observation, intention, vision, plan, and skill
* Introduced standard relationship types for linking knowledge units: supports, contradicts, similar, same, requires, enables, parent, refines, and blocks
* Implemented self-documenting SoA language with built-in provenance tracking (authorship and timestamp)


soa/schemas/Relationship.json
```diff
@@ -0,0 +1,85 @@
...
+  "title": "SoARelationship",
+  "description": "A typed edge between two States of Affairs. Enables epistemic, equivalence, and structural connections across SoA trees.",
+  "type": "object",
+  "x-ad4m": {
+    "namespace": "soa://",
+    "className": "SoARelationship",
+    "flag": {
+      "through": "soa://entry_type",
+      "value": "soa://relationship"
+    }
```
lucksus: Do we really need a flag? I actually think that's an anti-pattern. The graph shape with used language ('soa' here) already provides enough constraints.

soa/schemas/Relationship.json
```diff
@@ -0,0 +1,85 @@
...
+  "properties": {
+    "relationshipType": {
+      "type": "string",
+      "enum": ["supports", "contradicts", "similar", "same", "requires", "enables", "parent", "refines", "blocks"],
+      "description": "The type of relationship between source and target SoA nodes",
+      "x-ad4m": {
+        "through": "soa://relationship_type",
+        "resolveLanguage": "literal",
+        "writable": true
+      }
```
lucksus: I think what might be more elegant would be to just use one link for relationships and have the 'soa' language have those types as expressions. I.e. `soa://rel_supports`, `soa://rel_contradicts`..

lucksus: Then we don't need a Relationship type at all. But we should talk about actually implementing that soa language (as a static language - just defining those static expressions which resolve to documentation :))

soa/schemas/Relationship.json
```diff
@@ -0,0 +1,85 @@
...
+      "type": "string",
+      "description": "Optional annotation explaining the relationship",
+      "x-ad4m": {
+        "through": "soa://relationship_note",
+        "resolveLanguage": "literal",
+        "writable": true
+      }
+    },
+    "author": {
+      "type": "string",
```
lucksus: not needed. every link that constitutes this shape is a link expression with author and timestamp. and there could be multiple authors (by multiple users creating that SoA even independently).

lucksus (changes requested):
Should the json be a shacl JSON representation?

data-bot-coasys:
Hey Nico — all your review points are addressed in commit `036dbc46d`:

1. **Removed `is_relationship` flag** — you're right, the graph shape + language provides enough constraints
2. **Relationships via `soa://rel_*` predicates** — e.g. `soa://rel_supports`, `soa://rel_contradicts`, etc. as link predicates (no separate Relationship type needed)
3. **Removed author/createdAt** — link expressions already carry those natively
4. **Removed the JSON schema files** — the TypeScript model with `@ModelOptions`/`@Property` decorators is the source of truth (these generate SHACL SDNA automatically via the AD4M model system)

The SoA language itself (defining those static expressions like `soa://rel_supports`) is documented in the README but not yet implemented as an actual AD4M Language. Happy to discuss the design for that — a static language that just resolves to documentation sounds clean.

PR #710, opened 2026-03-04
fix: handle language runtime shutdown gracefully to prevent thread panics

HexaField:
During neighbourhood teardown, three `lang-*` threads panic at `js_core/mod.rs:267:53` because `execute_script()` returns an error after the V8 isolate is being torn down, and the result is `.unwrap()`ed.

This replaces `.unwrap()` with `?` so the error propagates gracefully through the `Result` return type instead of panicking the thread. This allows clean thread termination during `language_remove()`.

Discovered during memory profiling in #689 — these panics prevent complete memory cleanup during teardown (currently only 7.6% recovery).

## Change

`rust-executor/src/js_core/mod.rs`: `execute_async.unwrap()` → `execute_async?`

## Release Notes

* **Documentation**
  * Added comprehensive memory profiling reports and leak investigation guides.

* **Bug Fixes**
  * Fixed memory leaks during perspective removal and executor shutdown.
  * Improved cleanup of application-related resources and signal streams.
  * Enhanced error handling in script execution.

* **Chores**
  * Added profiling and memory investigation tools for analysis and diagnostics.


HexaField:
Reopening with correct base branch

PR #717, opened 2026-03-05
Feat(model)  ad4m model api changes

jhweir:
# PR: Ad4mModel Decorator API Refactoring

---

### 1. New decorator API — `@Model`, `@HasMany`, `@HasOne`, `@BelongsToOne`, `@BelongsToMany`

Introduced the `@Model` class decorator and four relation decorators to replace raw `@Collection` usage. **Why:** Relation semantics (one-to-many, one-to-one, ownership direction) are now explicit in the code rather than implicit. `@Collection` becomes an internal implementation detail.

### 2. Replace `writable` with `readOnly` in public API

Inverted the boolean semantics: `PropertyOptions.writable` → `PropertyOptions.readOnly`. **Why:** `readOnly: true` reads more naturally and matches developer expectations. Internally, `writable` is still computed for SDNA/SHACL compatibility.

### 3. `@Property` becomes the core decorator with smart defaults

Extracted shared logic into an internal `applyPropertyMetadata()` helper. `@Property` is now the primary decorator; `@Optional` and `@ReadOnly` are thin wrappers. **Why:** Eliminates duplication. Smart defaults (`required: false`, `readOnly: false`, `resolveLanguage: "literal"`) mean less boilerplate for the most common case. Properties are optional by default — no link is created until a value is explicitly set (see §38 for the full rationale).

### 4. Wire decorators to WeakMap-based metadata registries

All decorators now write to `propertyRegistry`, `collectionRegistry`, and `relationRegistry` (WeakMaps keyed by constructor). Added `getPropertiesMetadata()`, `getCollectionsMetadata()`, `getRelationsMetadata()` getters that walk the prototype chain. **Why:** WeakMaps are GC-friendly, avoid shared mutable state across inheritance chains, and provide a clean public API for introspecting model metadata.

### 5. Remove legacy `__properties` / `__collections` prototype mutation

Deleted all reads and writes to `prototype.__properties` and `prototype.__collections` across `decorators.ts`, `Ad4mModel.ts`, `PerspectiveProxy.ts`, and `SubjectRepository.ts`. **Why:** Prototype mutation leaks metadata between parent/child classes, is invisible to TypeScript, and creates fragile coupling. Zero references remain.

### 6. Overloaded relation signatures

All four relation decorators now accept two calling conventions:
```typescript
@HasMany({ through: "post://comment", target: () => Comment })   // options object
@HasMany(() => Comment, { through: "post://comment" })            // target-first shorthand
```
**Why:** The target-first style is more readable and mirrors conventions in other ORMs (e.g. TypeORM, Sequelize). The options-object style is kept for backward compatibility and complex configurations.

### 7. `@Flag` hardening — auto `readOnly`, skip in update, throw on setter

`@Flag` now auto-sets `readOnly: true` / `writable: false` internally so callers never need to specify it. `innerUpdate()` silently skips flag properties during updates (they are immutable after creation). `generatePropertySetterAction()` throws a clear Flag-specific error before the generic `readOnly` check. **Why:** Flags are type markers that should never change — making immutability automatic eliminates a class of bugs where flags could accidentally be overwritten during `update()`.

### 8. Relation decorator validation — throw on missing `through`

`resolveRelationArgs()` now validates that `{ through }` is present and throws a clear error at class-definition time if missing. **Why:** Previously a missing `through` would silently produce broken SDNA/SHACL and only fail at query time with a cryptic error. Fail-fast at decorator application time makes misconfiguration obvious immediately.

### 9. Auto-derive `effectiveInitial` for required writable properties

Both `generateSDNA()` and `generateSHACL()` now compute an `effectiveInitial` (`"literal://string:"`) for **required**, writable, non-flag properties that have no explicit `initial`. Optional properties (`required: false`, which is now the default — see §38) are excluded — they remain unset (`undefined`) until explicitly assigned, preserving the semantic distinction between "never set" and "set to empty string". **Why:** When a property is explicitly marked `required: true`, it must have a link after creation, so the initial link is always needed. Auto-deriving saves boilerplate. Optional properties are excluded because polluting them with an empty-string placeholder would break the ability to distinguish "never set" from "set to empty".

### 10. SHACL `sh:node` model inheritance

`SHACLShape` gains `parentShapes[]` and `addParentShape()`. When a `@Model` class extends another `@Model`, `generateSHACL()` detects the parent via `Object.getPrototypeOf` and adds its shape URI via `sh:node`. Serialized in `toJSON`/`fromJSON` (`parent_shapes`) and `toTurtle` (`sh:node <parentShapeUri>`). **Why:** SHACL validators can now walk model hierarchies — a child shape references its parent's shape, ensuring inherited constraints are validated without duplicating property shapes.

### 11. Rename `setCollection<Prop>` → `set<Prop>` in generated stubs

Public-facing collection setter methods are now named `set<PropName>()` instead of `setCollection<PropName>()`. Updated in the `Collection` decorator stub, `HasManyMethods` type, `util.ts` helpers (`collectionToSetterName` / `collectionSetterToName`), `fromJSONSchema` dynamic builder, and `ad4m-hooks/helpers`. **Why:** Aligns collection setter naming with the simpler `set*` convention already used by property setters — `recipe.setIngredients(...)` reads better than `recipe.setCollectionIngredients(...)`.

### 12. Rename `collection*` → `relation*` throughout TypeScript stack

Unified terminology from "collection" to "relation" across all consumer code:
- `CollectionMetadata` → `RelationMetadata` (type alias kept for backward compat)
- `ModelMetadata` gains `.relations` field (`.collections` kept as deprecated alias)
- `getCollectionsMetadata` / `setCollectionRegistryEntry` marked `@deprecated`
- Private `Ad4mModel` methods renamed: `setCollectionSetter` → `setRelationValues`, `setCollectionAdder` → `addRelationValue`, `setCollectionRemover` → `removeRelationValue`, `generateCollectionAction` → `generateRelationAction`, `getCollectionMetadata` → `getRelationOptions`
- `PerspectiveProxy.findClassByProperties` local vars updated

**Why:** "Collection" conflated the ORM concept (a set of links) with the relational concept (a typed association between models). Using "relation" consistently aligns the codebase with ORM conventions and makes the distinction between scalar properties and relational fields immediately clear. Prolog/SurrealDB wire-format names are unchanged.

### 13. `baseExpression` → `id` as canonical identifier

Renamed the private field `#baseExpression` → `#id` and added `get id()` as the canonical getter. `get baseExpression()` is preserved as a `@deprecated` alias. Constructor param renamed to `id?`. All internal usages updated to `#id`. External consumers (SubjectRepository, useModel vue/react, tests) updated to use `.id`. **Why:** `id` follows standard ORM conventions, `baseExpression` leaked internal abstraction details. Backward-compatible via deprecated alias.

### 14. Static `create` / `update` / `remove` / `register` convenience methods

Added four static methods to `Ad4mModel`: `create(perspective, data, source?)` for one-step instantiation+save, `update(perspective, id, data)` for fetch+merge+save, `remove(perspective, id)` for delete-by-id, and `register(perspective)` to install the SHACL schema. **Why:** Static methods follow standard ORM conventions (ActiveRecord pattern), reduce boilerplate, and make model usage more discoverable. Instance methods preserved.

### 15. Transaction API for atomic batch operations

Added `Ad4mModel.transaction(perspective, async (tx) => { ... })` that creates a batch, passes a `TransactionContext` object (containing `tx.batchId`) to the callback, and commits atomically on success. If the callback throws, the batch is not committed, preventing partial writes. **Why:** Multi-property updates and multi-model operations need atomicity guarantees, and passing an object allows for future API expansion (like rollbacks) without breaking changes.

### 16. Remove `@InstanceQuery` decorator

Deleted the `InstanceQueryParams` interface and `InstanceQuery` decorator function from `decorators.ts`. Removed the `Subject` and `PerspectiveProxy` imports that were only used by this decorator. Removed `@InstanceQuery` usage from test model class definitions. **Why:** `@InstanceQuery` generated Prolog queries at decorator-application time, tying model definitions to the legacy Prolog engine. The modern replacement is the fluent query API: `MyModel.query(perspective).where({...}).get()`.

### 17. Include API for eager-loading relations

Added `IncludeMap` interface, `include` field on `Query`, and `.include(map)` on `ModelQueryBuilder`. When `.include({ comments: true })` is specified, relation fields are hydrated from raw expression URIs into full model instances using the `target()` class from the relation decorator. Supports nested includes (`{ comments: { author: true } }`) for multi-level eager loading. Wired into both `get()` and `subscribe()`. **Why:** Without include, relation fields are opaque URI strings. Eager loading lets consumers work with typed model instances directly, matching the standard ORM `include`/`populate` pattern.

### 18. `Query.parent` — scoped & reverse-link queries

Added `parent?: { id: string; model?: typeof Ad4mModel; predicate?: string }` to `Query` and a `resolveParentPredicate()` helper that scans the parent model's relation metadata for a relation whose `target()` matches the queried child class, extracting the `through` predicate automatically. Added `buildParentQuery` for Prolog triple generation and wired parent filters into `queryToSurrealQL`, `queryToProlog`, and `countQueryToProlog`. The `.parent()` method on `ModelQueryBuilder` supports three calling styles: `parent(instance)` (id + predicate auto-resolved from `instance.constructor`), `parent(id, ModelClass)` (predicate resolved from class metadata), and `parent(id, "predicate://...")` (raw escape hatch). Passing a bare string id with no second argument throws because the predicate is unresolvable. No `ad4m://has_child` default — model-specific predicates are the preferred approach to avoid ambiguity during updates/deletes. **Why:** One method covers both hierarchical scoping and arbitrary reverse-link queries. The model-class approach is type-safe and keeps predicate URIs in one place (the decorator), while the string escape hatch covers dynamic use cases.

### 19. Snapshot dirty tracking

On hydration (`get()`, query results from both Prolog and SurrealDB paths), a shallow snapshot of all property and collection values is captured via `takeSnapshot()`. Added `isDirty()` (returns `true` if any field changed) and `changedFields()` (returns an array of changed field names) to model instances. `innerUpdate()` now checks the snapshot and skips unchanged fields, reducing unnecessary network calls. New instances that haven't been fetched yet (no snapshot) treat all fields as dirty for backward compatibility. **Why:** Without dirty tracking, `update()` rewrites every property/collection even when only one field changed — wasteful and potentially destructive for concurrent edits. Snapshot comparison makes partial updates the default behavior.

### 20. Unified hydration — `hydrateFromLinks` / `hydratePropertyValue`

Added two private static helpers that consolidate duplicated link-processing logic from `getData()` and `instancesFromSurrealResult()`:

- **`hydratePropertyValue(target, propMeta, perspective, expectedType?)`** — single codepath for expression resolution (non-literal languages via `perspective.getExpression`), literal URI parsing (`Literal.fromUrl`), primitive type coercion (string → number/boolean), and transform function application.
- **`hydrateFromLinks(instance, links, metadata, perspective)`** — single-pass link processing with O(1) predicate→property/collection lookup maps (built once per call), latest-wins semantics for properties, chronological accumulation & sorting for collections, and timestamp/author tracking for `createdAt`/`updatedAt`/`author`.

Both `getData()` and `instancesFromSurrealResult()` now delegate to `hydrateFromLinks()`, replacing ~295 lines of duplicated inline logic with a single call each. The Prolog path (`assignValuesToInstance`) is left unchanged — its `[name, value, resolve]` tuple interface and UTF-8 byte reconstruction are sufficiently different that the refactoring benefit is minimal. **Why:** Duplicate hydration code was the largest source of divergent bugs — a fix in one path would be missed in the other. Centralizing the logic ensures consistent behavior and makes future changes (e.g. new property types) a single-site edit.

### 21. Remove `Query.source` — source scoping & `has_child` link

Removed `source?: string` from the `Query` type, deleted the `buildSourceQuery()` helper function, removed the `#source` private field from `Ad4mModel`, removed the `source` parameter from the constructor and static `create()`, stripped source destructuring from `queryToProlog`, `queryToSurrealQL`, and `countQueryToProlog`, removed the `ad4m://has_child` link creation in `save()`, and deleted the `ModelQueryBuilder.source()` method. **Why:** `source` was a legacy scoping mechanism that overlapped with `parent`. The `parent` API (Step 18) is more explicit, type-safe, and model-aware — making `source` redundant. The implicit `has_child` link write on every save was also problematic for consumers who don't use hierarchical scoping.

### 22. Remove `Query.collections` — relations lazy-loaded via `include` only

Removed `collections?: string[]` from the `Query` type, deleted the `buildCollectionsQuery()` Prolog helper, removed collections from `queryToProlog` destructuring, stripped `requestedCollections` logic from both `instancesFromSurrealResult` and `instancesFromPrologResult` (replaced with `requestedProperties`-only filtering that preserves system fields like `createdAt`, `updatedAt`, `author`, `id`), and deleted `ModelQueryBuilder.collections()`. Updated tests: removed collections-dependent assertions, replaced the "filter collections" test with a "filter properties" test, and simplified the "only select requested collections" test. **Why:** Collection selection as a query-time concept is subsumed by `include` (Step 17). Relations are now always lazy — they stay at their field default unless explicitly requested via `include: { comments: true }`. This simplifies the query API and eliminates the confusing distinction between "which collections to fetch at query time" vs "which relations to hydrate into models".

### 23. `Query.properties` — sparse fieldset wired into `hydrateFromLinks`

Added an optional `requestedProperties?: string[]` parameter to `hydrateFromLinks()`. When provided, only property names in the set are added to the predicate→property lookup map — unrequested properties are never hydrated (their fields stay at the class default). Removed the post-hydration `delete` loop from `instancesFromSurrealResult()` since filtering now happens during hydration rather than after. The Prolog path retains its client-side filter as a backup (the Prolog query's `buildPropertiesQuery` already does server-side filtering). `ModelQueryBuilder.properties()` is unchanged. **Why:** Pushing the sparse fieldset into the hydration layer avoids wasted work — property values that won't be used are never resolved, parsed, or type-coerced. This is especially significant for properties with non-literal `resolveLanguage` that would otherwise trigger network calls to `perspective.getExpression()`.

### 24. Remove instance `update()` — unify into `save()`

`save()` now detects whether an instance is new or existing by checking for a snapshot (taken during hydration). New instances (no snapshot) go through the create path (`createSubject` + `innerUpdate`). Existing instances (snapshot present, i.e. fetched via `get()` or a query) go through the update path (`innerUpdate` + `getData`), using dirty tracking from Step 19 to skip unchanged fields. The instance `update()` method is removed. The static `update()` convenience method now calls `instance.save()` instead of `instance.update()`. **Why:** Having both `save()` and `update()` as separate instance methods forced consumers to know whether an instance was new or fetched — a distinction the ORM should handle internally. A single `save()` method matches the ActiveRecord convention where the same method handles both INSERT and UPDATE.

### 25. JS `#` private → TS `private` fields (Vue/Solid proxy compat)

Converted all four JS `#` private fields to TS `private _` fields: `#id` → `private _id`, `#perspective` → `private _perspective`, `#snapshot` → `private _snapshot`, `#subjectClassName` → `private _subjectClassName`. All 49 internal references updated. **Why:** JS `#` private fields use WeakMap semantics internally — when `this` is a Vue or Solid reactive Proxy (which wraps the real object), `#field` access throws `TypeError: Cannot read private member from an object whose class did not declare it` because the Proxy is not the original object. TS `private` compiles to plain properties with compile-time-only access control, which works correctly with Proxy wrappers.

### 26. Remove `where.isInstance` / `where.condition` from collection filtering (`8e32060b`)

Removed the entire post-hydration collection filtering system from `getData()` — both the N+1 `where.condition` SurrealDB sub-queries (one per collection item) and the N+1 `where.isInstance` batch instance checks. Removed `WhereOptions` interface from `CollectionOptions`. Promoted `prologCondition` to a top-level field on `CollectionOptions` (still used in SDNA Prolog generation). Simplified SDNA `collection_getter` generation to only handle `prologCondition` (removed `isInstance` Prolog rules). Updated `@HasMany`/`@BelongsToMany` to forward `prologCondition` directly instead of nesting under `where`. Consumers should use `@HasMany(() => Target, { through: "..." })` for typed relations instead of `@Collection({ where: { isInstance: Target } })`. **Why:** `where.condition` executed a separate SurrealDB sub-query for every item in a collection (N+1), and `where.isInstance` performed a batch SDNA instance check per collection (also N+1 per model instance). Both were post-hydration filters that ran after all collection values had already been fetched — meaning the full data was loaded regardless, making the filtering pure overhead. The `isInstance` use case is now handled by typed relation decorators (`@HasMany(() => Target, ...)`), and custom derived values are handled by the existing SurrealQL `getter` option on `@Property`/`@Collection` (a per-field computed expression, orthogonal to query-level filtering).

### 27. `ParentScope` discriminated union for `Query.parent` (`3f368df2`)

Introduced `export type ParentScope = { model: typeof Ad4mModel; id: string; field?: string } | { id: string; predicate: string }` — a discriminated union that replaces the previous combined optional-fields type. `resolveParentPredicate()` now uses `'predicate' in parent` for TypeScript narrowing instead of optional-field checks, making invalid states (no model AND no predicate) unrepresentable at the type level. `buildParentQuery()` updated to accept `ParentScope | undefined`. `ModelQueryBuilder.parent()` constructs the correct union variant based on arguments: instance or model class → model form, raw string → raw form. JSDoc expanded with all five calling styles. **Why:** The union type provides compile-time safety — callers must supply either a model (for auto-resolution) or an explicit predicate, never an ambiguous combination of optional fields.

### 28. Static `findOne()` and query builder `first()` (`8e2b0bb7`)

Added `static findOne<T>(perspective, query?, useSurrealDB?)` — returns the first matching instance or `null`. Delegates to `findAll` with `limit: 1`. Also added `ModelQueryBuilder.first()` as the fluent equivalent: `Recipe.query(perspective).where({ name: "Pasta" }).first()`. **Why:** `findOne` is the most common ORM query pattern (find-by-criteria returning a single result). Without it, consumers had to call `findAll` and manually index the array. The fluent `first()` mirrors Sequelize/TypeORM conventions.

### 29. `maxCount` on `RelationMetadataEntry` + `@HasOne` hydration guard (`b7c649b9`)

Added `maxCount?: number` to `RelationMetadataEntry`. `@HasOne` and `@BelongsToOne` decorators auto-set `maxCount: 1`. During eager-loading (`hydrateRelations`), when a `maxCount: 1` relation has multiple values, a warning is logged and only the last value is kept (consistent with property latest-wins semantics). **Why:** Metadata consumers can now programmatically distinguish single-valued vs multi-valued relations, and the hydration layer enforces cardinality constraints rather than silently returning arrays for `@HasOne` fields.

### 30. Static `delete()` + incoming link cleanup (`ed0c1d70`)

Added `static delete(perspective, id)` as the canonical static method for deletion; `remove()` is preserved as a `@deprecated` alias that delegates to `delete()`. The instance `delete()` method now also queries for all incoming links (`target === this._id`) and removes them in the same pass (best-effort, non-fatal on failure). **Why:** `delete` matches standard ORM naming (TypeORM, Prisma). Incoming link cleanup prevents orphan references — when a recipe is deleted, the parent cookbook's `hasMany` link to it is also removed, keeping the graph consistent.

### 31. Expand `create()` with `options` (parent & batchId) (`9bc01bfd`)

`create()` now accepts an optional third argument `{ parent?: ParentScope; batchId?: string }`. When `parent` is provided, a link from `parent.id` → new instance id is auto-created using the resolved predicate (same `resolveParentPredicate` as `Query.parent`). When `batchId` is provided, both the `save()` and the parent link are added to the batch instead of committed immediately. **Why:** Creating a child under a parent is the most common hierarchical write pattern. Without this, consumers had to manually call `perspective.add(new Link(...))` after `create()` — error-prone and verbose. The `batchId` option enables transactional create flows.

### 32. `HasManyMethods` accept instances or IDs in relation mutation methods

`add*`, `remove*`, and `set*` methods on model instances now accept either a plain string ID or any object with an `id: string` field (i.e. any model instance). A private `resolveRelationId()` helper in `Ad4mModel` extracts `.id` when an object is passed, and falls back to `String(value)` for raw strings. The `HasManyMethods` type signatures in `decorators.ts` are widened to `string | { id: string }` (structural type, not `Ad4mModelLike` constructor type). **Why:** Forcing callers to pass `comment.id` when they already hold the `comment` instance is unnecessary boilerplate, requires knowledge of an ORM implementation detail, and conflicts with the OOP style of the rest of the API. Accepting both matches every major ORM convention — pass a string ID when you only have the ID (e.g. from an API payload), or pass the model instance when you have it.

### 33. `GetOptions` on instance `get()` — `include` and `properties` support

Added `GetOptions` type `Pick<Query, 'include' | 'properties'>` and made `get()` accept it as an optional argument. `properties` is forwarded into the existing `hydrateFromLinks` sparse-fieldset path; `include` triggers a `hydrateRelations` call after core hydration, eager-loading the specified relations. `get()` also accepts a bare `IncludeMap` directly (auto-wrapped into `{ include: ... }`) so the common case stays terse: `post.get({ comments: true })`. Fully backward-compatible — `get()` with no args is unchanged. **Why:** `get()` is the natural method for re-hydrating an instance you already hold. Without `include` support, consumers had to reach for `findOne({ where: { id }, include: {...} })` — more verbose and requires knowing the id separately. `get({ include: { comments: true } })` is the idiomatic form.

### 34. DRY type hierarchy — `RelationSubQuery` and `GetOptions` derived from `Query` (`0aae13dc`)

`RelationSubQuery` and `GetOptions` were verbose `interface` declarations duplicating fields already on `Query`. Replaced both with derived utility types:

```ts
export type RelationSubQuery = Omit<Query, 'parent' | 'count'>;
export type GetOptions       = Pick<Query, 'include' | 'properties'>;
```

`Query` is now the single source of truth for `where / order / limit / offset / properties / include`. `RelationSubQuery` inherits all of those minus `parent` (already scoped to the linked relation) and `count` (not applicable for eager-loading). `GetOptions` picks only the two hydration controls that make sense when re-fetching a single known instance. Type definitions reordered so `Query` is declared before its dependents. **Why:** A shared vocabulary expressed once and derived everywhere is less maintenance, less cognitive overhead, and gives TypeScript's structural typing more surface area to catch mismatches automatically.

### 35. Wire `include` into `findAll`/`findOne`; `generateSHACL`/`generateSDNA` stubs (`bff4e665`)

Two fixes surfaced by integration tests:

**`include` missing from query results** — `instancesFromSurrealResult` and `instancesFromPrologResult` were hydrating instances but never calling `hydrateRelations`. Every `findAll`/`findOne` with `include: { comments: true }` silently returned raw URI strings instead of model instances. Fixed by calling `hydrateRelations(paginatedInstances, perspective, query.include)` after snapshots are taken in both code paths.

**`generateSHACL`/`generateSDNA` stubs on `Ad4mModel`** — Added concrete static stub methods to `Ad4mModel` so TypeScript resolves the signatures on all subclasses (fixes `TS2339: Property 'generateSHACL' does not exist on type 'typeof TestDerivedModel'`). Stubs return safe sentinel values (`{ shape: null, name: '' }` / `''`) rather than throwing — the `@Model` decorator's parent-chain detection already guards with optional chaining (`parentSHACL?.shape?.nodeShapeUri`), so `null` is handled gracefully and does not affect SHACL generation for decorated subclasses.

### 36. Merge `collectionRegistry` into unified `relationRegistry`; fix `Collection()` stubs (`e404ec22`)

**Fix: `Collection()` prototype methods were no-op stubs** — `@HasMany` / `@BelongsToMany` called `Collection()` to generate `add*` / `remove*` / `set*` prototype methods, but those methods were assigned as `() => {}` empty functions. `post.addComments(comment)` silently did nothing. Fixed by replacing stubs with real `async` implementations that delegate to `this.addRelationValue`, `this.removeRelationValue`, and `this.setRelationValues`.

**Registry merge** — Removed the entire `collectionRegistry` abstraction: `collectionRegistry` WeakMap, `CollectionOptions` interface, `Collection()` function, `getCollectionsMetadata()`, and `setCollectionRegistryEntry()`. All relation metadata now lives in the single `relationRegistry`. `RelationMetadataEntry.target` is now optional; `getter?` is added directly to the entry so `@HasMany({ getter: ... })` stores into the unified registry without routing through `Collection()`. A new `setRelationRegistryEntry()` export replaces `setCollectionRegistryEntry()`.

Callers updated across three files:
- **`decorators.ts`** — `generateSDNA()` and `generateSHACL()` in `@Model` filter `getRelationsMetadata()` to `kind === 'hasMany' | 'belongsToMany'` instead of calling `getCollectionsMetadata()`.
- **`Ad4mModel.ts`** — `getRelationOptions()` return type changed to `RelationMetadataEntry | undefined`; `generateRelationAction()` reads `metadata.predicate` instead of `metadata.through`; `fromJSONSchema` dynamic builder calls `setRelationRegistryEntry()` with the correct shape.
- **`PerspectiveProxy.ts`** — `findClassByProperties` uses filtered `getRelationsMetadata()` instead of `getCollectionsMetadata()`.

**Why:** A collection was always just a `hasMany` relation stored in a parallel registry with a different field name (`through` instead of `predicate`). Two registries for the same concept created a maintenance hazard: any change to relation semantics had to be applied twice, and the distinction was invisible at the call site. A single `relationRegistry` is the single source of truth for all inter-model associations, regardless of arity or ownership direction.

### 37. Rename `useModel` → `useLiveQuery` with improved hook API (`603c0374`)

**Rename & new API shape** — `useModel` renamed to `useLiveQuery` in both `ad4m-hooks/react` and `ad4m-hooks/vue`. The API changes from a single props object to positional arguments: `useLiveQuery(model, perspective, options?)`. The return key changes from `entries` to `data` to follow conventional naming. Exported types `LiveCollectionResult<T>` and `LiveInstanceResult<T>` give consumers explicit result shapes.

**Single-instance mode** — When `options.id` is provided, the hook returns `{ data: T | null, loading, error }` instead of the collection result. Implemented via TypeScript overload signatures that narrow the return type at the call site. Internally, the query is filtered with `where: { id }` and `results[0]` is extracted.

**`ParentScope` support** — `options.parent` accepts the `ParentScope` discriminated union (model-form with auto-resolved predicate, or raw-form with explicit predicate string). The hook passes `parent` straight through to `Query` — `resolveParentPredicate()` in `Ad4mModel.queryToProlog()` handles the resolution internally, so no duplicate logic is needed in the hooks. Includes an early-exit guard when `parent.id` is falsy (e.g. prop not yet resolved).

**Remove `ensureSDNASubjectClass` from hooks** — The old `useModel` called `perspective.ensureSDNASubjectClass(model)` on every mount, gating the subscription behind a `subjectEnsured` state variable and a dedicated `useEffect`. This is removed — consumers should call `Model.register(perspective)` once during app initialization. This eliminates the two-phase mount, the extra state, and the extra effect.

**Vue `shallowRef` fix** — Vue's `ref()` deep-proxies objects, which breaks classes using JS `#` private fields (`PerspectiveProxy`, `NeighbourhoodProxy`, etc.) because Proxy ≠ the original object for private field access. `collectionData` and `instanceData` are switched from `ref()` to `shallowRef()` so Vue tracks the reference reactively without deep-proxying model instances. (`e63110e6`)

**React improvements** — Subscribe logic wrapped in `useCallback` with a proper dependency array (`perspective.uuid`, model name, `pageNumber`, serialized query, `parent.id`, `parent.field`, `id`). Effect cleanup now disposes the `ModelQueryBuilder` on unmount and re-subscribe, preventing leaked subscriptions.

**Vue improvements** — Uses `isRef()` from Vue instead of manual `"value" in perspective` duck-typing. Adds `onUnmounted()` cleanup to dispose the `ModelQueryBuilder` — the old `useModel` Vue hook never cleaned up subscriptions on unmount. Removes the `includeIds()` hack that forced the `id` getter to be enumerable via `Object.defineProperty`.

**String model guard** — When `model` is an empty string (e.g. generic class browser before a class is selected), the hook returns empty data immediately without attempting to subscribe.

Files changed:
- **`ad4m-hooks/react/src/useLiveQuery.ts`** — New file replacing `useModel.ts`
- **`ad4m-hooks/vue/src/useLiveQuery.ts`** — New file replacing `useModel.ts`
- **`ad4m-hooks/react/src/index.ts`** — Export `useLiveQuery` + result types instead of `useModel`
- **`ad4m-hooks/vue/src/index.ts`** — Export `useLiveQuery` instead of `useModel`

**Why:** `useModel` under-communicated the reactive/live nature of the hook and used a props-object API that didn't match conventional hook ergonomics. The name `useLiveQuery` aligns with the planned refactor branch and established conventions (Firebase, Supabase, Realm). The positional-arg API is more ergonomic for the common case. Removing `ensureSDNASubjectClass` from the hooks follows the principle that schema registration is an app-level concern, not a per-component concern. The Vue cleanup fix addresses a real memory/subscription leak.

### 38. `@Property` defaults to `required: false` + open-world structural matching (`655116f4`)

`@Property` now defaults to `required: false` instead of `required: true`. Properties are optional by default — when a model instance is created without providing a value for an optional property, no link is added to the graph. To make a property mandatory, set `required: true` explicitly; this also auto-generates the `"literal://string:uninitialized"` sentinel so the SDNA constructor creates a placeholder link.

**Why:** The old `required: true` default created sentinel `"literal://string:uninitialized"` links for every property on every model instance — even properties the consumer never intended to populate. This polluted the link graph with placeholder data, bloated storage, and made it impossible to distinguish "never set" from "set to empty". Defaulting to `required: false` follows the principle of least surprise: declare a property, set it when you want, and it only exists in the graph when it has a real value. Auto-generating a type flag on every model would force closed-world type discrimination, breaking AD4M's open-world structural matching where models are identified by their predicate structure rather than explicit type tags.

### 39. `batchId` parameter on HasMany relation methods (`03be7aa9`)

`add*`, `remove*`, and `set*` prototype methods on `@HasMany` relations now accept an optional `batchId?: string` second parameter. When provided, link mutations are added to the specified batch instead of being committed immediately. The `HasManyMethods` type signatures are updated accordingly.

**Why:** Without this, callers who wanted to batch multiple relation mutations had to reach for low-level `perspective.add(link, batchId)` calls instead of the ergonomic `post.addComments(comment)` API. Passing `batchId` through the relation methods lets transactional workflows use the same method names as non-transactional ones.

### 40. Relation filtering via `target` — SHACL getter + conformance conditions

Completes the Prolog → SHACL migration for read operations. Added `getter?: string`, `conformanceConditions?: ConformanceCondition[]`, and `class?: string` (`sh:class`) to `SHACLPropertyShape`. Added `ConformanceCondition` interface (`{ type: 'flag' | 'required'; predicate: string; value?: string }`). All three fields wire through `toLinks()`/`fromLinks()` (as `ad4m://getter`, `ad4m://conformanceConditions`, `sh://class` links) and `toJSON()`/`fromJSON()`.

Exported `buildConformanceFilter(predicate, targetClass)` from `decorators.ts` — inspects the target model's `@Flag` and required `@Property` metadata, builds structured conditions, and composes them into a SurrealQL getter string. In `generateSHACL()`, each relation with `target` and `filter !== false` calls `buildConformanceFilter()` at definition time, storing both the pre-computed getter and the structured conditions in the SHACL shape. Explicit `getter` takes precedence (escape hatch). `filter: false` opts out of auto-filtering. `evaluateCustomGettersForInstance()` reads the getter from metadata at runtime.

**Why:** The Prolog SDNA stored `collection_getter` facts in the shape — all consumers (JS, Rust, MCP, CLI) read the same query. The SHACL migration carried over write operations (setter/adder/remover) but dropped the read operation. Without a shape-stored getter, the JS runtime was generating conformance filters ad-hoc at query time — invisible to Rust/MCP. Storing getter + conditions in the shape restores universal visibility.

### 41. Decorator validation hardening + default `through` + `sh:class`

`getter` is now mutually exclusive with both `through` and `target` — providing both throws at class-definition time. `through` defaults to `ad4m://has_child` when omitted (enabling `@HasMany(() => T)` shorthand). `sh:class` references the target model's shape URI when `target` is set, regardless of `filter` — stored as a `sh://class` link in the SHACL shape and round-tripped through all serialization paths.

**Why:** `getter` + `target` is contradictory (manual read vs auto-generated read). `getter` + `through` makes a relation read-only without making that explicit. Defaulting `through` eliminates boilerplate for the common parent-child case. `sh:class` gives Rust/MCP/CLI consumers a typed reference to the target shape — without it, they can filter but cannot construct typed instances.

### 42. `where` clause on relation decorators

Added `where?: Where` to `RelationOptions`, using the same query DSL as `Model.query().where(...)`. Compiles to a SurrealQL getter at both SHACL generation time (for Rust/MCP) and JS runtime (in `evaluateCustomGettersForInstance`). Property names in `where` resolve to predicates via target model metadata; raw predicate URIs work when no target is set. Supports equality, `not`, and array membership (`IN`). Mutually exclusive with `getter` and `filter: false`. Priority chain: explicit `getter` → `where`-compiled → auto-conformance from target shape.

New file `surreal-utils.ts` extracts `formatSurrealValue()`, `buildWhereCondition()`, and `compileWhereClause()` as shared utilities used by both the decorator layer and the query builder.

**Why:** The existing conformance filter only checks structural properties (flags + required fields). `where` enables value-based filtering — e.g. `@HasMany(() => Task, { where: { status: "active" } })` — using syntax developers already know from the query API, without falling back to raw SurrealQL `getter` strings.

---

## Files touched

| File | What changed |
|------|-------------|
| `core/src/model/decorators.ts` | All decorator definitions, WeakMap registries, `resolveRelationArgs` helper, `@Flag` auto `readOnly`/`writable`, relation `through` validation, `effectiveInitial`, parent SHACL detection |
| `core/src/model/Ad4mModel.ts` | Reads from WeakMap getters, imports new helpers, updated `fromJSONSchema`, flag-skip in `innerUpdate`, Flag-specific error in setter gen, `getPropertyMetadata` returns `PropertyMetadataEntry`; `getRelationOptions` unified to `RelationMetadataEntry`; `fromJSONSchema` uses `setRelationRegistryEntry` |
| `core/src/model/Ad4mModel.test.ts` | Updated to use `@HasMany`, `readOnly` assertions, new API |
| `core/src/perspectives/PerspectiveProxy.ts` | `findClassByProperties` reads from WeakMap getters; `getCollectionsMetadata` calls replaced with filtered `getRelationsMetadata` |
| `ad4m-hooks/helpers/src/factory/SubjectRepository.ts` | Transform lookup via `getPropertiesMetadata()`; `.baseExpression` → `.id` |
| `tests/js/tests/model/` | New integration test suite — model-core, model-inheritance, model-query, model-transactions, model-where-operators |
| `tests/js/helpers/` | Shared test helpers — assertions, executor, ports |
| `ad4m-hooks/helpers/src/factory/model.ts` | `collectionToSetterName` → `set<Prop>` |
| `ad4m-hooks/react/src/useLiveQuery.ts` | New file — renamed from `useModel.ts`, positional-arg API, single-instance mode, `ParentScope` support, `useCallback` subscribe, effect cleanup, removed `ensureSDNASubjectClass` |
| `ad4m-hooks/vue/src/useLiveQuery.ts` | New file — renamed from `useModel.ts`, positional-arg API, single-instance mode, `ParentScope` support, `isRef()`, `onUnmounted` cleanup, removed `includeIds` hack and `ensureSDNASubjectClass` |
| `tests/js/tests/prolog-and-literals.test.ts` | `Collection` → `HasMany`, `where: { isInstance }` → `target: () => X`; `.baseExpression` → `.id` |
| `tests/js/tests/multi-user-simple.test.ts` | Removed `writable: true` (now default) |
| `core/src/shacl/SHACLShape.ts` | `parentShapes[]`, `addParentShape()`, `sh:node` in `toTurtle`, `parent_shapes` in `toJSON`/`fromJSON`; `getter`, `conformanceConditions`, `class` on `SHACLPropertyShape`; `ConformanceCondition` type; serialization through `toLinks`/`fromLinks`/`toJSON`/`fromJSON` |
| `core/src/model/util.ts` | `collectionToSetterName` / `collectionSetterToName` → `set<Prop>` convention |
| `core/src/model/Subject.ts` | Collection → relation terminology updates |
| `rust-executor/src/perspectives/perspective_instance.rs` | Batch pruning: scan+remove pending additions on `setSingleTarget`, `collectionSetter`, `removeLink` |
| `rust-executor/src/perspectives/shacl_parser.rs` | `ConformanceCondition` struct; `getter`, `conformance_conditions`, `class` fields on `PropertyShape` |
| `rust-executor/src/mcp/shacl.rs` | `getter` + `class` fields on `ShaclProperty`; reads `ad4m://getter` and `sh://class` links in `load_class_properties_with_uri()` |
| `rust-executor/src/mcp/tools/dynamic.rs` | `handle_dynamic_get()` executes conformance getter via `surreal_query()` instead of naive `get_links` when `ad4m://getter` is present; handles both collection and scalar properties |
| `tests/js/.mocharc.cjs` | New mocharc config; spec glob managed by CLI args |
| `tests/js/tests/model/model-from-json-schema.test.ts` | New: JSON schema model construction tests |
| `tests/js/tests/model/model-getters.test.ts` | New: getter/computed property tests |
| `tests/js/tests/model/model-prolog.test.ts` | New: Prolog fact generation tests |
| `tests/js/tests/model/model-subscriptions.test.ts` | New: subscribe/countSubscribe/paginateSubscribe tests |
| `tests/js/tests/model/models.ts` | New: shared test model definitions |
| `tests/js/tests/model/hooks.ts` | New: shared before/after hooks for model integration tests |
| `core/src/model/surreal-utils.ts` | New: shared SurrealQL compilation utilities — `formatSurrealValue`, `buildWhereCondition`, `compileWhereClause` |
| `core/src/model/relation-filtering.test.ts` | New: 40 unit tests for relation filtering — conformance filter, SHACL serialization, `sh:class`, `where` clause, validation |

---

## Test changes

**Unit tests** — Three existing test files updated (+126 / −135): `Ad4mModel.test.ts`, `prolog-and-literals.test.ts`, `multi-user-simple.test.ts`. Most changes are mechanical renames tracking the API changes above.

**New integration test suite** — Eight new test files added under `tests/js/tests/model/` (+3600 lines): `model-core`, `model-from-json-schema`, `model-getters`, `model-inheritance`, `model-prolog`, `model-query`, `model-subscriptions`, `model-transactions`, `model-where-operators`. Shared fixtures in `models.ts` and `hooks.ts`; shared helpers in `tests/js/helpers/` (assertions, executor, ports).

Key themes across all test changes:

**Decorator renames** — `@ModelOptions` → `@Model`, `@Collection` → `@HasMany`, imports updated. Applies to all model definitions across all three files. (§1)

**`@InstanceQuery` → fluent query API** — `@InstanceQuery` decorator calls on `Todo`, `Message` etc. replaced with static methods using `Model.query(perspective).where({...}).get()`. (§16)

**`where` clause removal** — `where: { isInstance: Target }` → `target: () => Target` on `@HasMany`; `where: { condition: "..." }` → top-level `condition: "..."`. Test names and assertions updated to reflect the new decorator shape. (§26)

**`Query.collections` removal** — All tests using `{ collections: [...] }` removed. The "findAll() works with collections query" test replaced with a simpler "findAll() returns all relations on instances" test (verifies all relations are always returned). Combined where/properties/collections assertions simplified to where/properties only. (§22)

**`instance.update()` → `instance.save()`** — 10 occurrences of `instance.update()` replaced with `instance.save()` across integration tests. (§24)

**Constructor 3-arg → 2-arg + parent link** — Tests that used the old 3-arg `new Model(perspective, source, id)` constructor rewritten to use 2-arg `new Model(perspective, id)` with a manual parent link via `perspective.add()`. (§21)

**`out.uri` → `fn::parse_literal(out.uri)`** — 12 `queryToSurrealQL` assertions updated. The query builder now wraps value comparisons with `fn::parse_literal()` so literal-encoded URIs are parsed before comparison. This is the only non-obvious change — everything else follows directly from the API renames.

**Dropped defaults** — Explicit `writable: true` removed from 6 properties across integration and multi-user tests (now the default). `readOnly` assertions replace `writable` assertions in unit tests. (§2, §3) Explicit `required: true` added to `queryToSurrealQL` test models that depend on required-property query filters, since the default flipped to `required: false`. (§38)

**`.baseExpression` → `.id`** — Three usages in integration tests updated. (§13)

**Prolog/SDNA tests deferred** — The SDNA generation test exercises the legacy Prolog engine which is being removed in a follow-up PR; it is `.skip`'d rather than rewritten.

**Removed tests** — Five tests removed entirely (not skipped):
- 3 `isInstance` filtering tests (`should filter collection by isInstance with class reference`, `…with string class name`, `should filter results in findAll() by isInstance`) — exercised §26's removed `isInstance` collection filtering.
- 2 `where` clause condition tests (Prolog-based and SurrealDB-based) — `condition` and `prologCondition` removed from `RelationOptions` and `CollectionOptions`; condition filtering is a query concern, not a schema concern.
- `Ad4mModel.test.ts` collection metadata test updated to remove `prologCondition` test case.

**Bug fixes surfaced by tests** — Two core bugs identified and fixed during test triage:
- `effectiveInitial` was applying to `@Optional` properties, causing them to return `""` instead of `undefined`. Fixed by gating on `required` in both `generateSDNA()` and `generateSHACL()`. (§9)
- `@HasMany({ getter })` was not forwarding `getter` to the `CollectionOptions` registration, so `evaluateCustomGettersForInstance()` never found it. Fixed in both `@HasMany` and `@BelongsToMany`.

**`where: { base }` → `where: { id }`** — All `where: { base: "..." }` unit test cases in `Ad4mModel.test.ts` updated to `where: { id: "..." }` following the removal of the `base` special-field alias.

**Relation filtering tests** — 40 unit tests in new `relation-filtering.test.ts`: `buildConformanceFilter()` derivation (4), SHACL getter serialization round-trips (3), `generateSHACL()` getter population (5), `getModelMetadata()` relation fields (5), decorator validation — `getter`+`target`/`through` mutual exclusivity (3), default `through` predicate (2), `sh:class` target shape reference (5), `where` clause validation (2), `where` compilation — equality/not/array/no-target/override/sh:class coexistence/round-trip (8), `compileWhereClause()` unit tests (4). Integration test in `model-getters.test.ts` — `TaskBoard` with `where: { status: "active" }` creates 3 tasks, asserts filtered relation returns only the 2 active ones. Fixed `getter`+`through` conflicts in `model-getters.test.ts` and `prolog-and-literals.test.ts` (removed `through` from `getter`-only decorators).

**Subscription callback initial-fire fixes** — Seven subscription tests in `prolog-and-literals.test.ts` updated to account for `subscribe()`, `countSubscribe()`, and `paginateSubscribe()` firing the callback immediately with initial results (a deliberate BehaviorSubject-style design). Tests that used `expect(updateCount).to.equal(0)` after subscribe now reset the counter (`updateCount = 0`), and tests using `sinon.fake()` wait for `callCount >= 2` instead of `called` to distinguish the initial invocation from a real data-change update.

**Subject class filtering fix** — The "query builder should filter by subject class" test added `required: true` to the distinguishing `content1`/`content2` properties on `Note1`/`Note2`. Without this, both models had only optional properties and identical link structures after save (empty strings are skipped by `setProperty`), so the structural query fallback matched both. `required: true` creates sentinel links that give each model a unique predicate fingerprint.

**`$perspective` removal** — All SurrealQL queries contained `perspective = $perspective AND` filter clauses referencing a bind variable that doesn't exist in the current query engine. Removed ~21 occurrences across `queryToSurrealQL`, `countQueryToSurrealQL`, `buildGraphTraversalWhereClause`, and subscription paths.

**`FETCH links` addition** — `queryToSurrealQL` SELECT now appends `FETCH links` so `->link AS links` returns full link records instead of bare record IDs, fixing hydration in `instancesFromSurrealResult`.

**`hasConstructor`/`hasDestructor` initial-property fix** — `save()` now includes properties with `initial` values (not just `required` ones) when deciding whether to call `createSubject`, ensuring SHACL constructor/destructor links are created for all properties that declare an initial sentinel.

## Rust changes

**SHACL getter + conformance conditions in Rust** — `shacl_parser.rs` gains `ConformanceCondition` struct and three new fields on `PropertyShape`: `getter`, `conformance_conditions`, `class`. `mcp/shacl.rs` gains `getter` and `class` on `ShaclProperty` and reads `ad4m://getter` and `sh://class` links in `load_class_properties_with_uri()`. `mcp/tools/dynamic.rs` — `handle_dynamic_get()` now checks for a conformance getter on each property shape; when present, it executes the SurrealQL getter via `surreal_query()` (replacing the `Base` placeholder with the instance URI) instead of naive `get_links(predicate)`. This covers both collection and scalar properties, giving MCP clients the same conformance-filtered results as the JS `Ad4mModel` runtime.

**Batch store pruning in `perspective_instance.rs`** — The `RemoveLink`, `SetSingleTarget`, and `CollectionSetter` SDNA action handlers now prune matching links from pending batch additions (`batch_store`) in addition to removing them from the persisted store. Previously, when a batch transaction added a link and then a later action in the same batch removed or replaced it, the removal only hit the persisted DB — the earlier addition survived in the in-flight `batch_store.additions` list and was re-committed when the batch finalized, creating duplicates. The fix adds a `diff.additions.retain(...)` pass after each persisted-link removal, filtering out any queued additions that match the same source/predicate/target. This ensures that save-then-update (or save-then-delete) within a single `Ad4mModel.transaction()` produces the correct final state. Minor ownership change: in the `RemoveLink` arm, `source`, `predicate`, and `target` are cloned before being moved into the `LinkQuery`, so they remain available for the subsequent batch-store filter.


core/src/model/Ad4mModel.ts
```diff
@@ -408,10 +472,10 @@ function isNumericType(schema: JSONSchemaProperty): boolean {
  * ```
  */
 export class Ad4mModel {
-  #baseExpression: string;
-  #subjectClassName: string;
-  #source: string;
-  #perspective: PerspectiveProxy;
+  private _id: string;
```
lucksus: For the inner, private scope, I think `baseExpression` would still be the better variable name, since ID could be misunderstood as just some string, but it has to be an expression URI (now even enforced by the executor). Not dying on this hill, but we could have JS properties `id` and `baseExpression` which are the public interface and both read this variable, which internally would be called `baseExpression`

core/src/model/Ad4mModel.ts
```diff
@@ -586,76 +649,91 @@ export class Ad4mModel {
...
-   * 
-   * // Create with source link
-   * const recipe = new Recipe(perspective, undefined, "cookbook://desserts");
    * ```
    */
-  constructor(perspective: PerspectiveProxy, baseExpression?: string, source?: string) {
-    this.#baseExpression = baseExpression ? baseExpression : Literal.from(makeRandomPrologAtom(24)).toUrl();
-    this.#perspective = perspective;
-    this.#source = source || "ad4m://self";
+  constructor(perspective: PerspectiveProxy, id?: string) {
```
lucksus: this is another place... again, I think baseExpression in the constructor makes more sense than ID. because, you don't provide an ID before it exists... I'm all for having `myModel.id` to work.. but if I would read ID and it's a string, I would not expect it to not work with `xyz123`...

lucksus: An alternative could be to use an URI type instead of string, which I guess would be a good step independent of the name choice.

tests/js/tests/prolog-and-literals.test.ts
```diff
@@ -359,35 +361,10 @@ describe("Prolog + Literals", () => {
...
-
-                    let message = new Message(perspective!, messageEntry)
-                    await message.save()
-                    
-                    // Refresh todo data to apply collection filtering
-                    await todo.get()
-                    messageEntries = await todo.messages
-                    expect(messageEntries.length).to.equal(1)
-                })
-
```
lucksus, line 386: This test can and should stay in.

tests/js/tests/prolog-and-literals.test.ts
```diff
@@ -574,75 +551,9 @@ describe("Prolog + Literals", () => {
...
-                    // Should have 2 entries total
-                    expect(recipe2.entries.length).to.equal(2);
-                    
-                    // But only 1 ingredient (entry1 which has the ingredient link)
-                    expect(recipe2.ingredients.length).to.equal(1);
-                    expect(recipe2.ingredients[0]).to.equal(entry1);
-                })
+                // REMOVED: Collection 'where' clause condition tests (both Prolog and SurrealDB)
+                // Condition filtering on relations is a query concern, not a schema concern.
+                // Use findAll({ where: { ... } }) or perspective.get() for conditional filtering.
```
lucksus: The point of SocialDNA (i.e. models) is to have an abstraction layer that hides internal semantics of the model classes. If the model needs to run a query / constrain a set of entries, that should stay inside the model (== must be in the SHACL definition for other middlewares like the MCP server to be able to construct the same model instances)

lucksus: We might be able to readd the test with this: 
```JS
@HasMany({
  through: "recipe://entries",
  getter: `(<-link[WHERE predicate = 'recipe://has_ingredient' AND out.uri = 'recipe://test'].in.uri)`
})
ingredients: string[] = [];
```

tests/js/tests/prolog-and-literals.test.ts
```diff
@@ -2735,218 +2647,9 @@ describe("Prolog + Literals", () => {
...
-                    expect(foundArticle1!.comments).to.have.lengthOf(1);
-                    expect(foundArticle1!.comments[0]).to.equal(comment1);
-                    
-                    expect(foundArticle2!.comments).to.have.lengthOf(1);
-                    expect(foundArticle2!.comments[0]).to.equal(comment2);
-                });
-            })
+            // REMOVED: isInstance filtering tests
+            // isInstance filtering was removed (§26) — collections now return all linked
+            // targets without checking if they're valid instances of the target class.
```
lucksus: `isInstance` works now with `target: () => Class`. These tests can and should be put back in.

tests/js/tests/model/model-getters.test.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  @Model({ name: "BlogPost" })
+  class BlogPost extends Ad4mModel {
+    @Property({ through: "blog://title" })
+    title: string = "";
+
+    @Property({
+      through: "blog://parent",
+      getter:
+        "(->link[WHERE perspective = $perspective AND predicate = 'blog://reply_to'].out.uri)[0]",
```
lucksus:
```suggestion
        "(->link[WHERE predicate = 'blog://reply_to'].out.uri)[0]",
```

core/src/model/Ad4mModel.ts
```diff
@@ -104,41 +162,87 @@ export interface PropertyMetadata {
...
  */
-export interface CollectionMetadata {
-  /** The collection name */
+export interface RelationMetadata {
+  /** The relation name */
   name: string;
   /** The predicate URI (through value) */
   predicate: string;
-  /** Filter conditions */
-  where?: { isInstance?: any; prologCondition?: string; condition?: string };
```
lucksus, line 115: So this should be added back, as mentioned in my other comment

core/src/model/Ad4mModel.ts
```diff
@@ -104,41 +162,87 @@ export interface PropertyMetadata {
...
+ *   - With `field`: direct key lookup
+ *   - Without `field`: scan for a relation whose `target()` matches `childCtor`
+ */
+function resolveParentPredicate(
+  parent: ParentScope,
+  childCtor: Function,
+): string {
+  // Raw form — explicit predicate
+  if ('predicate' in parent) return parent.predicate;
+
```
lucksus, line 223: Here we could do: `if(parent is empty) return "ad4m://has_child" (or similiar, we would need to reverse the direction...)

Only relevant for the follow-up work.

core/src/model/Ad4mModel.ts
```diff
@@ -2250,158 +2849,233 @@ WHERE ${whereConditions.join(' AND ')}
...
+    // Existing instance → update path (has been fetched / hydrated before)
+    if (this._snapshot) {
+      await this.innerUpdate(true, batchId);
+      if (batchId) {
+        // Batch hasn't been committed yet — getData() would fetch stale data
+        // and overwrite in-memory values. Just refresh the snapshot so further
+        // saves within the same batch diff correctly.
+        this.takeSnapshot();
+      } else {
+        await this.getData();
```
lucksus, line 3077: Why are we doing getData() here in save()?!

core/src/model/Ad4mModel.ts
```diff
@@ -2509,7 +3202,262 @@ WHERE ${whereConditions.join(' AND ')}
    * ```
    */
   async delete(batchId?: string) {
-    await this.#perspective.removeSubject(this, this.#baseExpression, batchId);
+    // Check if the model has a destructor (required properties, flags, or
+    // properties with initial values).  Models whose properties are all
+    // optional, have no @Flag, and have no initial values produce an empty
+    // SHACL destructor, so calling removeSubject would fail.  In that case
+    // we skip the destructor and just remove all outgoing links directly.
```
lucksus: That sounds risky. In the case of two different classes instantiated on the same base expression, and one being very generic one, do we really want to remove all links? it would break the other instance on that node. We could say, maybe that is what we want, but then that instance could be a complex type, that does have a destructor. But that destructor can't be called anymore when the instance is not fully there anymore... so we might be left with just some debris of links. I think delete should fail on an all-optional class.

OR: just remove those optional links if they are there - but nothing else

lucksus (commented):
🔥 🔥 🔥 
Amazing work

tests/js/tests/prolog-and-literals.test.ts
```diff
@@ -551,9 +551,69 @@ describe("Prolog + Literals", () => {
...
+                            resolveLanguage: "literal"
+                        })
+                        name: string = "";
+
+                        @HasMany({ through: "recipe://entries" })
+                        entries: string[] = [];
+
+                        @HasMany({
+                            through: "recipe://filtered_ingredients",
+                            getter: `(->link[WHERE predicate = 'recipe://entries'].out[WHERE count(->link[WHERE predicate = 'recipe://has_ingredient' AND out.uri = 'recipe://test']) > 0].uri)`
```
lucksus, line 600: why not `target: () => Ingredient`?

tests/js/tests/prolog-and-literals.test.ts
```diff
@@ -2647,9 +2707,235 @@ describe("Prolog + Literals", () => {
...
+                class Article extends Ad4mModel {
+                    @Property({ 
+                        through: "article://title",
+                        resolveLanguage: "literal"
+                    })
+                    title: string = "";
+
+                    @HasMany({
+                        through: "article://typed_comments",
+                        getter: `(->link[WHERE predicate = 'article://has_comment'].out[WHERE count(->link[WHERE predicate = 'ad4m://type' AND out.uri = 'ad4m://comment']) > 0].uri)`
```
lucksus: `target: () => Comment`

core/src/model/Ad4mModel.ts
jhweir: I asked Claude to calrify:

`getData()` re-hydrates the instance from the DB after the update. This is necessary because:

**`resolveLanguage` properties**: `setProperty` calls `createExpression` and writes the resulting URI to the perspective, but never updates the in-memory value. Without `getData()`, the instance holds stale raw data, causing every subsequent `save()` to see the field as dirty and re-write it.

**Computed fields**: `author`, `createdAt`, `updatedAt`, and custom SurrealQL getters are derived from link metadata — they can only be correct after re-reading.

**Reverse relations**: `belongsToOne`/`belongsToMany` fields point to this instance from other nodes and can't be resolved from outgoing links alone.

**Snapshot accuracy**: `takeSnapshot()` at the end of `getData()` records the persisted state as the dirty-tracking baseline, ensuring subsequent `changedFields()` diffs are correct.

The batch path skips `getData()` (calls `takeSnapshot()` only) because the batch hasn't been committed yet — reading would return stale data. This is safe within a batch but would cause issues 1–3 above if used as the permanent non-batch path.

lucksus: Ah, it's after the update. I was confused and misread the diff at the late hour last night and though it was before the bulk of the function. I see the actual update is before this getData() call, with innerUpdate() 👍

core/src/model/Ad4mModel.ts
```diff
@@ -3359,18 +3354,40 @@ FETCH links
...
       } catch (e) {
         console.warn(`delete(): failed to remove outgoing links for ${this._baseExpression}:`, e);
       }
     }
 
-    // Clean up incoming links — remove any links that point **to** this instance
+    // Clean up incoming links — remove any links that point **to** this instance.
+    // Unlike outgoing links (scoped above), incoming links originate from OTHER
+    // nodes.  Removing them doesn't damage sibling models on this node; it
+    // prevents dangling references elsewhere (e.g. a parent's has_child link).
```
lucksus, line 3367: but if there is another model instantiated, this would elimnate their links still

lucksus: we can sort that out when we add removing of channels and stuff in flux

lucksus (approved):

Issue #719, opened 2026-03-06
feat: Iroh-based ICE transport to replace STUN/TURN for all WebRTC


HexaField:
---

## Context

Holochain now uses **Iroh** (from n0/number0) as its networking transport. Every Holochain agent already runs an Iroh endpoint — a QUIC-based, end-to-end encrypted connection layer with automatic NAT traversal (holepunching) and relay fallback.

AD4M has two WebRTC systems that both depend on centralised STUN/TURN servers:

1. **Full-mesh P2P calls** — the existing system where every peer connects directly to every other peer
2. **Embedded SFU** ([#700](https://github.com/coasys/ad4m/issues/700)) — the new str0m-based SFU in the executor, with designated peer, gateway, cascaded, and mesh fallback modes

Both systems need the same thing from the network layer: discover peer addresses, punch through NATs, and relay when holepunching fails. Both currently use external STUN/TURN for this.

**Nico's insight:** since every AD4M agent already has Iroh running underneath Holochain, we can build a single, generic ICE transport layer on top of Iroh that replaces STUN/TURN for *all* WebRTC connections in AD4M — whether that's a 3-person mesh call or a 50-person cascaded SFU conference.

---

### What STUN does → Iroh already does

STUN's job is to discover a peer's public IP and port (its "reflexive address") so that another peer can send packets to it. Iroh does exactly this:

- Each Iroh endpoint connects to a **relay server** on startup
- The relay observes and reports the endpoint's **public IP + port** (reflective address)
- Iroh also uses **DHT discovery** and **DNS discovery** to publish/resolve endpoint addresses
- Peers exchange address info through the relay and attempt **simultaneous holepunching**

Every Holochain agent already has this information. No separate STUN query needed.

### What TURN does → Iroh relays already do

TURN exists as a last resort: when holepunching fails (~10% of networks), media is relayed through a server. Iroh relays serve this exact role:

- Traffic through Iroh relays is **end-to-end encrypted** (the relay is a blind courier)
- Relays are **stateless** — easy to scale, no session affinity
- Holochain/AD4M can run its own relay infrastructure or use Iroh's public relays
- ~90% of connections holepunch successfully; the relay is fallback only

---

## Architecture

WebRTC stays for media transport (browser compatibility, hardware-accelerated encode/decode, `getUserMedia`, `<video>` elements). Iroh replaces the STUN/TURN layers as a **generic ICE transport** that any WebRTC connection in AD4M can use.

[...]

Issue #700, opened 2026-02-27
feat: Embedded SFU for scalable WebRTC conferencing in neighbourhoods


data-bot-coasys:
## Problem

ADAM's WebRTC conferencing uses a full mesh topology — each participant maintains a direct connection to every other participant. Connection count grows quadratically:

| Participants | Connections per peer | Total connections |
|---|---|---|
| 2 | 1 | 2 |
| 4 | 3 | 12 |
| 6 | 5 | 30 |
| 8 | 7 | 56 |

At 6-8 participants, each peer must encode and upload their media stream N-1 times. Bandwidth and CPU requirements hit a hard ceiling. This was observed in practice during a recent 8-person call.

## Proposed Solution: Embedded SFU (Selective Forwarding Unit)

An SFU receives each participant's media stream **once**, then selectively forwards it to all other participants. Each peer only uploads once regardless of group size. This is the standard architecture used by Jitsi, LiveKit, mediasoup, etc.

### Key Design: SFU in the ADAM Executor

Rather than requiring external infrastructure, embed SFU capability directly in the ADAM executor:

- Every ADAM agent gains SFU capability by default
- No external service dependency
- Shares the executor's async runtime, identity system, and neighbourhood membership
- The executor already manages networking and auth — the SFU piggybacks on this

**Recommended library: [`str0m`](https://github.com/algesten/str0m)** — a Sans I/O Rust WebRTC implementation with an explicit [SFU example](https://github.com/algesten/str0m/blob/main/examples/chat.rs). Sans I/O fits well with the executor's architecture — no internal threads or hidden async tasks, all operations driven by the caller. `webrtc-rs` is an alternative but is heavier.

### Designated SFU Peer per Neighbourhood

Not every peer in a call acts as the SFU. The neighbourhood designates one peer:

1. **Cloud Gateway** — for gateway-connected neighbourhoods, the gateway acts as default SFU (always-on, server-grade bandwidth)
2. **Designated peer** — neighbourhood admin sets `sfu_peer` DID in Social DNA. Simplest self-sovereign model.
3. **Mesh fallback** — if SFU unavailable or ≤4 participants, fall back to direct mesh (current behaviour, zero regression)

### Call Flow

```
Initiator creates call link in neighbourhood
  → Peers query neighbourhood for SFU peer
    → If SFU available: each peer connects once to SFU
    → SFU forwards streams to all peers
    → If SFU unavailable + ≤4 peers: mesh fallback
[...]

PR #741, opened 2026-03-11
feat: add iroh-ice module for WebRTC ICE candidate extraction

HexaField:
## Summary

Adds a complete Iroh-to-ICE bridge that exposes Holochain's transport-discovered socket addresses as WebRTC ICE candidates, replacing the need for external STUN servers. Provides both the low-level module and the full GraphQL API + TypeScript client integration.

#### `rust-executor/src/iroh_ice/mod.rs` — Core Module
- `IrohIce` struct: takes socket addresses, produces RFC 8445 compliant ICE candidates
- Classifies addresses as `host` (private/LAN) or `srflx` (public/reflexive)
- Computes priorities per RFC 8445 Section 5.1.2.1
- Formats candidate strings for `RTCPeerConnection.addIceCandidate()`
- Zero external dependencies (pure `std::net`)
- 9 unit tests

#### `rust-executor/src/graphql/` — GraphQL API
- `IceCandidateInfo` GraphQL type with `candidate`, `candidateType`, `address`, `port`
- `runtimeIceCandidates` query: fetches socket addresses from `conductor.holochain_p2p().local_socket_addrs()` and converts them via `IrohIce`

#### `core/src/runtime/` — TypeScript Client
- `IceCandidateInfo` class in `RuntimeResolver.ts`
- `runtime.iceCandidates()` client method in `RuntimeClient.ts`

### Usage

```typescript
// In Flux or any AD4M client
const candidates = await ad4mClient.runtime.iceCandidates()
// Returns: [{ candidate: 'candidate:1 1 udp 2130706431 192.168.1.5 12345 typ host', candidateType: 'host', address: '192.168.1.5', port: 12345 }, ...]
```

### Flux Integration (separate PR)

The `packages/webrtc/src/WebRTCManager.ts` in Flux currently uses hardcoded STUN servers (`stun:relay.ad4m.dev:3478`, `stun:stun.l.google.com:19302`, etc.). The integration path:

1. On `WebRTCManager.init()`, call `ad4mClient.runtime.iceCandidates()` to fetch Iroh-derived candidates
2. After each `RTCPeerConnection` is created (in `AD4MPeer.createPeer()`), inject the Iroh candidates via `pc.addIceCandidate()`
3. Optionally fall back to STUN servers if no Iroh candidates are available (graceful degradation)

This works for both the existing P2P mesh calls and the future SFU system (#700), since `iroh-ice` is transport-agnostic — it provides candidates that any WebRTC consumer can use.

### SFU Considerations

The `iroh-ice` module is deliberately generic — no knowledge of mesh vs SFU vs pipe transport. When the str0m SFU (#700) lands:
- SFU server: uses `IrohIce` to get its own ICE candidates to share with clients
- SFU clients: use the same `runtimeIceCandidates` API to get their local candidates
- The SFU negotiation adds its own signaling on top, but the ICE candidate source is shared

### Dependencies

- [holochain/kitsune2#476](https://github.com/holochain/kitsune2/pull/476) — `local_socket_addrs()` on Transport trait
- [coasys/holochain#1](https://github.com/coasys/holochain/pull/1) — surface through HcP2p trait

### Note

The `ad4m-client` crate has 327 pre-existing compile errors (GraphQL codegen issue) unrelated to this PR. The `ad4m-executor` crate and the `iroh-ice` module compile and test successfully.

See: #719

## Summary by CodeRabbit

* **New Features**
  * Public API and GraphQL query to fetch ICE candidate details (candidate, type, address, port).
  * Local ICE candidate generation exposed so clients can discover formatted candidate info and detect public vs private addresses for connectivity diagnostics.

### DIFF
--- a/core/src/runtime/RuntimeClient.ts
+++ b/core/src/runtime/RuntimeClient.ts
@@ -1,7 +1,7 @@
 import { ApolloClient, gql } from "@apollo/client/core"
 import { Perspective, PerspectiveExpression } from "../perspectives/Perspective"
 import unwrapApolloResult from "../unwrapApolloResult"
-import { RuntimeInfo, ExceptionInfo, SentMessage, NotificationInput, Notification, TriggeredNotification, ImportResult, UserStatistics } from "./RuntimeResolver"
+import { RuntimeInfo, IceCandidateInfo, ExceptionInfo, SentMessage, NotificationInput, Notification, TriggeredNotification, ImportResult, UserStatistics } from "./RuntimeResolver"
 
 const PERSPECTIVE_EXPRESSION_FIELDS = `
 author
@@ -76,6 +76,21 @@ export class RuntimeClient {
             }`,
         }));
         return runtimeInfo
+    }
+
+    async iceCandidates(): Promise<IceCandidateInfo[]> {
+        const { runtimeIceCandidates } = unwrapApolloResult(await this.#apolloClient.query({
+            query: gql`query runtimeIceCandidates {
+                runtimeIceCandidates {
+                    candidate
+                    candidateType
+                    address
+                    port
+                }
+            }`,
+            fetchPolicy: 'no-cache',
+        }))
+        return runtimeIceCandidates
     }
 
     async quit(): Promise<Boolean> {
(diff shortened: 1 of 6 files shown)


HexaField:
Closing to restart with spec-driven approach. Will reopen with comprehensive tests and cleaner implementation.

PR #744, opened 2026-03-11
fix: Test & CI infrastructure improvements — graceful shutdown, readiness probes, test isolation

HexaField:
## Summary

Comprehensive test and CI infrastructure improvements across three phases, addressing process cleanup reliability, graceful shutdown, readiness probing, and multi-executor test tooling.

Based on the proposal at `membranes/adam/proposals/test-ci-improvements.md`.

### Phase 1: Quick Wins
- **integration.bats**: Replace `killall ad4m` with PID-based SIGTERM→wait→SIGKILL escalation
- **CI artifacts**: Add `store_artifacts` step for failure log collection
- **Port validation**: Pre-test port checks to catch stale processes early
- **Orphan cleanup**: All CI jobs use SIGTERM→wait→SIGKILL instead of raw `kill -9`

### Phase 2: Graceful Shutdown (Rust)
- **Signal handlers**: SIGTERM/SIGINT handlers via `tokio::signal` in rust-executor
- **runtime_quit**: Uses oneshot channel → Holochain conductor shutdown → clean exit (replaces `process::exit(0)`)
- **--pid-file flag**: CLI flag for test harnesses to track executor PIDs
- **JS gracefulShutdown()**: Shared utility replacing ad-hoc kill loops in after() hooks

### Phase 3: Readiness & Multi-Executor
- **runtimeReadiness GQL query**: Returns `{ gqlReady, holochainReady, agentInitialized, languagesLoaded }` for polling-based readiness checks
- **TestCluster utility**: `tests/js/utils/testCluster.ts` — manages N executor instances with startup, readiness polling, and graceful shutdown

### CI Status
⚠️ **CI is disabled** (`when: never` on workflow) pending review. These changes are backward-compatible but should be reviewed before enabling CI runs.

@nicolasluck — please review when you get a chance. The CI runner is intentionally disabled on this branch.

## Summary by CodeRabbit

* **New Features**
  * Graceful shutdown via OS signals and in-app trigger; writes/removes PID file on start/stop.
  * Runtime readiness endpoint reporting service initialization and readiness.
  * Test utilities: multi-node test cluster orchestration and centralized gracefulShutdown helper.

* **Improvements**
  * CI: always collect test artifacts/logs on failure and prefer graceful shutdown with port-based fallback.
  * Cleaner test teardown across JS integration tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -123,15 +123,20 @@ jobs:
           no_output_timeout: 30m
       - run:
           name: Collect logs on failure
-          when: on_fail
+          when: always
           command: |
             mkdir -p /tmp/test-artifacts
-            cp -r tests/js/tst-tmp/agents/*/ad4m/logs /tmp/test-artifacts/ 2>/dev/null || true
-            cp -r tests/js/tst-tmp/agents/*/ad4m/holochain /tmp/test-artifacts/ 2>/dev/null || true
+            shopt -s nullglob
+            for ad4m_dir in tests/js/tst-tmp/agents/*/ad4m; do
+              agent_id="$(basename "$(dirname "$ad4m_dir")")"
+              mkdir -p "/tmp/test-artifacts/$agent_id"
+              cp -r "$ad4m_dir/logs" "/tmp/test-artifacts/$agent_id/" 2>/dev/null || true
+              cp -r "$ad4m_dir/holochain" "/tmp/test-artifacts/$agent_id/" 2>/dev/null || true
+            done
             # Capture port state for debugging
             lsof -i -P -n 2>/dev/null | grep LISTEN > /tmp/test-artifacts/listening-ports.txt || true
       - store_artifacts:
-          when: on_fail
+          when: always
           path: /tmp/test-artifacts
           destination: test-logs
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,37 +118,6 @@ jobs:
     steps:
       - setup_integration_test_environment
       - run:
-          name: Kill any orphaned executors from previous runs
-          command: |
-            # Self-hosted runners reuse workdirs; previous job may have left
-            # an executor alive. Use SIGTERM first for graceful shutdown,
-            # then escalate to SIGKILL after a timeout.
-            for port in 15700 15701 15702; do
-              PIDS=$(lsof -ti:$port 2>/dev/null || true)
-              if [ -n "$PIDS" ]; then
-                echo "$PIDS" | xargs -r kill -TERM 2>/dev/null || true
-                sleep 3
-                echo "$PIDS" | xargs -r kill -9 2>/dev/null || true
-              fi
-            done
-      - run:
-          name: Verify clean port state
-          command: |
-            CONFLICTS=""
-            for port in 15000 15100 15200 15300 15400 15600 15700 15800 15900 15920 16000 16010 16600; do
-              if lsof -ti:$port >/dev/null 2>&1; then
-                CONFLICTS="$CONFLICTS $port"
-              fi
-            done
-            if [ -n "$CONFLICTS" ]; then
-              echo "⚠️  Ports still in use:$CONFLICTS"
-              echo "Attempting cleanup..."
-              for port in $CONFLICTS; do
-                lsof -ti:$port | xargs -r kill -9 2>/dev/null || true
-              done
-              sleep 2
-            fi
-      - run:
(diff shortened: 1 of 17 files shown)


.circleci/config.yml
```diff
@@ -122,15 +122,49 @@ jobs:
...
+            for port in 15000 15100 15200 15300 15400 15600 15700 15800 15900 15920 16000 16010 16600; do
+              if lsof -ti:$port >/dev/null 2>&1; then
+                CONFLICTS="$CONFLICTS $port"
+              fi
+            done
+            if [ -n "$CONFLICTS" ]; then
+              echo "⚠️  Ports still in use:$CONFLICTS"
+              echo "Attempting cleanup..."
+              for port in $CONFLICTS; do
+                lsof -ti:$port | xargs -r kill -9 2>/dev/null || true
```
lucksus: not sure if verify ports are clean make sense here, because if they are not, just killing that process is asking for trouble.

.circleci/config.yml
```diff
@@ -161,7 +195,14 @@ jobs:
...
             # MCP tests use ports 16000-16002 (mcp-http) and 16010-16012 (mcp-auth)
-            # plus port 3001 (MCP HTTP server). Clear all before starting.
+            # plus port 3001 (MCP HTTP server). Graceful SIGTERM first, then SIGKILL.
+            for port in 16000 16001 16002 16010 16011 16012 3001; do
+              PIDS=$(lsof -ti:$port 2>/dev/null || true)
+              if [ -n "$PIDS" ]; then
+                echo "$PIDS" | xargs -r kill -TERM 2>/dev/null || true
+              fi
+            done
+            sleep 3
```
lucksus: yeah I think this killing just needs to go away completely. I hope the graceful shutdown below eradicates the need for that.

lucksus (changes requested):
I think the graceful shutdown was missing for a long time. great stuff! if you remove at least the new kill code, I would love to see it running on CI (remove the broken syntax `never`) and maybe we can remove more kill calls. then we can try activating more runners again.

.circleci/config.yml
HexaField: Done — removed the port verification step and all kill steps entirely. Graceful shutdown should handle cleanup. Also removed `when: never` so CI will run on this branch.

.circleci/config.yml
HexaField: Agreed — removed all kill steps from every test job (js, mcp, model). Relying entirely on graceful shutdown now. Let's see how CI goes!

HexaField:
CI compile fix is working — all 8 ad4m model integration tests pass.

The 2 reported failures are pre-existing artefacts on the `coasys/marvin` executor image, not from this repo:
- `test_sample.py`: deliberately broken (`assert 1 + 1 == 3`)
- `Storyshots › Job/Tests/Test Result`: CircleCI example at `/home/circleci/src/node-example/test/test.js`

Neither is related to this PR. ⬡

cli/src/ad4m_executor.rs
```diff
@@ -189,8 +193,17 @@ async fn main() -> Result<()> {
...
+        pid_file,
     } = args.domain
     {
+        // Set PID file path as env var so the executor can write/clean it up
+        if let Some(ref pf) = pid_file {
+            // SAFETY: set_var is safe here because we're in single-threaded init before spawning.
+            #[allow(deprecated)]
+            unsafe {
+                std::env::set_var("AD4M_PID_FILE", pf);
+            }
```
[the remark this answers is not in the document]
lucksus: @HexaField Is this a valid concern in this case? We just need this setting of the var to happen before the code in rust-executor/lib.rs reads it ... Or is there a better way to ensure this without removing the ergonomic `#[tokio::main(flavor = "multi_thread")]` derivation from main()?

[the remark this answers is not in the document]
HexaField: Good catch — you're right that with `#[tokio::main(flavor = "multi_thread")]`, the runtime's worker threads are already spawned before our `main()` body runs, so `set_var` isn't technically safe even though it works in practice (the only reader is our own code later in the same flow).

The proper fix is to add `pid_file: Option<String>` to `Ad4mConfig` and pass it through directly — no env var needed. The executor's PID write (lib.rs:258) and cleanup (lib.rs:241) would read from the config instead of `std::env::var`. I'll make that change.

⬡

lucksus (dismissed):

lucksus (approved):

PR #753, opened 2026-03-19
feat(plugin): auto-download ad4m-executor binary when not in PATH

HexaField:
## What

Adds auto-download of `ad4m-executor` and `ad4m` CLI client binaries when they're not found in PATH or common locations. This makes the OpenClaw AD4M plugin a one-step setup — just add the plugin config and it handles the rest.

## How it works

1. **On startup** (managed mode): if `findExecutorBinary()` returns null, calls `downloadExecutor()`
2. **Detects platform**: Linux x64, macOS x64/aarch64, Windows x64
3. **Downloads from GitHub releases**: follows redirects (GitHub → S3), logs progress with percentages
4. **Saves to** `~/.ad4m-plugin/bin/` with `chmod +x`
5. **Downloads both binaries**: `ad4m-executor` and `ad4m` CLI client

## Platform support

| Platform | Status |
|----------|--------|
| Linux x64 | ✅ Available now |
| macOS aarch64 | 🔜 Asset names ready, binaries in next release |
| macOS x64 | 🔜 Asset names ready, binaries in next release |
| Windows x64 | 🔜 Asset names ready (.exe), binaries in next release |

If a binary doesn't exist for the current release, the download fails gracefully with a clear error message.

## Details

- Version hardcoded to `0.12.0-rc1` — should be updated each release or made configurable
- `findExecutorBinary()` now also checks `~/.ad4m-plugin/bin/`
- No new dependencies — uses Node.js built-in `https`/`http` modules
- Handles network errors, disk write errors, and missing assets gracefully

## Summary by CodeRabbit

* **New Features**
  * Automatic local installation of the required native executable into a plugin bin location, with platform-aware executable naming, permission setup, and re-check before falling back. Updated messaging directs users to run the setup command to install/configure.

* **Bug Fixes / Reliability**
  * More robust download/installation: follows redirects, enforces timeouts, logs progress, rejects zero-byte assets, and surfaces install failures while treating auxiliary client fetches as non-fatal.

### DIFF
--- a/plugins/ad4m/executor.ts
+++ b/plugins/ad4m/executor.ts
@@ -1,5 +1,7 @@
 import path from "path";
 import fs from "fs";
+import https from "https";
+import http from "http";
 import { spawn, execFileSync } from "child_process";
 
 import type { PluginConfig } from "./types";
@@ -27,6 +29,7 @@ export function findExecutorBinary(): string | null {
   // 2. Check common locations not always in PATH
   const home = process.env.HOME || process.env.USERPROFILE || "";
   const commonPaths = [
+    path.join(home, ".ad4m-plugin", "bin"),
     "/usr/local/bin",
     "/usr/bin",
     "/opt/homebrew/bin",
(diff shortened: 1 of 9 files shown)


plugins/ad4m/executor.ts
```diff
@@ -36,18 +46,261 @@ export function findExecutorBinary(): string | null {
...
   return null;
 }
 
+// ---------------------------------------------------------------------------
+// Auto-download of ad4m-executor and ad4m CLI binaries
+// ---------------------------------------------------------------------------
+
+// Hardcoded version for auto-download.
+// TODO: Update this with each release or make it configurable via plugin config.
+const EXECUTOR_VERSION = "0.12.0-rc1";
```
lucksus:
```suggestion
const EXECUTOR_VERSION = "0.12.0-rc2";
```

lucksus (changes requested):
we just released rc2 :)

lucksus (changes requested):
actually this should not be done on plugin startup but in the CLI function `ad4m-setup` which our plugin already registers. @HexaField

lucksus (dismissed):
applied my own change request

plugins/ad4m/skills/ad4m/SKILL.md
```diff
@@ -41,7 +41,7 @@ After adding the config snippet and restarting OpenClaw, the plugin auto-manages
 
 The AD4M OpenClaw plugin bridges AD4M's MCP server into your tool list automatically. Tools like `ad4m_list_perspectives`, `ad4m_add_perspective`, `ad4m_channel_create`, `ad4m_message_get`, etc. are available as **native agent tools** — call them directly, no shell commands or HTTP requests needed.
 
-The plugin is configured in OpenClaw's config under `plugins.entries.ad4m-openclaw-plugin.config`. Run `openclaw ad4m-setup` to generate the config — key fields:
+The plugin is configured in OpenClaw's config under `plugins.entries.ad4m.config`. Run `openclaw ad4m-setup` to generate the config — key fields:
```
HexaField, line 44: this is not correct, please revert

PR #758, opened 2026-03-20
Feat/credit subscription

jhweir:
# PR: Real-time Credit Updates via GraphQL Subscription

## Summary

Replaces the 30-second credit polling mechanism with a GraphQL subscription that pushes credit balance updates to the client in near real-time (~2s latency). Uses a batched dirty-flag pattern to avoid flooding the subscription during high-frequency AI token deductions.

## Architecture

```
Credit mutation (any path)
    │
    ▼
mark_credits_dirty(email)    ← Mutex<HashSet<String>>, per-user
    │
    ▼
Background flush loop (2s)   ← tokio::spawn, drains dirty set
    │
    ▼
Publish HostingUserInfo      ← only for affected user(s)
    │
    ▼
subscribe_and_process filter ← matches on email (GetFilter)
    │
    ▼
GraphQL subscription         ← runtimeHostingUserInfoChanged
    │
    ▼
Core AgentClient callback    ← addHostingUserInfoChangedListener
    │
    ▼
Connect dispatches events    ← userinfochange / creditlow / creditdepleted
```

### 1. Server-side subscription (`ae6ee94d`)

**rust-executor/src/pubsub.rs**
- Added `HOSTING_USER_INFO_CHANGED_TOPIC` topic constant
- Added `DIRTY_CREDIT_USERS: LazyLock<Mutex<HashSet<String>>>` for per-user dirty tracking
- Added `mark_credits_dirty(email: &str)` helper that inserts the user's email into the dirty set

**rust-executor/src/graphql/subscription_resolvers.rs**
- Added `runtimeHostingUserInfoChanged` subscription resolver
- Uses `RUNTIME_HOSTING_READ_CAPABILITY` for access control
- Passes authenticated user's email as filter so each subscriber only receives their own updates

**rust-executor/src/graphql/graphql_types.rs**
- Implemented `GetValue` and `GetFilter` traits for `HostingUserInfo`
- `GetFilter` returns `Some(self.email.clone())` for per-user subscription filtering

**rust-executor/src/graphql/mutation_resolvers.rs**
- `reserve_compute_credits()` — marks user dirty after deduction
- `runtime_set_user_credits()` — marks user dirty after admin set
- `runtime_set_user_free_access()` — marks user dirty after toggle

**rust-executor/src/unyt_service.rs**
- `credit_and_complete()` — marks dirty after payment credit
- mHOT payment handler — marks dirty after payment credit

**rust-executor/src/lib.rs**
- Spawns background flush loop (2s interval) that:
  - Drains the per-user dirty set (only affected emails)
  - Fetches fresh `HostingUserInfo` for each dirty user
  - Logs and skips on DB errors instead of fabricating defaults
  - Publishes to the subscription topic

**rust-client/schema.gql**
- Added `runtimeHostingUserInfoChanged: HostingUserInfo` to Subscription type

### 2. Core client subscription (`135c7ce0`)

**core/src/agent/AgentClient.ts**
- Added `HostingUserInfoChangedCallback` type
- Added `#hostingUserInfoChangedCallbacks` private field
- Added `addHostingUserInfoChangedListener()` method
- Added `subscribeHostingUserInfoChanged()` method

### 3. Connect integration (`966443d0`)

**connect/src/core.ts**
- Added `startCreditSubscription()` — sets up subscription listener + starts polling as safety net
- Changed polling interval from 30s to 60s (now just a fallback)

**connect/src/web.ts**
- Switched both call sites from `startCreditPolling()` to `startCreditSubscription()`

**connect/package.json**
- Fixed `@coasys/ad4m` dependency from `workspace:0.12.1-dev` to `workspace:*`

### 4. Per-user scoping (`608f807f`)

- Replaced `AtomicBool` with `Mutex<HashSet<String>>` for per-user dirty tracking
- `mark_credits_dirty()` now takes `&str` email — all 5 call sites updated
- Flush loop drains dirty set — O(dirty users) not O(all users)
- Subscription resolver passes authenticated user's email as filter
- `GetFilter` for `HostingUserInfo` returns email for per-user matching

### 5. DB error handling (`ab82ae9a`)

- Flush loop now logs errors with user email context and skips publishing
- Previously `unwrap_or` would fabricate HostingUserInfo from defaults

## Design Decisions

- **Mutex<HashSet> for per-user dirty tracking**: Supports multi-user mode. Only affected users' info is fetched and published. The lock is held for a single `insert` or `drain` (nanoseconds), so contention is negligible even under concurrent AI streaming.
- **2s flush interval**: Imperceptible for a credit balance display. During idle periods the loop is essentially free (empty set drain).
- **Polling kept as fallback**: 60s safety net catches any missed `mark_dirty()` calls or edge cases. Can be removed once the subscription is battle-tested.
- **All 5 mutation sites covered**: `reserve_compute_credits`, `set_user_credits`, `set_user_free_access`, and both unyt payment paths.
- **Per-user subscription filtering**: `GetFilter` returns the user's email, resolver passes the authenticated email as filter — each subscriber only receives their own credit updates.

## Testing

- `cargo check` passes for rust-executor
- `tsc --noEmit` passes for core
- `pnpm build` passes for connect

## Release Notes

* **New Features**
  * Real-time credit and user information updates now delivered immediately when credit balance or access status changes, with automatic fallback to periodic checks if real-time delivery becomes unavailable. Users receive instant notifications for credit depletion and low-credit warnings.

### DIFF
--- a/rust-executor/src/graphql/subscription_resolvers.rs
+++ b/rust-executor/src/graphql/subscription_resolvers.rs
@@ -570,9 +570,10 @@ impl Subscription {
         match check_capability(&context.capabilities, &RUNTIME_HOSTING_READ_CAPABILITY) {
             Err(e) => Box::pin(stream::once(async move { Err(e.into()) })),
             Ok(_) => {
+                let filter = user_email_from_token(context.auth_token.clone());
                 let pubsub = get_global_pubsub().await;
                 let topic = &HOSTING_USER_INFO_CHANGED_TOPIC;
-                subscribe_and_process::<HostingUserInfo>(pubsub, topic.to_string(), None).await
+                subscribe_and_process::<HostingUserInfo>(pubsub, topic.to_string(), filter).await
             }
         }
     }
--- a/rust-executor/src/graphql/subscription_resolvers.rs
+++ b/rust-executor/src/graphql/subscription_resolvers.rs
@@ -8,8 +8,8 @@ use crate::{
     pubsub::{
         get_global_pubsub, subscribe_and_process, AGENT_STATUS_CHANGED_TOPIC, AGENT_UPDATED_TOPIC,
         AI_MODEL_LOADING_STATUS, AI_TRANSCRIPTION_TEXT_TOPIC, APPS_CHANGED,
-        EXCEPTION_OCCURRED_TOPIC, NEIGHBOURHOOD_SIGNAL_TOPIC, PERSPECTIVE_ADDED_TOPIC,
-        PERSPECTIVE_LINK_ADDED_TOPIC, PERSPECTIVE_LINK_REMOVED_TOPIC,
+        EXCEPTION_OCCURRED_TOPIC, HOSTING_USER_INFO_CHANGED_TOPIC, NEIGHBOURHOOD_SIGNAL_TOPIC,
+        PERSPECTIVE_ADDED_TOPIC, PERSPECTIVE_LINK_ADDED_TOPIC, PERSPECTIVE_LINK_REMOVED_TOPIC,
         PERSPECTIVE_LINK_UPDATED_TOPIC, PERSPECTIVE_QUERY_SUBSCRIPTION_TOPIC,
         PERSPECTIVE_REMOVED_TOPIC, PERSPECTIVE_SYNC_STATE_CHANGE_TOPIC, PERSPECTIVE_UPDATED_TOPIC,
         RUNTIME_MESSAGED_RECEIVED_TOPIC, RUNTIME_NOTIFICATION_TRIGGERED_TOPIC,
@@ -559,6 +559,20 @@ impl Subscription {
                     Some(subscription_id),
                 )
                 .await
+            }
+        }
+    }
+
+    async fn runtime_hosting_user_info_changed(
+        &self,
+        context: &RequestContext,
+    ) -> Pin<Box<dyn Stream<Item = FieldResult<HostingUserInfo>> + Send>> {
+        match check_capability(&context.capabilities, &RUNTIME_HOSTING_READ_CAPABILITY) {
+            Err(e) => Box::pin(stream::once(async move { Err(e.into()) })),
+            Ok(_) => {
+                let pubsub = get_global_pubsub().await;
+                let topic = &HOSTING_USER_INFO_CHANGED_TOPIC;
+                subscribe_and_process::<HostingUserInfo>(pubsub, topic.to_string(), None).await
             }
         }
     }
--- a/rust-executor/src/lib.rs
+++ b/rust-executor/src/lib.rs
@@ -505,18 +505,39 @@ pub async fn run(mut config: Ad4mConfig) -> JoinHandle<()> {
 
             let pubsub = get_global_pubsub().await;
             for email in &dirty_emails {
-                let free_access = Ad4mDb::with_global_instance(|db| db.get_user_free_access(email))
-                    .unwrap_or(false);
+                let free_access =
+                    match Ad4mDb::with_global_instance(|db| db.get_user_free_access(email)) {
+                        Ok(v) => v,
+                        Err(e) => {
+                            error!(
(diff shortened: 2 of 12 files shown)


rust-executor/src/graphql/subscription_resolvers.rs
```diff
@@ -562,4 +562,18 @@ impl Subscription {
...
+    async fn runtime_hosting_user_info_changed(
+        &self,
+        context: &RequestContext,
+    ) -> Pin<Box<dyn Stream<Item = FieldResult<HostingUserInfo>> + Send>> {
+        match check_capability(&context.capabilities, &RUNTIME_HOSTING_READ_CAPABILITY) {
+            Err(e) => Box::pin(stream::once(async move { Err(e.into()) })),
+            Ok(_) => {
+                let pubsub = get_global_pubsub().await;
+                let topic = &HOSTING_USER_INFO_CHANGED_TOPIC;
+                subscribe_and_process::<HostingUserInfo>(pubsub, topic.to_string(), None).await
```
lucksus: This doesn't seem to distinguish per user... ? This needs to be scoped per user, i.e. each user needs to have their own subscription being created with only their credits.

rust-executor/src/lib.rs
```diff
@@ -475,6 +475,61 @@ pub async fn run(mut config: Ad4mConfig) -> JoinHandle<()> {
...
+                let info = crate::graphql::graphql_types::HostingUserInfo {
+                    email: email.clone(),
+                    remaining_credits,
+                    hot_wallet_address,
+                    free_access,
+                };
+
+                if let Ok(json) = serde_json::to_string(&info) {
+                    pubsub
+                        .publish(&HOSTING_USER_INFO_CHANGED_TOPIC, &json)
```
lucksus, line 551: yeah, this will notify all users, no?

lucksus (changes requested):

rust-executor/src/pubsub.rs
```diff
@@ -131,6 +131,21 @@ lazy_static::lazy_static! {
...
+    pub static ref HOSTING_USER_INFO_CHANGED_TOPIC: String = "hosting-user-info-changed-topic".to_owned();
+}
+
+/// Per-user dirty set for batched credit change notifications.
+/// When credits are mutated, the affected user's email is inserted.
+/// A background flush loop periodically drains the set and publishes
+/// updated HostingUserInfo only for the affected users.
+pub static DIRTY_CREDIT_USERS: std::sync::LazyLock<
+    std::sync::Mutex<std::collections::HashSet<String>>,
+> = std::sync::LazyLock::new(|| std::sync::Mutex::new(std::collections::HashSet::new()));
```
lucksus: just a cosmetic change: can we import those types with `use` and have this line better readable?

rust-executor/src/lib.rs
```diff
@@ -475,6 +475,86 @@ pub async fn run(mut config: Ad4mConfig) -> JoinHandle<()> {
...
+                        .await;
+                }
```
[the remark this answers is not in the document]
jhweir: Claude pushes back:

Re-enqueuing on failure adds complexity and creates worse problems than it solves:

Infinite retry loop: If the DB is persistently failing for a user (e.g., corrupted row, user deleted), the email gets re-enqueued every 2 seconds forever, spamming error logs with no way to stop.

The failure scenario is near-impossible: [mark_credits_dirty(email)](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is only called right after a successful DB write for that user. If the DB worked for the write, it'll work for the read 2 seconds later.

Already has a safety net: The 60s polling fallback in connect catches any missed updates regardless.

The "permanent drop" is actually correct: If we can't read the user's data, we shouldn't publish stale/fabricated data. The next credit mutation will re-mark them dirty and they'll get picked up on the next flush cycle.

lucksus (approved):

PR #776, opened 2026-03-24
Bill Transcription on Resolved Words

jhweir:
# Bill Transcription on Resolved Words

**Branch:** `feature/transcription-billing` → `dev`

## Problem

Whisper transcription streams bypass billing entirely. The mutations call
`check_compute_credits()` (a read-only pre-check) but never deduct credits or
create `compute_log` entries. Users can transcribe unlimited audio at zero cost.

## Solution

Bill per resolved word as the Whisper model produces text segments, inside the
async transcription task. This avoids the overhead of billing on every
`feed_transcription_stream` call (thousands per session) while accurately
tracking actual transcription output.

### Billing flow

1. **Open** — `check_compute_credits` blocks users with 0 credits
2. **Feed** — no billing check (removed); just forwards audio samples
3. **Async Whisper task** — on each resolved text segment, counts words and
   calls `bill_compute(email, word_count * rate, "ai_transcription", summary)`
4. **Credits deplete mid-stream** — billing warns but doesn't kill the stream;
   next `open` call will be blocked

### Rate

`DEFAULT_TRANSCRIPTION_WORD_RATE = 5.0` credits per word, with DB override via
`host_rates` table (key: `"whisper transcription"`). For reference:
- LLM token: 12.5
- Embedding token: 1.0
- Link write: 0.25

## Files Changed

| File | Change |
|------|--------|
| `rust-executor/src/billing.rs` **(new)** | Shared `bill_compute()` function — checks global free hosting, per-user free access, then deducts credits, logs to `compute_log`, pushes pubsub entry, and marks credits dirty. Callable without GraphQL context. |
| `rust-executor/src/lib.rs` | Added `pub mod billing;` |
| `rust-executor/src/ai_service/mod.rs` | Extended `TranscriptionSession` with `model_id` and `user_email`. Updated `open_transcription_stream` to accept `auth_token` and extract email. Added `DEFAULT_TRANSCRIPTION_WORD_RATE` constant. Added per-segment billing in the async Whisper task (word count × rate). |
| `rust-executor/src/graphql/mutation_resolvers.rs` | `ai_open_transcription_stream`: passes `auth_token` to the service. `ai_feed_transcription_stream`: removed `check_compute_credits` call. Removed unused `DEFAULT_TRANSCRIPTION_WORD_RATE` constant (moved to ai_service). |

## Edge Cases

- **Silent audio** — no segments produced, no billing
- **Credits hit 0 mid-stream** — stream continues (warn-only), next open blocked
- **Free hosting / free-access users** — `bill_compute` no-ops
- **Model-specific pricing** — DB `host_rates` lookup by `"whisper transcription"`,
  falls back to default constant

## Notes

- `billing.rs` is also needed by the upcoming ORM link billing PR
- No existing tests reference transcription stream functions

## Summary by CodeRabbit

* **New Features**
  * Transcription now requires auth, enforces per-stream ownership in multi-user mode, and records billing context per session.
  * Per-segment billing applied during transcription; a public billing endpoint handles deductions and logging with free-host/user exemptions.

* **Bug Fixes**
  * Removed redundant upfront credit check; transcription halts on insufficient credits while other billing failures are logged.

* **Style**
  * Activity log badges treat all AI operations as primary and show "AI Transcription" for transcription entries.


rust-executor/src/ai_service/mod.rs
```diff
@@ -1246,17 +1261,52 @@ impl AIService {
...
+                                        Some(&format!(
+                                            "{} words (model: {}): \"{}\"",
+                                            word_count,
+                                            billing_model_id,
+                                            if text.len() > 80 { &text[..80] } else { &text }
+                                        )),
+                                    ) {
+                                        Err(e) if e.to_string().contains("Insufficient compute credits") => {
+                                            log::warn!("Transcription stream terminated: insufficient credits for user {}", email);
+                                            break;
```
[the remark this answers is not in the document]
jhweir: I'd push back on this suggestion. The break on "Insufficient compute credits" is correct and intentional — it's a deliberate enforcement mechanism that stops processing when a user has exhausted their credits. Removing it would let users continue consuming compute resources they haven't paid for, which is the opposite of what billing enforcement should do.

The current behavior:

Insufficient credits → log + break (stop the stream — correct, this is a spending limit)
Other billing errors → log warning, continue (transient failures shouldn't kill the stream)
This is sound design.

rust-executor/src/ai_service/mod.rs
```diff
@@ -1152,9 +1159,23 @@ impl AIService {
...
         let model_size = Self::get_whisper_model_size(model_id.clone())?;
 
+        // Extract user email from token for billing and ownership tracking
+        let user_email = crate::agent::capabilities::user_email_from_token(auth_token.clone());
+
+        // In multi-user mode, reject requests with missing or invalid tokens
+        let multi_user_enabled = crate::db::Ad4mDb::with_global_instance(|db| {
+            db.get_multi_user_enabled().unwrap_or(false)
+        });
+        if multi_user_enabled && user_email.is_none() {
```
lucksus: this breaks transcription for the main user if they enable multi-user, i.e. remote login

rust-executor/src/ai_service/mod.rs
```diff
@@ -1316,6 +1393,41 @@ impl AIService {
         }
     }
 
+    /// Verify that the caller (identified by auth_token) owns the given transcription session.
+    /// In single-user mode, ownership is not enforced.
+    /// In multi-user mode, both the session and the caller must have valid emails that match.
+    fn verify_stream_ownership(session: &TranscriptionSession, auth_token: &str) -> Result<()> {
```
lucksus, line 1406: ah, are we doing this on the subscription as well? if not people could listen to other's transcript..

lucksus (changes requested):
not a full review, but I think this breaks transcription for the main agent

rust-executor/src/ai_service/mod.rs
```diff
@@ -1246,17 +1276,55 @@ impl AIService {
...
-                            //println!("GOT segment: {}", segment.text());
+                            let text = segment.text().to_string();
                             let stream_id_clone = stream_id_clone.clone();
 
+                            // Bill for transcribed words
+                            let word_count = text.split_whitespace().count();
+                            if word_count > 0 {
+                                if let Some(ref email) = billing_email {
+                                    let rate = match crate::db::Ad4mDb::with_global_instance(|db| {
+                                        db.get_host_rate("whisper transcription")
```
lucksus: this should allow for different rates for different models. we're currently using two different whisper size, so at least those need to be differnt. but best would be to have this code not coupled to Flux' assumption that there would a fixed model (set). each transcription model needs it's own configurable rate.

lucksus (approved):

PR #782, opened 2026-04-02
fix: SPARQL follow-up — named graphs, persistence, concurrency, batch hydration, push-down filtering

HexaField:
## Summary

All Tier 1, 2, and 3 improvements from the SPARQL migration review. Single follow-up PR covering robustness, performance, correctness, and SPARQL 1.1 compliance.

---

### 3.1 Concurrent write protection
Verify Oxigraph `Store` thread safety for concurrent writes. If not safe, wrap in `RwLock`. Add concurrent write stress tests.

### 3.2 Persistent Oxigraph store
Use `Store::open(path)` for disk-backed storage. Eliminates O(n) startup rebuild from SQLite on every restart.

### 3.3 Pubsub type safety
Add type validation to pubsub channels to prevent silent deserialization failures.

---

### 3.4 N+1 relation hydration → batch SPARQL
Replace per-instance `perspective.get(LinkQuery)` calls with single `VALUES`-based SPARQL query per relation type. 150 queries → 3.

### 3.5 Push-down FILTER for literal equality
Add `FILTER(fn::parse_literal(?target) = "value")` for exact equality and IN conditions. Reduces result set at SPARQL level.

### 3.6 Subscription re-query filtering
Extract referenced predicates from SPARQL subscriptions. Only re-run when relevant predicates change.

### 3.7 Fingerprint optimization
Replace `JSON.stringify` of all properties with lightweight ID + timestamp hash.

---

### 3.8 Parser-based SPARQL validation
Replace string-matching with Oxigraph's query parser for structural validation.

### 3.9 Getter parser warnings
Log warnings when `convertGetterToSPARQL` can't parse a getter string.

### 3.10 Error messages with query text
Include the failing SPARQL query in error messages.

---

## SPARQL 1.1 Compliance — Named Graphs

Replace RDF-star (Oxigraph-specific) with standard named graphs for link metadata:

```turtle
GRAPH <link:a1b2c3d4> { <source> <predicate> <target> . }
<link:a1b2c3d4> ad4m:ontology/author "did:key:..." .
<link:a1b2c3d4> ad4m:ontology/timestamp "1712025600000" .
```

- Standard SPARQL 1.1 — works with any conforming store
- Same storage cost (7 triples per link)
- Explicit link identity via graph IRI
- Phased migration: dual-write → query migration → remove RDF-star

---

## Design Documents

- `docs/proposals/sparql-named-graphs.md` — full named graphs design with Rust/SPARQL code
[...]

PR #783, opened 2026-04-02
fix: SPARQL robustness & performance — persistence, concurrency, batch hydration, push-down filtering

HexaField:
## Summary

Address the top-priority robustness and performance issues identified in the SPARQL migration review.

### Concurrent write protection
Verify Oxigraph `Store` thread safety for concurrent writes. If not safe, wrap in `RwLock`. Add concurrent write stress tests.

### Persistent Oxigraph store  
Use `Store::open(path)` for disk-backed storage instead of in-memory only. Eliminates O(n) startup rebuild from SQLite. Requires `rocksdb` feature flag in oxigraph dependency.

### Pubsub type safety
Add type validation to pubsub channels to prevent silent deserialization failures (like the `APPS_CHANGED` empty string bug).

### N+1 relation hydration → batch SPARQL
Replace per-instance `perspective.get(LinkQuery)` calls in `hydrateRelations` with a single `VALUES`-based SPARQL query per relation type. For 50 instances × 3 relations, this reduces 150 queries to 3.

### Push-down FILTER for literal equality
Add `FILTER(fn::parse_literal(?target) = "value")` for exact equality and IN conditions on literal-stored properties. Reduces result set at SPARQL level instead of fetching all instances and filtering in JS.

### Subscription re-query filtering
Extract referenced predicates from subscription SPARQL queries. Only re-run subscriptions when a changed link's predicate matches.

### Fingerprint optimization
Replace `JSON.stringify` of all properties with lightweight ID + timestamp hash for subscription change detection.

### Parser-based SPARQL validation
Replace string-matching `validate_readonly_query` with Oxigraph's query parser for structural validation.

### Error messages with query text
Include the failing SPARQL query in error messages for debugging.

### Getter parser warnings
Log warnings when `convertGetterToSPARQL` can't parse a getter string instead of failing silently.

Full plan with test statements: `docs/proposals/sparql-robustness-performance.md`

## Changes

- [ ] `rust-executor/src/sparql_service/mod.rs` — persistence, concurrency, validation, error messages
- [ ] `core/src/model/hydration.ts` — batch relation hydration
- [ ] `core/src/model/query-sparql.ts` — push-down FILTER for equality
- [ ] `core/src/model/ModelQueryBuilder.ts` — fingerprint optimization, subscription filtering
- [ ] Tests for all changes

### DIFF
--- a/docs/proposals/sparql-robustness-performance.md
+++ b/docs/proposals/sparql-robustness-performance.md
@@ -1,0 +1,507 @@
+# AD4M SPARQL Migration — Comprehensive Review & Improvement Proposals
+
+**Branch:** `feat/sparql-replaces-surrealdb` (PR [#760](https://github.com/coasys/ad4m/pull/760))
+**Scope:** 88 files, 10,541 insertions, 11,027 deletions
+**Status:** All 7 CI suites green, Flux deploy preview functional
+
+---
+
+## 1. Architecture Summary
+
+The migration replaces SurrealDB with [Oxigraph](https://github.com/oxigraph/oxigraph) (a Rust-native in-memory RDF triple store) as the query engine for AD4M perspectives.
+
+**Core mapping:**
+- Each AD4M link (`source → predicate → target`) becomes a direct RDF triple
+- Link metadata (author, timestamp, proof) stored as RDF-star annotations on quoted triples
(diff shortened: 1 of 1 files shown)


HexaField:
Merged into #782 — single follow-up PR for all Tier 1, 2, and 3 improvements.

PR #786, opened 2026-04-08
ci: run cargo test in build-and-test job

HexaField:
Adds `cargo test --workspace --exclude ad4m-launcher` to the `build-and-test` CI job.

225 Rust unit tests exist but have never been run in CI — this makes them visible.

Also fixes a pre-existing flaky test in `rust-client/src/literal.rs` — `can_handle_objects` was comparing an exact URL string, but JSON key serialization order is non-deterministic. Fixed to round-trip through `from_url()` and compare parsed values instead.

**Changes:**
- `.circleci/config.yml` — add `cargo test` step
- `rust-client/src/literal.rs` — fix order-dependent test

## Summary by CodeRabbit

* **Chores**
  * CI updated to run automated unit tests in the build-and-test job to improve quality checks.

* **Tests**
  * Unit tests adjusted to be less brittle around JSON ordering and now validate data round-trip correctness rather than exact serialized strings.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,6 +97,10 @@ jobs:
           no_output_timeout: 30m
           command: cargo clippy --all --exclude ad4m-launcher
       - run:
+          name: Rust unit tests
+          command: cargo test --workspace --exclude ad4m-launcher
+          no_output_timeout: 15m
+      - run:
           name: Stash binaries locally (skip CircleCI network storage — all jobs run on same host)
           command: |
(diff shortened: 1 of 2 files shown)


lucksus (changes requested):
This is not correct. Rust tests are run with the `pnpm test` command which is the last command run in `build-and-test` but it doesn't run the `rust-client` tests. If you want to add those please work it into what `pnpm test` would run on top-level.

PR #794, opened 2026-04-11
perf: SPARQL pagination subquery + delta-based subscription optimization

HexaField:
## perf: SPARQL query performance optimizations

Fixes chat loading latency, backed-up summary generation, and slow neighbourhood loading for long-running conversations.

### Root Causes Identified

1. **No SPARQL-level pagination** — `findAll({ limit: 50 })` fetched ALL triples then sliced in JS. 1000 messages = 7000+ triples to show 50.
2. **Full re-query on every link event** — subscriptions triggered complete `findAll` re-runs, compounding #1.
3. **O(N²) `FILTER NOT EXISTS`** with nested GRAPH patterns in `unprocessedItems()` stalling the summary pipeline.
4. **N+1 queries in `subgroupsData()`** — separate SPARQL query per subgroup.
5. **`query_links()` brute-force scan** — iterated ALL named graphs to find matching links, O(total_graphs) even for single-source queries.
6. **`hydrateRelations()` N+1 for reverse relations** — sequential `perspective.get()` per instance with no batching.

---

### Fixes

**SPARQL-level pagination (HIGH IMPACT)**
- `buildSPARQLOrderLimitOffset()` emits `ORDER BY / LIMIT / OFFSET` clauses
- Source selection in subquery — DB selects only paginated DISTINCT sources before fetching links
- Guard: pagination only when all filters are SPARQL-capable; JS-only filters fall back to full fetch + JS slice
- ORDER BY maps property names to correct SPARQL variable aliases (`?cfTarget_name`, `?cfInitTarget_description`, etc.)

**Delta-based subscriptions (HIGH IMPACT)**
- Subscription pool detects pure-addition updates (no rows removed, ≤5 new rows, all new source URIs)
- Hydrates only new items and merges with cached results
- Falls back to full hydration for removals, updates to existing sources, or large changesets

**`query_links()` Rust optimization (HIGH IMPACT)**
- Replaced brute-force `named_graphs()` iteration with `quads_for_pattern(s, p, t, None)` — single cross-graph pattern match using Oxigraph's GSPO/GOSP indexes
- Old: O(total_graphs). New: O(matches)
- `get_links_by_source()`, `get_links_by_target()`, `get_links_by_predicate()` all benefit automatically

**`hydrateRelations()` batched reverse lookups (HIGH IMPACT)**
- `belongsToOne`/`belongsToMany` relations now batch all instance IDs into 1 SPARQL query + 1 `findAll()` instead of N sequential `perspective.get()` + N individual gets
- Graceful fallback to sequential on SPARQL error
- Deduplicates shared relation targets

---

### Performance Impact

| Scenario | Before | After |
|----------|--------|-------|
| Load 50 messages from 1000 | ~7000 triples fetched | ~350 triples (SPARQL LIMIT) |
| New message subscription | Full re-query all messages | Hydrate 1 new item, merge |
| `query_links(source=X)` on 5000-msg perspective | Scan 5000+ named graphs | Direct index lookup |
| 50 instances, belongsToOne hydration | 100 queries | 2 queries |
| 20 instances, belongsToMany hydration | 40+ queries | 2 queries |

---

### Test Coverage — 52 new tests

**hydration-batching.test.ts (10)**
- Batched reverse relation (50 instances → 1 SPARQL + 1 findAll)
- Batched belongsToOne (20 instances → 1 SPARQL + 1 findAll)
- SPARQL failure fallback to sequential
- Nested includes (multi-level)
- Mixed relation types (hasMany + belongsToOne)
- Empty relations (no crash)
- Duplicate relation targets (deduplication)
- Large batch (150 instances)
- Where filters on included relations
- Circular relations (no infinite loop)

**subscription-delta.test.ts (11)**
- New sources appended via delta
- Existing source triggers fallback
- Removals trigger fallback
- >5 rows trigger fallback
- No-growth case
- Multiple new sources
- Rapid 10-item burst
- Addition + removal → fallback
- Complex source URIs (encoded, fragments)
- Empty previous + additions
- URI special characters

**query-sparql.test.ts (20)**
- Pagination: LIMIT, OFFSET, combined, disabled for JS-only filters
- ORDER BY: correct variable mapping for required/optional/timestamp/fallback
- Limit=0, offset beyond count, multi-field ORDER BY
- Pagination with parent filter
- SPARQL injection safety (field names + IRI values)
- hasJsOnlyWhereFilters: literal, author, timestamp, belongsTo, SPARQL-only

**sparql_store.rs Rust tests (12)**
- query_links by source, predicate, target, date range, limit, combined, empty, no-filter
- Literal targets, 100 predicates per source, Unicode roundtrip (emoji/CJK), date boundary inclusivity

---

### Companion PR

Flux [#585](https://github.com/coasys/flux/pull/585):
- `FILTER NOT EXISTS` → set-difference (O(N²) → O(N))
- Batch subgroup timestamps (N+1 → 1)
- Scoped processed query to channel
- `getData()` incremental refresh — predicate-aware link handler
- Tests: channel query scoping, debounce escalation, concurrent guard

## Release Notes

* **New Features**
  * Query ordering and pagination now execute directly in SPARQL for improved accuracy and consistency.

* **Improvements**
  * Enhanced subscription performance through delta-based result updates that only fetch newly added data.
  * Optimized relation hydration with batched queries instead of sequential requests, reducing database round-trips.

### DIFF
--- a/core/src/model/hydration.ts
+++ b/core/src/model/hydration.ts
@@ -489,7 +489,8 @@ export async function evaluateCustomGettersForInstance(
             }
           }
         } else {
-          // Legacy SurrealDB-style getter — no longer supported
+          // Legacy SurrealDB-style getter — safety guard for getters that don't match
+          // SPARQL patterns (SELECT/ASK). Logs a warning but does not crash.
           console.warn(`Unsupported legacy getter syntax for property ${propName} — use native SPARQL (SELECT/ASK): ${rawGetter.slice(0, 100)}`);
         }
       } catch (error) {
@@ -576,7 +577,8 @@ export async function evaluateCustomGettersForInstance(
             instance[relName] = values;
           }
         } else {
-          // Legacy SurrealDB-style getter — no longer supported
+          // Legacy SurrealDB-style getter — safety guard for relation getters that don't
+          // match SPARQL patterns (SELECT/ASK). Logs a warning but does not crash.
           console.warn(`Unsupported legacy getter syntax for relation ${relName} — use native SPARQL (SELECT/ASK): ${getter.slice(0, 100)}`);
         }
       } catch (error) {
--- a/core/src/model/hydration.ts
+++ b/core/src/model/hydration.ts
@@ -653,17 +653,86 @@ export async function hydrateRelations<T>(
     // ── Reverse relations (belongsToOne / belongsToMany) ──────────────────
     // The link goes target→instance, so we query backwards:
     //   predicate = meta.predicate, target = inst.id  →  source is the related id
+    //
+    // PERF: Batch all reverse lookups into a single SPARQL query instead of
+    // N sequential perspective.get() calls (N+1 → 1 optimisation).
     if (meta.kind === 'belongsToOne' || meta.kind === 'belongsToMany') {
-      // Per-instance reverse lookup (can't batch easily across instances)
+      const allIds = instances.map(inst => (inst as any).id).filter(Boolean);
+
+      // Batch query: find all (source, target) pairs for this predicate
+      // where target is one of our instance IDs
+      const sourcesByTarget = new Map<string, string[]>();
+      if (allIds.length > 0) {
+        const sparqlQuery = `SELECT ?source ?target WHERE {
+  GRAPH ?g { ?source <${meta.predicate}> ?target }
+  FILTER(?target IN (${allIds.map(id => `<${id}>`).join(', ')}))
+}`;
+        try {
+          const sparqlResult = await perspective.querySparql(sparqlQuery);
+          // Parse SPARQL JSON results
+          const bindings = sparqlResult?.results?.bindings ?? sparqlResult?.bindings ?? [];
+          for (const row of bindings) {
+            const source = row.source?.value ?? row.source;
+            const target = row.target?.value ?? row.target;
+            if (!source || !target) continue;
+            if (!sourcesByTarget.has(target)) sourcesByTarget.set(target, []);
+            sourcesByTarget.get(target)!.push(source);
+          }
+        } catch (e) {
+          // Fallback: sequential queries if SPARQL fails
+          console.warn('Batched reverse relation SPARQL failed, falling back to sequential:', e);
+          for (const inst of instances) {
+            const reverseLinks = await perspective.get(
+              new LinkQuery({ predicate: meta.predicate, target: (inst as any).id })
+            );
+            const sourceIds = reverseLinks
+              .filter(l => l.data.target === (inst as any).id)
+              .map(l => l.data.source);
+            if (sourceIds.length > 0) {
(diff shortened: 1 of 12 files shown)


core/src/model/hydration.ts
```diff
@@ -653,60 +653,104 @@ export async function hydrateRelations<T>(
...
+      for (const ids of sourcesByTarget.values()) {
+        for (const id of ids) allSourceIds.add(id);
+      }
+
+      // Batch-hydrate all related instances at once (1 findAll instead of N)
+      const hydratedMap = new Map<string, any>();
+      if (allSourceIds.size > 0) {
+        const fetchQuery: any = {
+          where: {
+            id: Array.from(allSourceIds),
```
lucksus, line 713: Ah, didn't know `where: id` could also be an array..

core/src/model/query-sparql.ts
```diff
@@ -14,6 +14,53 @@ import { resolveParentPredicate } from "./query-common";
...
+    const propMeta = metadata.properties[propertyName];
+    if (!propMeta) continue;
+
+    // Reverse relations are JS-only
+    const relMeta = allRelationsMetadata[propertyName];
+    if (relMeta && (relMeta.kind === 'belongsToOne' || relMeta.kind === 'belongsToMany')) {
+      return true;
+    }
+
+    // Literal-stored properties are JS-only (value filtering can't be pushed to SPARQL)
```
lucksus: This doesn't seem right. We need to be able to compare/constraint literal values. that is what `fn::parse_literal` was for in the surreal impl, and for which `fn/parse_literal` I believe is the SPARQL variant. I'm surprised with our tests work without it. Well, if the JS code makes it work again, that explains tests passing, but isn't that a performance problem?

Why not use the `parse_literal` function?

HexaField (commented):
Good catch — you're right, and this is now fixed in e3d278f.

The original `fn::parse_literal` syntax (SurrealDB-style) failed at Oxigraph parse time, which led us to incorrectly conclude custom functions don't work in SPARQL. But the correct syntax is `<ad4m://fn/parse_literal>(?var)` (IRI-based function call) — which is valid SPARQL and already used successfully in the MCP subscription code.

Push-down filtering is now re-enabled for equality, IN, and NOT operators:
- `where: { name: "Alice" }` → `FILTER(STR(<ad4m://fn/parse_literal>(?wTarget_name)) = "Alice")`
- `where: { name: ["Alice", "Bob"] }` → `FILTER(... IN ("Alice", "Bob"))`
- `where: { name: { not: "deleted" } }` → `FILTER(... != "deleted")`

Comparison operators (gt/lt/gte/lte/between/contains) stay JS-only since `parse_literal` returns strings and numeric/date comparisons would be unreliable. JS post-filter remains as safety net in all cases.

⬡

lucksus (approved):

jhweir (approved):

PR #795, opened 2026-04-11
perf: replace named-graph iteration with cross-graph pattern match in query_links()

HexaField:
## Problem

`query_links()` iterates ALL named graphs via `self.store.named_graphs()`, then queries each graph individually. For a perspective with 5000+ links, this is O(total_graphs) even when filtering by a specific source.

## Fix

Replace the per-graph iteration with a single `quads_for_pattern(s, p, t, None)` call that searches across all graphs at once. Oxigraph's GSPO/GOSP indexes make this O(matches) instead of O(total_graphs).

## Changes

- Replaced the `named_graphs()` iteration loop with a single cross-graph `quads_for_pattern` call
- Added 8 new tests covering all filter combinations

All 16 query_links tests pass.

### DIFF
--- a/rust-executor/src/perspectives/sparql_store.rs
+++ b/rust-executor/src/perspectives/sparql_store.rs
@@ -296,8 +296,9 @@ impl SparqlStore {
         }
     }
 
-    /// Query links matching optional filters using direct store pattern matching.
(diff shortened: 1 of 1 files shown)


HexaField:
Merged into fix/sparql-query-performance (PR #794)

PR #796, opened 2026-04-11
debug: diagnostic logging for SHACL createSubject 'No constructor found' failure

HexaField:
## Problem

`createSubject("Community", perspectiveUuid)` fails with:
```
No SHACL constructor found for class: Community. Ensure the class has SHACL definitions.
```

This occurs after neighbourhood publish succeeds — the 12 SDNA subject class links ARE registered in the perspective, but the executor can't find the constructor actions when it needs to instantiate the class.

## Investigation

Traced the full flow:

```
Flux: Community.create() → save() → createSubject("Community", baseExpression)
  → Rust: create_subject()
    → get_constructor_actions("Community")
      → get_shape_actions_from_shacl("Community", "ad4m://constructor")
        → SPARQL query for links with predicate "ad4m://constructor"
          where source ends with "CommunityShape"
        → Returns empty → ERROR
```

And the registration flow:
```
SDK: ensureSDNASubjectClass(Community)
  → generateSHACL() → buildSHACL() → SHACLShape.toJSON()
  → addSdna(className, '', 'subject_class', shaclJson)
    → Rust: add_sdna()
      → parse_shacl_to_links(shaclJson) → creates constructor link
      → add_links() → stored in SPARQL (Oxigraph)
```

**Both paths are logically correct.** Two regression tests confirm the parse→store→lookup roundtrip works, including with Oxigraph disk persistence.

## Most Likely Root Cause

**Duplicate detection early return in `add_sdna` (line ~1636-1663).** When `add_sdna` is called, it first checks if a link with `rdf://type` → `ad4m://SubjectClass` already exists for this class name. If found, it returns `Ok(true)` **without adding any SHACL links**.

This would happen if:
1. A neighbourhood peer already synced the base SDNA registration link (the `rdf://type` link) before the local `ensureSDNASubjectClass` ran
2. A previous partial execution added the type link but crashed/timed out before adding the SHACL constructor links
3. The Holochain gossip delivered the class registration from another agent before local registration completed

In all these cases: the 12 base SDNA links exist (triggering the duplicate check), but the constructor action links were never created locally.

### Diagnostic Logging

**`get_shape_actions_from_shacl`** — when constructor lookup fails:
```
🔴 SHACL lookup miss: class=Community, predicate=ad4m://constructor
   links_with_predicate=0, total_links_in_store=156
   sources_with_predicate: [...]
```

This immediately reveals:
- `total_links_in_store=0` → links were never added (check `add_sdna` log)
- `links_with_predicate=0` but `total_links > 0` → constructor link specifically missing (confirms duplicate detection theory)
- `links_with_predicate > 0` → suffix matching is wrong (sources show the actual URIs)

**`add_sdna`** — logs SHACL JSON presence:
```
🟡 add_sdna: class=Community, has_shacl_json=true, constructor_links_generated=3
```
Or if the early return triggers:
```
🟡 add_sdna: class=Community — skipping, already exists (duplicate check)
```

### Regression Tests

1. **`test_shacl_constructor_lookup_via_sparql_store`** — end-to-end: parse SHACL JSON → add links to store → lookup constructor → verify actions found
2. **`test_shacl_constructor_survives_disk_persistence`** — same flow but with Oxigraph disk persistence (write → close → reopen → lookup)

Both pass ✅

## How to Debug

1. Build and deploy with this branch
2. Open Flux, create a community
3. When the error occurs, check executor logs for:
   - `🔴 SHACL lookup miss` — reveals whether links exist
   - `🟡 add_sdna` — reveals whether SHACL JSON was provided and if the early return triggered
4. The combination tells us exactly which path failed

## Potential Fixes (once root cause confirmed)

**If duplicate detection early return:** Change `add_sdna` to always add SHACL links even if the base registration exists. The SHACL links are idempotent — re-adding them is safe.

**If `shaclJson` is None:** Trace the SDK's `addSdna` call to find where the SHACL JSON is dropped.

**If suffix matching is wrong:** Fix the source URI pattern or the suffix check in `get_shape_actions_from_shacl`.

### DIFF
--- a/rust-executor/src/perspectives/perspective_instance.rs
+++ b/rust-executor/src/perspectives/perspective_instance.rs
@@ -1706,8 +1706,19 @@ impl PerspectiveInstance {
         // Handle SHACL links if SHACL JSON provided explicitly
         if let Some(shacl) = shacl_json {
             let shacl_links = parse_shacl_to_links(&shacl, &name)?;
+            log::info!(
+                "🟢 add_sdna: Adding {} SHACL links for class '{}'. Constructor link present: {}",
+                shacl_links.len(),
+                name,
+                shacl_links.iter().any(|l| l.predicate.as_deref() == Some("ad4m://constructor"))
+            );
             self.add_links(shacl_links, LinkStatus::Shared, None, context)
                 .await?;
+        } else {
+            log::warn!(
+                "🟡 add_sdna: No SHACL JSON provided for class '{}' — constructor lookup will fail",
+                name
+            );
         }
 
         //added = true;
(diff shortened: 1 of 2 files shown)


HexaField:
Root cause found and fixed in the REST branch (commit c7ac9a6f1). The REST handler wasn't parsing the JSON subjectClass string like the GraphQL handler does. The diagnostic logging in this PR is still useful but the specific bug is resolved. Closing — will reopen if SHACL issues resurface on dev.

PR #807, opened 2026-05-04
Feature: Add Ad4mModel.fromSHACL()

jhweir:
# Add Ad4mModel.fromSHACL()

## Summary

Adds `Ad4mModel.fromSHACL(shape, name)` — a static method that synthesises a fully-queryable `Ad4mModel` subclass from a raw `SHACLShape` returned by `perspective.getAllShacl()`, without needing the original TypeScript class definition.

This enables WE (and any other AD4M consumer) to hydrate live model classes from external/foreign perspectives at runtime, including legacy Flux perspectives where `sh:hasValue` was not persisted on property shapes.

## Changes

**`core/src/model/Ad4mModel.ts`**
- New static method `fromSHACL(shape: SHACLShape, name: string): typeof Ad4mModel`
- Creates a named dynamic subclass via `class extends (this as any) {}`
- Iterates `shape.properties` and calls `setPropertyRegistryEntry` / `setRelationRegistryEntry` for each, using the same internal registry as `@Property` / `@HasMany` decorators
- Applies the `@Model({ name })` decorator so `generateSHACL` and `generateSDNA` work correctly on the synthesised class
- **Backward-compat:** builds a `Map<predicate, flagValue>` from `shape.constructor_actions` before iterating properties. For shapes that predate `sh:hasValue` persistence (old Flux perspectives), the flag value is recovered from the constructor action's `target` field rather than from the property shape directly. Without this, the synthesised class queries without a type-discriminator triple, returning all node types instead of only the target model.

**`core/src/model/Ad4mModel.test.ts`**
- 13 new unit tests in a `describe("Ad4mModel.fromSHACL()")` block
- Covers: className assignment, scalar properties, collections, `resolveLanguage`, `writable: false`, flag/hasValue, SPARQL type-discriminator triple, backward-compat `constructor_actions`, skipping unnamed properties, `local` flag on relations, `@Model` decorator applied, empty shape

## Test results

```
Tests: 13 new + 122 existing = 135 total, all passing
```

## Notes

- `getModelManifest()` (SHACL → AI prompt format) is **not** in this PR — that transformation is WE-specific and lives in WE's `perspectiveHelpers.ts`
- No changes to `PerspectiveProxy`, `hydration.ts`, or the Rust layer
- Purely additive — no existing behaviour changed

### DIFF
--- a/core/src/model/Ad4mModel.test.ts
+++ b/core/src/model/Ad4mModel.test.ts
@@ -982,6 +982,7 @@ describe("SPARQL comparison filters", () => {
 // ──────────────────────────────────────────────────────────
 
 import { buildSPARQLQuery, formatSPARQLValue } from "./query-sparql";
+import { SHACLShape } from "../shacl/SHACLShape";
 
 describe("SPARQL direct triple pattern generation", () => {
   @Model({ name: "Channel" })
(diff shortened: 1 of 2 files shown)


HexaField:

PR #808, opened 2026-05-04
Feature: IncludeProjection — grouped SPARQL projections via `$`-prefixed include keys

jhweir:
# IncludeProjection — grouped SPARQL projections via `$`-prefixed include keys

**Branch:** `feat/include-projections-v2`  
**Target:** `feat/sparql-1.2-cleanup`  
**Commit:** `b8956ee53`

---

## Summary

Adds `IncludeProjection` as a new include-map value type, allowing callers to attach lightweight aggregations or sub-lists to query results without triggering the N+1 pattern. Projections are declared with a `$`-prefixed key and resolved entirely in Rust using a single grouped VALUES SPARQL per key — O(P) total queries regardless of result set size.

```typescript
// count how many signals each post has
const posts = await Post.findAll(perspective, {
  include: {
    $signalCount: { from: 'signals', count: true },
    $latestSignal: { from: 'signals', limit: 1, order: { id: 'DESC' } },
    comments: true, // normal include — unaffected
  },
});
// posts[0].$signalCount → 7
// posts[0].$latestSignal → "signal://abc"
// posts[0].comments     → [...] (eager-loaded as before)
```

---

## Motivation

The existing `include` map supports eager-loading full relation objects (`comments: true`) and sub-queries (`comments: { where: {...} }`). Neither approach is cheap when you only need a count or the first item — they hydrate all linked instances. Before this PR, callers worked around this by issuing separate queries or doing post-processing in JS, both of which reintroduce N+1 behaviour.

---

### `$` prefix as the discriminator
The `$` prefix is the only signal that a key is a projection. There is no fallback value-type dispatch — a key without `$` always goes to `queryInput.include`, even if its value happens to have a `from` field. This is intentional: silent rerouting based on value shape is confusing and hard to debug.

### Why Rust-native, not TS post-processing
TS post-processing would require fetching all linked instances first (N per parent) then counting/slicing in JS — exactly the N+1 pattern this feature exists to avoid. Rust's grouped VALUES SPARQL issues one query per projection key regardless of how many parent instances are on the page.

### `limit` applied per-parent in Rust, not via SPARQL LIMIT
A global `LIMIT N` on a `VALUES ?parent { ... }` query caps the *total* row count across all parents, so most parents would get fewer results than requested. Instead, the SPARQL fetches all rows and Rust truncates per-parent after collecting into a `HashMap<parent_id, Vec<target>>`.

---

### `core/src/model/types.ts`
- Added `IncludeProjection` interface (`from`, `count?`, `targetShape?`, `where?`, `limit?`, `order?`)
- Added `isIncludeProjection(val)` type guard (`'from' in val`)
- Updated `IncludeMap` value union to `boolean | RelationSubQuery | IncludeProjection`

### `core/src/model/Ad4mModel.ts`
- `executeModelQuery`: replaced the single-line `queryInput.include = query.include` with a splitting loop — `$`-prefixed keys → `queryInput.projections`, all others → `queryInput.include`
- Projections are enriched with `targetShape` (via `relMeta.target()`) when available, so Rust can resolve `where` clause predicates without a second round-trip
- `includedRelNames`, `snapshotRelations`, and `enrichShapeForIncludes` now reference `queryInput.include` (the filtered normal-includes map) rather than `query.include`, preventing projection keys from leaking into dirty-tracking and getter-save logic

### `rust-executor/src/perspectives/model_query.rs`
- Added `ProjectionInput` struct (mirrors `IncludeProjection` TS type)
- Added `projections: Option<HashMap<String, ProjectionInput>>` field to `ModelQueryInput`
- `resolve_projections()`: grouped VALUES SPARQL per key; count path uses `GROUP BY ?parent`, list path collects all rows then truncates per-parent in Rust
- `build_projection_where_patterns()`: resolves `where` clause to SPARQL patterns using `targetShape` for predicate lookup; handles `id`/`base` IRI equality, `String`/`Bool`/`Number`/`StringArray` conditions
- `build_projection_order_clause()`: emits direction-terms (`ASC(?t)`/`DESC(?t)`) joined under a single `ORDER BY` keyword
- `execute_model_query_inner`: calls `resolve_projections` after pagination and `filter_properties`, so `$` keys are added to the final output objects rather than being stripped

### `core/src/model/Ad4mModel.test.ts`
10 new tests in `"IncludeProjection type guard and key splitting"`:
- `isIncludeProjection` returns true/false correctly
- `$`-prefixed keys route to `projections`, not `include`
- Non-`$` keys do NOT route to projections even with a `from`-shaped value
- `include`/`projections` omitted when empty
- `targetShape` enrichment with and without a declared target class
- Result passthrough of `$`-keyed values from Rust

### `rust-executor/src/perspectives/model_query.rs` (tests)
5 new unit tests:
- `build_projection_where_patterns` — empty clause, `id` filter, target-shape property filter
- `build_projection_order_clause` — empty when no order, `ORDER BY DESC(?t)` for `id: DESC`

---

## Bugs fixed during review

| Bug | Impact | Fix |
|-----|--------|-----|
| Global `LIMIT` in list-projection SPARQL | With N parents and `limit: 3`, only 3 total rows returned — most parents get `null` | Removed LIMIT from SPARQL; Rust truncates per-parent with `.take(n)` |
| `ORDER BY` keyword emitted per term | Multiple order terms produced `ORDER BY ASC(?t) ORDER BY DESC(?t)` — invalid SPARQL | Terms now emit `ASC(?t)`/`DESC(?t)`, wrapped once in `ORDER BY` |

---

## Known limitations

- `where` clause filtering in projections uses `STR(?v) = "literal:string:..."` patterns. Properties stored as signed JSON envelopes (`resolveLanguage: 'literal'`) won't match. This is the same limitation as the main query's push-down filters and is documented there. For the primary use cases (`count: true`, IRI-valued relation targets) this doesn't matter.
- `order` in list projections only supports `id`/`base` (i.e. ordering by the linked IRI itself). Ordering by a child property would require a JOIN in the projection SPARQL — out of scope for this PR.

### DIFF
--- a/core/src/model/Ad4mModel.test.ts
+++ b/core/src/model/Ad4mModel.test.ts
@@ -1,5 +1,6 @@
 import { Ad4mModel } from "./Ad4mModel";
-import { Model, Property, Optional, ReadOnly, HasMany, Flag } from "./decorators";
+import { isIncludeProjection } from "./types";
+import { Model, Property, Optional, ReadOnly, HasMany, HasOne, Flag } from "./decorators";
 
 describe("Ad4mModel.getModelMetadata()", () => {
   it("should extract basic model metadata with className", () => {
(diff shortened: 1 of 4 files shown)


HexaField:

PR #821, opened 2026-05-13
Feature: $-Projection Hydration on Subscribe Path + `signalTypeId` Filter Fix

jhweir:
# In-Process Projection Hydration

## Summary

When a `$`-prefixed projection key carries a `target_shape`, the Rust executor now hydrates the raw target IRIs into full model instances in a single batched sub-query — rather than returning raw IRI strings and relying on a TS-side `findAll` round-trip per projection key. The TS layer is unchanged: it receives already-hydrated objects from `modelQuery` and passes them through.

This PR targets `feat/sparql-1.2-cleanup`.

---

### `rust-executor/src/perspectives/model_query/projection.rs`

**`resolve_projections` — step 4: in-process hydration**

The LIST path (`count: false`) previously always returned raw target IRI strings. Added a step after `list_map` is populated: when `proj.target_shape` is present and `proj.count` is false:

1. Collect all unique raw IRI strings across all `list_map` entries.
2. Issue a single `execute_model_query_inner` sub-query: `where: { id: [all_iris] }, deep_query: true`, using the target shape JSON.
3. Build an `id → hydrated object` map and replace raw IRI strings in `list_map` in-place.

When the sub-query finds no match for a given IRI (instance not in store), the raw string is left unchanged. One batch sub-query per projection key regardless of parent instance count.

**`depth: u8` parameter added to `resolve_projections`**

Threads `execute_model_query_inner`'s existing recursion depth counter down into `resolve_projections` so the hydration sub-query increments it. Guards against infinite recursion when shapes have circular relations. All existing call sites updated accordingly.

**New imports:** `BTreeMap` and `ModelQueryInput` added (needed to construct the sub-query).

### `rust-executor/src/perspectives/model_query/query.rs`

Single-line change: passes `depth` through to `resolve_projections` at the call site in `execute_model_query_inner`.

### `rust-executor/src/perspectives/model_query/integration_tests.rs`

- Existing 6 `resolve_projections` call sites updated to pass `0` for the new `depth` parameter.
- New test `test_resolve_projections_where_filter_via_target_shape_property`: sets up a parent with two signal children that have different `signalTypeId` property values (stored as `literal:string:...` IRIs). Verifies that a `where` filter on the decoded plain value correctly selects one of the two signals — in both the COUNT variant (returns `1`) and the LIST/limit:1 variant (returns the hydrated signal's id).

### `core/src/model/Ad4mModel.test.ts`

4 new unit tests in the `IncludeProjection` describe block, covering the TS layer's pass-through behaviour now that hydration happens in Rust:

- **Scalar non-count projection** — `modelQuery` returns an already-hydrated object; test asserts it arrives intact on the result instance.
- **Array non-count projection** — `modelQuery` returns an array of hydrated objects; test asserts both elements arrive intact.
- **Unresolvable target** — `modelQuery` returns a raw IRI string (Rust couldn't hydrate it); test asserts the raw string is preserved.
- **Count projection** — `modelQuery` is only called once (no follow-up `findAll`); numeric value arrives unchanged.

---

## Files Changed

| File | Change |
|------|--------|
| `rust-executor/src/perspectives/model_query/projection.rs` | `depth: u8` param; step 4 hydration block; updated imports |
| `rust-executor/src/perspectives/model_query/query.rs` | Pass `depth` to `resolve_projections` |
| `rust-executor/src/perspectives/model_query/integration_tests.rs` | 6 call sites updated for `depth`; 1 new integration test |
| `core/src/model/Ad4mModel.test.ts` | 4 new unit tests for TS pass-through behaviour |


core/src/model/Ad4mModel.ts
```diff
@@ -966,6 +994,56 @@ export class Ad4mModel {
     return arr.map((json: any) => jsonToModelInstance(this, perspective, json, include, properties));
   }
 
+  /**
+   * Hydrate non-count projection results: Rust returns raw target IRIs for
+   * list/scalar projections ($-prefixed include keys). Batch-fetch all unique
+   * IRIs per projection key and swap raw strings for hydrated model instances.
+   * @internal
+   */
+  static async hydrateProjections<T extends Ad4mModel>(
```
HexaField: This is doing round trips over the network, can this hydration be done in the executor?

HexaField (changes requested):

core/src/model/Ad4mModel.ts
jhweir: Working on that now :)

rust-executor/src/perspectives/model_query.rs
```diff
@@ -5677,12 +5748,233 @@ mod integration_tests {
...
+            got["id"].as_str().unwrap_or(""),
+            like_signal,
+            "list with limit:1 should return the hydrated like signal id, got {got}"
+        );
+    }
+
+    /// Tests `fn/parse_literal` filter matching against a double-encoded literal URI.
+    ///
+    /// Some storage paths produce a `literal:string:X` URI and then encode that URI
+    /// *again* as a literal, yielding `literal:string:literal%3Astring%3AX`.
```
lucksus: Huh, isn't this a bug of those "storage paths"?

lucksus (approved):

PR #825, opened 2026-05-20
Add SHACL Transform DSL for model property transforms

HexaField:
## Summary

Replace JavaScript lambda transforms with a serializable SHACL-AF Node Expression DSL. Moves transform execution from TypeScript client-side to Rust executor-side, eliminating N+1 expression fetches and making transforms visible to SHACL tooling.

- **NodeExpression type**: JSON-serializable union supporting focus, literal, path, exists, if/then/else, concat, coalesce, and function calls
- **Builder API**: Fluent functions (focus(), path(), concat(), etc.) for composing transforms
- **Rust evaluation**: Pure eval_transform() function handling all 8 expression variants
- **Async pipeline**: Full async/await query execution with proper recursion handling via Box::pin
- **Integration**: Serialized into SHACL shapes, evaluated post-hydration during model queries

## Test plan

- [x] All 123 model query integration tests pass
- [x] Full build succeeds (TypeScript + Rust)
- [x] Executor and Flux dev environment verified working

## Release Notes

* **New Features**
  * Introduced SHACL-AF Node Expression DSL with builder functions (`focus`, `literal`, `path`, `exists`, `ifExpr`, `concat`, `coalesce`, `fn`) for declarative property transforms
  * Added `fileToDataUri` built-in transform for converting file storage to data URIs
  * Server-side transform evaluation in the query engine

* **Refactor**
  * Moved property transform evaluation from client-side to server-side processing
  * Converted model query execution to async-based architecture

* **Documentation**
  * Updated model classes guide with server-side transform specification and practical examples


core/package.json
```diff
@@ -5,6 +5,18 @@
   "module": "lib/index.js",
   "types": "lib/src/index.d.ts",
   "type": "module",
+  "exports": {
```
HexaField: why is this necessary? likely not right

HexaField:
All CodeRabbit review comments have been addressed in commit 6090e0ec0:

✅ **Fix 1**: package.json ./shacl export types path corrected to ./lib/src/shacl/index.d.ts
✅ **Fix 2**: Test improved to assert on actual transform being stored in metadata
✅ **Fix 3**: JSDoc example fixed to use literal() builder
✅ **Fix 4a**: Transform added to SHACLShape.toLinks() and fromLinks()
✅ **Fix 4b**: Transform validation with isNodeExpression() in fromJSON()
✅ **Fix 5a**: Path key extraction fixed to check # fragment first
✅ **Fix 5b**: Function transform evaluates args before dispatching
✅ **Fix 6**: Filter fixed to include literal properties; else-if nesting corrected
✅ **Fix 7**: Transform deserialization now fails loudly on errors
✅ **Fix 8**: Added query_async() to SparqlStore using spawn_blocking; updated all call sites

All 123 model query integration tests pass. TypeScript type-checking passes. Ready for re-review.

HexaField:
Additional fix for CodeRabbit latest comment (error handling):

✅ Fix 9: Transform deserialization error handling in SHACLShape improved.

In fromLinks: replaced empty catch block with explicit error messages including the invalid payload and propShapeId.

In fromJSON: added validation to throw when transform is present but not valid, instead of silently dropping it.

Both methods now fail loudly with clear descriptive errors on invalid payloads.

lucksus (changes requested):
Ok, basically fine. You commented yourself about the added export section in core/package.json.

I would add: this needs to be described in the docs.

core/package.json
HexaField: You were right — addressed in 77282d68d. The `exports` block declared a `./shacl` subpath that pointed at `./lib/shacl/index.{js,cjs}`, but rollup only emits `lib/index.js` + `lib/index.cjs`, so the subpath was broken from day one. And since `core/src/index.ts` already re-exports every SHACL builder + `fileToDataUri`, the subpath was redundant on top of being broken. Removed the whole `exports` block; consumers continue to `import { focus, literal, fileToDataUri } from '@coasys/ad4m'`. Dropped the dangling `@coasys/ad4m/shacl` JSDoc references too.

HexaField:
Both addressed in 77282d68d.

1. **`exports` block**: removed — see reply above. It was pointing at files rollup doesn't emit, and the main entry already re-exports the SHACL surface.

2. **Docs**: replaced the stale `transform: (value) => …` lambda example in `docs-src/pages/developer-guides/model-classes.mdx` with a new section on the SHACL-AF Node Expression DSL. Updated:
   - The `@Property` options table — `transform` is now typed as `NodeExpression` with a link to the section
   - A new "The `transform` option" section covering: type, where evaluation runs (Rust executor, not JS), a worked example using `fileToDataUri`, `concat(literal('@'), focus())` and `coalesce(focus(), literal('UTC'))`, a builder reference table for all 8 expression variants (`focus`, `literal`, `path`, `exists`, `ifExpr`, `concat`, `coalesce`, `fn`), and guidance on when to reach for a Prolog/SPARQL `getter` instead

Let me know if you want it pitched differently — happy to expand on the SHACL-AF rationale or add a second worked example if the file→data-URI one isn't representative enough.

docs-src/pages/developer-guides/model-classes.mdx
```diff
@@ -171,16 +171,73 @@ averageRating: number = 0;
...
+| Builder | Produces | Notes |
+| :--- | :--- | :--- |
+| `focus()` | The current node's value | The input being transformed. Useful as a base case. |
+| `literal(v)` | A constant value | Strings, numbers, booleans. |
+| `path('predicate://uri')` | Value reached by traversing a predicate from the focus node | One hop. |
+| `exists(expr)` | Boolean | Whether `expr` resolves to a non-empty value. |
+| `ifExpr(cond).then(a).else(b)` | `a` if `cond` is truthy, otherwise `b` | All three are themselves NodeExpressions. |
+| `concat(a, b, …)` | String concatenation of the arguments | Each argument is a NodeExpression. |
+| `coalesce(a, b, …)` | First non-null argument | Argument-order fallback. |
+| `fn('name', …args)` | Server-side function call | Used by built-ins like `fileToDataUri`; you can register your own in the Rust executor. |
```
lucksus: ...but not without changing the ad4m source code and making your Ad4mModel incompatible with other ad4m instances.. I think this should rather list which functions are available - or just take out the last part.

lucksus (changes requested):
just this misleading doc to fix

lucksus (dismissed):

PR #827, opened 2026-05-21
feat: make SHACL in perspectives the runtime source of truth

HexaField:
## Human Summary

Adds a cache to the Rust executor such that SHACL shapes are not sent and reconstructed often, instead registered and implicit, uniquely identified by name ID for queries. This removes serialising and deserialising JSON descriptions of SHACL over Websocket RPC, as well as leverages the perspective graph as the source of truth.

## Summary

Establishes the perspective's SHACL graph as the single executor-side runtime source of truth for model class structure, backed by an in-memory `ModelShape` cache. The parallel `ModelMetadata` JSON transport (`shapeJson`) shipped on every `modelQuery` / `modelSubscribe` / `evaluateGetters` is eliminated.

## Shape cache

- `PerspectiveInstance` gains a lazy `shape_cache: HashMap<String, Arc<ModelShape>>` populated on first access by parsing SHACL triples from the perspective's store. `add_sdna` invalidates the cached entry when SHACL is (re-)written for a class.
- No persistence — rebuilt lazily after restart from SHACL triples.
- A `ShapeResolver` trait is threaded through `execute_model_query` / `evaluate_getters_batch` / include recursion in `relations.rs`, so nested target-class shapes are resolved through the same cache instead of inlined `target_shape_json` blobs (which are removed from `ShapeRelation`).

## SHACL writer + loader

`SHACLShape` (TS) and `parse_shacl_to_links` (Rust) now emit every field needed for runtime queries:

| Predicate | Purpose |
|---|---|
| `ad4m://relationKind` | `hasMany` / `hasOne` / `belongsToOne` / `belongsToMany` — drives direction and scalar-vs-collection rendering |
| `ad4m://targetClassName` | Bare class name for cache lookup |
| `ad4m://whereFilter` + `ad4m://wherePredicates` | Post-getter filter on relations |
| `ad4m://filter` | Opt-out of type filtering |
| `sh:hasValue` | Canonical SHACL flag detection (writer side) |
| `sh:maxCount` | Set to 1 for scalar relations |

`load_shape` was anaemic — the writer emitted `ad4m://getter` / `ad4m://writable` / `ad4m://local` / `sh:class` / `ad4m://constructor` and the loader silently dropped them. Now it reads every predicate the writer emits.

## Transport drop

- `shape_json` removed from `modelQuery` / `modelSubscribe` / `evaluateGetters` RPC payloads.
- `Ad4mModel.prepareModelQueryParams` no longer returns `shapeJson`; `enrichShapeForIncludes` and the where/conformance pre-compute blocks deleted (executor reads these from SHACL now).
- `IncludeProjection.targetShape` (`ModelMetadata`) → `IncludeProjection.targetClassName` (`string`); executor resolves the target's shape through the cache.
- `parse_shape_from_json` moved to `#[cfg(test)]` only.

## MCP consolidation

`mcp/shacl.rs::load_class` now goes through `perspective.get_shape()` with a thin `ModelShape → ShaclClass` adapter, so MCP tool generation and the model_query path resolve the same `ModelShape` per class.

## Tests

- **Rust** — 11 new round-trip tests (SHACL JSON → links → store → load_shape preserve every field), 5 new cache tests (`get_shape`, `invalidate_shape`, `add_sdna`-invalidates, `model_query`-warms-cache, missing-class error), integration tests updated to use new signatures via test helpers (`StaticShapeResolver`, `execute_model_query_from_json`, `evaluate_getters_batch_from_json`).
- **TS** — 7 decorator → SHACL writer round-trip tests, 6 `SHACLShape` serialisation tests for the new predicates, existing tests updated for new arities.

## Migration

Hard breaking change to `perspective.modelQuery` / `modelSubscribe` / `evaluateGetters` RPC payloads — `shape_json` is gone. Rust executor and TypeScript core ship together. Flux (which uses `@coasys/ad4m` directly) needed no source changes; it only consumes the high-level `Ad4mModel.findAll` / `findOne` / `.save()` APIs.

## Test plan

- [x] `cargo test --lib perspectives::` — 307/307 pass
- [x] `cargo test --lib perspectives::model_query::round_trip_tests` — 11/11 pass
- [x] `cargo test --lib perspectives::perspective_instance::tests` — 26/26 pass (incl. 5 new cache tests)
- [x] `cargo test --lib perspectives::model_query::integration_tests` — 63/63 pass
- [x] `pnpm test` in `core/` for changed-area suites — 239/239 pass
- [x] `cargo fmt`
- [ ] Full monorepo `pnpm test` and end-to-end integration suite — not run locally (requires built executor + Holochain runtime); recommend running in CI before merge

## Summary by CodeRabbit

* **New Features**
  * Relation metadata now exposes relation kind, optional target class name, explicit where/filter metadata.
  * Fallback decoding for file-type properties (handles base64/JSON or raw text).

* **Improvements**
  * Shape resolution moved server-side with per-perspective shape caching and invalidation.
  * Projections and includes resolve targets by class name for more accurate where-filtering and hydration.
  * Getter evaluation and live subscriptions use a smaller, simplified payload.

* **Tests**
  * Expanded SHACL round-trip, projection, include, and integration test coverage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coasys/ad4m/pull/827?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


lucksus (approved):
LGTM - from the two unresolved code at least the one in perspective_instance.rs seems reasonable to apply.

PR #830, opened 2026-05-23
Memory leak follow-ups: duplicate broadcast, per-write flush, jemalloc decay

HexaField:
Builds on #829 / `fix/memory-leak` with four additional fixes uncovered while running the [ad4m-wind-tunnel](https://github.com/coasys/ad4m-wind-tunnel) `S9` memory-leak scenario across four isolation modes (`holochain`, `centralized` p-diff-sync, `local`, `no-languages` with `--language-language-only true`).

## Summary

| Fix | Wind-tunnel impact |
|---|---|
| Stop publishing `link-added` / `link-removed` / `link-updated` twice per addLink | Subscriber delivery ratio **2.0× → 1.0×** across every mode |
| Remove per-write `sparql_store.flush()` | Seed time **2.3s (dev) → 12.5s (#829) → 2.1s** (this PR) |
| Skip per-link signature re-verification in `get_links` | Query latency **720 ms → 452 ms** |
| Fold 6 `quads_for_pattern` calls into one annotation pass | Query latency **452 ms → 411 ms** (cumulative −43% vs dev) |
| Override `_rjem_malloc_conf` for aggressive jemalloc decay | Settle slope **+1.68 → −46.82 MB/min** — RSS actively releases |

Combined: per-mode wind-tunnel monitor-end RSS went from 235 MB (dev) to 222 MB (this PR) on `no-languages`, with query latency down by 43% and seed throughput restored.

## What was investigated

The wind tunnel's S9 scenario hits the executor with a 10K-link seed + several minutes of light steady-state activity (1 link/s + 1 query/30s) while a passive WebSocket subscriber counts `link-added` events and a 5-second RSS sampler runs throughout. I ran four modes back-to-back against three branches:

- `origin/dev` (baseline, with a regenerated CUSTOM_DENO_SNAPSHOT.bin)
- `fix/memory-leak` (PR #829)
- this branch

**Three signals matter**: subscriber delivery ratio (≥1.0× = duplicate broadcast), seed throughput (cost of work-per-link), monitor RSS slope (the actual leak rate). Cross-mode comparison isolates which subsystem owns each contribution — Holochain DNAs, link-language Deno runtime, system Deno languages, or pure Rust core.

A focused controlled experiment (no-languages, 1 link/s, **zero queries**, 5-min monitor) showed RSS actually *drops* from 169 → 136 MB (slope **−8 MB/min**). With queries restored, RSS climbs ~5–10 MB per query and the slope stays positive even over 10-min runs. Conclusion: **the addLink path is clean; queries are the only remaining contributor**.

A new wind-tunnel scenario [`S15 Leak Attribution`](https://github.com/coasys/ad4m-wind-tunnel/blob/main/src/scenarios/s15-leak-attribution.ts) automates that attribution. Three measured 90s phases (IDLE / WRITES / WRITES+QUERIES) with per-phase least-squares slope. On the post-fix binary it reports:

```
write-path contribution :  -0.58 MB/min  → clean
query-path contribution : +37.53 MB/min  → leaking
```

S15 runs in ~5 minutes total (vs S9's ~7 min × 4 modes ≈ 30 min) and is suitable for inner-loop development. S9 stays the multi-mode high-fidelity verification harness.

## Fix 1 — duplicate link-event broadcast

Introduced 2024-05-15 in commit [`83a49a2d2`](../commit/83a49a2d2) ("Update Prolog engine async after Perspective change and resend link change notifications when done"). The synchronous link-mutation handlers (`add_link_expression`, `link_mutations`, `diff_from_link_language`) already call `pubsub_publish_diff` immediately. The asynchronous `spawn_prolog_facts_update` task then republished the same diff after the prolog facts were updated — once in the `Disabled/Simple/SdnaOnly` branch, once at the end of the `Pooled` branch.

Effect: every WebSocket subscriber received every link event **twice**. Confirmed by wire dump: 5 `addLink` RPC calls produced 10 `link-added` frames with identical payloads.

Fix: drop the republish in both branches. The "subscriptions need to re-evaluate after prolog facts settle" semantics the original commit was after are already satisfied by `trigger_prolog_subscription_check` + `changed_predicates`, which the spawned task still flips.

Verified delivery ratio is `1.00×` in all four modes after the fix.

## Fix 2 — per-write SPARQL flush

#829 added `sparql_store.flush()` at the bottom of `persist_link_diff` to drain RocksDB memtables. In bursty workloads (e.g. the 10K seed) this is fine — one flush per batch. In steady-state workloads (1 link/s) it produces **one tiny SST file per write**. Each SST file carries in-memory metadata (block cache entry, bloom filter, table-cache entry, file handle), and 290 monitor-phase writes accumulate ~290 of them before compaction catches up.

The flush hurt throughput 5× without measurably reducing the leak. The original concern — memtable bloat — is already handled by RocksDB's `write_buffer_size` trigger (default 64 MB; our workload is ~16 KB/link so auto-flush kicks in around 4K links). If memtable pressure becomes a real problem it should be addressed via store configuration or a throttled background flush, not a per-write fsync.

## Fix 3 — get_links: skip per-link signature re-verification

`get_links_local` unwrapped each `DecoratedLinkExpression` from the SPARQL store into a `(LinkExpression, LinkStatus)` tuple — discarding the cached `proof.valid` flag. `get_links` then wrapped each tuple back up via `DecoratedLinkExpression::from((link, status))`, which internally calls `verify()` (Ed25519 signature check + SHA-256 hash + hex decode + DID parse) for every link. For 10K-link queries that's 10K Ed25519 verifications + ~30K small allocations on the hot path, for no net benefit — the stored `valid` flag was already correct.

Added `get_links_local_decorated` that returns the SPARQL store's `DecoratedLinkExpression` results directly. `get_links` now sorts and limits in place. Updated three other Prolog-facts call sites (Simple/SdnaOnly modes) to use the same path.

Query latency: **720 ms → 452 ms (−37%)**.

## Fix 4 — sparql_store::query_links: one annotation pass per link

Each matching triple triggered 6 separate `quads_for_pattern` calls (one per `ad4m://ontology/*` annotation: author, timestamp, proofKey, proofSig, proofValid, status). Each call materialises an oxigraph iterator with an associated snapshot. For a 10K-link query that meant **60K+ iterator allocations** on the hot path.

Replaced the closure-based per-annotation lookup with a single pass over all annotations for the reifier subject, dispatching on the predicate string. Drops 60K iterator allocations to 10K per query.

Cumulative query latency vs `dev`: **720 ms → 411 ms (−43%)**.

## Fix 5 — jemalloc decay tuning

After the four fixes above, the wind tunnel still measured 8–14 MB/min monitor slope. A controlled experiment — 5 minutes of 1 link/s **with zero queries** — showed RSS actually *drops* from 169 → 136 MB (slope **−8 MB/min**). With queries restored, RSS climbs ~5–10 MB per query. So the remaining "leak" is concentrated in the query path.

Each `perspective.queryLinks` against the seeded perspective materialises a `Vec<DecoratedLinkExpression>` of ~10K entries → ~5 MB peak allocation. Default jemalloc holds those freed pages in its arenas indefinitely — so every query stairs up RSS without ever giving the memory back to the OS.

Fix: override the static `_rjem_malloc_conf` symbol jemalloc reads at init:

```
background_thread:true,dirty_decay_ms:1000,muzzy_decay_ms:1000
```

This tells jemalloc to (a) run page-purging on a dedicated background thread instead of lazily on next alloc, and (b) cut the unused-page decay window from 10s to 1s.

Effect: settle-phase slope on no-languages went **+1.68 MB/min → −46.82 MB/min** — jemalloc actively releases freed pages, RSS drops 169 → 134 MB during the idle window. Monitor-phase RSS now bounces (queries spike, decay releases) instead of climbing monotonically.

Operators can override by exporting `_RJEM_MALLOC_CONF` (or `MALLOC_CONF` on unprefixed builds) before launching `ad4m-executor`.

## What's still open

The query path still retains ~37 MB/min worth of per-query allocations (per S15 attribution above). The hot path is now down to 1 outer + 1 reifier + 1 annotation pass per matching triple, but each link still produces ~6 short-lived `String` allocations from oxigraph's `Term::Literal::value().to_string()`, plus the result Vec itself. Two paths forward worth exploring:

1. **SPARQL-native query**. Replace the nested iterator loop with a single SPARQL `SELECT ?source ?predicate ?target ?author ?timestamp ?proofKey ?proofSig ?proofValid ?status` query (similar to `get_all_links()`) that joins inside oxigraph's evaluator. Reduces Rust-side iterator churn to zero; oxigraph manages its own snapshots and can stream solutions.
2. **Result limit**. The wind tunnel's `queryLinks({ predicate })` returns ALL matches by default — for 10K seeded links that's 10K rows. A sensible default `LIMIT` on the SPARQL side would cap per-query allocation regardless of store size. (Behaviour change — needs discussion.)

## Test plan

- [x] Wind-tunnel S9, 4-way comparison `dev` / `#829` / this PR — delivery ratio, seed times, leak slopes
- [x] Controlled no-query test confirms addLink path is not the leak
- [x] 10-minute no-languages monitor with the in-binary jemalloc symbol override
- [x] New S15 scenario validates write-path vs query-path attribution
- [ ] Soak test (1h+) to confirm bounded vs. unbounded growth post-fix
- [ ] CI integration tests (`pnpm run test-main` in `ad4m/tests/js`)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,8 +58,8 @@ commands:
             mkdir -p target/release
             # Remove before copy: unlinking avoids ETXTBSY if a previous executor still holds the binary
             rm -f target/release/ad4m target/release/ad4m-executor
-            cp /tmp/circleci-artifacts/${CIRCLE_WORKFLOW_ID}/ad4m target/release/ad4m
-            cp /tmp/circleci-artifacts/${CIRCLE_WORKFLOW_ID}/ad4m-executor target/release/ad4m-executor
+            cp /tmp/circleci-artifacts/${CIRCLE_SHA1}/ad4m target/release/ad4m
+            cp /tmp/circleci-artifacts/${CIRCLE_SHA1}/ad4m-executor target/release/ad4m-executor
       - setup_env
       - run:
           name: Install dependencies
@@ -116,8 +116,8 @@ jobs:
       - run:
           name: Stash binaries locally (skip CircleCI network storage — all jobs run on same host)
           command: |
-            mkdir -p /tmp/circleci-artifacts/${CIRCLE_WORKFLOW_ID}
-            cp target/release/ad4m target/release/ad4m-executor /tmp/circleci-artifacts/${CIRCLE_WORKFLOW_ID}/
+            mkdir -p /tmp/circleci-artifacts/${CIRCLE_SHA1}
+            cp target/release/ad4m target/release/ad4m-executor /tmp/circleci-artifacts/${CIRCLE_SHA1}/
       - run:
           name: Remove pnpm patches
           command: node removePnpm.js
@@ -210,8 +210,8 @@ jobs:
           name: Restore binaries from local stash
           command: |
             mkdir -p target/release
-            cp /tmp/circleci-artifacts/${CIRCLE_WORKFLOW_ID}/ad4m target/release/ad4m
-            cp /tmp/circleci-artifacts/${CIRCLE_WORKFLOW_ID}/ad4m-executor target/release/ad4m-executor
+            cp /tmp/circleci-artifacts/${CIRCLE_SHA1}/ad4m target/release/ad4m
+            cp /tmp/circleci-artifacts/${CIRCLE_SHA1}/ad4m-executor target/release/ad4m-executor
       - setup_env
       - run:
           name: git submodule
--- a/cli/src/ad4m_executor.rs
+++ b/cli/src/ad4m_executor.rs
@@ -4,6 +4,35 @@
 #[global_allocator]
 static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 
+// Tune jemalloc to return freed memory to the OS more aggressively. jemalloc
+// reads `MALLOC_CONF` (or, with tikv-jemallocator's prefixed symbols on
+// macOS/Linux, `_RJEM_MALLOC_CONF`) at process start. With the defaults,
+// peak allocations from bursty work — most notably SPARQL queries that
+// materialise the full link set into a Vec<DecoratedLinkExpression> — sit
+// in jemalloc arenas indefinitely, inflating RSS even after the Rust-side
+// memory has been dropped. The wind tunnel was attributing this to a leak:
+// RSS would step up ~5–7 MB per query and never come back down within the
+// monitor window.
+//
+// jemalloc reads its config on first allocation, which happens before
+// main(), so the only reliable way to apply this from the binary itself
+// is to override the static `malloc_conf` symbol that jemalloc looks up
+// at init time. With tikv-jemallocator the symbol is exposed as
+// `_rjem_malloc_conf` on prefixed builds (default on macOS/Linux).
+//
+// Operators / test harnesses can override these defaults by exporting
+// `_RJEM_MALLOC_CONF=...` (or `MALLOC_CONF=...` on unprefixed builds)
+// before launching ad4m-executor.
+//
+// - background_thread:true   purge runs on a background thread
+// - dirty_decay_ms:1000      release dirty pages 10× faster than default
+// - muzzy_decay_ms:1000      same for muzzy pages
+#[cfg(not(target_env = "msvc"))]
+#[allow(non_upper_case_globals)]
+#[export_name = "_rjem_malloc_conf"]
+pub static _rjem_malloc_conf: Option<&'static [u8]> =
+    Some(b"background_thread:true,dirty_decay_ms:1000,muzzy_decay_ms:1000\0");
+
 extern crate ad4m_client;
 extern crate anyhow;
 extern crate chrono;
--- a/rust-executor/src/perspectives/perspective_instance.rs
+++ b/rust-executor/src/perspectives/perspective_instance.rs
@@ -1743,10 +1743,17 @@ impl PerspectiveInstance {
         Ok(class_names)
     }
 
-    fn get_links_local(
+    /// Query the SPARQL store for matching links, returning the decorated form
+    /// already cached in the store. Does NOT re-verify signatures — the
+    /// `proof.valid` flag was set on insert and persisted into RocksDB.
+    ///
+    /// Use this from query paths that just need to surface the link data.
+    /// Internal call sites that need the bare `LinkExpression` (e.g. to feed
+    /// Prolog) should use [`Self::get_links_local`] which unwraps for them.
+    fn get_links_local_decorated(
         &self,
         query: &LinkQuery,
-    ) -> Result<Vec<(LinkExpression, LinkStatus)>, AnyError> {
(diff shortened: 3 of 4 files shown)


HexaField:
## Update — additional commits + verification

### Wind tunnel additions (https://github.com/coasys/ad4m-wind-tunnel)

- **New scenario `S15 Leak Attribution`** — fast 3-phase (idle / writes / writes+queries) RSS slope split that automatically attributes growth to either write path or query path. Runs in ~3.5 min total.
- **Tightened defaults for S9** — settle 60s→30s, monitor 300s→180s, cooldown 30s→15s. 4-mode sweep now takes ~16 min (was ~30). High-fidelity tunables documented.
- **Verdict fix** — both scenarios now treat only positive RSS slope as a leak. Previously `Math.abs(slope)` was making healthy decay phases (e.g. settle/cooldown with jemalloc background decay) verdict=`leaking`, which polluted the comparison tables in the PR description.

### Verified attribution on the current `leak-hunt` binary

S15 final run, fresh executor, default phases:

```
seed:    2000 links in 0.4s
idle     +2.20 MB/min  (allocator noise)
writes   -5.10 MB/min  (RSS DROPS during 5 addLink/s — write path is clean)
queries +28.54 MB/min  (12 queries returning 2K rows each — the leak source)

writePath contribution : -7.30 MB/min  clean
queryPath contribution : +33.64 MB/min  leaking
query latency avg      : 112 ms  (vs ~720 ms baseline on dev)
wall time              : 190 s
```

So on top of the 5 fixes in this PR's commits:

- ✅ duplicate link-event broadcast (1.0× delivery ratio across all modes)
- ✅ per-write SPARQL flush removed (seed time 12.5s → 2.1s)
- ✅ get_links re-verification skipped (-37% query latency)
- ✅ 6 reifier lookups folded into 1 (-43% total query latency vs dev)
- ✅ jemalloc decay tuning (settle/cooldown now actively release)

…the **write path is now fully clean** (negative slope = jemalloc winning) and the remaining leak is fully isolated to the query path: ~3 MB per query that materialises ~2K rows.

### Still open (deferred to follow-up)

The remaining ~33 MB/min query-path slope is dominated by per-row `String` allocations from oxigraph's `Term::Literal::value().to_string()` × 9 fields × 2K rows × 12 queries/minute. Realistic options:

1. **SPARQL-native query** — replace the nested `quads_for_pattern` iterator with a single SPARQL `SELECT ... WHERE { ... }` so oxigraph's evaluator does the join. Cuts Rust-side iterator churn to zero.
2. **Default `LIMIT`** — `perspective.queryLinks` returns ALL matches by default. A sensible default cap would bound per-query allocation regardless of store size. Behaviour change; needs API discussion.
3. **Streaming response** — serialise solutions to JSON one at a time directly into the WS write buffer instead of building the full Vec first.

Each would need its own test plan + benchmark. They're independent so could land as separate PRs after this one.

lucksus (approved):

PR #831, opened 2026-05-25
feat: typed query DSL + remove dead engine plumbing

HexaField:
## Summary

Implements [`.specs/PLAN_6_TYPED_QUERIES.md`](https://github.com/coasys/ad4m/blob/feat/typed-queries/.specs/PLAN_6_TYPED_QUERIES.md) — the type-system half of the SPARQL 1.2 cleanup arc.

- **Typed query DSL** — `TypedQuery<T>` / `TypedWhere<T>` / `TypedOrder<T>` / `TypedIncludeMap<T>` infer compile-time constraints from each model's declared TS fields. No codegen, no runtime changes. Misspelled property/relation names, wrong value types, and string ops on numeric fields all become compile errors.
- **`$`-projection result inference** — `IncludeExtras<T,I>` propagates `$`-prefixed include keys into the result row type. `{ $signalCount: { from: 'signals', count: true } }` types `row.$signalCount` as `number`; `limit: 1` projections type as `T | null`; otherwise `T[]`. Closes the `(row as any).$x` gap We's docs were living with.
- **`fromSHACL` escape hatch** — `TypedQuery<Ad4mModel>` collapses to the loose `Query` shape automatically, so dynamic models (`Ad4mModel.fromSHACL`) opt out of strict typing by design.
- **Dead engine plumbing removed** — `_engine?` param dropped from `findAll`/`findOne`/`findAllAndCount`/`paginate` and `ModelQueryBuilder.engine()` deleted. Zero call sites in Flux or We pass these; the runtime already ignored them. Test fixtures updated.
- **`querySparql<T>` generic** — `PerspectiveProxy.querySparql` and `PerspectiveClient.querySparql` accept an optional binding type (default `any`).
- **New `SparqlBindings.ts`** — typed `parseLit` / `parseLitNumber` / `parseLitBoolean` / `parseSparqlCount` + `LinkBinding` / `CountBinding` interfaces for raw-SPARQL consumers.

Companion PRs:
- coasys/flux — typed bindings for the ~12 raw SPARQL methods, replaces local `parseLit` with the core re-export
- jhweir/we — doc updates for `$`-projection examples (drops `as any` and `count: true as const` workarounds)

## Test plan

- [x] `npx tsc --noEmit` clean (only pre-existing `AIClient.ts` error remains)
- [x] `npx jest --forceExit` — **434 tests pass** across 15 suites
- [x] New `core/src/model/types-typed-query.test.ts` adds 30+ compile-time assertions (`@ts-expect-error` markers on typos / wrong types, `expectExact`/`expectAssignable` on inference shapes, end-to-end inference through `findAll`/`findOne`)
- [x] Rebuilt `connect` (per CLAUDE.md), verified consumers (Flux api, We app-framework) type-check against the new types

## Notes for reviewers

- The mapped types iterate `[K in keyof T]` (not `[K in ModelDataKeys<T>]`) and re-narrow inside — works around a TS 4.9 quirk where complex computed type aliases collapsed to `string | undefined` in the key position, breaking relation extraction for the `AgentProfile`-shaped optional-HasOne pattern.
- The discriminated `TypedIncludeProjection<T>` union (`{count: true} | {limit: 1} | {limit?: number}`) is what preserves literal narrowing under TS 4.9 — `<const Q>` (TS 5.0+) would be cleaner but isn't available here.
- `_engine` removal is a breaking change at the *type* level only — the runtime ignored it. Surface in the release notes for any out-of-monorepo SDK consumer that was passing the positional arg.

## Release Notes

* **Refactor**
  * Updated Ad4mModel query methods (findAll, findOne, findAllAndCount, paginate, count, query) with improved type signatures and removed deprecated parameters
  * Removed deprecated engine() method from ModelQueryBuilder
  * Enhanced query parameter typing for where, order, properties, and include operations

* **New Features**
  * Added SPARQL binding utilities for parsing query results with type-safe helper functions

[Review Change Stack](https://app.coderabbit.ai/change-stack/coasys/ad4m/pull/831?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/ad4m-hooks/react/src/useLiveQuery.ts
+++ b/ad4m-hooks/react/src/useLiveQuery.ts
@@ -3,6 +3,7 @@ import {
   PerspectiveProxy,
   Ad4mModel,
   Query,
+  TypedQuery,
   ParentScope,
   PaginationResult,
   ModelQueryBuilder,
@@ -115,7 +116,11 @@ export function useLiveQuery<T extends Ad4mModel>(
   const [pageNumber, setPageNumber] = useState(1);
   const [totalCount, setTotalCount] = useState(0);
 
-  const modelQueryRef = useRef<ModelQueryBuilder<T | Ad4mModel> | null>(null);
+  // Widened to ModelQueryBuilder<Ad4mModel> because the typed chain methods
+  // (`where`, `order`, `include`) are contravariant in T — invariance prevents
+  // the per-branch `ModelQueryBuilder<T>` / `ModelQueryBuilder<Ad4mModel>` from
+  // unifying directly. Subscribe callbacks cast back to T[] at the boundary.
+  const modelQueryRef = useRef<ModelQueryBuilder<Ad4mModel> | null>(null);
 
   /** Build the full query from user options + pagination + parent + instance filter. */
   function buildQuery(): Query {
@@ -137,12 +142,18 @@ export function useLiveQuery<T extends Ad4mModel>(
     return q;
   }
 
-  /** Build a ModelQueryBuilder for either a class constructor or a string class name. */
-  function makeQueryBuilder(q: Query) {
+  /** Build a ModelQueryBuilder for either a class constructor or a string class name.
+   *  The cast to `TypedQuery<T>` is necessary because the hook is generic over an
+   *  arbitrary `T extends Ad4mModel` — TypeScript can't statically resolve whether
+   *  T has typed fields, so the deferred `TypedQuery<T>` conditional doesn't accept
+   *  the raw `Query` shape we build at runtime. End-user callers of the static API
+   *  retain strict typing; this cast is the documented escape hatch for generic
+   *  meta-utilities that build queries dynamically. */
+  function makeQueryBuilder(q: Query): ModelQueryBuilder<Ad4mModel> {
     if (typeof model === "string") {
-      return Ad4mModel.query(perspective, q).overrideModelClassName(model);
+      return Ad4mModel.query(perspective, q as TypedQuery<Ad4mModel>).overrideModelClassName(model);
     }
-    return (model as ModelCtor<T>).query(perspective, q);
+    return (model as ModelCtor<T>).query(perspective, q as TypedQuery<T>) as unknown as ModelQueryBuilder<Ad4mModel>;
   }
 
   function mergeEntries(oldEntries: T[], newEntries: T[]): T[] {
--- a/ad4m-hooks/vue/src/useLiveQuery.ts
+++ b/ad4m-hooks/vue/src/useLiveQuery.ts
@@ -4,6 +4,7 @@ import {
   PaginationResult,
   PerspectiveProxy,
   Query,
+  TypedQuery,
   ParentScope,
 } from "@coasys/ad4m";
 import {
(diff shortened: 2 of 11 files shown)


HexaField:
## CI status update

CodeRabbit comments addressed in b29e6e5d:
- `PropertyKeysOf` / `RelationKeysOf` / `RelatedModel` array branches now use `NonNullable<T[K]>`, picking up optional-array fields (`?: Foo[]`, `?: string[]`) that were previously dropped because `Foo[] | undefined extends Foo[]` evaluates false
- `PerspectiveClient.querySparql<T = any>` aligned with `PerspectiveProxy.querySparql<T = any>` so omitted type params land on `any` not `unknown`
- Added an `OptionalArrayLike` fixture to `types-typed-query.test.ts` covering the `?: Foo[]` / `?: string[]` regression — all 434 tests still pass

Companion hooks fix in ef5d4ac9:
- `useLiveQuery` (vue + react) casts at the `Ad4mModel.query` / `ModelCtor<T>.query` boundary since the generic-T hook can't statically resolve whether T has typed fields, so the deferred `TypedQuery<T>` conditional rejects the raw `Query` shape built at runtime
- Widened the `modelQuery` / `modelQueryRef` variable to `ModelQueryBuilder<Ad4mModel>` to side-step `ModelQueryBuilder<T>`'s contravariant chain-method variance (`where`/`order`/`include`)
- This fix unblocked the matching coasys/flux#599 — Flux CI is now fully green ✅

### Current check status

✅ build-and-test, plugin-ad4m-tests, integration-tests-email-verification, integration-tests-mcp, integration-tests-model — all passing

❌ `integration-tests-js`, `integration-tests-multi-user-simple` — failing on the same root cause: `Failed to unpack hApp: ... unknown field 'relay_url', expected one of 'name', 'description', 'roles', 'allow_deferred_memproofs', 'bootstrap_url', 'signal_url'`. This is a Holochain runtime/bundle manifest format mismatch — the test runtime rejects the `relay_url` field while the loaded hApp bundle includes it.

This branch is at the same commit base as `dev` (0 commits behind) and the failures reproduce on retry, so they're deterministic. But dev's last successful integration-tests-js / multi-user-simple run was 2026-05-22 — three days before this PR's first CI run — so the external state has had time to drift (Holochain CLI version on CI workers, cached hApp bundles from `https://esm.sh/@coasys/ad4m@0.5.0`, bootstrap server config).

All changes here are TypeScript-only and touch `core/src/model/*` / `core/src/perspectives/*` / `ad4m-hooks/*`. None of them affect hApp manifest generation, Holochain version, or test infrastructure. Happy to dig deeper if there's a connection I'm missing — but my read is these need a separate fix at the bundle/runtime layer.

HexaField:
## Local reproduction — failures are CI-environment-specific, not a code regression

Ran both failing suites locally against the same `ad4m-executor` binary the test runner uses:

| Test suite | CI on this branch | Local |
|---|---|---|
| `integration-tests-js` (\`pnpm run test-integration\`) | 14 failures, all cascading from \`Failed to unpack hApp: ... unknown field 'relay_url'\` | **66 passing, 0 cascade failures.** Single fail is unrelated: \`AssertionError: expected '0.13.0-test-2' to equal '0.13.0-test-6'\` in the Runtime > runtimeInfo test — local binary is from May 22 with the old version string, would clear after a \`cargo build --release --bin ad4m-executor\`. |
| `integration-tests-multi-user-simple` (\`pnpm run test-multi-user-simple\`) | 2+ failures, same \`relay_url\` cascade | **33 passing, 0 failing.** |

The \`relay_url\` hApp manifest decode errors do not reproduce locally. The local Holochain runtime + bundle config decodes the same bundles cleanly — so the CI workers have either a stale Holochain CLI version or cached hApp bundles from \`https://esm.sh/@coasys/ad4m@0.5.0\` that drifted between dev's last green run (2026-05-22) and this PR's first run (2026-05-25). My TypeScript-only changes don't touch hApp manifests, Holochain version, or test infrastructure.

Recommend either:
1. Bump/clear the CI worker cache for the integration-tests jobs and re-run, or
2. Land this PR as-is and address the bundle/runtime drift in a follow-up infra PR (since it likely affects dev too on the next CI trigger)

lucksus (approved):

PR #833, opened 2026-05-27
fix(p-diff-sync): bound validation retries for permanently-missing chunks

HexaField:
## Summary

Fixes a validation storm in AD4M's Holochain conductor where ops referencing chunked `PerspectiveDiffEntryReference` entries get stuck retrying app validation forever once their chunk ops become unreachable. One production incident showed hundreds of unique broken parent ops generating ~190 WARN log lines/sec from `app_validation_workflow`, indefinitely.

## The bug

`perspective_diff_sync_integrity::validate()` iterates `pdiff_ref.diff_chunks` and calls `must_get_valid_record(chunk_hash)` for each chunk. When the chunk record can't be fetched (cascade miss), it returns `ValidateCallbackResult::UnresolvedDependencies(...)`. Holochain's `app_validation_workflow` re-queues such ops on a cadence that floors at 100ms once `>29` ops are awaiting deps:

```rust
let interval = 2900u64.saturating_sub(outcome_summary.missing as u64 * 100) + 100;
```

There is no abandon-after-N-retries path — see the TODO in `validation_query.rs` near the `LIMIT 10000` query. The result: each stuck op is re-validated ~10x/sec forever.

The production trigger is "author committed a large diff (>500 link expressions, so the `CHUNKING_THRESHOLD` in `commit.rs`), parent gossipped, chunks didn't, then the author went offline / wiped state before the chunks propagated."

## The fix

Remove the chunk-presence check from `validate()`. The HDI cannot distinguish a transient cascade miss from a permanent miss inside a validation callback (all `must_get_*` helpers short-circuit the WASM with `UnresolvedDependencies` on absence, so we cannot fall through to `Invalid`), so the only options are loop-forever (current) or skip the check. Skipping is safe because chunk presence is already enforced at consumption time:

* `handle_broadcast` calls `load_diff_from_entry` before advancing `current_revision`, so peers cannot fast-forward onto a chunked reference whose chunks they cannot fetch. (Covered by the existing `test_chunked_broadcast_does_not_update_revision_on_failure`.)
* `pull` / `render` propagate chunk-fetch errors as zome-call errors.

A peer that holds the parent op but cannot resolve chunks simply ignores the data until (a) the chunks gossip in, or (b) the user resyncs.

**Tradeoff considered:** the chunk check was defense-in-depth — its removal means a peer will store the parent op in DHT even when chunks are missing. Coordinator-level surfaces already handle that gracefully. The alternative (Option B in the original plan) was to fix the workflow itself in the coasys holochain fork; that's a wider-impact change touching `app_validation_workflow.rs` and is deferred.

## Regression test

`bootstrap-languages/p-diff-sync/hc-dna/tests/sweettest/src/test_validation_storm.rs` reproduces the storm deterministically:

1. Spin up Alice + Bob conductors (networked).
2. Alice commits 35 chunked diffs (600 link expressions each).
3. Read the parent action+entry+signature from Alice's authored DB for each commit.
4. Inject `ChainOp::StoreRecord` ops for each parent directly into Bob's DHT db via `insert_op_dht`. Bob now holds N parent ops but no chunk ops.
5. Drop Alice's conductor so chunks become permanently unreachable.
6. Poll Bob's DHT for 25 seconds, summing `num_validation_attempts` across the injected ops and counting how many remain in validation limbo.

35 ops exceeds the "many missing → 100ms interval" threshold (>29) so the test exercises the pathological retry rate.

**Pass criteria:**
* Average attempts per op < 4.
* No ops still pending in validation limbo at the end of the window.

**Verified behavior:**
* With the fix applied: total 70 attempts across 35 ops (avg 2.00), every op reaches terminal state within 1s. ✅
* Without the fix: total 351 attempts (avg 10.03), 6 ops stuck in AwaitingAppDeps after 25s, assertion FAILS. ❌

## Other changes in this PR

* `chore(p-diff-sync): unbreak existing sweettests against current holochain` — the sweettest suite had bit-rotted (`SweetConductor::from_standard_config` → `::standard`, `conductors.iter_mut()` no longer compiles, `LinkExpression.timestamp` is `String` not `DateTime`). Minimal fixes so the whole crate compiles as a unit — required for the new test to even run.
* `chore: move p-diff-sync sweettest into the ad4m root workspace` — building from the nested `hc-dna/Cargo.toml` workspace misses the kitsune2 rev patches in the ad4m root `[patch.crates-io]` and the build fails with "missing transport field". WASM zomes stay in `hc-dna` so the WASM build pipeline is unchanged.

## Test plan

- [x] `cd bootstrap-languages/p-diff-sync/hc-dna && CARGO_TARGET_DIR=target RUSTFLAGS='--cfg getrandom_backend="custom"' cargo build --release --target wasm32-unknown-unknown -p perspective_diff_sync -p perspective_diff_sync_integrity && hc dna pack workdir && hc app pack workdir` — WASM zomes build cleanly.
- [x] `cargo test -p perspective_diff_sync_tests --test integration test_chunked_diff_with_missing_chunks -- --nocapture` — passes with fix (avg 2.00 attempts/op).
- [x] Revert the integrity zome change, rebuild WASM, re-run test — fails as expected (avg 10.03 attempts/op, 6 ops stuck).
- [x] `cargo fmt -p perspective_diff_sync_integrity -p perspective_diff_sync_tests` clean.
- [x] `cargo clippy -p perspective_diff_sync_tests --tests` — finishes with no errors.

## Open questions

* `app_validation_workflow.rs` retry interval and `validation_query.rs` abandon TODO are still open in the coasys holochain fork. The right long-term fix is conductor-side. This PR is a defensive zome-level workaround that addresses the immediate symptom.
* `transport-iroh` is forced on for the sweettest because tx5 requires `go` to be installed on the build host and iroh is the path ad4m's main builds already exercise. If CI uses tx5, the sweettest feature list will need to mirror that.

## Summary by CodeRabbit

* **Tests**
  * Added a regression test reproducing chunked-diff validation storms and updated test suite utilities and many tests for more robust/consistent behavior.
* **Bug Fixes**
  * Changed validation behavior to avoid re-queuing on missing chunk references, reducing repeated dependency re-checks and potential validation loops.

[Review Change Stack](https://app.coderabbit.ai/change-stack/coasys/ad4m/pull/833?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14358,6 +14358,43 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
+name = "perspective_diff_sync_integrity"
+version = "0.0.1"
+dependencies = [
+ "chrono",
+ "derive_more 0.99.20",
+ "hdi",
+ "hdk",
+ "holo_hash",
+ "holochain_serialized_bytes",
+ "serde",
+]
+
+[[package]]
+name = "perspective_diff_sync_tests"
+version = "0.1.0"
+dependencies = [
+ "chrono",
+ "futures",
+ "hdi",
+ "hdk",
+ "holochain",
+ "holochain_serialized_bytes",
+ "holochain_state",
+ "holochain_trace 0.7.0-dev.1",
+ "holochain_types",
+ "holochain_zome_types",
+ "perspective_diff_sync_integrity",
+ "pretty_assertions",
+ "rand 0.8.5",
+ "rusqlite",
+ "serde",
+ "serde_json",
+ "tokio",
+ "uuid 1.18.1",
+]
+
+[[package]]
 name = "pest"
 version = "2.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,8 @@ members = [
     "cli",
     "rust-client",
     "rust-executor",
-    "ui/src-tauri"
+    "ui/src-tauri",
+    "bootstrap-languages/p-diff-sync/hc-dna/tests/sweettest",
 ]
 
 [patch.crates-io]
--- a/bootstrap-languages/p-diff-sync/hc-dna/Cargo.lock
+++ b/bootstrap-languages/p-diff-sync/hc-dna/Cargo.lock
@@ -3,66 +3,6 @@
 version = 4
 
 [[package]]
-name = "addr2line"
-version = "0.25.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
-dependencies = [
- "gimli 0.32.3",
-]
-
-[[package]]
-name = "adler2"
-version = "2.0.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
-
-[[package]]
-name = "adler32"
-version = "1.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
-
-[[package]]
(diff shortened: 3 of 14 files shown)


lucksus (changes requested):
I don't think this is a good solution - just not having chunks be a dependency. I get how this resolves the validation storm, but it also means that we just missing diffs. When the toplevel entry gets processed as empty, perspective_instance is not going to reevaluate the link language later.

This does need a fix in Holochain I guess.

PR #854, opened 2026-06-08
feat(api): OpenAI-compatible /v1 endpoint surface (chat, embeddings, audio, realtime)

HexaField:
## Summary

Adds an industry-standard, OpenAI-compatible HTTP/WebSocket API surface mounted at `/v1`, alongside the existing AD4M WS-RPC transport.  Any tool that speaks OpenAI — the official `openai` SDKs, LangChain, LlamaIndex, the Vercel AI SDK, Cursor/Continue, Open WebUI, the `llm` CLI — can point its `base_url` at a running AD4M executor and just work.

All five phases from the design proposal land together:

| Phase | Endpoint(s) | Status |
|---|---|---|
| 1 | `GET /v1/models`, `POST /v1/chat/completions` (non-streaming), `POST /v1/completions` (legacy), `POST /v1/embeddings`, OpenAI error envelope | ✅ |
| 2 | `POST /v1/chat/completions` with `stream: true` — token-by-token SSE | ✅ |
| 3 | `POST /v1/audio/transcriptions` — multipart upload, batch Whisper | ✅ (WAV 16 kHz mono PCM first cut) |
| 4 | `POST /v1/audio/speech` — TTS | ✅ via OpenAI-compatible remote passthrough |
| 5 | `GET /v1/realtime` — WebSocket streaming STT (OpenAI Realtime subset) | ✅ |

The native WS-RPC `ai.*` surface, `core/AIClient`, Flux, and the launcher are unchanged.  Existing clients see no behaviour change.

### New module `rust-executor/src/api/openai_compat/`

| File | Role |
|---|---|
| `mod.rs` | module wiring |
| `router.rs` | axum router mounted at `/v1` and `/api/v1/openai/v1` |
| `types.rs` | serde structs mirroring OpenAI's wire schema |
| `errors.rs` | `{ "error": { "message", "type", "param", "code" } }` envelope + `IntoResponse` |
| `model_selector.rs` | resolve `model` strings → AD4M `model_id` (id → name → `"default"`) |
| `models.rs` | `GET /v1/models` |
| `chat.rs` | `POST /v1/chat/completions` (oneshot + SSE) and `POST /v1/completions` |
| `embeddings.rs` | `POST /v1/embeddings` — raw `f32` arrays |
| `audio.rs` | `POST /v1/audio/transcriptions` (multipart STT) + `POST /v1/audio/speech` (TTS) |
| `tts_passthrough.rs` | OpenAI-compatible upstream passthrough for `/v1/audio/speech` |
| `realtime.rs` | `GET /v1/realtime` WebSocket (subset of OpenAI Realtime transcription) |

### AIService surface additions

Two new public methods on `AIService` so the OpenAI layer can drive the existing per-model LLM threads without bypassing them:

- `prompt_messages(model_id, messages: Vec<(role, content)>) -> PromptResult` — builds an ephemeral in-memory `AITask` for the duration of the call (no DB writes), spawns it on the per-model thread, prompts, removes.  Stateless by design — backs `/v1/chat/completions` and `/v1/completions`.
- `prompt_messages_stream(model_id, messages) -> (mpsc::Receiver<String>, oneshot::Receiver<PromptResult>)` — same shape but the LLM thread forwards each Kalosm token through an mpsc channel.  Backs `stream: true`.

Plus:

- `transcribe_buffer(model_id, samples, auth_token) -> String` — open / feed / drain / close loop around the existing transcription engine.  Backs the batch `/v1/audio/transcriptions` endpoint.
- A new `LLMTaskRequest::PromptStream` variant in the per-model thread's command channel.  Local Kalosm calls `task.run(prompt).into_stream()` and forwards tokens; remote upstreams collapse to a single terminal chunk (a native streaming upstream client is a follow-up).

### Auth + billing

Reuses the existing JWT extractor (`AuthContext`) and capability framework verbatim.  The OpenAI layer changes the *shape* of requests, not the *policy*:

- `Authorization: Bearer <jwt>` (or `?token=` for the WebSocket realtime endpoint, same as `events_ws`)
- Capabilities: chat/completions/embeddings → `AI_PROMPT`, transcriptions + realtime → `AI_TRANSCRIBE`, speech → `AI_PROMPT`, models → `AI_READ`
- `bill_compute` deducts per-user credits with the existing operation labels (`ai_prompt`, `ai_embedding`, `ai_transcription`, `ai_tts`).
- `InsufficientCredits` → HTTP 429 with `type: "insufficient_quota"` matching OpenAI's quota-exhaustion shape so SDKs that retry on 429 do the right thing.

### Error envelope

Every error response carries the OpenAI envelope:

```json
{ "error": { "message": "...", "type": "invalid_request_error", "param": null, "code": "..." } }
```

With correct HTTP status codes: 400 / 401 / 403 / 404 / 429 / 500 / 501.

### Audio decode (Phase 3 scope note)

The batch transcription endpoint accepts **16 kHz mono PCM WAV** out of the box — the simplest container that requires no new dependencies and matches Whisper's reference format.  Other containers (mp3, m4a, ogg, flac) return a clear 400 pointing at the `Content-Type` so clients can transcode locally; full container support via `symphonia` + `rubato` is a self-contained follow-up that wasn't worth pulling into the same diff.

### TTS backend (Phase 4 scope note)

`/v1/audio/speech` forwards to a configured OpenAI-compatible upstream via `reqwest` (no new dependencies).  The proposal calls out passthrough as a valid Phase-4 deliverable alongside local Kokoro/Piper backends; the local engines pull in `ort` + voice model assets that deserve their own review and didn't make sense to ship in the same PR as the rest of the surface.

When no upstream is configured the endpoint returns 501 with a clear message pointing at the configuration path.

## Routing

The `/v1` surface mounts at two paths:

- `/v1/...` — the canonical OpenAI path
- `/api/v1/openai/v1/...` — alias for proxies that hard-code the `/api/v1` prefix from AD4M's native surface

Both share handlers and `AppState`.

## Test plan

- [ ] `cargo check -p ad4m-executor --lib` — clean
- [ ] `cargo test -p ad4m-executor --lib api::openai_compat` — handler unit tests
- [ ] Manual conformance: official `openai` Python SDK against a live executor:
  - [ ] `client.models.list()`
  - [ ] `client.chat.completions.create(stream=False)`
  - [ ] `client.chat.completions.create(stream=True)`
  - [ ] `client.embeddings.create()`
  - [ ] `client.audio.transcriptions.create(file=open("clip.wav", "rb"))`
  - [ ] `client.audio.speech.create()` (with TTS upstream configured)
  - [ ] Realtime WS via `openai.audio.transcriptions.stream()` (subset)
- [ ] Auth/billing parity: 401 on missing JWT, 403 on missing capability, 429 on `InsufficientCredits`
- [ ] Parity: `client.chat.completions.create` text matches `ai.prompt` over WS-RPC for the same model/prompt
- [ ] CircleCI green (`cargo fmt --check` + `build-and-test`)

## What's still deferred (not in this PR)

- **Full audio container support** (`symphonia` + `rubato` for mp3/m4a/ogg/flac decode + resample). Out of scope; clients should transcode to 16 kHz mono WAV today.
- **Local TTS backend** (Kokoro via `ort`, or Piper). `ModelType::TextToSpeech` enum variant, a dedicated TTS-typed model registration, and the local backend itself ship in a follow-up.
- **Tokenizer-exact `usage` counts.** Local Kalosm prompts still report `chars/4` estimates; remote upstreams already return exact counts and we forward those.  Wiring exact counts on the local path means reading the Kalosm tokenizer's encoded length per request — straightforward but better tracked alongside the per-token billing follow-up.
- **Per-model concurrency.** Each model is currently a single thread processing requests serially.  A public OpenAI endpoint invites concurrent load; a small worker pool per model + HTTP 429 backpressure is the obvious next step.

## Companion / follow-up artefacts

None for this PR — there's no client-side migration needed.  External tools point their `base_url` at the executor and that's it.  AD4M-native clients keep using `core/AIClient` against the WS-RPC surface unchanged.

cc @lucksus @data-bot-coasys for review.

## Summary by CodeRabbit

* **New Features**
  * Added an OpenAI-compatible API for model listing, chat and legacy completions, embeddings, speech generation, audio transcription, and realtime transcription.
  * Added streaming chat responses through server-sent events.
  * Added OpenAI-compatible error responses, model selection, usage reporting, and quota handling.
  * Added WAV audio transcription for 16 kHz mono PCM uploads.
* **Tests**
  * Added coverage for request parsing, audio decoding, streaming, input conversion, billing, and error formatting.


marvin-bot-coasys (commented):
Metering audit — posted by @marvin-bot-coasys at @lucksus's request as a review-assist against the Unyt integration's charging path on `dev`. All line refs are on the PR head (`03b724446b`) unless marked *dev*.

**Bottom line**: reuses `crate::billing::bill_compute` (good — same helper, not a re-implementation), but the *charging behavior* diverges from the WS-RPC `ai.*` path on `dev` in several ways, at least two of which look like bugs.

---

### 1. Prompt/embed billing regression cuts both ways

- **On `dev`** (`api/ai_ws.rs:239`, `:259`): WS-RPC `ai.prompt` and `ai.embed` never call `bill_compute`. Only `check_compute_credits_ws` (pre-check), then run. The only path that actually debits AI credits on `dev` today is the transcription worker in `ai_service/mod.rs:1341` — `word_count × host_rate(model_name)`, op `"ai_transcription"`.
- **On this PR**: every `/v1` endpoint calls `bill_compute(email, 1.0, "ai_prompt"|"ai_embedding"|"ai_tts")`. So the *same underlying `AIService::prompt_messages` call* debits 1.0 over `/v1/chat/completions` but 0 over WS-RPC `ai.prompt`. Two transports, same operation, different bills — needs a policy decision either way before this merges.

### 2. Transcription is double-billed on `/v1/audio/transcriptions`

`audio.rs:39-126` bills `("ai_transcription", 1.0)` after `transcribe_buffer` returns. But `transcribe_buffer` delegates to `open_transcription_stream`, whose worker (`ai_service/mod.rs:1341`) *also* bills `ai_transcription` per word × host-rate. Same request → two debits.

Fix: pick one. Either the shared worker is authoritative and the handler doesn't debit, or the handler bills a batch amount and the worker skips billing for this call path (needs a flag through).

### 3. Flat 1.0 credit isn't "metering," it's a token

Every endpoint charges `1.0` regardless of:
- Completion length (`chat.rs:113`, `:158`)
- Number of inputs or `total_tokens` (`embeddings.rs:20-78`)
- Characters synthesized (`audio.rs:128-174` speech)
- Upstream API cost (`tts_passthrough.rs:34` — hits real OpenAI, no upstream token accounting)

Author acknowledges this for streaming at `chat.rs:256-259` — comment says per-token skipped because Kalosm doesn't expose tokenizer counts. Fair for streaming as a v1, but nothing prevents `/v1/embeddings` from multiplying by `inputs.len()`/token estimate, or `/v1/audio/speech` from multiplying by input character count. Otherwise this becomes a per-request tax, not compute-metering.

### 4. `bill_compute` errors are silently dropped

Every handler uses `let _ = bill_compute(...);`. If the user runs out of credits between pre-check and debit (or the pre-check is skipped), the response ships anyway. Compare `ai_service/mod.rs:1350` in the transcription worker, which stops on `InsufficientCredits`.

With flat 1.0 billing the pre-check and the debit converge, so either trust the pre-check and skip the debit call entirely, or honor its return value.

### 5. Streaming path debits inside a non-cancel-safe forwarder

`chat.rs:261` fires the debit inside the tokio task forwarding tokens to the SSE client. If the client disconnects mid-stream, `token_rx.recv()` returns `None` and the debit fires anyway. Fine today because it's flat 1.0, but any move to per-token billing needs this to be cancel-aware.

### 6. Realtime WS auth via `?token=` query string

`realtime.rs` accepts JWT from `?token=` (matches `events_ws` — browsers can't set WS handshake headers). Two things worth flagging even though they're pre-existing:
- Tokens land in nginx access logs, `Referer`, browser history. Not new — but the surface is broader now.
- Only one capability check at upgrade (`realtime.rs:49`, `AI_TRANSCRIBE_CAPABILITY`). WS-RPC does per-RPC capability checks.

### 7. TTS passthrough model-picker is loose

`tts_passthrough.rs:109-117` falls back to any registered LLM-typed model with an OpenAI API key. Meaning any OpenAI key wired for chat can be silently consumed by `/v1/audio/speech`. Users setting up a chat key probably don't intend to authorize TTS spend against it — worth an explicit `type: "tts"` capability on the model config, or at minimum a config-side allowlist.

### 8. `errors.rs:78-87` — 429 shape ✓

`InsufficientCredits → HTTP 429 { "error": { "message": "...", "type": "insufficient_quota", "param": null, "code": "insufficient_quota" } }` is the correct OpenAI shape. Mapping is manual at each call site (`.map_err(|_| OpenAIError::insufficient_quota(...))`); a `From<BillingError> for OpenAIError` would DRY this up.

---

### Suggested changes before merge

1. Fix double-billing on `/v1/audio/transcriptions` (pick one debit site).
2. Decide the prompt/embed policy once: either WS-RPC starts billing too, or `/v1` stops. Don't ship two transports with two answers.
3. Stop discarding `bill_compute` errors — either propagate or don't bother calling.
4. Where amount-per-request is knowable (`embeddings` inputs, `speech` characters, transcription words), use it. Reserve flat 1.0 for streaming-chat only, with a TODO to plumb Kalosm tokens.

### Nice-to-haves

5. Per-call capability check in realtime WS.
6. `From<BillingError> for OpenAIError`.
7. Tighten TTS model-selector — explicit TTS capability on model configs.

lucksus (changes requested):
So, what @marvin-bot-coasys says - and: I think this should have tests. I know running large models is not the right thing for CI but testing the logic, esp. the billing logic, could be done with a very small model running on CPU in tests.

Since Marvin did find differences in who billing works, test could compare against the WS path..

marvin-bot-coasys (commented):
Round-2 re-review posted by @marvin-bot-coasys at @lucksus's request. Line refs are on head `79c084e091`.

**Bottom line: every actionable finding from the round-1 review is addressed. LGTM from the metering angle. Two follow-up notes — one on tokenization accuracy, one on missing tests.**

## Round-1 findings — status

| # | Finding | Response | Status |
|---|---|---|---|
| **1** | Prompt/embed billing divergence WS-RPC vs /v1 | Comment in `chat.rs:117` documents the WS-RPC gap and explicitly punts alignment to a separate PR against `dev`. /v1 chosen as canonical. | ✅ Documented + resolved |
| **2** | Double-billing on `/v1/audio/transcriptions` | Handler-level `bill_compute` removed in `audio.rs`. Only the worker in `ai_service/mod.rs` bills per-word now. | ✅ Fixed |
| **3** | Flat 1.0 credit isn't metering | Proportional billing where feasible: `chat.rs` non-stream + completions → `(prompt_tokens + completion_tokens) / 1000.0`; `embeddings.rs` → `data.len()` per vector; `audio.rs` speech → `chars / 1000.0` min 1.0; `chat.rs` stream still flat 1.0 with explicit TODO. | ✅ Substantially fixed |
| **4** | `let _ = bill_compute(...)` swallowing errors | Replaced with `bill_compute(...)?` via new `From<BillingError> for OpenAIError` impl. Stream path uses `if let Err(e) { log::warn!(...) }` — reasonable because the response body is already open. | ✅ Fixed |
| **8** | Suggested `From<BillingError> for OpenAIError` conversion | Landed exactly at `errors.rs:127-138`. Beautiful. | ✅ Fixed |
| 5 | Stream debit in non-cancel-safe forwarder | Fine while flat 1.0. Attach a TODO to the per-token migration. | ○ Deferred |
| 6 | Realtime `?token=` in query string | Pre-existing pattern from `events_ws`. | ○ Pre-existing |
| 7 | TTS passthrough model picker looseness | Nice-to-have. | ○ Nice-to-have |

## New note — tokenization accuracy (not blocking, worth a TODO)

Proportional billing math depends on `PromptResult::{prompt_tokens, completion_tokens}`. Those aren't real Kalosm tokens — they come from `estimate_token_count(text) = (chars + 3) / 4`, a heuristic assuming "≈ 4 chars per token."

That's OpenAI's English rule-of-thumb. Materially off in three common cases:

- **Non-Latin scripts** (CJK, Arabic, Hindi) — tokens often ~1 per char → billing **under-estimated ~4×**
- **Code / JSON** — often ~2 chars/token → billing **under-estimated ~2×**
- **Repetitive text** — BPE compresses well → billing **over-estimated**

Users on non-English workloads will feel it. Vastly better than flat 1.0, so not blocking, but worth a TODO at the definition of `estimate_token_count` (or a per-model correction factor) pointing at plumbing real tokens through when the Kalosm backend exposes them.

## Missing — tests for the metering changes

The metering behavior changed a lot in these commits, but `openai_compat/tests.rs` didn't grow: still the same 32 tests, none exercising the actual billing calls. What's missing to make the guarantees regression-proof:

1. **Proportional-billing math** — pure unit tests over `estimate_token_count` + the amount formulas per endpoint. Given fixed inputs, assert the expected credit amount.
2. **`From<BillingError> for OpenAIError`** — round-trip: construct `BillingError::InsufficientCredits`, convert, assert 429 status + `insufficient_quota` code. `BillingError::Other(...)` → 500 with `Billing operation failed`.
3. **Guard against double-billing regression on `/v1/audio/transcriptions`** — mocked `bill_compute` (or a tracing hook), assert **exactly one** debit per request.
4. **Per-endpoint contract test**: for each of `/v1/completions`, `/v1/chat/completions` (oneshot + stream), `/v1/embeddings`, `/v1/audio/speech` — assert the debit uses the correct operation label and the amount is proportional to input size.
5. **`?`-propagation test** — with a stub `bill_compute` that returns `InsufficientCredits` **after** the LLM call, the handler must reply 429 (not 200 with a body that took real compute).

(1) and (2) are pure-function unit tests, trivial to add. (3)–(5) need a tracing/mock seam, which is a design decision — could be done with a `#[cfg(test)]` counter on `bill_compute` or by extracting the metering into a trait injectable by the handler.

If any of these existed under a different path, sorry for the noise — happy to point at it.

lal-bot-coasys:
Test coverage for @marvin-bot-coasys's round-2 missing-tests ask is in `6648d6d73`, on top of a fresh `dev` merge (branch was BEHIND).

**Extracted the billing formulas** into a new `openai_compat/billing_amounts.rs` so the amount computation for each endpoint is a pure function that tests can pin — the handlers now route through `billing_amounts::{chat_or_completion_amount, stream_prompt_amount, embedding_amount, speech_amount}` instead of inlining `.max(0.001)` / `.max(1.0)` at the call sites. Same numbers, single source of truth per endpoint.

Coverage against Marvin's five items (32 → 51 tests):

1. **Proportional-billing math** — pure unit tests over `estimate_token_count` (empty, rounding, 100-char, unicode-scalar-not-bytes) and per-endpoint formulas: chat floors at `MIN_PROMPT_BILL`, scales linearly above 1000 tokens; embeddings bill per vector with floor of 1; speech bills chars/1000 with floor of 1; streaming stays flat 1.0 with the formula pinned so switching to per-token later is a deliberate change.
2. **`From<BillingError> for OpenAIError` round-trip** — `InsufficientCredits` → 429 with `insufficient_quota` type/code, `UserNotFound` and `Other` → 500 with generic `"Billing operation failed"`, plus explicit assertion that neither branch leaks the user email or internal error detail into the client body.
3. **Double-billing regression guard on `/v1/audio/transcriptions`** — the handler source is scanned at test time and asserted to contain no `bill_compute` call; if someone re-introduces one, the test fails with a message pointing at your round-1 finding #2. Also asserts no `transcription_amount` helper exists in `billing_amounts.rs` (a canary against the natural first step of re-adding handler billing). Kept structural rather than runtime-mock because a real seam (trait-based bill_compute or cfg(test) counter) is a design decision that touches `billing.rs` and belongs in its own change — happy to open that follow-up if you want the deeper coverage.
4. **Per-endpoint operation-label integrity** — the label strings `"ai_prompt"` / `"ai_embedding"` / `"ai_tts"` are contractual (they feed the compute_events log). Source-scan tests assert exact spelling per handler and count of occurrences per file, so a silent `"ai_promt"` typo or a missing bill site fails.
5. **`?`-propagation** — a helper function reproduces the exact `bill_compute(...)?` call-site shape used in every handler and asserts that a `BillingError::InsufficientCredits` return converts terminally to a 429 `OpenAIError`. Composes with (2) to give full flow coverage of the pattern without needing a runtime seam.

**Tokenization TODO** you flagged is now on `estimate_token_count` itself, with the specific failure modes (CJK/Arabic/Hindi under-estimate ~4×, code/JSON ~2×, repetitive text over-estimates) documented so future accuracy work has a clear place to land.

Not touched (deferred per your round-2 table): the streaming per-token migration (stream_prompt_amount stays flat 1.0 with a locked-in test), realtime `?token=` auth (pre-existing pattern), TTS passthrough model-picker looseness (nice-to-have).

Local: `cargo fmt` clean, `cargo test --lib openai_compat` = 51/51 pass. CircleCI running on the new commit.

marvin-bot-coasys:
Round-3 review posted by @marvin-bot-coasys at @lucksus's request. Line refs are on head `6648d6d7`.

**Bottom line: the billing-formula extraction and pin tests are good; the 4 source-scanning "static assertion" tests should be replaced with a real behavioural test before merge.**

---

## ✅ What lands cleanly

Lal's refactor addresses the round-2 ask directly:

- `billing_amounts.rs` — the pure-function extraction is exactly what round-2 wanted. Single source of truth for `chat_amount`, `chat_stream_amount`, `embeddings_amount`, `speech_amount`. Each handler now delegates.
- `estimate_token_count` promoted to `pub(crate)` so tests can pin the current `(chars + 3) / 4` heuristic — good. The `TODO` documenting where the heuristic is materially wrong (CJK, code, repetitive tokens) is a useful breadcrumb for future accuracy work.
- **15 of the 19 new tests are genuinely useful**: `estimate_token_count` pinning, per-endpoint amount helpers, `From<BillingError> for OpenAIError` round-trip (429 for `InsufficientCredits`, redacted 500 for `Other`), `?`-propagation through a handler-shaped call site. These lock in the actual credit-calculation contract and would catch a real math regression.

## ⚠️ Concerns

### 1. The 4 source-scanning tests are the wrong tool (blocking)

Files: `rust-executor/src/api/openai_compat/tests.rs` lines 586-596 (`no_ai_transcription_bill_in_transcriptions_handler`), 610-628 (`transcription_amount_helper_not_defined`), 645-659 (`chat_and_completions_use_ai_prompt_label`), 671-683 (`speech_uses_ai_tts_label`).

These tests `include_str!()` the sibling source file and then assert against substring matches on it. They're testing spelling, not behaviour.

The commit message is candid about why:
> "A full runtime test of 'exactly one debit' would need a mocked `bill_compute` (a trait seam or `cfg(test)` counter injected globally), which is a design decision that touches `billing.rs` itself."

That's the correct test, and it's the one that should land. What we have instead is:

**Concrete ways the static-scan defeats itself as a regression guard:**

- `use crate::billing::bill_compute as debit;` and then `debit(...)` inside the handler → double-bill regression, test passes.
- `crate::billing::bill_compute(...)` fully qualified → test passes.
- A macro that expands to `bill_compute(...)` → test passes.
- A future refactor that turns the handler into a table-driven dispatcher (`pub async fn transcriptions` becomes a thin shim, real work in `handle_transcription`) → the `body` slice from `pub async fn transcriptions` to `pub async fn speech` covers the shim only, and the actual billing site is in a helper that isn't scanned.

**Concrete ways they fail against non-regressions:**

- Add a comment in the transcriptions handler that mentions `bill_compute` (e.g. explaining why it's absent) → test fails with nothing broken. Ironic given the whole test exists to enforce a Do-Not-Add rule.
- Add a helper function between `pub async fn transcriptions` and `pub async fn speech` → the slice window silently grows to include unrelated code.
- Refactor `chat.rs` so oneshot + streaming + legacy completions share one bill site → `chat_and_completions_use_ai_prompt_label` fails on the `>= 3` count with correct code.

**What the test should be**: a `cfg(test)` counter in `billing.rs` (or a trait seam with an `Arc<AtomicUsize>` mock). ~15 lines in `billing.rs`, then in the transcription test:

```rust
let counter = billing::test_counter();
let _ = call_transcriptions_handler(...).await;
assert_eq!(counter.load(Ordering::SeqCst), 0,
    "handler must not bill — worker bills per word");
```

Same shape for the "exactly-one `ai_tts` bill" assertion. That's what "lock in the contract" actually means — the contract is behaviour, not source-code substring.

Recommendation: replace the 4 static-scan tests with the counter-seam version. If the seam feels heavy for one PR, drop the tests and open a follow-up issue; the static-scans give false confidence that's arguably worse than no test.

### 2. `transcription_amount_helper_not_defined` is the weakest of the four

Files: `tests.rs:610-628`.

This one asserts that `billing_amounts.rs` **does not contain** `fn transcription_amount`. The intent — "if a `transcription_amount` helper appears, someone probably wired handler-level billing back in" — is good.

But this test can't distinguish:
- Someone genuinely re-introduced the double-bill regression (bad — should fail)
- Someone added a `transcription_amount` helper because a legitimate second billing path is needed elsewhere (e.g. billing per-user-quota-check in a rate-limiter) (fine — should not fail)
- Someone added `fn transcription_amount_estimate` as a read-only helper for a monitoring endpoint (fine — but silently matches the substring)

Absence-of-a-symbol as a test is almost never the right shape. Prefer the counter approach or delete it.

### 3. Merge instead of rebase (minor)

Commit `3899c897` "Merge remote-tracking branch 'origin/dev' into lal/pr854-tests" pulls in 106 files of unrelated `dev` changes on top of the PR. A rebase onto current `dev` would keep the PR diff to the actual OpenAI-compat work. Not a blocker but adds noise to the review surface. Suggest `git rebase origin/dev && git push --force-with-lease` before merge.

### 4. The billing_amounts formulas are not covered end-to-end (minor)

The pure unit tests for `chat_amount`, `embeddings_amount`, `speech_amount` are good, but there's no test asserting that `chat.rs` actually **calls** the helper (as opposed to inlining the formula and reintroducing drift). One integration-shaped test per handler — `handler_uses_billing_amounts_helper` — would close the loop the static-scans were trying to close, and could be written today against the counter seam.

## Suggested resolution before merge

- [ ] Add a `cfg(test)` bill counter to `rust-executor/src/billing.rs` (small trait seam or `thread_local! AtomicUsize`).
- [ ] Replace `no_ai_transcription_bill_in_transcriptions_handler` and `speech_uses_ai_tts_label` with counter-based behavioural tests.
- [ ] Delete `transcription_amount_helper_not_defined` and `chat_and_completions_use_ai_prompt_label` (source-scan sniffing at label spelling is not a useful invariant).
- [ ] Rebase onto current `dev` to drop the merge commit.
- [ ] Optional: add one `handler_uses_billing_amounts_helper` integration test per endpoint using the counter seam, to guarantee handlers don't drift from the extracted formulas.

Everything else in the commit is good — the formula extraction + per-formula unit tests + `From<BillingError>` conversion tests are exactly the coverage round-2 asked for. Once the source-scan tests are replaced (or dropped with a follow-up issue), this is ready.

marvin-bot-coasys:
Round-4 update posted by @marvin-bot-coasys, taking over from @lucksus's request. Head is now `8bdfbcba7`.

**Bottom line: took the PR over the finish line — replaced the source-scan tests with a proper counter seam, added `/v1` integration tests, fixed a real wire-shape bug the earlier reviews missed. 55/55 tests pass locally.**

## Diff summary

6 files, +586 / −101:

| File | What |
|---|---|
| `rust-executor/src/billing.rs` | New `#[cfg(test)] pub mod test_seam` — thread-local counter of every `bill_compute` call + `ForcedResult` injector for exercising `?`-propagation. Zero-cost in release. |
| `rust-executor/src/api/openai_compat/errors.rs` | New `OpenAIJson<T>` extractor + `From<JsonRejection> for OpenAIError`. Fixes a real wire-shape bug (see below). |
| `rust-executor/src/api/openai_compat/{audio,chat,embeddings}.rs` | Every /v1 handler that takes a JSON body switched from `Json<T>` → `OpenAIJson<T>`. |
| `rust-executor/src/api/openai_compat/tests.rs` | Deleted the 5 source-scan tests. Added 8 behavioural tests (seam self-tests + router-driven integration). |

## Wire-shape bug found (and fixed)

While writing the OpenAI-shape contract test I discovered axum's default `Json<T>` extractor rejects malformed / missing-field request bodies with **422 UNPROCESSABLE_ENTITY + a plain-text body** — not the OpenAI envelope. So any OpenAI SDK client sending an invalid request against the current PR would get:

```
HTTP/1.1 422 Unprocessable Entity
content-type: text/plain

Failed to deserialize the JSON body into the target type: missing field `model` at line 1 column 2
```

… which the SDK can't parse. Fixed by wrapping the extractor: `OpenAIJson<T>` catches `JsonRejection` and maps it to `OpenAIError::invalid_request(...)` → 400 with the correct envelope. Confirmed by the new `chat_completions_invalid_json_returns_400` / `embeddings_invalid_json_returns_400` / `error_envelope_shape_matches_openai` tests.

## Tests replaced

Gone (source-scan sniffing):
- `no_ai_transcription_bill_in_transcriptions_handler`
- `transcription_amount_helper_not_defined`
- `chat_and_completions_use_ai_prompt_label`
- `embeddings_uses_ai_embedding_label`
- `speech_uses_ai_tts_label`

Kept (pure unit tests that DO lock behaviour, from Lal's commit):
- `estimate_token_count_*`
- per-endpoint `*_amount` formula pins
- `From<BillingError> for OpenAIError` conversion
- `?`-propagation

New (behavioural, via counter seam + axum router):
- `seam_records_calls_and_resets` — seam self-test
- `seam_can_force_insufficient_credits` — forced-result injection works
- `transcriptions_handler_does_not_bill` — the round-1 finding #2 double-bill regression guard, now behavioural: drives a real multipart request through the router and asserts `bill_compute` call_count() == 0 after the response. Any regression that re-adds handler-level billing fires this test with a clear message.
- `chat_completions_invalid_json_returns_400`
- `embeddings_invalid_json_returns_400`
- `transcriptions_missing_model_returns_400`
- `speech_missing_input_returns_400`
- `error_envelope_shape_matches_openai` — every fallible /v1 response has the `{ "error": { "message", "type", "param", "code" } }` shape
- `chat_completions_stream_does_not_bill_at_setup` — `stream: true` defers billing to end-of-stream; setup path must not debit up front (companion to round-1 note #5)

## Not in this commit (deliberate)

Three router-driven billing tests I initially wrote (`speech_handler_bills_once_with_ai_tts_label`, `insufficient_credits_from_bill_compute_returns_429`, `other_bill_error_is_redacted_500`) were removed after realising the speech handler gates billing on `user_email_from_token(auth_token).is_some()` — which requires a real JWT + Wallet setup. Driving those through the router with an admin credential (as I was) never reaches the billing code because there's no user email.

The contract is instead covered by three layered assertions already in the file:
1. `speech_amount_per_thousand_chars` pins the exact amount formula.
2. `insufficient_quota_error` + `handler_shape_bill_compute_question_mark_propagates_insufficient_credits` pin the `From<BillingError>` conversion at 429 with the `insufficient_quota` code.
3. `seam_records_calls_and_resets` proves the seam records what the handler would call.

An e2e binary with Wallet + DB + AIService bootstrapped could compose those into a real end-to-end billing test in a follow-up — the seam and the router harness are ready for it.

## Round-1 / round-2 / round-3 findings — final status

| # | Finding | Status |
|---|---|---|
| 1 | WS-RPC vs /v1 billing divergence | documented + resolved (79c084e0) |
| 2 | /v1/audio/transcriptions double-bill | **fixed + behavioural regression test lands here (8bdfbcba7)** |
| 3 | flat 1.0 credit isn't metering | fixed (79c084e0 — per-endpoint proportional) |
| 4 | bill_compute errors dropped | fixed via `?` + `From<BillingError>` (79c084e0) |
| 5 | stream debit in non-cancel-safe fwd | TODO stays in chat.rs::chat_stream (documented) |
| 6 | realtime `?token=***` query string | pre-existing pattern, unchanged |
| 7 | TTS passthrough loose model picker | nice-to-have, unchanged |
| 8 | `From<BillingError>` conversion | landed (79c084e0) |
| — | **JsonRejection wire shape (new)** | **fixed here (8bdfbcba7)** |
| — | **Static source-scan tests (round-3)** | **replaced with counter-seam behavioural tests here (8bdfbcba7)** |

## Verification

- `cargo fmt --check`: clean
- `cargo check --release --tests -p ad4m-executor`: 0 errors
- `cargo test --release --lib -p ad4m-executor api::openai_compat -- --test-threads=1`: **55 tests, 0 failed**

Ready for merge from my end. If Josh (@HexaField) wants to look over the counter-seam design or the `OpenAIJson` extractor before merging that's welcome, but nothing here should be blocking.

rust-executor/src/ai_service/mod.rs
```diff
@@ -851,6 +884,144 @@ impl AIService {
...
                             },
 
+                            LLMTaskRequest::PromptStream(stream_request) => match model {
+                                LlmModel::Remote((ref mut remote_client, ref model_string)) => {
+                                    // Remote upstreams use the non-streaming
+                                    // chat call today; we deliver the full
+                                    // response as a single token chunk so
+                                    // SSE consumers still see the "stream"
+                                    // protocol (one delta + final usage).
+                                    // A native streaming upstream client is
```
lucksus, line 894: ah, @data-bot-coasys, I think we need this for the upcoming tool-calling harness, no?

lucksus (approved):

PR #873, opened 2026-07-16
feat: Graph-Expression Duality

HexaField:
## Summary

**Graph-Expression Duality with Diff-DAG Commit Addressing.** Graph-backed subjects can now be exported as commit-addressed Expressions (Merkle diff-DAG) and mounted back into a perspective. The addressing model converges with p-diff-sync: each commit is `graph://<hash(diff, parents)>`, carrying a `GraphDiff` (additions/removals as canonical N-Triple lines), parent commit IRI(s), and a snapshot hash for integrity verification. The content hash (whole-graph SHA-256) is demoted to a snapshot integrity field — not the primary address.

> **Stacked PR.** Base is `feat/subject-class-named-graphs` (#812), not `dev`. Review #812 first — this PR is the net diff on top of it. It should be merged after #812 lands (or retargeted if #812 is squashed).

## What's in this PR

**Diff-DAG Commit Addressing**
- Primary address: `graph://<hex-sha256(canonical(diff, parent_iris))>` — a Merkle diff-DAG node. Commit carries a `GraphDiff { additions, removals }` and parent commit IRI(s).
- Each graph-backed subject tracks a **head pointer** (latest commit IRI) that advances on export. Genesis commit has no parents; all triples are additions.
- **Snapshot hash** (whole-graph content hash) is carried as an integrity field on the commit, used for bootstrapping and tamper detection.
- Three properties gained over content-hash-only: **scale** (hash the delta, not the graph), **historicity** (parent chain), **merge** (multiple parents for concurrent edits).

**Internal Partitioning**
- Each subject's triples live in a named partition `ad4m://graph/<base>` (`make_graph_iri`). This internal IRI is never surfaced — callers address by subject base or by `graph://<hash>`.
- The graph label is excluded from hashing, so identical triples address identically regardless of partition IRI.

**Reifiers**
- Link metadata is carried as RDF 1.2 quoted triples: `<link:HASH> rdf:reifies <<(source predicate target)>>`. Every graph-backed subject carries reifier triples, and both the commit hash and snapshot hash cover direct triples + reifier triples + reifier metadata.

**Commit Proofs**
- Export attaches a signature over `"<commitIri>\n<timestamp>"` by the calling agent's DID. `mount_expression` verifies the proof **before** loading, then verifies the snapshot hash (when present) and **rolls back on mismatch**.
- A proof is simply "signed by a DID". AD4M treats individual agents and groups (graphs that carry their own DID) identically — there is no separate agent/graph signer role.

**Rust**
- `commit.rs` (new): `canonical_commit_payload`, `commit_hash`, `commit_iri`, `compute_diff`, `HeadTable` for per-subject head tracking.
- `perspective_instance`: `get_graph_as_expression` (computes diff from previous head, creates commit, advances head), `mount_expression` (verifies over commit IRI, checks snapshot hash), `graph_content_hash`, `get_graph_expression`, `unmount_graph`, `mounted_graphs`.
- `snapshot_proof` (build/verify proof bundles over commit IRI) and `mount_table` (trust-tracked mount registry with lineage).
- `sparql_store`: `GraphDumpFormat`, `make_graph_iri`, `content_hash`, `dump_graph`, `canonical_ntriples` (now `pub(crate)`).
- Domain types: `GraphDiff`, `SnapshotProof`, `TrustLevel`, `MountedGraphEntry` (with `parents`, `snapshotHash`), `ExpressionRendered` (with `graphDiff`, `parents`, `snapshotHash`) — all ts-rs exported.
- WS-JSON-RPC handlers for the six graph operations.

**Core (TS)**
- `PerspectiveProxy` / `PerspectiveClient` gain the six graph methods and `GraphExportOptions`.
- `ExpressionRendered` gains `address?`, `snapshotProofs?`, `graphDiff?`, `parents?`, `snapshotHash?`.
- `GraphDiff` type exported from `@coasys/ad4m`.

## Design decisions worth a look

**JSON-LD export is deferred (not a bug).** oxigraph 0.5.x's JSON-LD serializer errors with *"JSON-LD does not support RDF 1.2 yet"* on any reifier-bearing graph. Export ships **three** formats — `nquads-canonical`, `nquads`, `turtle` — and defers `jsonld` until oxigraph supports RDF 1.2 in its JSON-LD writer.

**Test-harness fix.** Signing tests share a process-global agent + wallet. This PR makes `setup_wallet` idempotent and guards `init_global_test_instance` to initialise once, and fixes a pre-existing `perspectives::test` `setup()` that relied on cross-test global-state pollution and failed in isolation.

## Testing

Local verification (this repo has no PR-triggered CI). Rust suite runs single-threaded (`--test-threads=1`) because signing/perspective tests share process-global state:

- **Rust lib suite:** 689 passed, 0 failed, 7 ignored.
- **`cargo fmt`**: clean. **`cargo clippy`**: new modules have zero warnings.
- **Core TS build:** 0 `tsc` errors; rollup emits `lib/index.cjs` + `lib/index.js`.
- Commit-addressing tests: genesis commit structure, determinism, different diffs/parents produce different hashes, parent-order normalization, head-advancement + lineage, compute_diff (additions/removals/empty-to-full), HeadTable get/set.
- Duality round-trip tests: commit IRI ≠ content-hash IRI, snapshot hash integrity, diff + parents on ExpressionRendered, proof verify-before-load, rollback-on-snapshot-hash-mismatch, unsigned commit rejection, unmount idempotency.

## Heads-up for the base branch (not introduced here)

`core/pnpm-lock.yaml` on the base contains unresolved git merge-conflict markers, introduced by an upstream merge. The root workspace lockfile is clean and is what a filtered root install uses, so core builds fine that way — but the core lock should be regenerated at its source.

## Reviewers

@lucksus @data-bot-coasys


core/src/generated/api/GraphSignBy.ts
```diff
@@ -0,0 +1,8 @@
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
+
+/**
+ * Who a graph snapshot is signed by (SPEC_8 §3.3). Plan 8 supports `agent`
+ * only; `graph` and `both` require a `did:graph` signing key (Plan 9) and are
+ * rejected at export time.
+ */
```
HexaField: GraphSignBy is redundant, the idea with ad4m is that agents and groups (graphs with DIDs) are isomorphic and can be treated identically.

core/src/generated/api/SnapshotProofRole.ts
```diff
@@ -0,0 +1,6 @@
+// This file was generated by [ts-rs](https://github.com/Aleph-Alpha/ts-rs). Do not edit this file manually.
+
+/**
+ * The signer role carried by a single `SnapshotProof` (SPEC_8 §3.3).
+ */
+export type SnapshotProofRole = "agent" | "graph";
```
HexaField: Redundant

HexaField (commented):
Mention of "Spec 8" should be completely removed, it is an internal planning spec and the repo code and comments must stand on its own.

core/src/generated/api/GraphSignBy.ts
HexaField: Agreed — deleted. Since an agent and a group are both just DIDs, a proof is now simply "signed by a DID": export always signs with the caller's DID, and verification resolves each proof against its own `signerDid` regardless of whether that DID denotes an individual or a group. No signer-role parameter survives in the Rust API, the WS-RPC surface, or the TS SDK. — Hex ⬡

core/src/generated/api/SnapshotProofRole.ts
HexaField: Agreed — deleted. `SnapshotProof` no longer carries a role; it's `{ signerDid, signerKeyId?, signature, timestamp }`. Verification treats agent-DIDs and group/graph-DIDs identically, so the distinction had no meaning. — Hex ⬡

HexaField:
Done — all internal-planning references removed from code and comments ("Plan 8/9", the `SPEC_8` tags, `§X.Y` pointers, and the opaque `REQ-*` IDs); explanatory prose stays. I left the two pre-existing public-spec citations (`Spec §5.2`, `SHACL-AF §6`) untouched since those reference published specs, not the internal plan — say the word if you'd like those reconsidered too. Commit message and PR body reworded to match. — Hex ⬡

PR #881, opened 2026-08-06
feat(interpretation): runInterpretation over WebSocket + TS client + integration test

data-bot-coasys:
# runInterpretation over WebSocket + TS client + integration tests + overlay accept/reject

**Top of the stack.** Exposes the generic interpreter over the WebSocket API + a TypeScript client + full JS integration tests, so a client runs interpretation into a perspective's own SHACL subject classes with one call — and can **accept/reject** the interpretation overlay's suggestions.

### The stack at a glance

```mermaid
flowchart LR
  A["#879 · base engine<br/>hint → LLM → create_subject"] --> B["#883 · tree-aware<br/>upsert · relations · grouping"] --> C["#886 · provenance overlay<br/>+ human-divergence gate"] --> D["#889 · auto-processor<br/>watch · elect · claim · cursor"] --> E["#881 · WS API + TS client<br/>+ accept/reject overlay"]
  style E fill:#1f6f3f,stroke:#0d3d22,color:#fff
```

| PR | branch | owns |
|----|--------|------|
| **#889** | `feature/interpretation-auto-processor` | Neighbourhood auto-processor: per-perspective watch-loop, debounce/batch, authorship election + synced processing-claim, durable processed-turn cursor on `InterpretationRun.sources` (exactly-once across peers and restarts). |

### One-shot interpretation

```typescript
const bases = await perspective.runInterpretation(
  [{ speaker: "did:key:alice", text: "I'll ship the docs" }],
  "soa://ext/",           // base prefix for new instance URIs
  ["Task", "Belief"],     // optional class filter; omit for all registered classes
  { id: "soa://project/42", predicate: "soa://has-item" },  // optional existingScope: dedup lookup
  { id: "soa://project/42", predicate: "soa://has-item" },  // optional mintScope: link new bases as children
);
// → string[] of created/updated base URIs
```

The executor resolves the perspective's registered SHACL shapes server-side, runs the LLM with the transcript as context, and writes typed instances via `create_subject`/`update_subject`. The §4 provenance gate applies: if a human already owns a value, the LLM's suggestion is staged as an overlay instead of overwriting it.

`existingScope` and `mintScope` mirror the AutoProcessor semantics: `existingScope` constrains the dedup lookup to instances under a subtree; `mintScope` links every freshly-created base as a child of `mintScope.id` via its predicate (upserts of pre-existing instances are NOT re-parented — same rule as the watcher).

### Auto-processor (standing watch)

```typescript
await perspective.addAutoProcessor({
  processorId: "flux-channel",
  sourceScopeQuery: `SELECT ?speaker ?text ?timestamp WHERE {
    ?m <ns://body> ?text .
    ?r <http://www.w3.org/1999/02/22-rdf-syntax-ns#reifies> <<( ?m <ns://body> ?text )>> .
    ?r <ad4m://ontology/author> ?speaker .
    ?r <ad4m://ontology/timestamp> ?timestamp .
  } ORDER BY ?timestamp`,
  interpretationClasses: ["ns://ConversationSubgroup"],
  debounceMs: 500,        // quiet window after last new item
  batchMin: 2,            // wait for at least N turns (default 1)
  batchMax: 32,           // cap per pass
  claimTtlMs: 600_000,    // 10 min — size ≳ sync latency + pass runtime
  sourceWindowMs: 604_800_000,  // optional: 7 days; omit for no window
  existingScope: { id: "soa://project/42", predicate: "soa://has-item" },  // optional
  mintScope:     { id: "soa://project/42", predicate: "soa://has-item" },  // optional
});
```

The executor's watch loop polls every 500ms, gathers new turns via the scope query, debounces, and runs one interpretation pass per ready batch. Multiple peers coordinate via a synced `ProcessingClaim` (min-DID wins) so exactly one peer processes each batch. Already-processed turns are tracked on `InterpretationRun.sources` — durable across restarts, shared via link sync.

**Scope query requirements:** Must bind `?speaker`, `?text`, and `?timestamp`. Author and timestamp come from the body link's **reifier** (`rdf:reifies` → `ad4m://ontology/author` + `ad4m://ontology/timestamp`), not an app-level `ns://author` predicate. A query missing `?timestamp` fails the gather.

### Events

```typescript
await perspective.addAutoProcessorEventListener((event) => {
  console.log(event.step, event.agentDid, event.itemIds);
});
```

| Step | Meaning |
|------|---------|
| `batchReady` | Debounce elapsed; a batch is about to be claimed |
| `claimed` | This peer won the claim and will process |
| `backedOff` | Another peer holds the claim; this peer stood down |
| `awaitingAuthor` | No batch author is online; waiting |
| `notCandidate` | An earlier-in-order online author was elected; this peer stood down |
| `gatheringTranscript` | Preparing the batch payload for the LLM |
| `runningInterpretation` | LLM call in progress |
| `processed` | Pass complete; `bases` has the written instance URIs |
| `shapesMissing` | A configured class's SHACL shape hasn't synced yet; will retry |
| `emptyTranscript` | Batch was empty (all turns already processed) |

### Accept / reject suggestions

When the §4 gate stages an LLM suggestion (instead of overwriting a human-owned value), it appears as an overlay:

```typescript
const pending = await perspective.interpretationOverlays();
// → [{ base, kind: "create"|"update", run, inferred: [[predicate, value], ...] }]

// Accept: LLM's staged value becomes the real value; overlay is deleted.
await perspective.acceptInterpretation(base);           // whole base
await perspective.acceptInterpretation(base, "ns://title");  // one property

// Reject: suggestion dropped. A rejected "create" deletes the instance;
// a rejected "update" keeps the existing real value.
await perspective.rejectInterpretation(base);
await perspective.rejectInterpretation(base, "ns://title");
```

Once an overlay is accepted (deleted), the §4 gate treats the base as human-owned — later runs will not silently overwrite it.

### `@Model` classes for graph querying

Apps can query interpretation state directly off the perspective graph — no knowledge of raw link predicates required. Three classes mirror the Rust SDNA exactly (same predicates, same identity property as class discriminator) and round-trip against what the executor writes:

```typescript
import { InterpretationOverlay, InterpretationRun, AutoProcessorConfig } from "@coasys/ad4m";

// Register once (idempotent) so the perspective's SHACL registry knows the shapes.
await InterpretationOverlay.register(perspective);
await InterpretationRun.register(perspective);
await AutoProcessorConfig.register(perspective);

// Query.
const pending = await InterpretationOverlay.findAll(perspective);  // all nodes carrying ad4m://interp/kind
const runs    = await InterpretationRun.findAll(perspective);      // all completed passes
const configs = await AutoProcessorConfig.findAll(perspective);    // registered processors
```

`InterpretationOverlay.findAll()` is the primary Flux/James target: it returns every node currently carrying an `ad4m://interp/kind` link, i.e. every pending proposal awaiting human accept/reject.

Note: the plain-data DTO returned by `perspective.interpretationOverlays()` is now exported as `InterpretationOverlayInfo` to avoid a class/interface name clash — the `@Model` class `InterpretationOverlay` is the one used for graph querying.

**Class discrimination.** These three classes carry no `rdf://type`/`@Flag` discriminator. Conformance is by required-property presence — a node with `processor_id`+`source_scope_query`+`interpretation_class`+`debounce_ms`+`batch_max`+`claim_ttl_ms` IS an `AutoProcessor`; anything else is not. Same pattern for `InterpretationRun` (`run_id` identity) and `InterpretationOverlay` (`kind` identity). Type flags were dropped as an anti-pattern for subject classes (Nico 2026-08-19).

## Call flow (high level)

```mermaid
sequenceDiagram
  participant C as TS client
  participant WS as Executor WS-RPC
  participant I as Interpretation engine
  participant O as Overlay (#886)
  C->>WS: perspective.runInterpretation(uuid, transcript, basePrefix)
  WS->>WS: capability check + resolve the perspective's SHACL shapes
  WS->>I: run_interpretation
  I->>O: apply_with_overlay (create/update + provenance)
  O-->>WS: created / updated base URIs
  WS-->>C: result
  Note over C,O: later — a human reviews pending suggestions
  C->>WS: acceptInterpretation(base[,prop]) / rejectInterpretation(...)
  WS->>O: copy inferred→real & clear · or delete base / drop suggestion
```

## Stacking / this revision

Merged the rebased **#889** (which carries **#886**) down, including the durable processed-turn cursor and timestamped transcript turns; reconciled the interpretation API (`plan_interpretation_ops_resolved`, `apply_with_overlay`, `existing_relation_links`) and dropped the now-removed `llm_model` from the WS `AddAutoProcessor` handler. Scope queries updated to the reifier pattern (`?speaker`/`?text`/`?timestamp` via `rdf:reifies`).

Latest additions (post-dev merge):

- **Dev merged in** — includes PR #874 typed RDF literals + fn cleanup. All `AutoProcessor` / `InterpretationRun` / `InterpretationOverlay` scalar properties now land on the fast deterministic typed-literal path (Rust SDNA declares no `resolve_language`; TS `@Model` declarations mirror this — the earlier `resolveLanguage: "literal"` decorations were dropped, since they would have selected #874's signed-envelope storage that the Rust writer never produces).
- **Type-flag discriminator removed** — Rust `AUTO_PROCESSOR_SDNA` + `INTERP_RUN_SDNA` no longer declare `rdf://type`/`ad4m://type` scalars; class conformance is by required-property presence. TS `@Model` classes match. Cursor SPARQL (`processed_ids_query`) now keys on `interp/run_id` (identity property) so a bare node carrying only `interp/processor`+`interp/sources` cannot masquerade as a run.
- **`runInterpretation` gains `existingScope`+`mintScope`** — one-shot API now matches AutoProcessor for parent-scope dedup + mint-scope child linking. Reuses the watcher's `partition_created` + `write_mint_scope_links` helpers.
- **`LazyDid` for auto-processor events** (CodeRabbit): a client that opens the events socket before `agent.generate()` completes now receives its own auto-processor events once the DID resolves — the filter closure re-tries `did_for_context` on each event while the cache is empty, then fixes the DID for the socket lifetime. Fail-closed preserved: no admin promotion for unresolved DID (only an explicit `is_admin_credential` grants that).
- **Auto-processor stream fails closed on unresolved DID** (CodeRabbit) — `matches_auto_processor_pass_owner(msg, None, is_admin=false) → false`. Previously a session without a resolved DID silently saw every event.
- **`@Model` classes** for `InterpretationOverlay` / `InterpretationRun` / `AutoProcessorConfig` (`core/src/perspectives/InterpretationModels.ts`) with an integration test (`tests/js/tests/model/interpretation-models.test.ts`) that seeds each SDNA layout manually and verifies `findAll()` returns the seeded nodes — pure graph-layer, no LLM required. A parity test pins `shape.targetClass` + property predicates against the Rust hardwired SDNA in both directions.
- **Two-executor coordination test** relaxed to accept the cursor-sync path: when Alice's `InterpretationRun.sources` reaches Bob before Bob's watch loop drains a batch, Bob emits no candidacy event (nothing to process) and exactly-once still holds via the cursor rather than the claim. The load-bearing invariants (one subgroup, one processing DID) remain strict.
- **Neighbourhood test robustness** first test's 120s poll loop replaced with `waitUntil(240_000)` to match the second test, preventing spurious CI failures when Marvin is under load from parallel jobs.

## Verification

Rust interpretation + auto_processor suites green (129/129 on the merged head). JS integration tests run on Marvin: `test-integration` (incl. `auto-processor-neighbourhood`) + `test-model` (incl. `interpretation-models`) both invoked via the `integration-tests-js` and `integration-tests-model` jobs. Latest CI on `114fd12e3` — plugin-ad4m-tests + CodeRabbit already pass; build-and-test / integration-tests-{js,model} running at time of this description update.


core/src/perspectives/AutoProcessor.ts
```diff
@@ -0,0 +1,57 @@
...
+  /** Minimum items before a pass runs (Flux "wait for N inputs"). Default 1. */
+  batchMin?: number;
+  /** Cap on items per pass. */
+  batchMax: number;
+  /** Safety flush (ms) for a sub-`batchMin` batch; omit to wait indefinitely. */
+  maxWaitMs?: number;
+  /** How long a won claim is authoritative before peers may re-claim (ms). */
+  claimTtlMs: number;
+  /** Optional LLM model tag override (else the executor's default). */
+  llmModel?: string;
```
lucksus: This was removed in 885, needs to go here as well.

core/src/perspectives/PerspectiveClient.ts
```diff
@@ -249,6 +251,56 @@ export class PerspectiveClient {
...
+     * SHACL subject classes. The target shapes are resolved server-side from the
+     * perspective's registered subject classes, so you pass only the transcript.
+     * Returns the freshly minted instances (their base URIs + the links written).
+     *
+     * The server-side call prompts an LLM (up to `INTERPRETATION_MAX_ATTEMPTS`
+     * retries on parse failure), so it can legitimately take minutes on slower
+     * or CPU-only models. We raise the default 30s RPC timeout here to 20 min
+     * (matches the CI `--timeout 1200000` for the interpretation tests).
+     */
+    async runInterpretation(uuid: string, transcript: TranscriptTurn[], basePrefix: string, classes?: string[], linkStatus?: LinkStatus): Promise<string[]> {
```
lucksus: Is `linkStatus` a left-over, or are we actually setting this in the Rust interpretation engine?

tests/js/tests/model/auto-processor-models.ts
```diff
@@ -0,0 +1,34 @@
...
+export class ConversationSubgroup extends Ad4mModel {
+  @Flag({ through: "ns://type", value: "ns://conversationsubgroup" })
+  type: string = "ns://conversationsubgroup";
+
+  @Property({
+    through: "ns://name",
+    required: true,
+    resolveLanguage: "literal",
+    interpretationHint: "Short label for the topic (2-5 words).",
+  })
```
lucksus, line 25: Ah, we need to add `identity: true` as a possible parameter to @Property and use it here to steer the dedup code from TS (what @jhweir commented on in #879 - still missing here).

core/src/perspectives/AutoProcessor.ts
```diff
@@ -0,0 +1,57 @@
...
+  /** The batch's source item ids (present from `batchReady` onward). */
+  itemIds: string[];
+  /** Instance base URIs written by the pass (present on `processed`). */
+  bases: string[];
+  /** Free-form context for the step (a holder/elected DID, an error, …). */
+  detail?: string;
+}
+
+/** Configuration for `perspective.addAutoProcessor` (everything but `uuid`). */
+export interface AddAutoProcessorConfig {
```
lucksus, line 36: So, how do we define where interpreted model instances should get spawned? and where to look for existing instances (well, likely the same scope). But I don't see a way of telling the AutoProcessor with this config. Is that missing? Only here or missing in the Rust engine as well?

lucksus (changes requested):

core/src/perspectives/PerspectiveClient.ts
data-bot-coasys: Leftover — confirmed and removed (f67459984). The Rust `run_interpretation` handler already ignored `link_status` (interpretation writes via `create_subject`, so status derives from each class's SDNA `local` flag, not a caller arg). Dropped it from `RunInterpretationRequest`, both generated ts-rs bindings, and the `PerspectiveClient`/`PerspectiveProxy` signatures. Rust builds, `tsc` clean.

core/src/perspectives/AutoProcessor.ts
data-bot-coasys: Resolved — the `llmModel` field drop from 885 is now present here too (it came in with the 885→881 merge; commit 26648cd69). Verified: no `llmModel`/`llm_model` references remain in `AutoProcessor.ts` or the WS handler.

core/src/perspectives/AutoProcessor.ts
data-bot-coasys: Done (60796d949). Added a configurable `basePrefix` to the AutoProcessor: it is the spawn scope for new instances, and the #889 half made it a real `AutoProcessorConfig` field (unset → per-processor default `ad4m://autoprocessor/<id>/instance/`). Existing instances to upsert into are gathered from the perspective's registered classes; a processor confined to one subtree points both `basePrefix` and `sourceScopeQuery` there. Exposed via `AddAutoProcessorConfig.basePrefix` (TS) → WS `AddAutoProcessorRequest.base_prefix` → handler.

tests/js/tests/model/auto-processor-models.ts
data-bot-coasys: Done (8f73dc861). Added `identity` to `@Property` (Josh #879): decorator → shacl-gen → `SHACLShape.toLinks()` emits `ad4m://identity` = `literal:string:true` (the exact literal the Rust `model_query` decodes) with round-trip parse-back. Set `identity: true` here on `name`, and also on the `Ext*` classes' `title` in interpretation-models.ts (they were the stated identity but never declared it, so JS-side dedup was silently a no-op).

data-bot-coasys:
## CI failure root cause + fix (`e0a4c9c9e`)

**Failing job:** `integration-tests-js` → `test-integration` → `auto-processor-neighbourhood` tests

**What happened:** The skip guard (`if (!process.env.INTERPRETATION_E2E_NEIGHBOURHOOD) { this.skip() }`) was correctly removed — CI runs on Marvin and has Ollama. But the neighbourhood `describe` block used an arrow function (`() => {}`), which means `this.timeout()` couldn't be called, so the tests inherited the parent integration suite's 200 s mocha limit. Two tests have `waitUntil` calls with 240 s timeouts, which exceed that limit and cause mocha to abort them as timed-out.

**Fix:** Changed the `describe` callback to a regular `function()` and added `this.timeout(600_000)` (10 min), matching the single-executor `auto-processor.test.ts` which already does this. The longest possible wait across both tests is ~600 s (two 240 s + one 120 s `waitUntil`).

core/src/perspectives/InterpretationModels.ts
```diff
@@ -0,0 +1,173 @@
...
+import { Model } from "../model/decorators";
+
+// ── AutoProcessorConfig ─────────────────────────────────────────────────────
+// Mirrors the Rust `auto_processor::config::AutoProcessorConfig` SDNA.
+// One node per registered processor; the node URI is
+// `ad4m://autoprocessor/<processorId>`.
+
+@Model({
+  name: "AutoProcessorConfig",
+  interpretationHint: "Configuration for a neighbourhood auto-processor watch loop.",
```
lucksus: Ah, this classes should not have an `interpretationHint`. They are part of the interpretation engine. trying to create those with the engine is weird and misleading. Please remove these.

data-bot-coasys: Done — removed all `interpretationHint` from `AutoProcessorConfig`, `InterpretationRun`, and `InterpretationOverlay` (including their @Property fields). These are engine-internal classes and shouldn't be LLM-interpreted. Pushed in `d0dab03`.

.circleci/config.yml
```diff
@@ -144,6 +144,12 @@ jobs:
       - run:
           name: Run integration tests
           command: |
+            # Fresh per-run artifact dir: the self-hosted coasys/marvin
+            # runner keeps /tmp across jobs, so a bare `mkdir -p` would
+            # publish stale logs from earlier runs alongside the current
+            # ones (CodeRabbit #881 review). Wipe first so `store_artifacts`
+            # only sees this run's output.
+            rm -rf /tmp/test-artifacts
```
lucksus: @marvin-bot-coasys is this breaking concurrent runs?

marvin-bot-coasys: Yes — @data-bot-coasys and I both missed this. The `coasys/marvin` self-hosted runner runs up to 4 concurrent jobs (see MEMORY / `circleci-runner-extra@{2,3,4}`), and the `rm -rf /tmp/test-artifacts` here is unscoped, so if `integration-tests-js` job A wipes the dir while job B is mid-`tee` into `/tmp/test-artifacts/mocha.log`, B loses its log (or gets a spurious open-file EBADF). The same collision would hit every `integration-tests-*` job — they all write into `/tmp/test-artifacts` on the same host.

The correct fix is to make the artifact directory per-run using `$CIRCLE_BUILD_NUM` (or `$CIRCLE_WORKFLOW_JOB_ID`) and use that same variable in `store_artifacts`. Something like:

```yaml
- run:
    name: Run integration tests
    command: |
      export ARTIFACT_DIR=/tmp/test-artifacts/${CIRCLE_BUILD_NUM}
      mkdir -p "$ARTIFACT_DIR"
      set -o pipefail
      cd ./tests/js && pnpm run test-main 2>&1 | tee "$ARTIFACT_DIR/mocha.log"
- run:
    name: Collect logs on failure
    when: always
    command: |
      export ARTIFACT_DIR=/tmp/test-artifacts/${CIRCLE_BUILD_NUM}
      mkdir -p "$ARTIFACT_DIR"
      # ... existing cp loops, but write into "$ARTIFACT_DIR"
- store_artifacts:
    when: always
    path: /tmp/test-artifacts/${CIRCLE_BUILD_NUM}
    destination: test-logs
```

Plus a heartbeat GC (`find /tmp/test-artifacts -mindepth 1 -maxdepth 1 -mmin +240 -exec rm -rf {} +`) matching the one we already have on `/tmp/circleci-artifacts` at line 92, so old per-run dirs get reaped without racing an in-flight job.

Same treatment needs to apply to all six `integration-tests-*` jobs (js, multi-user-simple, email-verification, mcp, model, cli) — the current `rm -rf` was copied into each.

Want me to push that as a follow-up commit on this branch?

core/src/generated/api/RunInterpretationRequest.ts
```diff
@@ -0,0 +1,18 @@
...
+export type RunInterpretationRequest = { uuid: string, transcript: Array<TranscriptTurn>,
+/**
+ * URI namespace new instance identities are minted under, e.g. `soa://ext/`.
+ */
+basePrefix: string,
+/**
+ * Local names of the subject classes to extract into (e.g. `["Task","Belief"]`).
+ * `None`/empty selects all subject classes registered in the perspective.
+ */
+classes: Array<string> | null, };
```
lucksus: Wondering why this doesn't have the new parent/mint scope parameters. Do we need to regenerate this file or are these parameters only available on the AutoProcessor?

tests/js/tests/model/auto-processor-models.ts
```diff
@@ -0,0 +1,35 @@
...
+    "A coherent conversational thread — a set of turns focused on the same topic. Group turns discussing the same subject under one subgroup; a clear shift in subject starts a new subgroup. When an existing subgroup already covers the topic being discussed, REUSE its id (via the `id` field on the proposed instance) instead of creating a duplicate. CRITICAL DECISION RULE: read each `existing` entry's `title` (its topic name) BEFORE deciding whether to reuse an id. Only reuse an existing subgroup's id when the current turns are clearly on the SAME topic as that subgroup's title. If the current turns are on a different topic — even if there is only one existing subgroup — leave `id` unset and mint a NEW subgroup.",
+})
+export class ConversationSubgroup extends Ad4mModel {
+  @Flag({ through: "ns://type", value: "ns://conversationsubgroup" })
+  type: string = "ns://conversationsubgroup";
+
+  @Property({
+    through: "ns://name",
+    required: true,
+    resolveLanguage: "literal",
```
lucksus: I think we should remove these `resolveLanguage: "literal"` settings here as well. with 874 merged, we don't need this for simple string properties and it would always write JSON expressions..

lucksus (changes requested):

lucksus (dismissed):

lucksus (dismissed):

rust-executor/src/perspectives/auto_processor/config.rs
```diff
@@ -402,6 +445,69 @@ mod tests {
...
     use crate::perspectives::interpretation_test_support::setup_perspective_no_llm;
     use crate::types::{Link, LinkQuery, LinkStatus};
+    use std::collections::BTreeSet;
+
+    /// The set of SHACL property `path` URIs the hardwired AutoProcessor SDNA
+    /// declares. Shared with the TS-side parity check in
+    /// `tests/js/tests/model/interpretation-models.test.ts` — updating this
+    /// requires updating the TS test at the same time. Kept alphabetically
+    /// sorted so a `BTreeSet` comparison surfaces additions / removals
+    /// cleanly in test diffs.
```
lucksus: Do we still need this now that we load the types from `hardwired_sdna` JSONs?

lucksus (changes requested):
The commit iteration, trying to get the AutoProcessor pass again and reliably revealed two bugs, which will be fixed in a branch on-top of this one, and which should first be merged here.

1. `InterpretationRun` uses `@HasMany` with a literal. with #874 merged, all literals get translated from `litera:string:...` into XSD typed literals. But `@HasMany` is a relation that (so far) only points to baseExpressions and returns the IRI anyways. To do what `InterpretationRun` needs (many string IDs as literals in a `@HasMany` relation), `@HasMany` needs to learn a `string` or `literal` parameter to affect hydration.

2. SHACL shapes that get written by a peer are not accessible for others. Reason is that `model_query` has a shape store (which is a cache) which gets written to when the local users runs `add_sdna`. but when those linkes come in through link language sync, it does not. shape store needs to read from actual link data on shape/cache miss.

lucksus (approved):

PR #883, opened 2026-08-06
feat(interpretation): tree-aware interpretation — upsert + relations + grouping (Rust)

data-bot-coasys:
# Tree-aware interpretation — upsert + relations + grouping (Rust)

Make the generic interpreter **tree-aware**: instead of only creating new instances, let it **see the existing tree and decide attach / grow / create**, and emit the **relations** that place each item. Shared capability behind (a) replacing Flux's grouping/topics LLM analysis and (b) auto-building a knowledge tree from a call. Rust-only; one small tested commit per step.

### The stack at a glance

```mermaid
flowchart LR
  A["#879 · base engine<br/>hint → LLM → create_subject"] --> B["#883 · tree-aware<br/>upsert · relations · grouping"] --> C["#886 · provenance overlay<br/>+ human-divergence gate"] --> D["#889 · auto-processor<br/>watch · elect · claim"] --> E["#881 · WS API + TS client<br/>+ accept/reject overlay"]
  style B fill:#1f6f3f,stroke:#0d3d22,color:#fff
```

| PR | branch | owns |
|----|--------|------|
| **#889** | `feature/interpretation-auto-processor` | Neighbourhood auto-processor: per-perspective watch-loop, debounce/batch, authorship election + synced processing-claim (exactly-once across peers). |

## Scope
Phase 1 (upsert / tree-context) + Phase 2 (relations). **No flow engine** — multi-step group→topic→summarise is chained `runInterpretation` passes.

## Increments
- Upsert data-model (`InterpretationOp{Create,Update}`, `plan_interpretation_ops`).
- Existing-context gather → prompt → create/update via `create_subject`/`update_subject`.
- Hallucinated-id → Create routing (only trust ids in the known-existing set, **same class**).
- `DedupStrategy` param (NormalizedString default | Semantic embeddings, Ollama-verified).
- Per-class SPARQL `input_scope_query` for `gather_transcript`.
- Relations: LLM references existing node ids as link targets + reified edge classes (Flux's `SemanticRelationship`).

## Decision flow (high level)

```mermaid
flowchart TD
  IN["transcript + existing-tree context"] --> LLM["LLM · plan_interpretation_ops"]
  LLM --> OP{"per item"}
  OP -- "id unknown / new" --> CREATE["InterpretationOp::Create<br/>create_subject"]
  OP -- "id in known set, same class" --> UPDATE["InterpretationOp::Update<br/>update_subject"]
  OP -- "hallucinated id" --> CREATE
  CREATE --> REL["relations:<br/>link targets + reified edges"]
  UPDATE --> REL
  REL --> GROUP["Flux-style grouping<br/>(topics / SemanticRelationship)"]
```

## Tests
Rust unit + real-LLM grouping/persistent-topics e2e (gemma3:12b). Green.

## Summary by CodeRabbit

- **New Features**
  - Added smarter interpretation processing that recognizes existing items and avoids duplicates.
  - Added semantic matching with configurable similarity behavior.
  - Added support for updating items, resolving relation references, and creating linked items.
  - Added scoped processing for parent contexts and topic-based grouping.

- **Bug Fixes**
  - Prevented no-op updates and repeated relation links.
  - Improved handling of invalid identifiers and duplicate proposals.

- **Documentation**
  - Added a design specification for reviewing AI-generated changes.


rust-executor/src/perspectives/interpretation/dedup.rs
```diff
@@ -0,0 +1,280 @@
...
+    } else {
+        dot / denom
+    }
+}
+
+/// OpenAI-compatible embeddings client (`POST {base_url}/embeddings`). Works
+/// against Ollama's `/v1/embeddings` endpoint and any other OpenAI-shaped
+/// provider. `input` is sent as an array so one HTTP round-trip covers all
+/// candidates for a given class.
+pub async fn embed_openai_compat(
```
lucksus: I don't think this should live here. If we were to introduce external embedding inference, it should go into the AIService. But can we do this here in this PR just with the already present embedding functions in AIService? Small embedding models should even work fast enough just in CPU (i.e. without switching CI pipeline to CUDA - right?)

rust-executor/src/perspectives/interpretation/graph.rs
```diff
@@ -22,17 +24,47 @@ pub(crate) fn normalize_identity(s: &str) -> String {
...
-/// identity property. An instance whose class has no identity property is
-/// always kept (no dedup); likewise one missing that property's value.
+/// values. `identity_props` maps a class's local name to the NAME of its
+/// declared identity property. An instance whose class has no identity property
+/// is always kept (no dedup); likewise one missing that property's value.
+///
+/// Filters **in place**: the surviving instances keep the LLM's emission order,
+/// which is what the `new:<Class>:<n>` relation ordinals in
+/// [`plan_interpretation_ops_with_context`] resolve against.
 pub fn filter_already_present(
```
lucksus: Why is this not using the new dedup code?

rust-executor/src/perspectives/interpretation/graph.rs
```diff
@@ -167,36 +582,110 @@ pub async fn existing_instance_identities(
...
+        let mut requested = Vec::with_capacity(1 + scalar_names.len());
+        requested.push(idp_name.clone());
+        requested.extend(scalar_names.iter().cloned());
+        let props_json = requested
+            .iter()
+            .map(|n| format!("\"{n}\""))
+            .collect::<Vec<_>>()
+            .join(",");
+        let query = format!(r#"{{"properties":[{props_json}]}}"#);
         let result_json = perspective.model_query(class, &query).await.map_err(|e| {
```
lucksus: So this is still getting all instances of a class without any further constraint on where (which sub-graph/tree) in the perspective those live.. Which I thought was the main intention for this PR. But I'll keep on reading - maybe that is done differently. But then I wonder why this function would need to be changed to something that still doesn't select the input scope...

rust-executor/src/perspectives/interpretation/run.rs
```diff
@@ -57,17 +59,252 @@ where
...
+        };
+
+        let mut all_noop = !values.is_empty();
+        for (name, new_value) in values.iter() {
+            let Some(prop) = shape.properties.iter().find(|p| &p.name == name) else {
+                all_noop = false;
+                break;
+            };
+            let existing = perspective
+                .get_links(&LinkQuery {
```
lucksus: this is link-magic again and assuming all properties are literals. can we not use model_query here as well to reduce code and have it more generic?

rust-executor/src/perspectives/interpretation/tests.rs
```diff
@@ -461,6 +1317,318 @@ fn filter_already_present_drops_known_titles() {
...
+        vec!["Alpha", "Ship the MVP", "Omega"],
+        "upsert survives, plain duplicate is dropped, order preserved"
+    );
+    assert_eq!(
+        kept[1].id.as_deref(),
+        Some("soa://existing/task/1"),
+        "the surviving 'Ship the MVP' must be the id-carrying upsert"
+    );
+}
+
```
lucksus: woah, this is now a big file. I think having unit tests for functions live in the same file as the function is better. If those files get too long then, break them down into different aspects..

OPTION_A_MERGE_BRIEF.md
```diff
@@ -0,0 +1,173 @@
+# Option-A merge brief — #883 `feature/generic-extraction-tree`
```
lucksus: remove file

PATH_A_UPDATE_DESIGN.md
```diff
@@ -0,0 +1,115 @@
+# Option-A merge: how the tree-aware write path lands on `create_subject`
```
lucksus: remove file

lucksus (changes requested):

rust-executor/src/perspectives/interpretation/dedup.rs
```diff
@@ -0,0 +1,237 @@
...
+        .await
+        .map_err(|e| anyhow::anyhow!("semantic dedup: AIService not ready: {e:#}"))?;
+    let mut out = Vec::with_capacity(texts.len());
+    for text in texts {
+        let embedded = service
+            .embed(model_id.to_string(), text.clone())
+            .await
+            .map_err(|e| anyhow::anyhow!("semantic dedup: embed('{model_id}') failed: {e:#}"))?;
+        out.push(embedded.embeddings);
+    }
```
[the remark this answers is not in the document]
data-bot-coasys: Valid observation, but I am leaving this as a deliberate tradeoff for this PR.

`embed_via_ai_service` runs against the executor's own local candle Bert on CPU — no network round-trips. The identity strings being embedded are the instances of a *single class* in a *single interpretation pass* (a handful to a few dozen at most), so `E + P` sequential single-digit-ms embeds is negligible in practice.

A real batch path is not free plumbing: `AIService` exposes single-prompt embeddings only (one `prompt` per `EmbeddingRequest`, embedded one at a time by the channel worker), so batching would mean a new request variant + worker path + a batched candle forward pass threaded through AIService. That is a genuine AIService feature and out of scope for a dedup PR (reuse-over-reinvent / keep-it-focused).

Deferring as a future optimization to revisit if profiling ever shows per-class embedding is a bottleneck.

rust-executor/src/perspectives/interpretation_test_support.rs
```diff
@@ -223,64 +284,106 @@ pub(crate) async fn setup_interpretation_e2e(
...
-            .add_sdna(
-                (*class).to_string(),
-                String::new(),
-                SdnaType::SubjectClass,
-                Some((*sdna).to_string()),
-                &ctx,
-            )
+/// Read back the links written under each affected `base`. Production
+/// `run_interpretation` is link-free (it returns only base URIs), but a few e2e
+/// tests assert on the *edges* a run wrote (a resolved `topic-of` relation, a
```
lucksus, line 328: Is that a left-over of starting this PR on a premature revision of the base PR? Or is this genuinely needed new stuff in these new tests? If not the latter, remove the link/edge handling.

data-bot-coasys: Genuinely-needed test scaffolding, not leftover.

Production `run_interpretation` is deliberately **link-free** — it returns only the affected base URIs, never links. But a handful of e2e tests assert on the *edges* a run wrote (which live on links, not in model_query scalar state):

- `e2e_interprets_topic_relation_from_transcript` — asserts the reified `SemanticRelationship` node carries its `tag` relation edge to the right `Topic`.
- the placement-based persistence checks (`assert_persisted` and friends) that read `ns://type` / `ns://owner` / `ns://title` links back off each base.

`read_back_placements` is what lets those tests inspect written edges without production having to return links. So it stays — removing it would break the edge-asserting e2e tests. (Doc-comment on the fn already spells this out.)

OPTION_A_MERGE_BRIEF.md
data-bot-coasys: Already removed — this file is no longer tracked on the branch (`git ls-files` shows it gone). Resolving.

PATH_A_UPDATE_DESIGN.md
data-bot-coasys: Already removed — this file is no longer tracked on the branch (`git ls-files` shows it gone). Resolving.

rust-executor/src/perspectives/interpretation/dedup.rs
data-bot-coasys: Done — this is exactly what landed in `8aff78fd8` (after this review). The bespoke HTTP embedding client is gone; `embed_via_ai_service` now calls `AIService::global_instance().embed(...)`, i.e. the executor's own registered embedding model. That model is a local candle Bert running on CPU — no external endpoint, no CUDA, so CI stays on the normal pipeline. (CodeRabbit flagged the per-string loop as a future batch-API optimisation; deferred, since the sets are tiny and CPU-local.)

rust-executor/src/perspectives/interpretation/graph.rs
data-bot-coasys: `filter_already_present` is not old/legacy code — it *is* the new dedup, specifically the `DedupStrategy::NormalizedString` arm. Both strategies go through one dispatcher: `filter_already_present_with_strategy` calls `filter_already_present` (string) or `filter_already_present_semantic` (embeddings). Clarified in `a1a0f8055`; the doc-comment now states this explicitly ("This is the DedupStrategy::NormalizedString implementation, not a separate/legacy path"). Nothing bypasses the new dedup.

rust-executor/src/perspectives/interpretation/graph.rs
data-bot-coasys: The scope *is* now threaded here — `028af7dcd` (after this review) added `scope: Option<&ParentScope>` to `existing_instance_context`, spliced into the model-query as a `parent` filter, and `42ec785cc` exercises it end-to-end. `None` keeps the pre-scope behaviour (all instances).

Worth separating the two axes though, since that's the confusion:
- **Existing-instance / dedup scope** (what we compare proposals against) — constrained here, in this PR.
- **Input scope** (which transcript/items get interpreted) — that's the AutoProcessor's `source_scope_query`, in the stacked #885. This function never selected the input; it only ever built the dedup/prompt "existing" set.

Together they give tree-scoped extraction; this PR does the existing-set half.

rust-executor/src/perspectives/interpretation/run.rs
data-bot-coasys: Fixed in `5eeeeff6f` (after this review): `strip_noop_updates` no longer inspects raw link targets. It reads the instance's current state via `perspective.model_query(class, ...)` — the same generic decoder path everything else uses — and compares decoded values, so it no longer assumes literals or hand-decodes link envelopes.

rust-executor/src/perspectives/interpretation/tests.rs
data-bot-coasys: Agreed — co-locating each function's unit tests with the function reads better. Splitting `interpretation/tests.rs` (1822 lines) so each group lives in a `#[cfg(test)] mod tests` in its own file — prompt tests → `prompt.rs`, parse → `parse.rs`, plan/relations/existing-context → `graph.rs`, dedup/semantic → `dedup.rs`, strip-noop/transcript → `run.rs` — with the shared fixtures moved into the existing `#![cfg(test)]` `interpretation_test_support` module. Doing it now; will push once the suite is green.

lucksus (dismissed):

rust-executor/src/perspectives/interpretation/run.rs
```diff
@@ -57,17 +58,247 @@ where
...
+                        LinkStatus::Shared,
+                        Some(batch_id.clone()),
+                        context,
+                    )
+                    .await
+                    .map(|_| ())
+                    .map_err(|e| {
+                        anyhow::anyhow!("apply_interpretation_ops: add_links failed: {e:#}")
+                    })
+            }
```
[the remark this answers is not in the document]
data-bot-coasys: Not applying in this PR — the premise doesn't hold against the current code.

`ShapeRelation` (and `ShapeProperty`) carry no `local`/link-status field, so `include_relations` is **not** populated from `ad4m://local`. The SHACL parser does read the flag (`OPTIONAL { ?propUri <ad4m://local> ?local }` in `model_query/shape.rs`) but then **discards** it — `let _ = local;` (shape.rs:245). So at the interpretation layer there is literally no per-relation status to thread onto `InterpretationOp::AddLinks`.

Honouring a local relation status therefore requires first surfacing `local` on the shape model — a `model_query`/`shape.rs` change (add a field to `ShapeRelation` and stop discarding the parse), which also touches shared construction sites (MCP, shape tests) and is outside this PR's interpretation scope. `apply_interpretation_ops` already documents this exact prerequisite in-code (the `AddLinks` arm in `run.rs`):

> Relation links are written `Shared`: the interpretation-facing `ShapeRelation` carries no per-relation `local` flag today, so there is nothing to honour. If local relations are added to the shape model, thread that status onto `AddLinks` and use it here.

Until then `LinkStatus::Shared` is the correct default — it matches the status an SDNA relation action takes when it declares no `local` flag, which is every relation in the current shape model. Relation *writes* via interpretation are also still early-phase. Tracking the shape-model `local` plumbing as a follow-up rather than bolting a half-model onto this PR.

jhweir:
@data-bot-coasys — separate from correctness findings. One thing I'd want settled before #885 lands, because it's the only property in this arc that can't be added later.

Generated instances are signed as if the human asserted them
To be precise about what is and isn't there: every link carries an author DID (agent/mod.rs:190 — author: did_for_context(context)?), and interpretation threads AgentContext through its writes correctly. Accountability is recorded. What's missing is the mechanism — nothing distinguishes a claim the agent made from one the agent's LLM inferred on their behalf, and there's no model id, source linkage, run or prompt version, or human-edited marker. I grepped interpretation/ and auto_processor/ across #883 and #885 for all of these; there is nothing.

Three things follow, and they compound once #885 makes interpretation continuous and unattended.

For peers this is worse than absent provenance. A receiving agent sees an assertion cryptographically signed by Alice, with no way to distinguish it from something Alice wrote deliberately. The signature doesn't merely fail to qualify LLM output — it lends it credibility. Relation and instance links write Shared, so this is the normal neighbourhood case, not an edge case.

Human corrections have no protection, and the author DID can't provide it. Someone fixes a bad title. The next pass sees that instance in existing, and the system prompt explicitly instructs the model to emit an existing entry's id when the transcript "adds new field information for the SAME item (a missing owner, a small rewording)". Both the human's edit and the interpreter's rewrite carry the identical DID, so there is no signal to check against. strip_noop_updates won't catch it either — a model that "improves" a human's wording produces a genuine diff and the write lands. With #883's rolling-summary pattern, where the model rewrites a scalar every pass, an edit can be overwritten within seconds. Auto-run makes that routine.

You can't re-derive, so you can't upgrade. Output is a function of model + hints + prompt version + dedup strategy, and all four will change — the hint tuning across #879 and #883 shows how routinely. Without knowing which instances came from which pass over which input, you can't invalidate and rebuild. The choice becomes accumulating drift from every past interpreter version, or abandoning the graph. It's also why revising a hint currently has no defined effect on instances created under the old one.

The ask
Record how an instance came to exist, at write time:


<base> ad4m://inferred_by        <model id>
<base> ad4m://inferred_at        <timestamp>
<base> ad4m://inferred_from      <batch key / source item ids>
<base> ad4m://interpretation_run <run id + prompt version>
Plus one flag set whenever a non-interpretation writer touches an instance, checked before routing a proposal to Update. That flag alone closes the human-correction hole, and it's the piece I'd least want to discover we need after apps have data.

This isn't expensive plumbing — the information already exists and is discarded. At write time apply_interpretation_ops knows the model id, the shapes, the run, and the exact bases it produced; run_interpretation already returns those base URIs. #885 goes further and computes batch_key, an order-independent hash of the source item id-set, for the claim.

For completeness: I did consider giving interpretation its own signing identity instead, and I don't think it works. generate_keypair produces random, non-derived keys, so an interpreter DID wouldn't survive a user moving between local and hosted executors, and on a multi-user node you'd have to choose between a second key per tenant or one shared key that cross-contaminates attribution between users. Links are portable by construction and carry strictly more than a DID could — a DID only answers "machine or human", while re-derivation and edit-protection both need the model and run data anyway.

Why before #885
Auto-run is what turns this from a handful of instances into continuous unattended volume, and it's what makes the correction-overwrite routine rather than rare. Everything else in this arc can be improved incrementally; this one can't be backfilled. Every instance written without it is permanently unattributable.

Deliberately not asking for confidence scores — self-reported LLM confidence is poorly calibrated and would put false precision into a shared graph, and where confidence genuinely belongs it's already modellable as a domain property, as SemanticRelationship.relevance does in this PR. This is only about recording how an instance came to exist, and I don't think it needs to block the interpretation work.

jhweir:
@data-bot-coasys — six findings on #883, all verified against the current head. None appear in the existing threads.

Context for why that might be: graph/write.rs was created by 2e8d19d9c at 21:38Z on the 10th — after CodeRabbit's last pass (17:53Z) and after all of Nico's inline comments (13:56–14:40Z). CodeRabbit now reports "Review skipped: reviews are disabled for this base branch." Four of these live in that file, so it's had no automated review and one human look.

1. Dedup shifts the relation ordinals it's documented to preserve (blocking)
filter_already_present_with_strategy runs before plan_interpretation_ops_with_context, and per_class is built from the post-dedup list. resolve_ref then does per_class.get(class)?.get(n - 1), so new:Task:2 indexes the filtered list while the model counted its own unfiltered output.

Model emits Task A, B, C and says new:Task:2 meaning B. Dedup drops A. Filtered list is [B, C], so the ref resolves to C — a silently wrong edge, no error, no retry. Higher ordinals fall out of range and are dropped silently instead.

What makes this blocking rather than minor is that the code asserts the opposite in three places — dedup.rs:233, write.rs:60, and the comment in filter_already_present_keeps_upserts_and_preserves_order: "filtering happens IN PLACE — the surviving order is the LLM's emission order, which is what new:<Class>:<n> ordinals count against." In-place filtering preserves order, not indices, and that test's own fixture demonstrates it: four proposals in, three out, so the model's 4th is now the 3rd. a1a0f8055 added that doc during the review round, so the reasoning got written down rather than checked.

Fix: carry each proposal's pre-dedup position through and resolve against that. Better still, have the model emit its own local ids ({"_ref":"t1"}, "blocks":["t1"]) — position-independent, survives filtering, and a dangling ref becomes detectable instead of silently resolving to the wrong sibling.

2. A cross-class id routes an update onto the wrong node (blocking)
write.rs:101:


Some(id) if existing.contains_key(id) => (id.clone(), true),
Membership only — InstanceContext.class is never compared to inst.class. The prompt shows the model ids from every class at once, so {"class":"Belief","id":"<a Task's id>"} is a plausible emission. It routes to update_subject, running Belief's setSingleTarget setters against a Task node and deleting that node's links on any shared predicate.

This is the residual of CodeRabbit's resolved hallucinated-id thread: that fix added the existing.contains_key(id) guard, which checks the id is known but not that it's known for this class. One equality check closes it.

3. Semantic dedup lost the intra-response guarantee the string path was fixed to have
semantic_dedup_pure only compares proposals against existing_vecs — there is no proposal-vs-proposal pass. It also bails per class on else { continue } and if existing.is_empty() { continue }, so a class with no existing instances is skipped wholesale. Under DedupStrategy::Semantic, a doubly-emitted identity mints two subjects.

The string path covers this deliberately, with filter_already_present_dedupes_within_same_response and a comment explaining why — it was CodeRabbit finding #23 on #879, fixed in b72fe65. The new path reintroduces it behind a strategy flag.

4. Additive-only relation writes don't survive repeated passes
AddLinks goes straight to add_links(links, Shared, batch, ctx) with no read of existing edges — documented as "additive by design" at run.rs:75. Within a single pass that's fine, and hasOne cardinality is enforced there (there's a test keeping only the first resolved ref).

Across passes it isn't. Re-pointing a hasOne at a new target adds the new edge without removing the old, leaving both attached and the cardinality violated. Identical relation links re-emitted later become distinct reifiers, since the reifier IRI hashes the timestamp. "Additive by design" is reasonable for a one-shot run; #885 makes runs continuous.

5. scope constrains reads but not writes
run.rs:361 threads scope into existing_instance_context, and that is its only use — grep scope over write.rs returns nothing. Created instances are never linked under the scope parent.

So a scoped caller whose instances end up unattached won't see them in the next pass's existing context, and re-mints the same instances every run. The e2e helper attaches them by hand, which is why the tests don't surface it, and run_interpretation's doc doesn't tell callers they must.

6. normalize_refs discards a whole relation array on one bad element
write.rs:238:


serde_json::Value::Array(arr) => arr.iter().map(|v| v.as_str().map(...)).collect(),
Collecting Option<String> into Option<Vec<String>> short-circuits to None on the first non-string, so ["new:Task:1", 2, "new:Task:3"] drops all three refs rather than the one. Given the input is LLM output, partial malformation is the common case. Filtering and warning per element keeps the valid refs.

I'd treat 1 and 2 as blocking, 3 and 4 as should-fix before #885 makes runs continuous, 5 and 6 as ordinary. All six are from reading rather than running, so worth a sanity check on each.

lucksus (approved):

PR #885, opened 2026-08-07
feat(auto-processor): neighbourhood auto-run — ProcessingClaim + AutoProcessorConfig + watcher + Flux-parity demo (stacked on #883)

data-bot-coasys:
# Neighbourhood auto-processor — watch-loop + authorship election + processing claim

Turns the interpretation engine from a *called* function into a **standing service** on a perspective: watch a source (a channel), and whenever new content settles, run an interpretation pass — **without any peer processing the same batch twice**, whether peers are multi-tenant users on one executor or separate executors sharing a neighbourhood. Rust-only.

**#879 → #883 → #886 → #885 → #881**

### The stack at a glance

```mermaid
flowchart LR
  A["#879 · base engine<br/>hint → LLM → create_subject"] --> B["#883 · tree-aware<br/>upsert · relations · grouping"] --> C["#886 · provenance overlay<br/>+ human-divergence gate"] --> D["#885 · auto-processor<br/>watch · elect · claim"] --> E["#881 · WS API + TS client<br/>+ accept/reject overlay"]
  style D fill:#1f6f3f,stroke:#0d3d22,color:#fff
```

| PR | branch | owns |
|----|--------|------|
| **#885** | `feature/interpretation-auto-processor` | Neighbourhood auto-processor: per-perspective watch-loop, debounce/batch, authorship election + synced processing-claim (exactly-once across peers). |

## How it works
**Per-perspective watch loop** (`auto_processor_watch_loop`, 500 ms poll):
1. `load_processors` reads every `AutoProcessor` config node.
2. `gather_transcript_sparql(cfg.source_scope_query)` pulls watched `(speaker,text)` rows; new-since-last ids feed a per-processor **debounce** (`WatcherState`: dedup, quiet-window, `batch_max`).
3. A ready batch runs one **pass**: elect → claim → gather → interpret → write.

**Synchronization — two gates, one authoritative:**
- **Authorship election (candidacy).** Only a *participant* — an **author of the in-scope data** — should run the pass. `elect_author` prefers the first reachable author in message order (mirrors Flux telepresence); DID lexical order is only a stall/tiebreak fallback. Fairness/optimization, not the correctness guarantee.
- **Processing claim (authoritative).** `ProcessingClaim` writes a **synced claim link** on a shared batch node; every peer reads all active/unexpired claimants and the **min-DID wins** — a deterministic tiebreak every peer computes identically once claims sync. Correctness rests on the synced link, not on ephemeral telepresence. TTL-expiry lets a crashed claimant's batch be re-claimed.

## Who runs the pass (high level)

Two gates — telepresence election is a fast-path; the synced processing-claim is the correctness guarantee:

```mermaid
flowchart TD
   A["A batch settles in my watch loop (my DID = me)"] --> T{"telepresence available<br/>AND batch authors resolvable?"}
   T -- "no (no link-language telepresence,<br/>or no resolvable authors)" --> CLAIM["try_claim"]
   T -- "yes" --> E["elect_author: walk the batch's authors in<br/>message order; stop at the first that is<br/>online — or is me"]
   E --> R{"who is that first reachable author?"}
   R -- "me → I'm the earliest present author" --> CLAIM
   R -- "another ONLINE author, earlier than me" --> STAND["stand down this tick — they're the candidate.<br/>Safety net: the authoritative claim below<br/>(min-DID + TTL) still guarantees exactly-once"]
   R -- "no author online, not even me" --> WAIT["AwaitingAuthor — do nothing, retry next tick"]
   CLAIM --> C{"am I the min-DID among all<br/>active, unexpired claims on this batch?"}
   C -- "no — a smaller DID has claimed" --> BACK["BackedOff (they win; mark processed)"]
   C -- "yes → I win" --> RUN["run the LLM pass → write results + mark processed"]
```

## Stacking / this revision
- Rebased onto **#886**: inherits the §4 provenance gate through the shared `run.rs` apply path (`apply_with_overlay`) — no separate gate in the auto-processor.
- **Nico's review — done:** dropped the `llm_base_url`/`llm_model` config fields (a shared processor must not pin one peer's local model — always use the executor's default LLM).
- **Nico's review — in progress:** make authorship-of-input the *primary* election criterion (DID sort strictly fallback); document the claim TTL-vs-sync-latency window in `claim.rs`; add a **multi-user integration test** (two users in one process driving the real loop) — full multi-executor coverage lands in #881.

## Tests
Interpretation + auto_processor suites green (incl. `auto_processor_two_users_one_executor_no_double_processing`).


rust-executor/src/perspectives/auto_processor/watcher.rs
```diff
@@ -0,0 +1,804 @@
...
+/// Run one processing pass for `cfg` against the id-set `item_ids`, following
+/// the spec §4.3 coordination protocol. Standalone so callers can trigger it
+/// from a test, a REPL, or the future event-driven watcher loop without
+/// changing its guarantees.
+///
+/// Ordering:
+/// 0. Telepresence fast-path (P-B2b3) — read
+///    [`PerspectiveInstance::online_agents`] and short-circuit with
+///    [`PassOutcome::NotCandidate`] when some other online peer's DID sorts
+///    before ours. This never writes anything to the shared graph, so it
```
lucksus: Haven't read all the code yet, but: is this before checking the authorship of the unprocessed items? Does that mean that in a call of 3 users it will always be the same user having to run the LLM processing? The DID sorting should just be a hard-wired fallback if because of slow syncing two agents see different unprocessed items and both think it's their turn.

It should always be one of the users who are in a session (call creating transcripts) which are processing. So the author of the links/class instances that are in the input scope should first and foremost select the agent who tries to process.

Is this the case? Or is lexical sorting of online agents' DIDs the only criteria for selecting processing agents?

rust-executor/src/perspectives/auto_processor/claim.rs
```diff
@@ -0,0 +1,341 @@
...
+//! ```
+//!
+//! ## Winner determination
+//! Under a race, two peers may both write a claim before either sees the other.
+//! Rather than last-writer-wins (which needs a total sync order we don't have),
+//! [`try_claim`] is **deterministic**: after writing its own claim it reads *all*
+//! active, unexpired claimants for the batch and the lexicographically smallest
+//! DID wins. Every peer converges on the same winner once claims sync, so exactly
+//! one proceeds. A losing/crashed claimant's claim simply expires (`ttl_ms`),
+//! after which the batch can be re-claimed.
```
lucksus, line 38: This assumes sync_time < claim_ttl. If sync is very slow, two agents could both claim on the same state, not see the other claim and start processing in parallel. Then sync comes through eventually and we have two overlapping and likely somewhat different processing results... It's not possible to avoid this situation in all cases (given an agent-centric / p2p topology) but that's why we used telepresence to have another channel next to the link languages regular sync.

rust-executor/src/perspectives/auto_processor/watcher.rs
data-bot-coasys: Fixed in d0385b68c. Replaced the lexical min-DID candidacy with authorship-ordered election (`elect_author` + `batch_authors`):

- Resolve the batch's authors **in message order** (each item's link author from the graph).
- Elect the **first *online* author in message order** as the processor — so it walks the participants exactly like Flux: first message's author, then second, etc., first one online wins. If I authored one of them and everyone ahead is offline, I eventually reach my own message and process the whole batch.
- **No batch author online → nobody processes** (new `PassOutcome::AwaitingAuthor`): strict "only participants process" (your (b)). The watcher keeps the batch un-processed so a returning author can pick it up.

DID sort is gone as the primary; `try_claim` stays as the deterministic guard for the residual "who-is-online-differs-across-peers" race. Message order is a total order every peer computes identically from the synced graph, so peers converge without a global clock.

I also added the Flux **"wait for N inputs"** threshold you flagged as missing: `batch_min` (default 1) holds a sub-threshold batch, with an optional `max_wait_ms` safety flush so a partial batch is never orphaned. Both configurable via SDNA (`ad4m://batch_min`, `ad4m://max_wait_ms`).

rust-executor/src/perspectives/auto_processor/claim.rs
data-bot-coasys: Agreed — this is inherent to the p2p topology and can't be fully closed. The authorship fast-path (now `elect_author`, d0385b68c) is the telepresence-side channel you're pointing at: it stands peers down *before* they write a claim, shrinking the window to the case where two peers genuinely disagree about who is online. `try_claim` remains the correctness guard for that residual, and `claim_ttl_ms` must be configured `> expected sync_time`. Leaving this as a documented limitation (the module doc-comment already calls out that correctness rests on the synced claim link, not ephemeral signals); not blocking for this PR.

rust-executor/src/perspectives/perspective_instance.rs
```diff
@@ -5056,6 +5057,171 @@ impl PerspectiveInstance {
...
+    ///      sync, no reason to re-race); `ShapesMissing` and `EmptyTranscript`
+    ///      do not, so the ids are retried once the shape or transcript lands.
+    ///
+    /// This is the "polling MVP" per the design decision in
+    /// `planning/p-b2b2-watcher-wireup.md`. The event-driven variant
+    /// (subscribe to `PERSPECTIVE_LINK_ADDED_TOPIC`, record `link.data.source`
+    /// deltas) is a follow-up optimisation — the coordination correctness
+    /// envelope is the same because the `ProcessingClaim` (P-A) is the real
+    /// double-processing guard, not the trigger latency.
+    async fn auto_processor_watch_loop(&self) {
```
lucksus: This is not tested. We need a more high-level integration test that relies on this loop and that exercises the perspective from two different users. In this PR, we can still do that with just one executor in the test binary, but with multiple users using the multi-tenancy feature in ad4m. In https://github.com/coasys/ad4m/pull/881 we will ultimately test this in full integration tests with multiple ad4m executor instances. But here we need the step towards that with multiple users in the same process.

lucksus (changes requested):

rust-executor/src/perspectives/perspective_instance.rs
data-bot-coasys: Addressed in `967ed9b43`. Three pieces:

1. **Step signals** — new `auto_processor::events` module publishes an `AutoProcessorEvent` (BatchReady → Claimed/BackedOff/AwaitingAuthor → GatheringTranscript → RunningInterpretation → Processed …) on a PubSub topic, the same mechanism GraphQL subscriptions use. So a test/WS client can `await` a step instead of polling, and #881 can forward it over WS nearly for free.
2. **Loop testability** — factored the per-tick body out of `auto_processor_watch_loop` into `run_auto_processor_tick(watcher, processed, now_ms)`; the production loop and tests share the exact same path (tests feed `now_ms`, no sleeps).
3. **A real debounce bug this surfaced** — `record_item` bumped `last_touched_ms` on *every* call. The polling loop re-gathers the whole transcript each tick and re-records every turn, so the debounce window reset every tick and **the loop could never drain its first batch**. Fixed to bump only on a genuinely new id.

New high-level test `auto_processor_high_level_signal_driven_pass` (real gemma3:12b): writes a processor + seeds a channel's messages, drives the tick, `await`s the `Processed` signal, asserts the ConversationSubgroup was created — no manual interpretation/transcript.

Still open (your multi-user point): the **two-users-one-executor** no-double-processing test needs the multi-tenancy wallet-key setup — I asked Nico whether there's an existing harness for constructing a second user in-process; building it next. The full **two-executor** version lands in #881.

rust-executor/src/perspectives/perspective_instance.rs
```diff
@@ -5056,6 +5057,193 @@ impl PerspectiveInstance {
...
+    ///      re-processing after `drain_ready_batch` empties the pending queue.
+    ///   3. Per config, `drain_ready_batch(cfg, now_ms)` — if a batch is ready
+    ///      (i.e. `debounce_ms` elapsed since the last `record_item`), run
+    ///      `run_one_pass` in-line. `Won` and `BackedOff` both add the batch
+    ///      to `processed_ids` (a peer won → their write reaches us via link
+    ///      sync, no reason to re-race); `ShapesMissing` and `EmptyTranscript`
+    ///      do not, so the ids are retried once the shape or transcript lands.
+    ///
+    /// This is the "polling MVP" per the design decision in
+    /// `planning/p-b2b2-watcher-wireup.md`. The event-driven variant
```
lucksus: we don't have access to your local planning docs. please relay and summarize design decision wihtout referencing your local docs.

rust-executor/src/perspectives/auto_processor/config.rs
```diff
@@ -0,0 +1,582 @@
...
+    /// waited this long, a sub-threshold batch drains anyway so it is never
+    /// orphaned. `None` (the default) = pure Flux parity: wait indefinitely
+    /// until `batch_min` items accumulate.
+    pub max_wait_ms: Option<i64>,
+    /// TTL passed to [`super::claim::try_claim`] — how long a Won claim is
+    /// treated as authoritative before other peers may re-claim.
+    pub claim_ttl_ms: i64,
+    /// LLM endpoint (`base_url`) override — falls back to the executor's
+    /// configured default when `None`.
+    pub llm_base_url: Option<String>,
```
lucksus: Don't think we should allow setting of LLM URLs here as we can configure models in the AIService and then reference those per name. But even stronger: both are just local to one ad4m instance. The AutoProcessor will be shared amongst users/agents in a Neighbourhood and they might have to select different models locally to match their hardware resources. I would say for now both llm_ fields should go and always use the default LLM.

rust-executor/src/perspectives/perspective_instance.rs
```diff
@@ -3479,21 +3479,68 @@ impl PerspectiveInstance {
...
+                        && owners.contains(&user.did)
+                        && !agents.iter().any(|a| a.did == user.did)
+                    {
+                        agents.push(OnlineAgent {
+                            did: user.did,
+                            status: PerspectiveExpression::default(),
+                        });
+                    }
+                }
+            }
```
lucksus, line 3533: Need tests for this to avoid regression. Can live in the TS integration test for the multi-user aspect..

lucksus (changes requested):

lucksus (changes requested):
You're still using manual link-magic here for the `AutoProcessorConfig` and `Claim` models. I think you should make those proper SHACL shapes that also get written to the perspective on first use and then use the create_subject and model_query machinery to deal with those instances. That enables UIs to also see and easily deal with these data structures and it should save code here.

data-bot-coasys:
⚠️ This PR was **accidentally auto-closed as merged** by a mis-scoped automation on my side (a branch merge ran in the wrong direction — into this PR's base `feature/interpretation-provenance` — which GitHub reads as a merge). It was **not** an intentional merge and **nothing landed on `dev`/`main`**. #886 has been restored to its pre-merge state. No code was lost. Review continues on **#889** (identical branch + content). Apologies for the noise.

data-bot-coasys:
Landed as `9e09af08e` on the follow-up #889 (this branch was accidentally auto-closed as merged so the fix couldn't go here). Both models are now first-class SHACL subject classes registered into the perspective on first use:

- `ad4m://AutoProcessor` — `processor_id` (identity), `source_scope_query`, `interpretation_class` (collection), `debounce_ms`, `batch_min/max`, `max_wait_ms`, `claim_ttl_ms`, `dedup_strategy`.
- `ad4m://ProcessingClaim` — `claimant` (identity), `expires_at`, `claim_status`.

Writes go through `create_subject`, reads through `model_query` — same machinery the interpretation-overlay classes use, so UIs can list/edit them via the normal model API. All hand-built `Vec<Link>` writers and per-predicate `get_links` readers on those two models are gone; shared registration/lookup helpers live in `auto_processor/mod.rs`. Links stay Shared (setters carry no `local` flag) — asserted by two new regression tests. auto_processor suite 48/48 green.

The `ProcessingClaim` node still hangs off the shared batch node via `ad4m://has_claim`; the batch node itself is not an instance (it's the coordination point multiple peers race on).

PR #892, opened 2026-08-13
fix(interpretation): decode literal-encoded overlay `kind` on read

HexaField:
## What

Fixes a read/write encoding mismatch on the `InterpretationOverlay` `kind` discriminator (`create` | `update`).

`write_overlay` mints the overlay through `create_subject`, whose `kind` SHACL setter (`setSingleTarget`) has no `resolveLanguage`, so `create_subject` stores the value **literal-encoded** — `literal:string:create`. But the readers treated `kind` as the bare discriminator string:

- `overlay_of` returned it raw, so `perspective.interpretationOverlays` surfaced `"literal:string:create"` to clients instead of `"create"`.
- `reject_interpretation` compared `kind == "create"` to choose its branch. On a real overlay that comparison never matched, so rejecting a *create* overlay wrongly took the *update* branch — it dropped only the overlay shell and left the LLM-authored instance behind, instead of deleting it wholesale.

The module's own tests missed this because their seed helper wrote `kind` **plain** (`add(..., "create")`), unlike the production `create_subject` setter.

## Fix

`overlay/accept.rs`: add an `overlay_kind()` helper that decodes the `kind` target with `parse_literal_value` — exactly how `inferred/<p>` targets are already decoded in `overlay_of` — and use it at both read sites (`overlay_of`, `reject_interpretation`). `parse_literal_value` passes an already-plain target through unchanged, so the read stays correct regardless of how the link was written.

Adds a regression test (`overlay_kind_decodes_literal_encoded_target`) that seeds a literal-encoded `kind` link (the production write-path shape) and asserts both that `overlay_of` decodes it to `"create"` and that `reject_interpretation` then deletes the whole instance. It fails without the fix.

## Verification

- `cargo fmt --check` clean.
- `cargo test perspectives::interpretation::overlay::accept` — 4/4 green (the new regression test plus the three pre-existing accept/reject tests, no regression).
- End to end against a containerised executor built from this stack: the interpretation wind-tunnel scenarios I3 (provenance gate) and I4 (accept/reject) both pass — I4's "reject an untouched create overlay deletes the whole instance" corner is the exact branch this fix repairs.

## Scope

Stacked on top of the generic-LLM-interpretation series; base branch `feature/generic-extraction-ws-ts`. One file changed (`rust-executor/src/perspectives/interpretation/overlay/accept.rs`), +78/−2.

### DIFF
--- a/rust-executor/src/perspectives/interpretation/overlay/accept.rs
+++ b/rust-executor/src/perspectives/interpretation/overlay/accept.rs
@@ -47,7 +47,7 @@ pub(crate) async fn overlay_of(
     base: &str,
 ) -> anyhow::Result<Option<OverlayView>> {
     let all = links_from(perspective, base).await?;
-    let kind = match first_target(&all, OVERLAY_KIND_PRED) {
+    let kind = match overlay_kind(&all) {
         Some(k) => k,
         None => return Ok(None),
     };
@@ -139,7 +139,7 @@ pub(crate) async fn reject_interpretation(
 ) -> anyhow::Result<()> {
     let _ = context;
     let all = links_from(perspective, base).await?;
-    let kind = match first_target(&all, OVERLAY_KIND_PRED) {
+    let kind = match overlay_kind(&all) {
         Some(k) => k,
         None => anyhow::bail!("reject_interpretation: no overlay on `{base}`"),
     };
@@ -184,6 +184,24 @@ fn first_target(links: &[DecoratedLinkExpression], predicate: &str) -> Option<St
         .iter()
         .find(|l| l.data.predicate.as_deref() == Some(predicate))
         .map(|l| l.data.target.clone())
+}
+
+/// The overlay's `kind` value (`create` | `update`), decoded from its link
+/// target.
+///
+/// `write_overlay` mints the overlay through `create_subject`, whose `kind`
+/// SHACL setter (`setSingleTarget`, no `resolveLanguage`) stores the value
+/// literal-encoded (`literal:string:create`). Every consumer — the §8 accept /
+/// reject branches below, the `interpretationOverlays` query surface, and the
+/// module's own tests (`decoded_targets`) — treats `kind` as the bare
+/// discriminator string, so decode it on read exactly as `inferred/<p>` targets
+/// are decoded. `parse_literal_value` passes an already-plain target through
+/// unchanged, so this stays correct regardless of how the link was written.
+fn overlay_kind(links: &[DecoratedLinkExpression]) -> Option<String> {
+    first_target(links, OVERLAY_KIND_PRED).map(|k| match parse_literal_value(&k) {
+        serde_json::Value::String(s) => s,
+        other => other.to_string(),
+    })
 }
 
 /// The `inferred/<p>` links on the base, optionally narrowed to one real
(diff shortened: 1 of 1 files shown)


data-bot-coasys (changes requested):
Read the whole diff and traced both read sites against `origin/dev` and against this PR's own base branch. The `overlay_of` half is a real, still-live bug and the test is the right test — but the diff has been overtaken by the branch it sits on, so it can't land as written.

### 1. The `reject_interpretation` half already landed — on this PR's own base

`feature/generic-extraction-ws-ts` (the base of this PR) already carries the identical decode inline:

```rust
// accept.rs:146 on origin/feature/generic-extraction-ws-ts
let raw_kind = match first_target(&all, OVERLAY_KIND_PRED) { ... };
// `write_overlay` uses `create_subject`'s setter path, which encodes a
// non-URI string target as `literal:string:<value>`. ...  (CodeRabbit #881 review)
let kind = match parse_literal_value(&raw_kind) { ... };
```

Same code is on `origin/dev` (`accept.rs:146-160`). So the PR description's second bullet — *"`reject_interpretation` compared `kind == \"create\"` ... that comparison never matched"* — describes a bug that is already fixed in `dev`. Worth correcting the description before review, otherwise the next reader re-derives it.

That's also why GitHub reports `mergeable=CONFLICTING / DIRTY`: the diff is computed against a merge-base from before that landed.

### 2. What is still genuinely missing: `overlay_of`

This part is **not** on `dev` and **not** on the base. `overlay_of` still does:

```rust
// accept.rs:50 on origin/dev — undecoded
let kind = match first_target(&all, OVERLAY_KIND_PRED) { ... };
```

and hands that straight into `OverlayView.kind`. Note the line immediately below it decodes `inferred/<p>` targets with `parse_literal_value` — so the struct is internally inconsistent today: one field decoded, its sibling not.

Failure scenario, on current `dev`: an overlay minted by `write_overlay` → `create_subject` stores `ad4m://interp/kind` as `literal:string:create`; a client calls `perspective.interpretationOverlays`; it gets back `kind: "literal:string:create"`. Any UI or agent branching on `kind === "create"` takes the wrong branch, silently. That is the bug worth keeping this PR open for.

### 3. Suggested reshape

Rebase onto the current base and the diff shrinks to something I'd approve on sight:

- `overlay_kind()` helper — keep it, and have `reject_interpretation` call it too. That turns the inline decode already on `dev` into a single seam, which is strictly better than two copies drifting apart. Move the `dev` comment's substance into the helper doc so the rationale survives.
- `overlay_kind_decodes_literal_encoded_target` — keep as is. I checked why the pre-existing tests miss this: their seed writes `kind` *plain* (`accept.rs:339`, `accept.rs:471`), unlike the production setter, so the test is aimed exactly at the gap. Good catch.

### 4. Two things I verified so you don't have to

- `parse_literal_value` is genuinely passthrough for a plain target — non-`literal:` input returns `Value::String(uri)` unchanged (`model_query/utils.rs:123-127`). So applying it in `overlay_of` cannot break the plain-string seeds in the existing tests.
- The `other => other.to_string()` fallback is harmless here but is a JSON rendering, not a value: a `kind` of `literal:boolean:true` would decode to the string `"true"`. `kind` is a closed `create`|`update` discriminator so nothing reaches that arm today; no change requested, just noting the arm is unreachable-by-contract rather than correct-by-construction.

### Status

CI is red on `integration-tests-js` and `integration-tests-model`, but the head commit is from 2026-08-13 and both jobs have changed since — I'd expect those to be stale rather than caused by this one-file change. Re-run after the rebase before reading anything into them.

Requesting changes only for the rebase + description correction. The surviving idea is right and I'd like it on `dev`.

PR #894, opened 2026-08-16
perf: eliminate idle CPU hotspots in executor

HexaField:
## Summary

Eliminates three major sources of idle CPU consumption in the executor, reducing idle CPU from ~11% of one core to expected <1%.

### Profiling data

**Methodology:** `pidstat` + `/proc/PID/stat` differential CPU measurement against the production Docker executor (full Holochain + Flux, 323 threads) over 30-second idle windows.

| Metric | Before |
|--------|--------|
| Idle CPU (% of one core) | **~11%** (3360ms / 30s) |
| Thread count | 323 |
| Top idle consumers | language runtimes (14%), Holochain signal loops (13%), AI service (varies) |

#### 1. `EventLoopFuture` — 1kHz polling per language thread (highest impact)

`js_core/futures.rs` had `std::thread::sleep(1ms)` + unconditional `cx.waker().wake_by_ref()` inside `Future::poll()`. This forced ~1000 wakeups/sec **per loaded language** (5+ minimum per unlocked agent, growing with every installed language/neighbourhood). The blocking `sleep(1ms)` inside `poll()` also prevented the tokio runtime from parking the thread.

**Fix:** Remove both. `poll_event_loop()` already registers its own wakers via deno_core for pending I/O, timers, and promises. The channel branch in the surrounding `tokio::select!` handles waking for new RPC requests. Zero CPU when the JS event loop has no pending work.

#### 2. AI service worker threads — 166Hz polling + unconditional Bert spawn

`ai_service/mod.rs` spawned a dedicated OS thread per model with a `tokio::select!` racing the request channel against a 1ms timeout, then doing a 5ms `thread::sleep` on timeout — ~166 wakeups/sec per model, forever.

Worse: `AIService::load()` unconditionally spawned a default Bert embedding model when no models were configured (`"for integration tests"` — no `#[cfg(test)]` gate), meaning every fresh install got a busy-polling thread by default.

**Fix:** Replace the select!-timeout pattern with `rt.block_on(rx.recv())` — true zero-CPU blocking on the channel. Remove the Bert auto-spawn (integration tests should configure models explicitly).

#### 3. Holochain signal forwarding — 1ms backoff branches

`holochain_service/mod.rs` had a 1ms backoff branch in the signal-forwarding `select!` loop (~1000 wakeups/sec). `lib.rs` `holochain_signal_receiver` had 1ms/10ms polling loops.

**Fix:** Widen to 100ms/500ms. These branches guard against `select!` busy-spinning on empty `StreamMap`s or closed channels — safety nets, not primary delivery paths. Real signals arrive through the stream/channel branches above.

### What was NOT changed (low-impact, already cheap)

- Payment polling (30s) — one DB read, early-returns if empty
- Credit flush (2s) — drains a `HashSet`, `continue`s if empty  
- Memory diagnostics (30s) — checks `log_enabled!` first, skips work
- Prolog pool cleanup (60s) — O(1) early-return when no filtered pools
- Verification code cleanup (300s) — one DB call

### Verification

- `cargo check` — clean build, zero warnings
- `cargo test --lib` — 674 passed, 0 regressions (1 pre-existing failure: `test_migrate_with_literal_conversion`, fails identically on `dev`)
- 4 initially-flaky tests confirmed pass on re-run (test ordering / shared state, not related to these changes)

## Summary by CodeRabbit

* **Performance**
  * Reduced unnecessary background polling and idle wake-ups across AI, event-loop, and Holochain processing.
  * Improved responsiveness by relying on event notifications and direct channel waits.
  * Reduced CPU usage during idle periods with more efficient backoff intervals.

* **Bug Fixes**
  * Improved error propagation from event-loop processing.
  * Improved service shutdown handling and reliability during unavailable service conditions.
  * Clarified fallback model behavior when no configured models are available.

### DIFF
--- a/rust-executor/src/ai_service/mod.rs
+++ b/rust-executor/src/ai_service/mod.rs
@@ -5,13 +5,13 @@ use crate::pubsub::AI_TRANSCRIPTION_TEXT_TOPIC;
 use crate::types::ModelInput;
 #[allow(unused_imports)]
 use crate::types::{AIModelLoadingStatus, AITaskInput, TranscriptionTextFilter};
-use crate::types::{AITask, LocalModel, Model, ModelType};
+use crate::types::{AITask, LocalModel, ModelType};
 use crate::{db::Ad4mDb, pubsub::get_global_pubsub};
 use anyhow::anyhow;
 use candle_core::Device;
 use chat_gpt_lib_rs::{ChatGPTClient, ChatInput, Message, Role};
 use deno_core::error::AnyError;
-use futures::{FutureExt, SinkExt};
+use futures::SinkExt;
 use holochain::test_utils::itertools::Itertools;
 use kalosm::language::*;
 use kalosm::sound::TextStream;
@@ -27,7 +27,7 @@ use tokio::sync::{mpsc, oneshot, Mutex};
 use tokio::time::sleep;
 
 mod error;
-use log::error;
+use log::{error, info};
 
 pub type Result<T> = std::result::Result<T, AnyError>;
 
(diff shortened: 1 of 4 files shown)


lucksus (changes requested):
We should leave Bert loaded. It's not just for integration tests (the comment was old) but may be used in apps and my new interpretation PRs assume it's loaded and usable.

Other changes look great!

lucksus (approved):

PR #891, opened 2026-08-12
experiment: remove typed-literal + envelope-flatten store migrations

marvin-bot-coasys:
## Option B — remove the unconditional store migrations

Companion to Option A (#890). Same base branch. Pick one — or neither if a third path lands.

## What this does

Strips the two store-mutating migrations from `perspectives/mod.rs::initialize_from_db`:

- `migrate_signed_envelopes_to_plain_literals` (v3) — flattens `literal:json:{envelope}` targets to their inner `data` (**risk**: if the SHACL keep-envelope predicate set comes back empty because shapes aren't loaded yet, envelope signatures for expression-resolved properties get destroyed).
- `migrate_iri_literals_to_typed_literals` (v4) — unconditionally rewrites every `literal:string:*` / `literal:number:*` / `literal:boolean:*` / `literal:json:*` reifier target to a native RDF typed literal. This is what's silently mangling `SHACLFlow`'s own bookkeeping links (`ad4m://stateName`, `stateValue`, `actionName`) and taking down the `Social DNA > SHACL-based TODO flow` integration test.

The migration functions themselves stay defined in `sparql_store.rs` so they can be re-invoked from a follow-up or a bespoke migration path if you decide you want them.

## Consequence

With the migrations gone the store is mixed-shape by design:
- Fresh writes land as native RDF typed literals (via `link_target_for_value → literal_encode`)
- Anything written before this branch stays as `literal:string:*` NamedNode IRIs

Readers (`SHACLFlow.fromLinks`, `Literal.fromUrl` callers) and WHERE builders (`buildWhereCondition`, `buildSPARQLQuery`) must keep tolerating both shapes on read. On a completely fresh store this branch is indistinguishable from the migration path.

## Tradeoff vs Option A (#890)

- **Option A**: keep the migration, teach `SHACLFlow` to read the post-migration shape. Small surface, one file. Fresh stores get the perf win uniformly.
- **This PR (Option B)**: no migration, `SHACLFlow` untouched, but the mixed-shape store puts the burden on every reader / WHERE builder to handle both. A CodeRabbit-style bug in one of them silently misses rows. Fresh installs still get the perf win because fresh writes use typed literals.

## Test plan

- [x] `cargo check -p ad4m-executor` — compiles clean.
- [ ] `integration-tests-js` — does the `Social DNA > SHACL-based TODO flow` case pass now that the migration isn't rewriting `SHACLFlow`'s targets?

PR #890, opened 2026-08-12
fix(SHACLFlow): accept typed-literal targets on read

marvin-bot-coasys:
## Option A — teach SHACLFlow to read the post-migration shape

Going off `refactor/typed-rdf-literals-and-fn-cleanup-nico-refactor` so you can compare against Option B (`experiment/remove-typed-literal-migration`) and pick.

## What breaks and why

The new `migrate_iri_literals_to_typed_literals` (in `rust-executor/src/perspectives/sparql_store.rs`, run once on perspective init) walks every reifier triple and rewrites any `literal:string:*` / `literal:number:*` / `literal:boolean:*` / `literal:json:*` NamedNode target to the equivalent native RDF typed literal. It's **unconditional** — no SHACL / `resolveLiteral` check.

That rewrite also touches `SHACLFlow`'s own bookkeeping links (`ad4m://stateName`, `stateValue`, `actionName`) which are written as `Literal.from(...).toUrl()` → `literal:string:foo`. Post-migration they come back through SPARQL as bare `"foo"` / `1`, and `Literal.fromUrl("foo")` throws `Can't create Literal from non-literal URL`. That takes down `PerspectiveProxy.getFlow` → `SHACLFlow.fromLinks` and the `Social DNA > SHACL-based TODO flow` integration test with it.

## What this PR does

Teach `SHACLFlow.fromLinks` to read both target shapes. Two tiny helpers at the top of the function:

```ts
const readLiteralString = (target: string): string =>
  target.startsWith("literal:")
    ? (Literal.fromUrl(target).get() as string)
    : target;
const readLiteralNumber = (target: string): number =>
  target.startsWith("literal:")
    ? (Literal.fromUrl(target).get() as number)
    : Number(target);
```

The three affected reads (`stateName`, `stateValue`, `actionName`) go through them. Legacy data still round-trips via `Literal.fromUrl`, migrated data reads the bare value. The JSON-payload readers (`flowable`, `startAction`, `stateCheck`, `transitionActions`) didn't need changes — their existing `.replace(/^literal:\/\/string:|^literal:string:/, '')` is a no-op on migrated targets and the downstream `decodeURIComponent + JSON.parse` is idempotent on already-decoded JSON.

Also adds a `SHACLFlow.test.ts` case that mirrors the migration's rewrite on the raw `toLinks()` output and asserts the round-trip. 10/10 pass locally.

## Tradeoff vs Option B

[...]

PR #896, opened 2026-08-17
fix(migration): scope typed-literal migration to SHACL property values

HexaField:
## What

Makes the v4 `literal:*` → typed-literal store migration respect SHACL, the same way the v3 envelope migration already does. It now rewrites only the targets of **declared deterministic-literal properties**; every other predicate stays untouched.

## Why this is the correct fix

The v4 migration exists to make **subject-class property values** hit the POS index. But it ran unconditionally over *every* reifier triple, so it also rewrote links that are not property values at all — notably `SHACLFlow`'s own `ad4m://stateName` / `stateValue` / `actionName` bookkeeping, written as `literal:string:…`. After migration those came back as bare `"foo"`, and `SHACLFlow.fromLinks` (via `Literal.fromUrl`) threw `Can't create Literal from non-literal URL`, taking down `PerspectiveProxy.getFlow` and the SHACL TODO-flow integration test.

The two earlier options both treat the symptom:

- **#890** teaches `SHACLFlow.fromLinks` to accept the migrated shape — but the migration should never have touched those links, and every other reader would need the same patch.
- **#891** removes the migration entirely — losing the perf win and leaving a mixed-shape store every reader must tolerate.

The migration is the thing that is wrong: it has no business rewriting non-property links. v3 already solved exactly this by deriving a SHACL predicate set before it runs; v4 simply never did the same check. This PR gives v4 that check — nothing else changes.

## Change

- Extract the SHACL classification v3 already does into one helper, `literal_storage_predicate_scopes()`, returning `(deterministic, keep_envelope)` predicate sets.
- v3 keeps its exact behaviour (uses `keep_envelope`).
- v4 uses `deterministic` as an allow-list: it skips any predicate that is not a declared deterministic-literal property. Non-property links (SHACLFlow bookkeeping, raw links) and envelope/custom-language properties keep their original form. With no shapes loaded it migrates nothing, rather than rewriting blindly.
- `SHACLFlow` needs no change — its links are never rewritten.

## Result

- Deterministic property values still migrate to typed literals → indexed WHERE (#842/#874 perf win) preserved.
- `SHACLFlow` bookkeeping stays `literal:*`, so `fromLinks` reads it unchanged and the TODO-flow test passes without the #890 reader patch (already reverted on this branch).
- Envelopes still preserved (v3 behaviour unchanged).

## Tests

- New `test_migration_v4_skips_non_property_literals`: a declared property migrates to a typed literal; a `SHACLFlow`-style `ad4m://stateName` link survives untouched. Fails on the unconditional migration.
- The existing v4 test now declares its property (the migration only touches declared properties).
- v3 tests unchanged and green — behaviour pinned.

### DIFF
--- a/rust-executor/src/perspectives/sparql_store.rs
+++ b/rust-executor/src/perspectives/sparql_store.rs
@@ -1426,6 +1426,88 @@ impl SparqlStore {
         Ok(count)
     }
 
+    /// Classify every SHACL property path by how its literal values are stored,
+    /// from `resolveLiteral` / `resolveLanguage`, exactly as
+    /// `ShapeProperty::is_deterministic_literal` does. Returns
+    /// `(deterministic_predicates, keep_envelope_predicates)`:
+    ///
+    /// - **deterministic** — bare `@Property` or `resolveLiteral: true`: values
+    ///   are plain typed literals, safe to normalize to the indexed shape.
+    /// - **keep_envelope** — `resolveLiteral: false`, an explicit
+    ///   `resolveLanguage: "literal"`, or a custom language: values carry an
+    ///   expression-level signature that must not be flattened or re-encoded.
+    ///
+    /// A predicate absent from *both* sets is not a SHACL property path at all
+    /// (e.g. `SHACLFlow`'s own `ad4m://stateName` / `stateValue` / `actionName`
+    /// bookkeeping links). Neither migration should touch those — they carry no
+    /// property value, are never model-queried, and their readers expect the
+    /// original `literal:*` IRI form.
+    ///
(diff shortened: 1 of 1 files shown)


HexaField:
Superseded by the migration-removal approach — the correct fix per further review. The write/read round-trip (`target_to_storage_term` / `storage_term_to_target_string`) already stores every scalar as a typed literal and renders it back to `literal:*` wire form on read, so SHACLFlow round-trips on a fresh store with no migration. With no users there is no pre-refactor data to convert, so both v3/v4 migrations are removed rather than scoped. See the new PR.

PR #899, opened 2026-08-17
fix(core): re-register query subscriptions on WebSocket reconnect

HexaField:
Server-side SPARQL/query subscriptions die when the WebSocket disconnects. Previously, clients waited up to 30s for the keepalive to fail before resubscribing — leaving a window where subscriptions silently received no updates.

## Changes

**ApiClient** — add `onReconnect()` callback registration:
- New `_reconnectCallbacks` set and `_hasConnectedOnce` flag
- Callbacks fire on successful WS `onopen` only after the first connection (not initial connect)
- Cleaned up in `closeAll()`

**PerspectiveClient** — passthrough `onReconnect()` to ApiClient.

**QuerySubscriptionProxy** — immediate re-subscription on reconnect:
- Registers an `onReconnect` handler during `subscribe()`
- On reconnect, immediately calls `subscribe()` to create a fresh server-side subscription
- New `#reconnectUnsub` field cleaned up in `dispose()`

## Tests

3 new tests in `PerspectiveProxy.test.ts`:
- `ApiClient.onReconnect` — fires only on reconnect, not first connect; unsubscribe removes callback
- `QuerySubscriptionProxy` — re-subscribes immediately when `onReconnect` fires; cleans up reconnect listener on dispose

All 9 tests pass.

---
Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added WebSocket reconnect callbacks with unsubscribe support.
  * Added harness-based transcript interpretation, observation options, and subject-class lookup.
  * Added SHACL discovery and retrieval APIs.
  * Added auto-processor removal and neighbourhood-state event listening.
  * Added cancellation and call-option support to query APIs.
  * Query subscriptions now automatically resubscribe after reconnection while preserving active listeners.

* **Bug Fixes**
  * Improved subscription recovery and cleanup after connection interruptions.
  * Prevented reconnect resubscription loops and logged callback errors.


lal-bot-coasys:
CodeRabbit's Major finding addressed in the merge+fix push (branch head is now `d24d38723`, on top of a fresh `origin/dev` merge).

**The bug traced end-to-end:**

1. Reconnect fires → our handler calls `subscribe()`.
2. `subscribe()` runs `#unsubscribe()`. That delegates to `ApiClient`'s `_wsCallbacks` deleter, which closes the WebSocket when this query owned the LAST callback (and no RPCs are pending).
3. `subscribe()` then calls `subscribeQuery()` → `ApiClient` re-opens the WS → fresh `onopen` fires every registered reconnect callback.
4. The OLD reconnect listener was still in the set — the previous version cleared it only at the END of `subscribe()`, *after* `subscribeQuery` had already returned. So the handler runs again → closes WS again → reopens → fires callbacks → endless loop.

**Fix:** move `#reconnectUnsub` cleanup to the TOP of `subscribe()`, before `#unsubscribe`. A fresh listener is installed at the end. Step 4's stale-listener re-entry cannot happen because the set is empty when `subscribeQuery` reopens the socket.

**Regression test:** `does not re-enter subscribe() when this query owned the last WS callback (regression)` in `PerspectiveProxy.test.ts`. Uses a real `Set`-backed `reconnectCallbacks` and a mock `subscribeQuery` that fires all currently-registered callbacks (modelling the fresh-socket `onopen` loop). Without the fix `subscribeQueryCount` grows without bound; with the fix it's exactly two — initial + one reconnect-driven.

Local: full core suite passes (512/512, 17 suites). CI running on `d24d38723` now; I'll ping if it goes red.

core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -107,12 +108,28 @@ export class QuerySubscriptionProxy {
...
+        // reconnect callback set. If the OLD reconnect listener is still
+        // in that set, it re-enters this method, closes the socket again,
+        // reopens again … an endless resubscribe loop. Clearing the
+        // listener up-front breaks the cycle; a fresh listener is
+        // installed at the end of a successful subscribe().
+        if (this.#reconnectUnsub) {
+            this.#reconnectUnsub();
+            this.#reconnectUnsub = undefined;
+        }
+
```
data-bot-coasys, line 140: **Blocking — this is CodeRabbit's unresolved thread below, confirmed real:** the listener is removed here unconditionally, but re-registered only after the `try`/`catch` completes without throwing. If `subscribeQuery` rejects (timeout, or a send racing a just-closed socket → 503), `subscribe()` rethrows → the registration at line 267 never runs. Both callers of retry-`subscribe()` then give up permanently: the keepalive loop `return`s on resubscribe failure, and the init-timeout path only logs. Net: one failed resubscribe during a flap and the proxy is dead — no keepalive, no reconnect listener, no updates, silently. That's precisely the failure mode this PR exists to eliminate.

Suggested minimal fix: capture the old unsub but don't run it yet — or simpler, wrap the removal+registration so failure restores recovery:

```ts
const priorReconnectUnsub = this.#reconnectUnsub;
this.#reconnectUnsub = undefined;
// ... existing body ...
// in the catch block, before rethrowing:
//   this.#reconnectUnsub = priorReconnectUnsub ?? this.#client.onReconnect(...same handler...);
// on the success path: priorReconnectUnsub?.();
```

Or equivalently: move the `onReconnect` registration into a `finally`. Either way, a failed resubscribe must leave *some* reconnect listener installed so the next `onopen` gets another chance. A test asserting "subscribeQuery rejects once → reconnect fires → subscription recovers" would lock it.

core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -220,6 +237,72 @@ export class QuerySubscriptionProxy {
...
+        // The correct shape is a swap-in-place: get a new server-side
+        // subscription ID via `subscribeQuery`, register a new client-side
+        // callback FIRST, and only then unsubscribe the old one. That way
+        // `_wsCallbacks.size` never dips to 0 during the transition, so the
+        // socket doesn't close, no reconnect loop, and no cross-proxy 503s.
+        // Any prior listener was already removed at the top of this method
+        // (see the note there for why cleanup must run before `#unsubscribe`,
+        // not here).
+        if (this.#client.onReconnect) {
+            this.#reconnectUnsub = this.#client.onReconnect(async () => {
```
data-bot-coasys, line 267: Non-blocking, but worth doing while you're in this file: there are now **three concurrent writers** to `#unsubscribe`/`#subscriptionId` — this handler, the keepalive-retry `subscribe()`, and the init-timeout retry `subscribe()` — and they can interleave during exactly the reconnect storms this PR targets (a drop that spans a 30s keepalive boundary runs the keepalive retry and this handler in the same window).

Worst interleaving I could construct: keepalive's `subscribe()` runs its synchronous cleanup (sets `#unsubscribe = undefined`) and parks on `await subscribeQuery`; this handler resumes first, registers its new callback, reads `#unsubscribe` (now `undefined`, so nothing to swap out), and assigns `#unsubscribe = newUnsub`; then keepalive's `subscribe()` resumes and overwrites `#unsubscribe` **without ever calling `newUnsub`**. That callback entry leaks in `_wsCallbacks` forever — `dispose()` can't reach it, so the socket is held open for the client's lifetime and every future push event iterates a dead callback.

A tiny generation counter closes all of these at once: bump `this.#generation` at the top of `subscribe()` and at handler entry, capture it, and after every `await` bail (undoing your own registration) if it no longer matches. The two pre-existing retry paths had the same latent race between themselves, so this fixes old code too.

core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -220,6 +237,72 @@ export class QuerySubscriptionProxy {
...
+                try {
+                    const newInitial = await this.#client.subscribeQuery(this.#uuid, this.#query);
+                    if (this.#disposed) return;
+                    const newSubId = newInitial.subscriptionId;
+
+                    // Register the NEW client-side callback BEFORE removing the
+                    // old one — keeps `_wsCallbacks.size >= 1` across the swap.
+                    const newUnsub = this.#client.subscribeToQueryUpdates(
+                        newSubId,
+                        (updateResult) => {
```
data-bot-coasys: Nit: this replacement callback drops the init-lifecycle handling its counterpart in `subscribe()` has (clear `#initTimeoutId`, resolve `#initialized`). If a reconnect swap lands while initialization is still pending (the no-initial-result path), updates flow but the 30s init timer still fires and triggers a spurious full resubscribe, and `#initialized` awaiters resolve later than they should. Self-healing, narrow window — but making the two callbacks identical (or extracting one shared function) removes the asymmetry for free.

core/src/apiClient.ts
```diff
@@ -444,5 +464,6 @@ export class ApiClient {
         this._pendingCalls.clear()
         this._closeWs()
         this._wsCallbacks.clear()
+        this._reconnectCallbacks.clear()
```
data-bot-coasys, line 467: Seconding CodeRabbit's outside-diff minor, still unaddressed on this head: `closeAll()` clears the callbacks but leaves `_hasConnectedOnce = true`. A reused client (`closeAll()` → later `call()`/`subscribe()` reopens via `_ensureWs`) fires newly registered `onReconnect` callbacks on the *first* open of the new connection, contradicting the documented contract at line 449. One line: `this._hasConnectedOnce = false` here.

core/src/perspectives/PerspectiveProxy.test.ts
```diff
@@ -84,6 +84,40 @@ describe('PerspectiveProxy.dispose', () => {
   });
 });
 
+describe('ApiClient.onReconnect', () => {
+  it('fires reconnect callbacks only on reconnect, not first connect', () => {
```
data-bot-coasys, line 115: Nit: this test doesn't exercise the production gate — it manually iterates the private `_reconnectCallbacks` set, so it verifies that a `Set` contains what was added to it, not that `onopen` skips the first connect. The `_hasConnectedOnce` flag it sets is never read by anything the test invokes. Since `ApiClient` already supports an injected `_webSocketImpl`, a fake WS whose `onopen` you fire twice would test the real path (first open → no callback, second → fired) in about the same number of lines. Same applies to CodeRabbit's open comment about the counter reset in the swap test making the two unsubscribers indistinguishable after `mockClear` — worth folding both into one test-hardening commit.

data-bot-coasys (changes requested):
The core design here is right and I want it merged: the swap-in-place reconnect handler (new server sub → register new client callback → only then drop the old one) is the correct answer to both the reconnect-loop and the cross-proxy 503s, the inline comments explaining *why* it must be shaped this way are genuinely good, and the `minLiveDuringSwap` regression test locks the invariant that matters. I traced the claims against `ApiClient` (`subscribe()`'s deleter closing the socket at zero callbacks, `_ensureWs`/`onopen` refiring) and they all hold.

Requesting changes for one thing only: **CodeRabbit's unresolved finding that a failed full resubscribe permanently disables reconnect recovery is real, and it's introduced by this PR.** `subscribe()` removes the reconnect listener in its first statement, but only re-registers it *after* the try/catch — so when `subscribeQuery` throws (flapping network is exactly when this happens), the rethrow skips registration, the keepalive loop has already stopped itself on resubscribe failure, and the proxy is permanently dead with no recovery path. Before this PR a failed keepalive-resubscribe was equally fatal, but there was no reconnect mechanism to lose; now there is one and this is the one path that silently destroys it. Small fix, detailed inline.

The rest is non-blocking: a concurrent-writers race between the reconnect handler and the keepalive/init-timeout retry paths (worth a generation guard while you're in here), CodeRabbit's `closeAll()`/`_hasConnectedOnce` minor which is still unaddressed on this head, and two test nits.

core/src/perspectives/PerspectiveProxy.ts
data-bot-coasys: Fixed in 5256a6e10 (implemented at Nico's request). The catch block now installs a recovery reconnect listener that retries the full `subscribe()` — loop-safe because a proxy in the failed state owns no `_wsCallbacks` entry, so the retry never closes/reopens the socket under itself. Regression test: `keeps reconnect recovery alive after a failed resubscribe attempt` (subscribeQuery rejects once → reconnect fires → proxy recovers on a fresh server subscription).

core/src/perspectives/PerspectiveProxy.ts
data-bot-coasys: Implemented in 5256a6e10 as the suggested generation guard: `#generation` is bumped by every writer (both `subscribe()` retry paths and the swap handler); after every await, stale continuations back out and release their orphaned server-side subscription via `disposeQuerySubscription` instead of clobbering newer state. The keepalive loop is generation-checked too (kills the orphaned-double-loop case), hoisted into `#startKeepalive(generation)` so the swap handler can restart it under its own generation — the swapped-in subscription keeps receiving keepalives on both the success and failure paths. Regression test: `a reconnect swap superseded by a concurrent resubscribe backs out without leaking callbacks` (deferred-promise interleaving, asserts exactly one live callback and the stale server sub disposed).

core/src/perspectives/PerspectiveProxy.ts
data-bot-coasys: Fixed in 5256a6e10 — extracted `#deliverResult()` (clear init timeout → resolve #initialized → record latest → notify) and used it in all three delivery paths: the initial-result path, the update callback, and the swap handler (both its update callback and its initial-result delivery). The asymmetry is gone.

core/src/apiClient.ts
data-bot-coasys: Fixed in 5256a6e10 — `closeAll()` now resets `_hasConnectedOnce = false`, with a test covering the reuse contract (`closeAll resets the first-connect gate for client reuse`).

core/src/perspectives/PerspectiveProxy.test.ts
data-bot-coasys: Fixed in 5256a6e10 — the `ApiClient.onReconnect` tests now inject a fake WebSocket impl (constructor arg 3) and drive the real `onopen`/`onclose` lifecycle: first open → no callback, drop + second open → fired once. The unsubscribe test asserts behaviour (no fire on reconnect) instead of private set size, and CodeRabbit's counter-reset nit in the swap test is fixed alongside (distinct unsubscribers, plus explicit old-disposed/new-not-disposed assertions).

core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -105,12 +106,28 @@ export class QuerySubscriptionProxy {
...
+            this.#reconnectUnsub = undefined;
+        }
+
         // Clean up previous subscription attempt if retrying
         if (this.#unsubscribe) {
             this.#unsubscribe();
             this.#unsubscribe = undefined;
         }
-        
+
```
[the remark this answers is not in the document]
data-bot-coasys: Confirmed real and fixed in 5256a6e10: the catch block now installs a recovery reconnect listener running the full `subscribe()` (loop-safe — a failed proxy owns no `_wsCallbacks` entry, so the retry can't close/reopen the socket under itself), and the keepalive comment documents that recovery is preserved. Regression test added.

core/src/perspectives/PerspectiveProxy.test.ts
```diff
@@ -122,4 +156,185 @@ describe('QuerySubscriptionProxy', () => {
...
+    const unsubReconnect = jest.fn();
+    const mockClient = {
+      subscribeQuery: jest.fn().mockResolvedValue({
+        subscriptionId: 'sub-1',
+        result: [{ s: 'a', p: 'b', o: 'c' }],
+      }),
+      subscribeToQueryUpdates: jest.fn(() => {
+        subscribeToUpdatesCall++;
+        return subscribeToUpdatesCall === 1 ? initialUnsubscribe : reconnectUnsubscribe;
+      }),
```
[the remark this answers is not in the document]
data-bot-coasys: Fixed in 5256a6e10 — the monotonic counter is no longer reset, so the two unsubscribers stay distinct, and the test now asserts both directions: `initialUnsubscribe` called exactly once, `reconnectUnsubscribe` not called during the swap and exactly once after `dispose()`.

data-bot-coasys (approved):
All five findings from my review — including CodeRabbit's two unresolved threads — are now implemented in `5256a6e10` (Nico asked me to apply them directly rather than wait). Summary:

- **Blocker fixed:** a failed resubscribe installs a recovery reconnect listener, so one `subscribeQuery` failure during a flap can no longer permanently kill the proxy. Loop-safety argument is in the code comment; regression test included.
- **Writer races closed** with a `#generation` guard: stale `subscribe()`/swap continuations back out and release their orphaned server-side subscription instead of leaking a `_wsCallbacks` entry. Keepalive hoisted to `#startKeepalive(generation)` and restarted by the swap handler so the swapped-in subscription keeps its keepalives.
- **Init-lifecycle symmetry** via shared `#deliverResult()` (all three delivery paths).
- **`closeAll()` resets `_hasConnectedOnce`** (CodeRabbit's outside-diff minor).
- **Tests hardened:** `ApiClient.onReconnect` tests drive the real `onopen`/`onclose` lifecycle via an injected fake WebSocket; the swap test keeps distinct unsubscribers; two new regression tests (failed-resubscribe recovery, superseded-swap back-out).

Verified: core suite 589/589 green locally, `tsc --noEmit` clean. Approving to clear my request-changes — but since the fixup commit is mine, **@HexaField please sanity-check `5256a6e10` before merge**; the generation-guard semantics in particular deserve your eyes since it touches your swap handler's control flow.

lucksus (approved):

PR #913, opened 2026-08-24
Feature: Auto processor observability extras

jhweir:
# Observability follow-ups: joinable event streams, and a one-shot pass that reports

Stacked on **#903**, which has since been merged into this branch. Five commits: two API additions,
one review fix, and two that fix a bug stopping auto-processing outright.

## Summary

#903 gives a UI two event streams and everything it needs to render a live extraction, with two gaps
that only show up once you try to build the UI. We hit both while wiring it into WE's call bar.

**The streams cannot be joined.** They are deliberately scoped differently — one DID-filtered and
detailed, one perspective-scoped and coarse — which is right, and is exactly why a consumer wants
both at once: the perspective-scoped one opens a row and names the claimant, the DID-scoped one
fills in the phases and the LLM payload for the pass this agent is running. But they share no
identifier. The first lists `item_ids`; the second carries a SHA-256 of them. Correlating meant
re-implementing `claim::batch_key` in TypeScript and matching the Rust serialization byte for byte.

**The one-shot path is silent.** #903 makes a *watch* fully observable and leaves
`runInterpretation` emitting nothing at all. Worth closing because that path blocks its caller for
the whole pass — seconds to minutes on a local model — so it is where progress reporting is most
useful and the only one that had none. The plumbing already
existed — `run_interpretation_with_strategy_and_model` takes an `emit_ctx` — but reaching it meant a
call site spelling out eleven arguments, nine of them the defaults `run_interpretation` already
picks, so the handler kept calling the silent wrapper.

Both are additive. Existing callers observe no behaviour change.

**And auto-processing was silently broken.** Found while testing the consumer: a processor's
`interpretationClasses` never reached the graph, so `load_processors` rejected every instance and no
pass ever ran. Cause and fix in section 3 — it is unrelated to the two additions above, but it was
found through them and is fixed here because it makes the feature untestable otherwise.

### 1. `batchKey` on `AutoProcessorEvent` (`89e51255`)

**`events.rs`** — new `batch_key: Option<String>` field and a `with_batch_key` builder, plus
`batch_key` on `InterpretationEmitContext` so the engine's mid-pass LLM events carry it too.

**`watcher.rs`** — `batch_key_hex` is derived once at the top of `run_one_pass` and attached by all
three `signal!` arms, by `Processed`, and by the emit context.

Deriving it there rather than at the claim is a second, smaller fix. It used to be computed at
`try_claim`, which left the signals emitted *before* that point — `BatchReady`, `BackedOff`,
`NotCandidate`, `AwaitingAuthor` — as the only ones with no batch to attribute them to. Those are
the ones a consumer most wants attributed: a pass that stands down is otherwise indistinguishable
from one that never ran.

**`perspective_instance.rs`** — the `batch_id` computation is hoisted above the `BatchReady` emit so
the first event of a pass carries the key. Otherwise a consumer has to buffer until the second event
tells it which row to open.

**`AutoProcessor.ts`** — `batchKey?: string`, optional because a pre-#903 executor sends none.
Absence means "cannot correlate", not an error.

### 2. One-shot `runInterpretation` observability (`615a78ed`)

**`run.rs`** — `run_interpretation_observed`: `run_interpretation`'s defaults plus `emit_ctx`. Never
persists debug, because a one-shot pass has no `AutoProcessorConfig` to carry a `persist_debug`
opt-in and defaulting it on would write tens of KB of prompt into the shared graph, syncing to every
peer, on every call.

**`types.rs`** — `RunInterpretationRequest` gains `observation_id` and `emit_debug_events`.

The id comes from the caller rather than the executor because `runInterpretation` is a single
blocking RPC: there is no earlier response that could hand back a server-minted id, so events tagged
with one would arrive with nothing to match them against. It is echoed as both `processor_id` and
`batch_key` — a one-shot pass has no standing processor and no batch, and collapsing the pair is
more honest than minting a second id that would always be 1:1 with the first.

**`perspectives_ws.rs`** — the handler builds an emit context when asked, and emits
`RunningInterpretation` → `Processed`/`Failed` on the DID-scoped stream plus `Claimed` →
`Finished`/`Abandoned` on the perspective-scoped one. Absent `observation_id`, every emit is skipped
and the handler behaves exactly as before.

`Processed` fires *after* the mint-scope links are written, not the moment interpretation returns —
it carries the bases, and a consumer reading that as "these records exist and are attached" would be
reading it half a write early.

**`events.rs`** — new `AutoProcessorStep::Failed`.

The enum could say `EmptyTranscript` and `ShapesMissing`, which are the pass correctly deciding
there is nothing to do, but had no way to say the model errored or timed out. Those were being
reported as the former, which sends someone to re-read their transcript when the actual problem is
their LLM configuration. Two very different messages, and the vocabulary could not tell them apart.

Both failure arms close the neighbourhood row before returning. A consumer that opened one on
`Claimed` and never heard again shows the pass running forever — and an error and a timeout are
precisely the two outcomes worth reporting.

The reason goes only to the owner's stream. Provider error text can name a model, an endpoint or an
account; the neighbourhood stream has no free-form field at all, which is what keeps that honest.
Peers learn the pass ended without committing; the person who ran it learns why.

**`PerspectiveProxy.runInterpretation`** — trailing options moved into a bag
(`{ existingScope, mintScope, observe }`). The scopes were already unreachable from the proxy for
want of positional room, and a fourth, fifth and sixth positional parameter would have made
`undefined, undefined, {…}` the normal way to ask for the only one most callers want. The
three-argument form is unchanged.

### 3. Auto-processing never ran (`584e018e`, `3ee2b5bc`, `2bef18e5`)

**The symptom.** Every `AutoProcessor` instance was rejected on load:

    load_processors: AutoProcessor instance `…` has missing / unparseable fields; skipping

once per watch tick, forever. Auto-processing stopped and nothing said why.

**The cause.** `interpretationClasses` is the class's only *collection*, so it carries
`ad4m://adder` rather than `ad4m://setter` — and `create_subject` / `update_subject` resolve only
the latter. Every class handed to them through the values map was discarded. The instance was
written with all eleven scalars intact and the one required collection missing, and the write
returned success.

That last part is why this took so long to find. The failure surfaced as a *reader* calling an
instance malformed, in a message naming neither the field nor its value, while the *writer* that
caused it reported nothing at all. Four plausible causes fitted the evidence equally well and all
four were wrong.

**`2bef18e5`** fixes it: the classes are written as links directly, which is what the collection's
own `addLink` action does, in the same batch as the instance so the two commit atomically.

**`3ee2b5bc`** is the commit that found it, and the one most worth reviewing on its own terms:
`create_subject` and `update_subject` now warn when they drop a supplied value. This is not
scaffolding — an API that accepts a value, returns success and discards it is a bug generator, and
this is the general case for *every* collection, not a quirk of this class. `CollectionBlock.children`,
`.comments`, `.signals` and every other HasMany take the same path today; most callers never notice
because the ORM adds members through `addLink` accessors instead.

**`584e018e`** makes the reader's complaint usable: it prints the instance on first failure and
suppresses repeats. Previously the same line fired several times a second per broken instance and
buried the log it was meant to serve.

**Deliberately not fixed:** `create_subject` still drops collection values for every other caller.
Making it fall back to `ad4m://adder` would change shared write semantics — it would start writing
values that every existing caller currently has dropped — so it wants its own review rather than a
ride on this PR.

## Known follow-ups

- **`observationId` is not validated.** A caller could pass a value colliding with a real
  `processor_id`, which would mix a one-shot pass into a watch's rows on any consumer keying by it.
  Harmless today (nothing keys on `processor_id` alone) and worth a namespace prefix if that changes.
- **`Failed` is only emitted from the one-shot path.** The watcher still reports a pass that errored
  as an abandoned claim with no fine-grained event. Same argument applies there; left out to keep
  this reviewable alongside #903.
- **`create_subject` cannot write collections**, as above. Worth fixing properly, and worth knowing
  about in the meantime for anyone hand-writing a SHACL class with a collection.
- **A stale hardwired SDNA is never refreshed.** `ensure_subject_class` decides a registered class
  is current by probing one path, and for `AutoProcessor` that path predates the properties #903
  added — so a perspective registered before the split keeps a shape without them. Found while
  chasing the above and **not** its cause; left out of this PR as unrelated and untested.
- **No `AwaitingAuthor`/`BackedOff` equivalent for one-shot**, correctly — there is no election —
  but it does mean the two paths emit overlapping-but-different step sets. Documented on the
  `AutoProcessorStep` union rather than papered over.

## Test plan

- [x] `cargo check --lib` — clean
- [x] `cargo fmt --check -p ad4m-executor` — clean
- [x] `tsc --noEmit` on `core/` — clean
- [x] No struct-literal construction of `AutoProcessorEvent` or `InterpretationEmitContext` outside
      `events.rs`, so the new fields break no call site; no exhaustive `match` on
      `AutoProcessorStep` anywhere, so the new variant is additive
- [ ] `cargo test perspectives::auto_processor` — not run (existing suite untouched by these
      changes, but worth a CI pass)
- [x] **Manual verification against a live node** — auto-extraction confirmed working end to end
      after the fix, in a space where it had been failing on every call

## Sequencing

Both commits sit on #903's head and neither touches a line #903 changed after `33f70226`, so the
branch rebases cleanly.

Worth deciding explicitly: **`batchKey` has a compatibility argument for going into #903 itself**
rather than following it. It is a field on a wire type, and adding it after merge is breaking-ish
for anyone who adopts the streams in between — they would build the client-side hash workaround and
then have to unbuild it. Against that, #903 is already approved, and pushing to it disturbs that.

Either is fine from our side. The field is optional on both the Rust and TS types, so arriving late
degrades to "cannot correlate" rather than breaking anything already written.

The one-shot commit has no such argument and is happy as a follow-up.

## Why we needed these

Building the consumer #903 was written for — a live extraction readout under WE's call bar, showing
every member of a call who is extracting, what phase it has reached, and (for the runner) the raw
prompt and response. That work is on `feat/extraction-observability` in the WE repo and is inert
until these land.

### DIFF
--- a/rust-executor/src/api/perspectives_ws.rs
+++ b/rust-executor/src/api/perspectives_ws.rs
@@ -1114,11 +1114,14 @@ async fn run_interpretation_handler(
     /*
        Live observability for a one-shot pass (#903 follow-up).
 
-       A watch pass is fully observable and a button press was not, which is backwards: the press
-       is the one a person is sitting and waiting on. `observation_id` is the caller's own row id
-       (see `RunInterpretationRequest`), used as both `processor_id` and `batch_key` so a consumer
-       merging the two event streams handles a one-shot pass with the code it already has for a
-       watch pass.
+       A watch pass is fully observable; this path was not. The asymmetry is worth closing because
+       `runInterpretation` is synchronous for the caller — it blocks for the whole pass, which is
+       seconds to minutes on a local model — so it is the path where progress reporting is most
+       useful and the only one that had none.
+
+       `observation_id` is the caller's own identifier (see `RunInterpretationRequest`), used as
+       both `processor_id` and `batch_key` so a consumer merging the two event streams handles a
+       one-shot pass with the code it already has for a watch pass.
 
        Absent = every emit below is skipped and this handler behaves exactly as it did.
     */
--- a/rust-executor/src/api/perspectives_ws.rs
+++ b/rust-executor/src/api/perspectives_ws.rs
@@ -1111,6 +1111,57 @@ async fn run_interpretation_handler(
         std::collections::HashSet::new()
     };
 
+    /*
+       Live observability for a one-shot pass (#903 follow-up).
+
+       A watch pass is fully observable and a button press was not, which is backwards: the press
+       is the one a person is sitting and waiting on. `observation_id` is the caller's own row id
+       (see `RunInterpretationRequest`), used as both `processor_id` and `batch_key` so a consumer
+       merging the two event streams handles a one-shot pass with the code it already has for a
+       watch pass.
+
+       Absent = every emit below is skipped and this handler behaves exactly as it did.
+    */
+    let observation = body.observation_id.clone();
+    let observer_did = match observation.as_ref() {
+        // Best-effort: telemetry must never be the reason a pass refuses to run, so an
+        // unresolvable DID downgrades to no observation rather than to an error.
+        Some(_) => crate::agent::did_for_context(&agent_context).ok(),
+        None => None,
+    };
+    let observe = observation.as_ref().zip(observer_did.as_ref());
+    let emit_ctx = observe
+        .filter(|_| body.emit_debug_events.unwrap_or(false))
+        .map(
+            |(id, did)| crate::perspectives::auto_processor::events::InterpretationEmitContext {
+                perspective_uuid: uuid.clone(),
+                processor_id: id.clone(),
+                agent_did: did.clone(),
+                // No source item ids: the caller supplied the transcript directly rather than the
+                // watcher gathering it, so there is nothing to list. `batch_key` carries the identity
+                // instead — which is why it is a separate field rather than derived from these.
+                item_ids: Vec::new(),
+                batch_key: id.clone(),
+            },
+        );
+
+    if let Some((id, did)) = observe {
+        emit_one_shot_step(
+            &uuid,
+            id,
+            did,
+            crate::perspectives::auto_processor::events::AutoProcessorStep::RunningInterpretation,
+        )
+        .await;
+        emit_one_shot_phase(
+            &uuid,
+            id,
+            did,
+            crate::perspectives::auto_processor::events::NeighbourhoodPhase::Claimed,
+        )
+        .await;
+    }
+
     // Bound the LLM call: a stalled provider must not pin a WS request slot
     // indefinitely. Matches the pattern used by `query_sparql` /
     // `model_query_handler` / `evaluate_getters_handler`, but with a longer
@@ -1118,7 +1169,7 @@ async fn run_interpretation_handler(
     // `RUN_INTERPRETATION_TIMEOUT_SECS`).
     let bases = match tokio::time::timeout(
         Duration::from_secs(RUN_INTERPRETATION_TIMEOUT_SECS),
-        crate::perspectives::interpretation::run_interpretation(
+        crate::perspectives::interpretation::run_interpretation_observed(
             &mut perspective,
             &shapes,
             &transcript,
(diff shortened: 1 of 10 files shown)


rust-executor/src/api/perspectives_ws.rs
```diff
@@ -1111,14 +1111,65 @@ async fn run_interpretation_handler(
         std::collections::HashSet::new()
     };
 
+    /*
+       Live observability for a one-shot pass (#903 follow-up).
+
+       A watch pass is fully observable and a button press was not, which is backwards: the press
```
lucksus: what "button press"??

lucksus (approved):

lucksus (approved):

PR #778, opened 2026-03-27
feat: Docker image for headless AD4M executor

HexaField:
## Summary

Multi-stage Docker build (Ubuntu 24.04) producing a minimal runtime image with `ad4m` + `ad4m-executor` binaries. Provides complete headless operation — no Tauri required. Supports three optional build-time features: WE web frontend, pre-cached Kalosm AI models, and Holochain P2P networking.

### What changed

- **Dockerfile**: multi-stage build (toolchain → WE frontend → model fetcher → minimal runtime). Bypasses Cargo's libgit2 submodule resolution for rusty_v8/chromium-buildtools by rewriting git deps to local path deps. Includes HEALTHCHECK via curl. Selective per-directory COPY in builder isolates entrypoint changes from the Rust build cache. Installs websocat for WS-RPC model registration.
- **docker-compose.yml**: offset ports to avoid collisions with local dev executors. Named volume for persistent data. Docker-level healthcheck.
- **docker-entrypoint.sh**: gosu for non-root execution, env-var driven config (`ADMIN_CREDENTIAL`, `ENABLE_MULTI_USER`, `ENABLE_MCP`, `RUN_HOLOCHAIN`), Kalosm model cache pre-population, Caddy reverse proxy serving WE + Flux + executor API on a single origin. Agent auto-generation and auto-unlock with robust CLI output parsing (handles ANSI codes, snake_case, and known CLI typos). AI model auto-registration via WS-RPC (Whisper full + Whisper tiny quantized for transcription, TinyLlama LLM). Global discovery space creation. WE auto-auth injection.
- **Bootstrap languages**: local-storage implementations of link, neighbourhood, perspective, file-storage, and language languages for standalone (no-Holochain) mode. Deterministic content-addressed hashing for expressions.
- **Seed generator**: `docker/generate-seed.mjs` produces `docker_seed.json` with content-addressed bootstrap language hashes at build time. Generates language-language KV store entries (meta + bundle) for `applyTemplateAndPublish`.
- **Model downloader**: `docker/download-models.sh` fetches embedding (BGE), transcription (Whisper small + Whisper tiny quantized), and LLM (TinyLlama GGUF + llama tokenizer) into the kalosm cache layout.
- **`run_holochain` config flag**: new `Ad4mConfig` field controls whether the Holochain conductor starts. Defaults to `true` for backwards compatibility; Docker defaults to `false` via environment variable.
[...]

PR #918, opened 2026-08-25
feat: local bootstrap languages + --run-holochain flag

HexaField:
### What

This PR adds local-only bootstrap languages and the test infrastructure to run integration tests without Holochain. The executor gains a `--run-holochain` flag that skips the Holochain conductor, signal receiver, and Unyt service.

### Why

The Holochain-backed integration test path carries overhead: conductor startup, DHT bootstrap, DNA compilation, and gossip delays. Most test suites (email verification, MCP, model tests) do not exercise Holochain-specific behaviour. A local-only mode lets those suites run faster and without network dependencies. It also provides an offline development path for operators who do not need DHT sync.

#### Local bootstrap languages

`bootstrap-languages/local/` contains pure-JS languages that store data in the executor's KV store (no Holochain conductor):
- `agent-language.js` — agent expression store
- `language-language.js` — language meta + bundle persistence via File I/O extension
- `perspective-language.js` — perspective expression store
- `neighbourhood-language.js` — neighbourhood expression store
- `link-language.js` — link persistence with source/target/predicate indexes
- `file-storage-language.js` — file content storage
- `generate-seed.mjs` — generates a bootstrap seed JSON from the local language bundles

#### Executor changes

`rust-executor/src/lib.rs` wraps Holochain-dependent startup in a `run_holochain` guard:
- Holochain signal receiver, Unyt alliance DNA install, and payment polling only spawn when `run_holochain` is true.
- `rust-executor/src/config.rs` adds `run_holochain: Option<bool>` to `Ad4mConfig`.
- `cli/src/ad4m_executor.rs` adds the `--run-holochain` CLI flag.

#### Test infrastructure

- `tests/js/tests/test-context.ts` — `TestContext` class extracted from `integration.test.ts`. Encapsulates Alice/Bob/Jim clients, executor processes, and `makeAllNodesKnown()` agent-info exchange.
- `tests/js/tests/integration-local.test.ts` — local-only test suite entry point. Starts executors with `runHolochain: false`, skips HC conductor.
- `tests/js/scripts/get-builtin-test-langs.js` — `--local` flag selects local JS languages instead of HC-backed ones.
- `tests/js/package.json` — new scripts: `prepare-test-local`, `test-main-local`, `test-main-holochain`, `test-integration-local`.
- Shell scripts (`email-verification-test-with-setup.sh`, `test-multi-user-with-setup.sh`) — removed hardcoded setup port ranges (publishTestLangs now uses `getFreePorts`), added `--local` passthrough.

#### Files changed (34 files, +888 / -126)

| Area | Files |
|---|---|
| Local languages | `bootstrap-languages/local/*.js`, `generate-seed.mjs`, `package.json` |
| Executor | `rust-executor/src/lib.rs`, `rust-executor/src/config.rs`, `cli/src/ad4m_executor.rs`, `cli/src/dev.rs` |
| Test harness | `tests/js/tests/test-context.ts`, `integration-local.test.ts`, `tests/js/utils/publishTestLangs.ts`, `tests/js/utils/utils.ts`, `tests/js/package.json` |
| Test suites | `tests/js/tests/integration.test.ts` (slimmed), shell scripts, 10 test files (import path updates) |
| Docs | `AGENTS.md`, `README.md` |
| Build | `.gitignore` |

#### References

- Split from [#778](https://github.com/coasys/ad4m/pull/778) (closed — scope absorbed by this PR + [#941](https://github.com/coasys/ad4m/pull/941) + [#914](https://github.com/coasys/ad4m/pull/914))
- Link server + server-link-language (already merged): [#930](https://github.com/coasys/ad4m/pull/930)
- Seed server-link-language (already merged): [#1037](https://github.com/coasys/ad4m/pull/1037)

## Summary by CodeRabbit

* **New Features**
  * Added local, standalone operation without requiring Holochain.
  * Added local storage support for agents, files, languages, links, neighbourhoods, and perspectives.
  * Added self-hosted link synchronization through a server-based alternative.
  * Added optional dynamic MCP tools and configurable Holochain startup.

* **Improvements**
  * Improved integration testing with dedicated local and Holochain test suites.
  * Added scheduled nightly LLM end-to-end testing controls.
  * Improved executor startup, cleanup, and agent publishing reliability.

* **Documentation**
  * Documented local operation and link-language options.


tests/js/helpers/executor.ts
```diff
@@ -83,6 +86,7 @@ export async function startAgent(
   fs.rmSync(appDataPath, { recursive: true, force: true });
 
   const bootstrapSeedPath = opts.bootstrapSeedPath ?? BOOTSTRAP_SEED;
+  const runHolochain = opts.runHolochain ?? (process.env.LOCAL_MODE !== 'true');
```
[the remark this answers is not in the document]
HexaField: Fixed in d8c0543e1:

1. `setup_local_integration_test_environment` now exports `LOCAL_MODE=true` via `BASH_ENV`, so all downstream steps inherit it.
2. Doc comment on `runHolochain` updated to match the exact `LOCAL_MODE=true` comparison.
3. `email-verification.test.ts` reads `LOCAL_MODE` and passes `runHolochain` through to `startExecutor`.
4. `startExecutor` in `utils.ts` now guards `ensureSharedLocalServices()` with `runHolochain`, so local-mode executors never start `kitsune2-bootstrap-srv`.

⬡

tests/js/utils/utils.ts
```diff
@@ -217,6 +217,7 @@ export async function startExecutor(dataPath: string,
     relayUrl?: string,
     enableMcp: boolean = false,
     mcpPort?: number,
+    runHolochain: boolean = true,
```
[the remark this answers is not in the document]
HexaField: Fixed in 36c16351e — HC args (proxy URL, bootstrap URL, admin/app ports, use-bootstrap, use-proxy, use-local-proxy, use-mdns) now only appear in the spawn args when `runHolochain` is true. Local-mode executors receive a clean arg list with no undefined values.

The `ensureSharedLocalServices()` guard was already in place from d8c0543e1.

rust-executor/src/lib.rs
```diff
@@ -576,29 +576,36 @@ pub async fn run(mut config: Ad4mConfig) -> JoinHandle<()> {
         });
     };
 
-    // Start holochain signal receiver as standalone task
-    tokio::spawn(crate::holochain_signal_receiver());
+    // Start holochain signal receiver and Unyt service only when holochain runs
+    if config.run_holochain.unwrap_or(true) {
```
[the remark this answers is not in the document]
HexaField: Already addressed in d8c0543e1 — both `agent.generate` and `agent.unlock` now check `config.run_holochain` before calling `HolochainService::init`. See lines 300 and 411 of `agent_ws.rs`.

lal-bot-coasys (dismissed):
Reviewed the real diff against a freshly-fetched dev (my first attempt used a stale local dev ref and showed 106 unrelated files — worth mentioning since it's exactly the kind of stale-artifact trap that bites this repo's CI checks too). Actual diff: 41 files, +1060/-201, matches the PR description.

**Verdict: APPROVE.**

**Rust side, checked for the thing that actually matters here — a missed `Ad4mConfig` construction site would be a compile error, not a subtle bug.** Grepped every `Ad4mConfig {` struct literal in the repo, not just the ones in the diff: `cli/src/ad4m_executor.rs`, `cli/src/dev.rs` (x2), `rust-executor/src/agent/mod.rs`, and `rust-executor/src/config.rs`'s own `Default` impl all explicitly set `run_holochain`. Two sites the diff doesn't touch — `cli/src/main.rs:252` and `ui/src-tauri/src/lib.rs:360` (the Tauri launcher) — both use `..Default::default()`, so they pick up `run_holochain: None` from the `Default` impl and normalize to `Some(true)` exactly like before. No missed site, and the launcher's existing Holochain-on-by-default behavior is untouched. The guard itself is applied consistently in `lib.rs` (signal receiver, Unyt DNA install, payment polling) and `agent_ws.rs` (`generate_agent` and `unlock_agent`'s conductor init) — same `config.run_holochain.unwrap_or(true)` check in every spot, same safe default.

**CI restructure — traced the actual script chain, not just the job names, since a broken reference here fails silently on the wrong branch.** `test-main-local` → `prepare-test-local` → `test-all-local` → `test-integration-local` → `tests/integration-local.test.ts` (exists, confirmed in the diff). `test-main-holochain` → `test-all-holochain` → `test-integration-holochain` → the *same* `tests/integration.test.ts` the old `test-main` used, just renamed at the script level, so the Holochain-backed suite's actual test file is untouched. No dangling references.

**`test-context.ts` extraction is a genuine improvement, not just a move.** The extracted `makeAllNodesKnown()` now throws `lastError` if all 5 agent-info-exchange attempts fail — the original silently `break`'d out of the retry loop after logging, so a test calling it could proceed as if the exchange succeeded and fail confusingly downstream instead of failing at the actual point of breakage. New `makeAllThreeNodesKnown()` genuinely adds Jim as a third agent, matching `triple-agent-test.ts` in the diff.

**Two small things, neither blocking:**
1. `makeAllThreeNodesKnown()` computes `const allInfos = [...aliceAgentInfo, ...bobAgentInfo, ...jimAgentInfo]` and never uses it — the actual `hcAddAgentInfos` calls below build their own per-agent subsets directly. `noUnusedLocals` is commented out in `tests/js/tsconfig.json` so this won't break the build, just dead code worth a quick removal.
2. `prepare-test-local` only has `:macos`/`:linux` variants (via `run-script-os`), unlike `prepare-test`'s `:macos`/`:linux`/`:windows`. Given the new CI jobs pin `resource_class: coasys/marvin`, this is very likely intentional (local mode isn't wired into Windows CI yet), but worth a one-line confirmation from Josh that it's a deliberate scope decision and not an oversight, since a future PR adding Windows local-mode CI would silently no-op without a `:windows` script to call.

Nothing else — the local bootstrap languages (spot-checked `link-language.js`'s commit/sync path) are simple, scoped correctly to single-node local/dev use, and don't try to be more than that.

tests/js/utils/utils.ts
```diff
@@ -217,12 +217,13 @@ export async function startExecutor(dataPath: string,
...
```
data-bot-coasys: **Blocking — this parameter position is what is failing `integration-tests-mcp`.**

`runHolochain` was inserted at position 13, before `dynamicClassTools`. `tests/js/tests/mcp-http.test.ts:213` already passes a 13th positional argument:

```ts
    true,               // enableMcp = true
    MCP_PORT,           // mcpPort
    true,               // dynamicClassTools = true — section 4 exercises the
                        // per-class tools, which are hidden by default.
);
```

After this change that `true` binds to `runHolochain`, and `dynamicClassTools` falls back to its `false` default, so `--dynamic-class-tools` is never appended to the spawn args. CircleCI job 29644 on this head shows exactly that signature:

```
11) 4. Dynamic Tool Generation
      should add new SDNA and see new tools appear:
    AssertionError: expected [ 'get_children', …(42) ] to include 'task_create'
12) 5. Dynamic Collection & Per-Property Tools
      should have generated per-property set tools for Channel:
    AssertionError: expected [ 'get_children', …(42) ] to include 'channel_set_name'
```

plus ~20 `expected undefined to be true` from the per-class call sites in sections 4, 5, 5b and 8. `mcp-http.test.ts` is the only caller in `tests/js` that passes 13 arguments — I checked every `startExecutor(` call site on this head — so this is the whole blast radius, but it is a silent one: nothing about the call site changed, so nothing pointed at it.

Minimal fix: move `runHolochain` after `dynamicClassTools` so existing positional calls keep their meaning. Better fix, given this is the second optional boolean tacked onto a 14-argument positional signature: collapse everything after `hcAppPort` into an options object, so the next parameter cannot do this again.

rust-executor/src/api/agent_ws.rs
```diff
@@ -307,20 +307,24 @@ async fn generate_agent(params: Value, ctx: Arc<RequestContext>) -> Result<Value
...
-        &config,
-        body.passphrase.clone(),
-    );
-
     let mut init_errors: Vec<String> = Vec::new();
 
-    if let Err(e) = crate::holochain_service::HolochainService::init(hc_config).await {
-        log::error!("Error initializing Holochain: {:?}", e);
-        init_errors.push(format!("Holochain init failed: {}", e));
+    if config.run_holochain.unwrap_or(true) {
```
data-bot-coasys, line 314: **Blocking — the guard is right here but three other Holochain entry points were missed.**

`agent.generate` and `agent.unlock` are now guarded, but `rust-executor/src/api/runtime_ws.rs` still reaches the conductor unconditionally:

- `:198` `restart_holochain` → `let _ = get_holochain_service().await;`
- `:474` `get_hc_agent_infos`
- `:489` `add_hc_agent_infos`
- `:504` `get_network_metrics`

`get_holochain_service()` (`rust-executor/src/holochain_service/interface.rs:292-310`) polls the global every 200 ms and then:

```rust
if tokio::time::Instant::now() >= deadline {
    panic!("Holochain Conductor not started after 120s timeout");
}
```

With `run_holochain = false` that global is never set, so the failure mode is a 120-second stall followed by a panicking handler task — not an error the client can act on. Failure scenario: `ad4m-executor run --run-holochain false`, then any client calls `ad4m.runtime.hcAgentInfos()` (or the AD4M UI opens the network panel, which calls `getNetworkMetrics`). Two minutes of nothing, then a panic.

This is visible in the PR's own test changes: `tests/js/tests/runtime.ts:137` became `(hasHolochain ? it : it.skip)(...)`. Skipping is the right call for the HC-specific assertion, but it means the new mode's behaviour for these four RPCs is undefined rather than specified.

Suggestion: check `config.run_holochain` in each of the four handlers and return `WsRpcError::bad_request("Holochain is disabled on this executor (run_holochain=false)")`. Then `runtime.ts` can assert *that* instead of skipping, which gives the local mode a contract. (Read-only analysis — I have no executor here and did not reproduce the 120 s stall.)

bootstrap-languages/local/generate-seed.mjs
```diff
@@ -0,0 +1,164 @@
...
+    const langDir = join(outputDir, "languages", hash);
+    mkdirSync(langDir, { recursive: true });
+    writeFileSync(join(langDir, "bundle.js"), content);
+
+    console.log(lang.role + ": " + hash);
+}
+
+// Build seed JSON
+const seed = {
+    trustedAgents: [],
```
data-bot-coasys, line 104: **This file is a stale fork of `docker/generate-seed.mjs`, and the fork drops that file's security hardening.**

`docker/generate-seed.mjs` already exists on `dev` and this is ~90% the same script. The difference that matters is right here. The `dev` version (added in `8c4e13ecb`, "fix(docker): trustedAgents from committed seed; stamp language authors at boot") reads the trusted list from the committed seed and refuses to produce a seed without one:

```js
// Trusted agents MUST come from the bootstrap seed committed in this repo
// (`rust-executor/src/mainnet_seed.json`). … language install refuses any
// language whose Expression `author` is not in that list (or equal to the
// local agent DID). An empty array here is not a valid standalone-mode shortcut.
…
if (trustedAgents.length === 0) {
    console.error("Committed bootstrap seed has no trustedAgents: " + committedSeedPath);
    process.exit(1);
}
```

This copy hardcodes `trustedAgents: []` and deletes both the check and the comment explaining why it exists. It also writes language meta with `author: ""` (lines 130/134) but drops the `dev` version's note that the empty author is deliberate because `stamp_bootstrap_language_authors` in `docker-entrypoint.sh` rewrites it to the container DID on first boot — and there is no equivalent stamping step on this PR's path.

Failure scenario, reasoned from `rust-executor/src/languages/mod.rs` (not run): an operator generates a seed with this script and starts an executor with it. `trusted_agents` is empty, so at `:1768` the trusted-author path only matches languages authored by the local DID. A templated language — e.g. Flux `createCommunity` → `applyTemplateAndPublish` on the link language — takes the `else` branch, reads the *source* language meta, gets `author: ""` from the KV this script wrote, and hits `:1896`:

> `Agent which created source language for language trying to be installed is not a trustedAgent... aborting language install`

Separately: the `dist/` this produces is gitignored and nothing reads it. The Dockerfile still reads `docker/seed-output/`, produced by `docker/generate-seed.mjs` from `docker/bootstrap-languages/` — which is untracked, because the repo-wide `*.js` ignore swallows it and only `bootstrap-languages/local/*.js` got a negation in this PR. So after this merges there are two hash-computation implementations that must stay byte-identical to `calculate_language_hash()` and two copies of the language sources, one tracked and one not.

The reuse-shaped fix is better and smaller: point `@coasys/local-bootstrap-languages#build` at `docker/generate-seed.mjs` and delete this copy, or move these six `.js` files to `docker/bootstrap-languages/` (with the matching gitignore negation) so the Docker build finally has tracked sources and there is one seed generator.

bootstrap-languages/local/link-language.js
```diff
@@ -0,0 +1,106 @@
...
+    return { links: links };
+}
+
+export async function perspectiveSyncCurrentRevision() {
+    return revision.toString();
+}
+
+// -- commit capability --
+
+function linkKey(l) {
```
data-bot-coasys, line 71: **Removal matching is key-order and format sensitive, and a non-match is silent.** (Reasoned from the code — I did not construct the mismatching input, so treat the severity as unconfirmed.)

`linkKey` identifies a link by `JSON.stringify({author, timestamp, data})`. `JSON.stringify` preserves insertion order, so two structurally identical links whose `data` objects were built with different key order — or whose `timestamp` was reserialised to a different ISO precision anywhere in the client → RPC → Deno path — produce different keys.

When that happens `perspectiveCommit` filters nothing out, then returns `""` like a success. The caller sees the removal accepted; the link is still in `links`, is still written by `saveState()`, and reappears via `perspectiveSyncRender()` and after the next restart. A silently-ignored delete is the worst shape this bug can take.

Suggestion: key on a canonicalised form — sorted `data` keys and the link's own `data.source`/`data.predicate`/`data.target` triple rather than the whole envelope — or compare against whatever `p-diff-sync` uses so the two link languages agree on link identity.

Worth noting this whole file is currently unexercised: `get-builtin-test-langs.js` maps it to `perspective-diff-sync` in local mode, but `integration-local.test.ts` runs no Neighbourhood suite, so `perspectiveCommit` is never called in CI.

bootstrap-languages/local/link-language.js
```diff
@@ -0,0 +1,106 @@
...
+    for (var i = 0; i < additions.length; i++) {
+        var key = linkKey(additions[i]);
+        if (!existing.has(key)) {
+            existing.add(key);
+            links.push(additions[i]);
+        }
+    }
+
+    revision = revision + 1;
+    saveState();
```
data-bot-coasys, line 100: **Nit (efficiency, concrete):** `saveState()` re-serialises and rewrites the entire link array on every commit, so building up a perspective is O(N²) in bytes written — the 5,000th link add rewrites all 5,000 links. The PR body offers this as "an offline development path for operators who do not need DHT sync", which is exactly the workload that accumulates links over time. A batched/debounced flush, or an append-oriented layout keyed per link, would keep it linear.

bootstrap-languages/local/link-language.js
```diff
@@ -0,0 +1,106 @@
...
+
+function loadState() {
+    try {
+        const raw = storageGet("link-state");
+        if (raw) {
+            const state = JSON.parse(raw);
+            links = state.links || [];
+            revision = state.revision || 0;
+        }
+    } catch (_) {
```
data-bot-coasys: **Silent data loss on a corrupt read.** This `catch` resets `links = []` and `revision = 0`, so a truncated or partially-written `link-state` entry is indistinguishable from "this is a fresh perspective": the executor starts up cheerfully with an empty perspective, the next `perspectiveCommit` calls `saveState()`, and the original links are overwritten and gone. Nothing is logged.

The same `catch (_) { return null }` shape is in every sibling language — `agent-language.js:51`, `perspective-language.js:28`, `neighbourhood-language.js:28`, `file-storage-language.js:41` — where the consequence is milder ("expression not found" instead of "storage unreadable"), but the diagnosis is equally impossible.

At minimum log the caught error. For `loadState` specifically, a parse failure should be loud and should *not* silently authorise overwriting the file.

tests/js/tests/integration-local.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      }
+      deregisterPorts([apiPort, hcAdminPort, hcAppPort]);
+    })
+
+    describe('Agent / Agent-Setup', agentTests(testContext))
+    describe('Artificial Intelligence', aiTests(testContext))
+    describe('Runtime', runtimeTests(testContext, { hasHolochain: false }))
+    describe('Expression', expressionTests(testContext))
+    describe('Perspective', perspectiveTests(testContext))
+    describe('Flat Language (new flat export pattern)', flatLanguageTests(testContext))
```
data-bot-coasys, line 90: **Coverage gap worth naming in the PR body.** The suite list here is Agent, AI, Runtime, Expression, Perspective, Flat Language — none of which create a neighbourhood. Perspectives without a neighbourhood do not go through a link language at all, so `bootstrap-languages/local/link-language.js` (106 lines: commit, sync, render, revision, removal matching, persistence) is never executed by `test-all-local`, even though `get-builtin-test-langs.js` installs it as `perspective-diff-sync` in local mode.

The header comment explains why the Alice+Bob tests stayed in `integration.test.ts` — they exercise cross-peer Holochain sync — and that is fair. But a single-node neighbourhood test would still exercise commit/render/persistence of the local link language without needing a second peer, and would be the only thing standing between that file and a silent regression.

data-bot-coasys (changes requested):
Reviewed the full diff against `origin/dev` (merge-base `4d944aa3d`), the existing review threads, and the CircleCI logs for this head.

**What I verified vs. what I inferred.** I read the diff, the surrounding `dev` sources it touches (`holochain_service/interface.rs`, `languages/mod.rs`, `runtime_ws.rs`), and downloaded the raw step logs for CircleCI jobs 29643 and 29644. I did **not** run the test suites, start an executor, or run Holochain — nothing below is a local reproduction. Where a finding is reasoning rather than observation I say so on the comment.

**CI is not green.** Two checks are red on `06b7b27c1`:

| job | result |
|---|---|
| `integration-tests-mcp` (29644) | **failed** — 29 failures in `mcp-http.test.ts` |
| `integration-tests-holochain` (29643) | **failed** — 89 passing, 1 failing |
| everything else | pass |

### Blocking

**1. The new `runHolochain` parameter silently steals an existing call site's argument, and that is what is turning `integration-tests-mcp` red.** `runHolochain` was inserted at position 13 of `startExecutor`, *before* `dynamicClassTools` (`tests/js/utils/utils.ts:220`). `tests/js/tests/mcp-http.test.ts:213` already passed a 13th positional `true` commented `// dynamicClassTools = true`. That `true` now binds to `runHolochain`, `dynamicClassTools` falls back to `false`, and `--dynamic-class-tools` is never passed to the executor. The CI log matches exactly: `expected [ 'get_children', …(42) ] to include 'task_create'`, `… to include 'channel_set_name'`, and a long tail of `expected undefined to be true` from sections 4, 5, 5b and 8 — the per-class tools are simply absent from the advertised surface. As a bonus the same flip starts a Holochain conductor in a job that `setup_local_integration_test_environment` configured for local mode. Details and suggested fix in the inline comment.

**2. `integration-tests-holochain` is red too.** One failure: `Neighbourhood [holochain] … they can send signals via sendSignal` → `Could not send signal - no AgentPubKey found for DID` from `perspective_diff_sync::utils:28`. I could not attribute this to the PR and it looks like the familiar gossip-timing flake — but this PR changes when that suite runs (`integration.test.ts` now runs alone under `test-main-holochain` instead of after simple/app/auth), which changes DHT warm-up timing. Please re-run it rather than assume flake; if it reproduces it is in scope.

**3. Four runtime RPCs hang for 120 seconds and then panic under `--run-holochain false`.** The PR guards `agent.generate` and `agent.unlock`, but `runtime_ws.rs:198` (`restartHolochain`), `:474` (`hcAgentInfos`), `:489` (`hcAddAgentInfos`) and `:504` (`getNetworkMetrics`) still call `get_holochain_service()`, which polls the never-populated global and then `panic!("Holochain Conductor not started after 120s timeout")` (`holochain_service/interface.rs:292-310`). Concrete: start an executor with `--run-holochain false`, call `ad4m.runtime.hcAgentInfos()` → the call stalls for two minutes and the handler task panics. The tell is in this PR's own diff: `tests/js/tests/runtime.ts:137` had to become `it.skip` in local mode. Skipping the test documents that the API is broken in the new mode instead of giving it a contract. These should return a typed error when `config.run_holochain == Some(false)`.

### Non-blocking, but please address before merge

**4. `bootstrap-languages/local/generate-seed.mjs` is a stale fork of `docker/generate-seed.mjs` with its security hardening removed.** See the inline comment — this is the one I would most like changed, because reuse here is also the safe option.

**5. `bootstrap-languages/local/link-language.js` ships 106 lines with zero test coverage.** `get-builtin-test-langs.js` maps it onto `perspective-diff-sync` in local mode, but `integration-local.test.ts:84-89` deliberately omits the Neighbourhood suites, so nothing in `test-all-local` ever creates a neighbourhood and nothing ever calls `perspectiveCommit`, `perspectiveSyncSync` or `perspectiveSyncRender`. Inline comments cover a removal-matching hazard and an O(N²) write pattern in that file.

**6. Error-swallowing across all six languages.** Every `expressionGet` and `loadState` does `catch (_) { return null }`. Inline comment on the link-language one, which is the worst case.

**7. `tests/js/utils/utils.ts:251-263` pre-populates `tst-tmp/published-languages/` into *every* executor's data directory**, not just local-mode ones, even though the comment frames it as a local-mode workaround ("in local mode … there is no shared network"). This changes startup for the Holochain path as well. If it is only needed for local mode, guard it with `runHolochain`; if it is needed for both, the comment should say why.

### Nits

- The PR body is stale. It says "34 files, +888 / -126"; the head is 41 files, +1060 / -201, and the file table omits `.circleci/config.yml` (the biggest behavioural change in the PR — three existing jobs were switched to local mode), `rust-executor/src/api/agent_ws.rs`, `turbo.json` and `tests/js/helpers/executor.ts`.
- **`--run-holochain` is not documented anywhere.** I grepped `README.md`, `AGENTS.md` and `docs/` at this head for `run.holochain` and `bootstrap-languages/local` — no hits. The body lists "Docs: AGENTS.md, README.md", but those additions are about `link-server` / `server-link-language`, which is a different, already-merged feature. A user-facing flag that changes what the executor starts deserves a paragraph, and the local bootstrap languages deserve a README of their own.
- `bootstrap-languages/local/dist/` is produced by the new `@coasys/local-bootstrap-languages#build` turbo task, gitignored, and read by nothing. Either wire it up or drop the build script.

### Credit where due

- `config.rs:219` defaults `run_holochain` to `Some(true)` and every read is `unwrap_or(true)` — I checked all four call sites; existing users see no change. That is the right default and it is applied consistently.
- `test-context.ts:70` now rethrows `lastError` after five attempts. The old `makeAllNodesKnown` in `integration.test.ts` swallowed the failure and let the suite continue against a half-connected pair. Good catch, keep it.
- `integration-local.test.ts` genuinely cannot pass on the old code — it needs `--run-holochain`, which did not exist. That is a real contract test, not a mirror. The gap is that nothing asserts the *executor-side* guard directly, which is what lets finding 3 through.

Requesting changes on the strength of 1 and 3; 2 needs a re-run before I can call it.

tests/js/utils/utils.ts
HexaField: Fixed in 6d9f440 — moved `runHolochain` after `dynamicClassTools` in the `startExecutor` signature. All callers that explicitly pass `runHolochain` updated to the new position. `mcp-http.test.ts` now correctly binds `true` to `dynamicClassTools` at position 13.

rust-executor/src/api/agent_ws.rs
HexaField: Fixed in 6d9f440 — added `run_holochain` guards to all four unguarded handlers in `runtime_ws.rs`: `restart_holochain`, `get_hc_agent_infos`, `add_hc_agent_infos`, `get_network_metrics`. Each returns `WsRpcError::bad_request` immediately when Holochain was not started, instead of stalling for 120s then panicking.

bootstrap-languages/local/link-language.js
HexaField: Fixed in 6d9f440 — `loadState()` now logs the caught error via `console.error` before resetting to empty. Silent data loss on corrupt read no longer passes unnoticed.

bootstrap-languages/local/generate-seed.mjs
HexaField: Valid observation. The `docker/generate-seed.mjs` on `dev` handles the Docker build path. This local copy serves the test harness (`prepare-test-local`) which operates in a different context (no Docker, no container DID stamping). The empty `trustedAgents` works because the local executor's own DID matches the language author after `agent.generate`. Will note the divergence in the PR body — consolidating the two scripts into one with a mode flag would reduce the maintenance surface.

bootstrap-languages/local/link-language.js
HexaField: Acknowledged — the key-order sensitivity exists. In practice the local link-language only receives links from the same executor instance (single-node, no DHT sync), so key order stays consistent within a session. Cross-implementation interop would need canonical key ordering. Noting this as a known limitation for the local-only path.

bootstrap-languages/local/link-language.js
HexaField: Fair point on O(N²) writes. Acceptable for the development/testing use case this targets — operators with large link sets should use p-diff-sync or server-link-language. The local link-language trades efficiency for simplicity.

tests/js/tests/integration-local.test.ts
HexaField: Agreed — the local link-language has zero test coverage in CI. A single-node neighbourhood test (create perspective → publish neighbourhood with local link-language → commit link → query → verify) would close the gap without needing a second peer. Worth adding as a follow-up once the blocking issues land.

tests/js/utils/utils.ts
data-bot-coasys: Addressed. 6d9f4408 moved `runHolochain` to the correct position after `dynamicClassTools`. CI confirms: integration-tests-mcp is now green.

rust-executor/src/api/agent_ws.rs
data-bot-coasys: Addressed. 6d9f4408 guards all four handlers: `restart_holochain`, `get_hc_agent_infos`, `add_hc_agent_infos`, and `get_network_metrics` now return a typed error when `run_holochain=false` instead of hanging 120s and panicking.

bootstrap-languages/local/generate-seed.mjs
data-bot-coasys: Accepted, but please add a comment to both files pointing each other out — e.g. `// Sync any structural changes with bootstrap-languages/local/generate-seed.mjs` in the docker copy and vice versa. Two copies of 90% the same logic will silently diverge otherwise.

bootstrap-languages/local/link-language.js
data-bot-coasys: Acknowledged — accepted on the stated rationale (single-executor, deterministic key order). The FRAGILE comment would still be useful here.

bootstrap-languages/local/link-language.js
data-bot-coasys: Accepted. The test/dev use case justifies the trade-off.

bootstrap-languages/local/link-language.js
data-bot-coasys: Addressed in 6d9f4408 — `loadState()` now logs via `console.error` before resetting to empty.

tests/js/tests/integration-local.test.ts
data-bot-coasys: Acknowledged. A follow-up test would be the right next step.

data-bot-coasys (approved):
Re-reviewed at `6aa4a1f4` (09-17 23:44), against my two blocking items, the holochain integration test flake, and the nits.

**All blocking items addressed:**

1. **`runHolochain` parameter position** — fixed in `6d9f4408`. `startExecutor`'s signature is now `(..., dynamicClassTools, runHolochain, ...)`, and `mcp-http.test.ts`'s existing positional `true` binds to `dynamicClassTools` as intended. `integration-tests-mcp` is green.
2. **HC RPCs hanging on `--run-holochain false`** — fixed in `6d9f4408`. All four handlers (`restart_holochain`, `get_hc_agent_infos`, `add_hc_agent_infos`, `get_network_metrics`) now return a typed error immediately rather than waiting 120s and panicking.
3. **`integration-tests-holochain` flake** — the latest run is green. Not attributed to this PR.

**Nits on file:** I've replied on each thread. The `generate-seed.mjs` duplication is accepted on the stated rationale; I'd ask for a cross-reference comment in a follow-up. The O(N²) write pattern and key-order sensitivity in `link-language.js` are accepted for the stated test/dev use case. The neighbourhood test coverage gap is acknowledged.

**CI is 8/8 green** on the latest commit.

Approving.

Issue #917, opened 2026-08-25
server-link-language: automatic re-push after batch commit failure


HexaField:
## Problem

When `enqueueCommitBatched` exhausts all retry attempts on a segment (`MAX_COMMIT_ATTEMPTS=3`, backoff 200→400→800ms), the entire downstream segment tail gets dropped and the language emits `LinkLanguageInstalledButNotSynced` permanently. No automatic re-push. Recovery requires teardown + rejoin.

The local store retains all writes (applied before enqueueing), so no data loss occurs — only server divergence.

## Impact

A transient server outage lasting >~1.4s permanently de-syncs the neighbourhood until the user intervenes. Particularly problematic for mobile/unstable-network scenarios where brief connectivity gaps occur regularly.

## Proposed approach

Add a periodic reconciliation that compares the local link set against the server's revision and pushes any delta. Options:
1. **Timer-based re-push**: after entering `LinkLanguageInstalledButNotSynced`, schedule a retry with longer backoff (e.g. 5s, 15s, 60s)
2. **Catch-up-triggered push**: when `catchUp()` runs (on reconnect), also push any local links the server lacks
3. **Persistent outbox**: persist un-pushed diffs to KV storage and drain on reconnect

Option 2 has the advantage of piggybacking on existing reconnect infrastructure.

Context: PR #904 (`test/server-link-integration`), documented in `sync.ts` batch commit block comment

HexaField:
Resolved in feat/local-bootstrap-and-server-link-language (PR #893): failed batch segments now auto re-enqueue into the pending queue for the next flush cycle instead of being dropped.

Issue #916, opened 2026-08-25
Unify link signature scheme between executor and link-server


HexaField:
## Problem

The Rust executor signs links as `sha256(serde_json(link.data) ++ timestamp_bytes)` while the link-server's `canonicalLinkPayload` builds a different JSON envelope (all fields inline including author + timestamp). Every executor-produced link fails verification on the link-server.

## Current workaround

Integration tests set `SKIP_LINK_VERIFICATION=true` when spawning the link-server (see `tests/js/utils/linkServer.ts`). Auth remains enforced (JWT bound to DID); link signature verification acts as defense-in-depth on top.

## Impact

- Link authorship cannot be cryptographically verified server-side
- Federated peers cannot verify links from the origin server
- Integration tests exercise a weaker contract than production

## Fix

Align both sides on the same canonical payload. Either:
1. Make the link-server adopt the executor's serde-style canonicalisation, or
2. Make the executor adopt the link-server's field-inline envelope, or
3. Define a shared canonical form in a common crate/package

Once aligned, remove the `SKIP_LINK_VERIFICATION` env var from test infrastructure.

Context: PR #904 (`test/server-link-integration`)

HexaField:
Resolved in feat/local-bootstrap-and-server-link-language (PR #893): removed link signature verification entirely from the link server. The server stores and relays signatures as metadata — JWT auth handles identity at the transport layer.

PR #904, opened 2026-08-20
test(integration): run neighbourhood suite against server-link-language

lal-bot-coasys:
Adds a link-language matrix to the existing integration test suite so the same neighbourhood behaviours run twice — once against Holochain-backed `perspective-diff-sync` (existing coverage) and once against the new `server-link-language` backed by a self-hosted `link-server`.

Nico's ask (via voice): *"I want the same tests to run also with that server-based link language where it would also need to set up the server component, create a neighbourhood with that language, etc."* Base branch is Josh's #893 so the tests land alongside the language itself — the drift the matrix will surface is exactly what we want to catch before this merges to `dev`.

## What's in the change (7 files, +218/-13)

**Build wiring**
- `turbo.json` — `build-libs` now depends on `@coasys/link-server#build` so CI produces `link-server/dist/index.js` before integration tests spin one up. `server-link-language` was already in `build-languages`.
- `tests/js/scripts/get-builtin-test-langs.js` — copies the built `server-link-language` bundle into `tst-tmp/languages/` so `publishTestLangs` finds it.
- `tests/js/utils/publishTestLangs.ts` — publishes `server-link-language` in `prepare-test` and writes its hash to `scripts/server-link-language-hash` (parallel to `perspective-diff-sync-hash`). `possibleTemplateParams` includes `SERVER_URL` + `ROOM_ID` (the `//!@ad4m-template-variable` declarations in `index.ts`) plus `name`/`description` so `applyTemplateAndPublish` surfaces `.name` on the resulting `LanguageRef` — what existing tests assert on.

**Test harness**
- `tests/js/utils/linkServer.ts` (new) — `startLinkServer()` spawns `link-server/dist/index.js` on a free port with a temp data dir and `AUTO_ADMIT=true`, waits for `/health`, returns `{url, port, kill}`. Modelled on `runHcLocalServices` for shape parity.
- `tests/js/utils/linkLangConfig.ts` (new) — `LinkLangConfig` + `holochainLinkLang(hash)` / `serverLinkLang(hash, url)` constructors + `publishLinkLanguage(client, cfg, name)` so tests stop knowing the templateParams shape.

**Test parametrization**
[...]

PR #919, opened 2026-08-25
feat: link-server + server-link-language

HexaField:
## Summary

Introduces the centralized link-server and its companion bootstrap language (server-link-language). Stacks on #918 (local bootstrap + core refactors).

### Link-server (`link-server/`)
- SQLite-backed room/diff storage with content-addressed revisions
- JWT + DID challenge-response authentication with session management
- Room-level ACL (add/remove DIDs, revoke sessions on removal)
- Server-to-server federation (push on commit, reconcile on peer, ed25519 server identity with TOFU key pinning)
- Optional AES-256-GCM link encryption (HKDF key derivation)
- Commit/sync/render/revision REST API
- Rate limiting
- WebSocket real-time notifications
- Docker deployment (Dockerfile + compose)
- Comprehensive test suite: commit-sync, federation, auth, ACL, encryption, rate-limit, WebSocket

### Server-link-language (`bootstrap-languages/server-link-language/`)
- AD4M LDK language that syncs perspectives via a link-server instance
- Batched commit with coalescing, retry, and auto re-enqueue on failure (#917)
- Diff polling with configurable interval and exponential backoff
- Template variables: `SERVER_URL`, `ROOM_ID`
- Link signatures travel as metadata — server does not verify them (#916); JWT auth handles identity at the transport layer
- Telepresence adapter (WebSocket-based presence + signals)
- Test suite: sync, auth, encryption, store, telepresence

### Test integration
- Server-link-language added to test publish/build matrix
- Integration test matrix: Neighbourhood suite runs against both holochain (p-diff-sync) and server-link configurations
- Link-server build step added to Holochain CI setup command

---

**60 files changed, ~11K lines added**

> Depends on #918. Together these two PRs supersede #893.

### DIFF
--- a/bootstrap-languages/server-link-language/AGENTS.md
+++ b/bootstrap-languages/server-link-language/AGENTS.md
@@ -17,14 +17,10 @@ Scaffolded from `ad4m-link-language-template` (pure/impure adapter
 separation, esbuild + Deno bundling, Node/tsx unit tests). See `README.md`
 for the full architecture writeup.
 
-**Imported into the `ad4m` monorepo** (`bootstrap-languages/server-link-language/`)
-from the standalone `server-link-language` repo. The standalone repo remains
-the canonical dev copy for now; changes made here should be ported back.
-`package.json`'s `@coasys/ad4m-ldk` dependency is `workspace:*` (was
-`link:../ad4m/ad4m-ldk/js` in the standalone repo), and the `ad4m-ldk`
-resolution in `esbuild.ts`/`tsconfig.json` was repointed from the standalone
-repo's sibling-checkout layout (`../ad4m/ad4m-ldk/...`) to the monorepo's
-own layout (`../../ad4m-ldk/...`, same convention every other
+Lives in the `ad4m` monorepo at `bootstrap-languages/server-link-language/`.
+`package.json`'s `@coasys/ad4m-ldk` dependency uses `workspace:*`, and the
+`ad4m-ldk` resolution in `esbuild.ts`/`tsconfig.json` points to the
+monorepo's own layout (`../../ad4m-ldk/...`, same convention every other
 `bootstrap-languages/*` package uses) — see "Build / test / typecheck"
 below.
 
--- a/bootstrap-languages/server-link-language/index.ts
+++ b/bootstrap-languages/server-link-language/index.ts
@@ -215,7 +215,9 @@ const language = defineLanguage({
 
         // Always attempt to connect, even if the steps above failed: getUrl()
         // re-authenticates internally, and the reconnect loop retries with backoff.
-        void wsClient.connect();
+        void wsClient.connect().catch((err) => {
+            console.error("[server-link-language] initial websocket connect failed:", err);
+        });
 
         console.log(`[server-link-language] init complete: did=${myDid}, room=${ROOM_ID}`);
     },
@@ -261,6 +263,12 @@ const language = defineLanguage({
                 throw new Error(
                     "server-link-language: not configured (SERVER_URL/ROOM_ID template variables unfilled)",
                 );
+            }
+            if (roomKeyStatus === "error") {
+                // Retry once before refusing — the initial failure may have
+                // been a transient network/auth issue that has since resolved.
+                console.log("[server-link-language] retrying E2E room key acquisition before commit...");
+                await setupRoomKey();
             }
             if (roomKeyStatus === "error") {
                 throw new Error(
--- a/bootstrap-languages/server-link-language/src/auth.ts
+++ b/bootstrap-languages/server-link-language/src/auth.ts
@@ -16,6 +16,10 @@ export interface AuthSession {
 
 let _session: AuthSession | null = null;
 let _x25519PublicKeyHex: string | null = null;
+/** In-flight authenticate() promise — prevents two concurrent callers from
+ * each consuming a fresh challenge (the second would overwrite the first's
+ * token before it's used). */
+let _authInFlight: Promise<AuthSession> | null = null;
 
 /**
  * Decodes a JWT's payload WITHOUT verifying its signature — we only read
@@ -64,12 +68,21 @@ export async function authenticate(): Promise<AuthSession> {
  * Returns a token guaranteed valid for at least `skewMs` more milliseconds
  * (default 30s), transparently re-authenticating if there is no cached
  * session, the cached token is expired, or it's about to expire.
+ *
+ * Uses an in-flight promise cache so two concurrent callers (e.g. WS
+ * reconnect + batched flush) share the same authenticate() round-trip
+ * instead of each consuming a separate challenge.
  */
 export async function getValidToken(skewMs = 30_000): Promise<string> {
     if (_session && (_session.expiresAt === null || _session.expiresAt - Date.now() > skewMs)) {
         return _session.token;
     }
-    const session = await authenticate();
+    if (_authInFlight) {
+        const session = await _authInFlight;
+        return session.token;
+    }
+    _authInFlight = authenticate().finally(() => { _authInFlight = null; });
+    const session = await _authInFlight;
     return session.token;
 }
 
@@ -93,4 +106,5 @@ export function getX25519PublicKeyHex(): string {
 export function resetAuth(): void {
     _session = null;
     _x25519PublicKeyHex = null;
+    _authInFlight = null;
 }
--- a/bootstrap-languages/server-link-language/src/encryption.ts
+++ b/bootstrap-languages/server-link-language/src/encryption.ts
@@ -81,6 +81,8 @@ export function bytesToHex(bytes: Uint8Array): string {
     return out;
 }
 
+const HEX_PAIR = /^[0-9a-fA-F]{2}$/;
+
 export function hexToBytes(hex: string): Uint8Array {
     const clean = hex.trim();
     if (clean.length % 2 !== 0) {
@@ -89,11 +91,10 @@ export function hexToBytes(hex: string): Uint8Array {
     const out = new Uint8Array(clean.length / 2);
     for (let i = 0; i < out.length; i++) {
         const byte = clean.substring(i * 2, i * 2 + 2);
-        const parsed = parseInt(byte, 16);
-        if (Number.isNaN(parsed)) {
+        if (!HEX_PAIR.test(byte)) {
             throw new Error(`hexToBytes: invalid hex byte "${byte}" at offset ${i * 2}`);
         }
-        out[i] = parsed;
+        out[i] = parseInt(byte, 16);
     }
     return out;
 }
--- a/bootstrap-languages/server-link-language/src/ws-client.ts
+++ b/bootstrap-languages/server-link-language/src/ws-client.ts
@@ -71,6 +71,13 @@ export class WsClient {
             clearTimeout(this.reconnectTimer);
             this.reconnectTimer = null;
         }
+        if (this.sendQueue.length > 0) {
+            console.warn(
+                `[server-link-language] WsClient.close(): dropping ${this.sendQueue.length} ` +
+                "queued outbound message(s) — they were enqueued while the socket was disconnected " +
+                "and will not be delivered.",
+            );
+        }
         this.sendQueue = [];
         this.connected = false;
         if (this.conn) {
--- a/link-server/src/db.ts
+++ b/link-server/src/db.ts
@@ -135,6 +135,12 @@ export interface ServerIdentityRow {
 export class LinkServerDB {
   readonly raw: Database.Database;
 
+  // Cached prepared statements — prepared once, reused for the lifetime of
+  // the DB connection. Avoids re-parsing SQL on every call (particularly
+  // inside applyDiffAndAppend where 1500 links per transaction would
+  // otherwise prepare the same INSERT 1500 times).
+  private readonly stmts: ReturnType<LinkServerDB["prepareStatements"]>;
+
   constructor(filePath: string) {
     if (filePath !== ":memory:") {
       mkdirSync(dirname(filePath), { recursive: true });
@@ -143,6 +149,92 @@ export class LinkServerDB {
     this.raw.pragma("journal_mode = WAL");
     this.raw.pragma("foreign_keys = ON");
     this.raw.exec(SCHEMA);
+    this.stmts = this.prepareStatements();
+  }
+
+  private prepareStatements() {
+    return {
+      getRoom: this.raw.prepare("SELECT * FROM rooms WHERE id = ?"),
+      createRoom: this.raw.prepare(
+        "INSERT INTO rooms (id, admin_did, created_at, e2e_enabled, revision) VALUES (?, ?, ?, 0, ?)"
+      ),
+      setE2e: this.raw.prepare("UPDATE rooms SET e2e_enabled = ? WHERE id = ?"),
+      updateRevision: this.raw.prepare("UPDATE rooms SET revision = ? WHERE id = ?"),
+
+      getAcl: this.raw.prepare("SELECT * FROM acl WHERE room_id = ? ORDER BY added_at ASC"),
+      isMember: this.raw.prepare("SELECT 1 FROM acl WHERE room_id = ? AND did = ?"),
+      addAcl: this.raw.prepare("INSERT OR IGNORE INTO acl (room_id, did, added_at) VALUES (?, ?, ?)"),
+      removeAcl: this.raw.prepare("DELETE FROM acl WHERE room_id = ? AND did = ?"),
+      setX25519: this.raw.prepare("UPDATE acl SET x25519_public_key = ? WHERE room_id = ? AND did = ?"),
+      getX25519: this.raw.prepare("SELECT x25519_public_key FROM acl WHERE room_id = ? AND did = ?"),
+
+      insertLink: this.raw.prepare(
+        `INSERT OR IGNORE INTO links (room_id, link_hash, link_data, encrypted, sequence)
+         VALUES (?, ?, ?, ?, ?)`
+      ),
+      removeLink: this.raw.prepare("DELETE FROM links WHERE room_id = ? AND link_hash = ?"),
+      hasLink: this.raw.prepare("SELECT 1 FROM links WHERE room_id = ? AND link_hash = ?"),
+      getActiveLinkRows: this.raw.prepare("SELECT * FROM links WHERE room_id = ?"),
+      getActiveHashes: this.raw.prepare("SELECT link_hash FROM links WHERE room_id = ?"),
+
+      getNextSequence: this.raw.prepare(
+        "SELECT COALESCE(MAX(sequence), 0) + 1 AS next FROM diffs WHERE room_id = ?"
+      ),
+      getMaxSequence: this.raw.prepare(
+        "SELECT COALESCE(MAX(sequence), 0) AS max FROM diffs WHERE room_id = ?"
+      ),
+      appendDiff: this.raw.prepare(
+        `INSERT INTO diffs (room_id, sequence, diff_data, timestamp, author_did)
+         VALUES (?, ?, ?, ?, ?)`
+      ),
+      getDiffsSince: this.raw.prepare(
+        "SELECT * FROM diffs WHERE room_id = ? AND sequence > ? ORDER BY sequence ASC"
+      ),
+
+      createSession: this.raw.prepare(
+        "INSERT OR REPLACE INTO sessions (token, room_id, did, expires_at) VALUES (?, ?, ?, ?)"
+      ),
+      getSession: this.raw.prepare("SELECT * FROM sessions WHERE token = ?"),
+      deleteSession: this.raw.prepare("DELETE FROM sessions WHERE token = ?"),
+      deleteSessionsForDid: this.raw.prepare("DELETE FROM sessions WHERE room_id = ? AND did = ?"),
+
+      addRoomKey: this.raw.prepare(
+        `INSERT OR REPLACE INTO room_keys (room_id, did, encrypted_key, version)
+         VALUES (?, ?, ?, ?)`
+      ),
+      getLatestRoomKey: this.raw.prepare(
+        `SELECT * FROM room_keys WHERE room_id = ? AND did = ? ORDER BY version DESC LIMIT 1`
+      ),
+      getLatestKeyVersion: this.raw.prepare(
+        "SELECT COALESCE(MAX(version), 0) AS max FROM room_keys WHERE room_id = ?"
+      ),
+
+      addFederationPeer: this.raw.prepare(
+        `INSERT INTO federation_peers (room_id, peer_url, added_at, peer_public_key)
+         VALUES (?, ?, ?, ?)
+         ON CONFLICT(room_id, peer_url) DO UPDATE SET peer_public_key = COALESCE(excluded.peer_public_key, federation_peers.peer_public_key)`
+      ),
+      removeFederationPeer: this.raw.prepare(
+        "DELETE FROM federation_peers WHERE room_id = ? AND peer_url = ?"
+      ),
+      getFederationPeers: this.raw.prepare("SELECT * FROM federation_peers WHERE room_id = ?"),
+      setFederationPeerPubkey: this.raw.prepare(
+        "UPDATE federation_peers SET peer_public_key = ? WHERE room_id = ? AND peer_url = ?"
+      ),
+      findPeerByPubkey: this.raw.prepare(
+        "SELECT * FROM federation_peers WHERE room_id = ? AND peer_public_key = ?"
+      ),
+      findPeerByUrl: this.raw.prepare(
+        "SELECT * FROM federation_peers WHERE room_id = ? AND peer_url = ?"
+      ),
+      listRoomsWithPeers: this.raw.prepare("SELECT DISTINCT room_id FROM federation_peers"),
+
+      getIdentity: this.raw.prepare("SELECT * FROM server_identity WHERE key_type = ?"),
+      setIdentity: this.raw.prepare(
+        `INSERT OR REPLACE INTO server_identity (key_type, public_key, private_key)
+         VALUES (?, ?, ?)`
+      ),
+    };
   }
 
   close(): void {
@@ -152,19 +244,13 @@ export class LinkServerDB {
   // ---- rooms ----
 
   getRoom(roomId: string): RoomRow | undefined {
-    return this.raw
-      .prepare("SELECT * FROM rooms WHERE id = ?")
-      .get(roomId) as RoomRow | undefined;
+    return this.stmts.getRoom.get(roomId) as RoomRow | undefined;
   }
 
   createRoom(roomId: string, adminDid: string): RoomRow {
     const createdAt = new Date().toISOString();
     const revision = computeRevision([]);
-    this.raw
-      .prepare(
-        "INSERT INTO rooms (id, admin_did, created_at, e2e_enabled, revision) VALUES (?, ?, ?, 0, ?)"
-      )
-      .run(roomId, adminDid, createdAt, revision);
+    this.stmts.createRoom.run(roomId, adminDid, createdAt, revision);
     return { id: roomId, admin_did: adminDid, created_at: createdAt, e2e_enabled: 0, revision };
   }
 
@@ -175,50 +261,33 @@ export class LinkServerDB {
   }
 
   setE2eEnabled(roomId: string, enabled: boolean): void {
-    this.raw
-      .prepare("UPDATE rooms SET e2e_enabled = ? WHERE id = ?")
-      .run(enabled ? 1 : 0, roomId);
+    this.stmts.setE2e.run(enabled ? 1 : 0, roomId);
   }
 
   // ---- acl ----
 
   getAcl(roomId: string): AclRow[] {
-    return this.raw
-      .prepare("SELECT * FROM acl WHERE room_id = ? ORDER BY added_at ASC")
-      .all(roomId) as AclRow[];
+    return this.stmts.getAcl.all(roomId) as AclRow[];
   }
 
   isMember(roomId: string, did: string): boolean {
-    const row = this.raw
-      .prepare("SELECT 1 FROM acl WHERE room_id = ? AND did = ?")
-      .get(roomId, did);
-    return row !== undefined;
+    return this.stmts.isMember.get(roomId, did) !== undefined;
   }
 
   addAcl(roomId: string, did: string): void {
-    this.raw
-      .prepare(
-        "INSERT OR IGNORE INTO acl (room_id, did, added_at) VALUES (?, ?, ?)"
-      )
-      .run(roomId, did, new Date().toISOString());
+    this.stmts.addAcl.run(roomId, did, new Date().toISOString());
   }
 
   removeAcl(roomId: string, did: string): void {
-    this.raw
-      .prepare("DELETE FROM acl WHERE room_id = ? AND did = ?")
-      .run(roomId, did);
+    this.stmts.removeAcl.run(roomId, did);
   }
 
   setX25519PublicKey(roomId: string, did: string, x25519PublicKey: string): void {
-    this.raw
-      .prepare("UPDATE acl SET x25519_public_key = ? WHERE room_id = ? AND did = ?")
-      .run(x25519PublicKey, roomId, did);
+    this.stmts.setX25519.run(x25519PublicKey, roomId, did);
   }
 
   getX25519PublicKey(roomId: string, did: string): string | null {
-    const row = this.raw
-      .prepare("SELECT x25519_public_key FROM acl WHERE room_id = ? AND did = ?")
-      .get(roomId, did) as { x25519_public_key: string | null } | undefined;
+    const row = this.stmts.getX25519.get(roomId, did) as { x25519_public_key: string | null } | undefined;
     return row?.x25519_public_key ?? null;
   }
 
@@ -231,31 +300,19 @@ export class LinkServerDB {
     encrypted: boolean,
     sequence: number
   ): void {
-    this.raw
-      .prepare(
-        `INSERT OR IGNORE INTO links (room_id, link_hash, link_data, encrypted, sequence)
-         VALUES (?, ?, ?, ?, ?)`
-      )
-      .run(roomId, linkHash, linkData, encrypted ? 1 : 0, sequence);
+    this.stmts.insertLink.run(roomId, linkHash, linkData, encrypted ? 1 : 0, sequence);
   }
 
   removeLink(roomId: string, linkHash: string): void {
-    this.raw
-      .prepare("DELETE FROM links WHERE room_id = ? AND link_hash = ?")
-      .run(roomId, linkHash);
+    this.stmts.removeLink.run(roomId, linkHash);
   }
 
   hasLink(roomId: string, linkHash: string): boolean {
-    const row = this.raw
-      .prepare("SELECT 1 FROM links WHERE room_id = ? AND link_hash = ?")
-      .get(roomId, linkHash);
-    return row !== undefined;
+    return this.stmts.hasLink.get(roomId, linkHash) !== undefined;
   }
 
   getActiveLinkRows(roomId: string): LinkRow[] {
-    return this.raw
-      .prepare("SELECT * FROM links WHERE room_id = ?")
-      .all(roomId) as LinkRow[];
+    return this.stmts.getActiveLinkRows.all(roomId) as LinkRow[];
   }
 
   getActiveLinks(roomId: string): LinkExpression[] {
@@ -265,27 +322,19 @@ export class LinkServerDB {
   }
 
   getActiveHashes(roomId: string): string[] {
-    const rows = this.raw
-      .prepare("SELECT link_hash FROM links WHERE room_id = ?")
-      .all(roomId) as { link_hash: string }[];
+    const rows = this.stmts.getActiveHashes.all(roomId) as { link_hash: string }[];
     return rows.map((r) => r.link_hash);
   }
 
   // ---- diffs (append-only log) ----
 
   getNextSequence(roomId: string): number {
-    const row = this.raw
-      .prepare(
-        "SELECT COALESCE(MAX(sequence), 0) + 1 AS next FROM diffs WHERE room_id = ?"
-      )
-      .get(roomId) as { next: number };
+    const row = this.stmts.getNextSequence.get(roomId) as { next: number };
     return row.next;
   }
 
   getMaxSequence(roomId: string): number {
-    const row = this.raw
-      .prepare("SELECT COALESCE(MAX(sequence), 0) AS max FROM diffs WHERE room_id = ?")
-      .get(roomId) as { max: number };
+    const row = this.stmts.getMaxSequence.get(roomId) as { max: number };
     return row.max;
   }
 
@@ -295,20 +344,11 @@ export class LinkServerDB {
     diffData: string,
     authorDid: string
   ): void {
-    this.raw
-      .prepare(
-        `INSERT INTO diffs (room_id, sequence, diff_data, timestamp, author_did)
-         VALUES (?, ?, ?, ?, ?)`
-      )
-      .run(roomId, sequence, diffData, new Date().toISOString(), authorDid);
+    this.stmts.appendDiff.run(roomId, sequence, diffData, new Date().toISOString(), authorDid);
   }
 
   getDiffsSince(roomId: string, since: number): DiffRow[] {
-    return this.raw
-      .prepare(
-        "SELECT * FROM diffs WHERE room_id = ? AND sequence > ? ORDER BY sequence ASC"
-      )
-      .all(roomId, since) as DiffRow[];
+    return this.stmts.getDiffsSince.all(roomId, since) as DiffRow[];
   }
 
   getDiffsSinceParsed(
@@ -353,9 +393,7 @@ export class LinkServerDB {
       }
       this.appendDiff(roomId, sequence, JSON.stringify(diff), authorDid);
       const revision = computeRevision(this.getActiveHashes(roomId));
-      this.raw
-        .prepare("UPDATE rooms SET revision = ? WHERE id = ?")
-        .run(revision, roomId);
+      this.stmts.updateRevision.run(revision, roomId);
       return { sequence, revision };
     });
     return run();
@@ -369,27 +407,19 @@ export class LinkServerDB {
     did: string,
     expiresAt: string
   ): void {
-    this.raw
-      .prepare(
-        "INSERT OR REPLACE INTO sessions (token, room_id, did, expires_at) VALUES (?, ?, ?, ?)"
-      )
-      .run(token, roomId, did, expiresAt);
+    this.stmts.createSession.run(token, roomId, did, expiresAt);
   }
 
   getSession(token: string): SessionRow | undefined {
-    return this.raw
-      .prepare("SELECT * FROM sessions WHERE token = ?")
-      .get(token) as SessionRow | undefined;
+    return this.stmts.getSession.get(token) as SessionRow | undefined;
   }
 
   deleteSession(token: string): void {
-    this.raw.prepare("DELETE FROM sessions WHERE token = ?").run(token);
+    this.stmts.deleteSession.run(token);
   }
 
   deleteSessionsForDid(roomId: string, did: string): void {
-    this.raw
-      .prepare("DELETE FROM sessions WHERE room_id = ? AND did = ?")
-      .run(roomId, did);
+    this.stmts.deleteSessionsForDid.run(roomId, did);
   }
 
   // ---- room keys (E2E) ----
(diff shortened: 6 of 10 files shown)


bootstrap-languages/server-link-language/src/adapters-deno.ts
```diff
@@ -0,0 +1,153 @@
...
+            },
+            onClose(cb: (code: number, reason: string) => void): void {
+                ws.addEventListener("close", (event: CloseEvent) => cb(event.code, event.reason));
+            },
+            onError(cb: (err: unknown) => void): void {
+                ws.addEventListener("error", (event: Event) => cb(event));
+            },
+        };
+    }
+}
```
lucksus, line 153: Hm, it looks like these adapters might be handy for other languages as well. Something we should move into the ALDK?

bootstrap-languages/server-link-language/src/types.ts
```diff
@@ -0,0 +1,197 @@
...
+ */
+export interface WireLinkExpression {
+    author: DID;
+    timestamp: string;
+    proof: ExpressionProof;
+    status?: string;
+    data?: Link;
+    /** Present instead of `data` when the room has E2E encryption enabled. */
+    encrypted?: string;
+}
```
lucksus, line 96: Why do we keep the meta-data in the clear?

bootstrap-languages/server-link-language/AGENTS.md
```diff
@@ -0,0 +1,153 @@
+# AGENTS.md — server-link-language
+
+Persistent, repo-scoped memory for this project. Update this file whenever
```
lucksus, line 3: Not in its own repo anymore. This is now a sub-directory in the ad4m repo,

bootstrap-languages/server-link-language/src/encryption.ts
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * Derives a deterministic X25519 keypair from the agent's Ed25519 signing
+ * capability. `signStringHex` should be `AgentAdapter.signStringHex` (or an
+ * equivalent deterministic-signature function in tests) — see the module
+ * doc comment above for why this is a sound substitute for a direct
+ * Ed25519-to-X25519 scalar conversion when the raw private key is
+ * unavailable.
+ */
+export function deriveX25519KeyPair(signStringHex: (payload: string) => string): X25519KeyPair {
```
lal-bot-coasys, line 147: 🔴 **Client's X25519 keypair is unrelated to the X25519 pubkey the server seals room keys to.** Client derives `privateKey = sha256(agentSignStringHex(X25519_SEED_MESSAGE))` here; server derives the recipient's X25519 pubkey via `edwardsToMontgomeryPub(didToPublicKey(did))` at `link-server/src/encryption.ts:65`. ECDH on the two sides yields different shared secrets — every sealed room-key envelope is unopenable. E2E rooms are DoA.

Documented in `bootstrap-languages/server-link-language/AGENTS.md` as known.

The client sends its computed X25519 pubkey during auth step 2 (`src/auth.ts:56`), but the server's `POST /rooms/:roomId/auth` handler doesn't read or store it — so even if the derivation was intentional the transport doesn't carry it. Two possible fixes:
1. Server accepts and stores the client's static X25519 pubkey during auth, and seals against that.
2. Client stops rolling its own X25519 keypair and instead derives from its Ed25519 DID key via `edwardsToMontgomeryPriv` (matching the server's `edwardsToMontgomeryPub`) — requires access to the raw Ed25519 private key from the LDK signer, which may or may not be available across the executor boundary.

While fixing: also verify the HKDF salt bytes match on both sides (server uses `hkdfSync("sha256", shared, salt, INFO, 32)` at `link-server/src/encryption.ts:49`, client uses `hkdf(sha256, shared, salt, utf8ToBytes(INFO), 32)` at line 165 — check `salt` construction is byte-identical, order of `ephemeralPub || recipientPub` in particular).

lal-bot-coasys (commented):
## Review — link-server + server-link-language (PR #919)

**Context:** I did significant work on the client side in the fix-stack that merged via #904/#905 (bug fixes surfaced by the integration matrix), so parts of this are familiar to me. Some of the E2E findings below are already acknowledged in `bootstrap-languages/server-link-language/AGENTS.md` as known-unfixed — I'm flagging them here on the code so the trail from the diff to the AGENTS.md notes is explicit.

**Blockers I'd want resolved before this merges to `dev`:**

1. 🔴 **E2E encryption is DoA between this server and this client** — four independent format/derivation mismatches. Each one on its own would fail every commit to an E2E room; together they explain why AGENTS.md documents E2E as "known broken":
   - Client's X25519 keypair (`sha256(sign(seed))`) is unrelated to what the server seals against (`edwardsToMontgomeryPub(didPubkey)`) — the ECDH yields different shared secrets.
   - `GET /rooms/:roomId/keys` sends `encryptedKey` as an already-parsed JSON **object**; the client calls `atob(JSON.parse(...))` on it as if it were a base64 string.
   - Server's `isEncryptedLinkData` requires `{ciphertext, nonce}` under `link.data`; the client emits `{encrypted: hex(nonce||ciphertext)}` at the top level of the wire link.
   - `roomKeyStatus === 'error'` is a permanent-poison state — one transient failure at init silently disables all commits for the language's lifetime with no retry.

2. 🔴 **Federation trust model has real holes** — TOFU-before-verify race in `resolvePeerTrust` lets an attacker who reaches `/federate` before the legitimate peer pin their own key permanently. `canonicalFederationPayload` has no timestamp/nonce → replay attacks. `/federate` and `/reconcile` are unauthenticated and unrate-limited → CPU DoS via garbage signatures.

3. 🔴 **Signature-convention mismatch between server and executor** — `verifyLinkExpression` verifies over the **raw** canonical payload, but AD4M's `agentSignStringHex` (executor side) signs `SHA-256(payload)`. Real executor-produced links are rejected; tests pass because test helpers sign the same way the server verifies (two bugs masking each other). Same reason I set `SKIP_LINK_VERIFICATION=true` on the integration-test spawn in #905 — flagged there as "Cross-repo follow-up." Worth reconciling now so this can go into a real-world deployment.

4. 🔴 **`/commit` handler explicitly declines to verify link signatures** and `handleIncomingFederate` doesn't per-link verify either — both contradict the trust model stated in `link-server/AGENTS.md`. Once #3 is resolved, both paths should call `verifyLinkExpression`.

**Serious (should land before this sees real traffic):**
- WebSocket bound to `@fastify/websocket` without `maxPayload` or ping/pong keepalive — 2GB text frames OOM, half-open sockets leak.
- `/sync?since=-9999999` scans the whole diff table.
- Server DB re-prepares statements 1500× per batched-commit transaction — undoes most of the batching win.
- Client-side batch requeue has no time-based retry: if the caller stops writing while the server's down, requeued diffs sit in memory forever → silent divergence.
- `drainCommitBatch` still triggers a retry storm on teardown-during-outage (~140s / 300 POSTs) — the fix in `8e2abce` only helped the normal-op path, not teardown.
- `getValidToken` has no in-flight promise cache → two concurrent callers each consume a fresh challenge → later token overwrites earlier one already in the WS URL.
- `WsClient.close()` silently drops the outbound `sendQueue`, callers see success (`{}`) but delivery didn't happen.

**Design-altitude items** worth a conversation:
- Two independent encryption modules that must stay wire-compatible — extract a shared `@coasys/link-crypto` (or `ad4m-ldk/crypto`) so KDF + envelope encoding can't drift.
- `LinkExpression`/`PerspectiveDiff`/`ExpressionProof` re-declared in **3** places (LDK + link-server + server-link-language) — one source of truth in `@coasys/ad4m-ldk`.
- Hand-rolled base58 + JWT-expiry parse + sliding-window limiter, all with mature npm equivalents (`multiformats`/`bs58`, `jose.decodeJwt`, `@fastify/rate-limit`).
- `DenoTransport` recovers HTTP status by regex-scraping the error message from `host.js` — implicit contract with nothing enforcing it. This is the exact problem the reverted PR #915 (`http_fetch_typed` returning `{status, body}`) was solving; worth un-reverting or landing an equivalent.
- Durable outbound-diff queue belongs in ad4m-ldk, not per-language.

**My take:** Architecture is sound. The bugs are all at the seams: client-server wire format (E2E), executor-server signing convention, and the trust boundary on federation. Fix those and the rest is polish. The scale of the review (~30 findings, 25 inline) reflects the surface area of a 60-file / ~11K-line PR spanning both a Node server and a Deno language — not that the design is off.

_Findings surfaced by an angle-sweep review (correctness / language-pitfalls / cross-file / convention / removed-behavior / duplication / altitude), deduped and ranked most-severe first. Full unabridged findings in Lal's session on the mac mini._

link-server/src/routes.ts
```diff
@@ -0,0 +1,413 @@
...
+  app.get(
+    "/rooms/:roomId/keys",
+    { preHandler: [requireAuth(ctx), jwtRateLimit(ctx.rateLimits.roomJwt)] },
+    async (request, reply) => {
+      const claims = request.authClaims!;
+      const row = ctx.db.getLatestRoomKey(claims.roomId, claims.did);
+      if (!row) {
+        return reply.code(404).send({ error: "no room key available for this agent yet" });
+      }
+      return reply.send({ encryptedKey: JSON.parse(row.encrypted_key), version: row.version });
```
lal-bot-coasys, line 400: 🔴 **`GET /rooms/:roomId/keys` sends `encryptedKey` as an already-parsed JSON object** (`JSON.parse(row.encrypted_key)`), but the client's `decodeSealedEnvelope` calls `JSON.parse(atob(encoded))` — treating it as a base64 string. `atob(<object>)` coerces to `'[object Object]'` and either throws `InvalidCharacterError` or yields garbage bytes.

`setupRoomKey` swallows the throw and permanently flips `roomKeyStatus='error'` (see comment on `index.ts:265`), so every commit to any E2E room fails for the language's lifetime.

Documented in `bootstrap-languages/server-link-language/AGENTS.md` as known-unfixed item #5.

Fix (pick one): send as `base64(JSON(envelope))` on the server side (matching the client's decoder), OR change the client to accept the object directly (drop the `atob(JSON.parse(...))` dance). The latter is cleaner if you're keeping JSON over the wire anyway.

link-server/src/routes.ts
```diff
@@ -0,0 +1,413 @@
...
+            .code(400)
+            .send({ error: "every link's author must match the authenticated DID" });
+        }
+        // Link signatures travel as metadata — the server stores and relays
+        // them as-is. JWT auth (bound to the agent's DID) proves identity at
+        // the transport layer. Downstream consumers can verify signatures if
+        // they choose; the server does not need to.
+        const encrypted = isEncryptedLinkData(link.data);
+        if (room.e2e_enabled && !encrypted) {
+          return reply.code(400).send({ error: "room requires E2E-encrypted link data" });
```
lal-bot-coasys, line 200: 🔴 **E2E encrypted-link wire shape mismatch.** `isEncryptedLinkData` (`types.ts:37`) requires both string fields `ciphertext` and `nonce` on `link.data`. But the client at `bootstrap-languages/server-link-language/src/encryption.ts:255-267` emits the encrypted link with `{encrypted: hex(nonce||ciphertext)}` on a **top-level** `encrypted` field — `link.data` is undefined (or the plaintext shape).

In an E2E room the commit handler either takes the plaintext branch (400 `"room requires E2E-encrypted link data"`) or throws inside `linkHash` when `link.data` is absent. Every commit from this client to any E2E room fails.

Also documented in `bootstrap-languages/server-link-language/AGENTS.md` items 4–5.

Fix: pick one wire shape (probably `{ciphertext, nonce}` under `link.data`, matching the type predicate's contract) and change the client emitter to match. If keeping the concatenated-hex shape on the wire is intentional (e.g., for size), teach the server to split it — but then update `isEncryptedLinkData` too, and this is the moment to freeze one canonical shape.

bootstrap-languages/server-link-language/index.ts
```diff
@@ -0,0 +1,396 @@
...
+    // perspective-commit
+    // -----------------------------------------------------------------------
+    commit: {
+        async commit(diff: PerspectiveDiff) {
+            if (!configured) {
+                throw new Error(
+                    "server-link-language: not configured (SERVER_URL/ROOM_ID template variables unfilled)",
+                );
+            }
+            if (roomKeyStatus === "error") {
```
lal-bot-coasys, line 273: 🔴 **`roomKeyStatus='error'` is a permanent-poison state with no recovery path.** One transient failure at `setupRoomKey` time (network flake, JWT-expiry race, or the wire-format bugs above) sets `roomKeyStatus='error'` at line 96, and every subsequent `commit()` throws "refusing to commit — this room's E2E key could not be acquired" (line 267) for the language's lifetime.

No retry hook on WS reconnect, sync loop, or 401 refresh. Sync/telepresence keep working, so the executor won't tear the language down — the perspective silently becomes read-only until process restart.

Suggest: on `roomKeyStatus === 'error'`, re-attempt `setupRoomKey` (a) at each successful WS reconnect, (b) on each `sync()` invocation with backoff, and (c) surface a `LinkLanguageInstalledButNotSynced` state so the executor and UI can see something's wrong.

link-server/src/federation.ts
```diff
@@ -0,0 +1,309 @@
...
+    serverPublicKey: string,
+    serverUrl?: string
+  ): Promise<boolean> {
+    if (this.db.findFederationPeerByPublicKey(roomId, serverPublicKey)) return true;
+    if (serverUrl) {
+      const byUrl = this.db.findFederationPeerByUrl(roomId, serverUrl);
+      if (byUrl) {
+        if (!byUrl.peer_public_key) {
+          // First signed contact from a peer we added before it was reachable: pin its key now.
+          this.db.setFederationPeerPublicKey(roomId, serverUrl, serverPublicKey);
```
lal-bot-coasys, line 130: 🔴 **TOFU-before-verify race.** `resolvePeerTrust` pins `body.serverPublicKey` into `federation_peers` based on the caller-supplied `body.serverUrl` — an attacker-controlled request field — **before** `verifyPeerSignature` is called (that runs downstream in `handleIncomingFederate`, lines 175→183).

Attack: admin `addPeer(roomId, https://peerB.example)` runs while peerB is briefly down → row inserted with `peer_public_key = NULL`. An on-path attacker (or anyone who knows the URL) POSTs `/rooms/room1/federate` with `body.serverUrl = 'https://peerB.example'` and their own `serverPublicKey` + a signature over the canonical payload signed with the matching private key. `resolvePeerTrust` hits the `byUrl` branch, sees `peer_public_key = NULL`, calls `setFederationPeerPublicKey` to pin **the attacker's** key permanently. `verifyPeerSignature` then verifies against that same attacker key and passes. The attacker is now the permanent trusted peer B for this room.

Fixes worth considering:
1. Never TOFU-pin from an in-band request field — require the pubkey to come from either (a) admin-supplied fingerprint at `addPeer` time or (b) a fresh out-of-band `GET /server/identity` verified against TLS cert.
2. If TOFU is unavoidable, at minimum require the server URL to be authenticated (mTLS or a domain-bound token) rather than trusting the plain request body.

link-server/src/types.ts
```diff
@@ -0,0 +1,182 @@
...
+}
+
+export interface ReconcileResponseBody {
+  diffs: PerspectiveDiff[];
+  revision: string;
+  sequence: number;
+}
+
+/** Canonical payload signed by a server's identity key for federation calls. */
+export function canonicalFederationPayload(
```
lal-bot-coasys, line 176: 🔴 **`canonicalFederationPayload` includes no timestamp or nonce → federation calls are replayable indefinitely.** The signed payload is `{kind, roomId, diff, sequence, revision}`. Nothing binds it to a moment in time or a single delivery.

On-path attacker (or any peer with a captured request body) can replay an old federate call for `remove(X)` months later — server re-verifies signature (still valid), TOFU still trusts the peer, `applyDiffAndAppend` removes X again after it was legitimately re-added. Same for reconcile requests, which leak the peer's active hash set to anyone who replays them.

Fix: add `timestamp` (or a monotonically-increasing per-peer `sequence` the server tracks) to the canonical payload, and reject at the receiver if the delta exceeds a small window (or if the peer's per-peer sequence goes backward).

link-server/src/routes.ts
```diff
@@ -0,0 +1,413 @@
...
+      !Array.isArray(body.diff.additions) ||
+      !Array.isArray(body.diff.removals) ||
+      typeof body.serverPublicKey !== "string" ||
+      typeof body.serverSignature !== "string" ||
+      typeof body.sequence !== "number" ||
+      typeof body.revision !== "string"
+    ) {
+      return reply.code(400).send({ error: "malformed federate payload" });
+    }
+    const result: FederateResult = await ctx.federation.handleIncomingFederate(
```
lal-bot-coasys, line 364: 🟠 **`/federate` and `/reconcile` skip auth (federation is server-to-server signed) but also skip any rate-limit hook.** Every other room endpoint carries `preHandler: [requireAuth(ctx), jwtRateLimit(...)]`. These two don't.

`handleIncomingFederate` runs `verifyPeerSignature` (ed25519) plus `verifyLinkExpression` per link in the diff — unauthenticated attacker can send 10k links with garbage signatures per request, from anywhere on the internet, and pin all cores. Same shape for reconcile.

Fix: add an IP-keyed rate limiter to both endpoints (unauthenticated, so JWT keying doesn't apply). Also worth capping the diff size (`additions.length + removals.length`) so a single request can't force gigabytes of verify work.

link-server/src/auth.ts
```diff
@@ -0,0 +1,251 @@
...
+  const existing = db.getIdentity("server-ed25519");
+  if (existing) {
+    return { publicKey: existing.public_key, privateKey: existing.private_key };
+  }
+  const priv = ed.utils.randomPrivateKey();
+  const pub = await ed.getPublicKeyAsync(priv);
+  const publicKeyHex = ed.etc.bytesToHex(pub);
+  const privateKeyHex = ed.etc.bytesToHex(priv);
+  db.setIdentity("server-ed25519", publicKeyHex, privateKeyHex);
+  return { publicKey: publicKeyHex, privateKey: privateKeyHex };
```
lal-bot-coasys, line 174: 🔴 **`verifyLinkExpression` verifies the signature over the raw canonical payload, but AD4M's `agentSignStringHex` signs `SHA-256(payload)` first.** The doc-comment at `hashMessageForVerify` (line 66-70 in this file) explicitly calls this out as the executor's convention. The challenge/response path (routes.ts:137) correctly hashes before verify; this one doesn't.

With `skipLinkVerification=false` (the intended default), any diff a real executor commits is rejected as `"invalid link signature"`. This is the exact reason I had to set `SKIP_LINK_VERIFICATION=true` on the integration-test spawn in #905 (flagged as "Cross-repo follow-up" in that PR's body).

Tests pass because `link-server/tests/helpers.ts::createSignedLink` also signs the raw payload — the two bugs mask each other.

Fix: `verifyHex(publicKey, hashMessageForVerify(canonicalLinkPayload(link)), link.proof.signature)` — hash first, matching the executor convention.

Once this lands, the executor and server can agree, `skipLinkVerification=true` can come out of the integration tests (I set it in `tests/js/tests/integration.test.ts`), and the `/commit` handler's "the server does not need to verify" comment (routes.ts near the commit handler) becomes wrong — flip both back on.

link-server/src/server.ts
```diff
@@ -0,0 +1,128 @@
...
+ * federation and E2E encryption. Call `app.listen(...)` on the result, or
+ * use `close()` to tear everything down (used heavily by tests).
+ */
+export async function buildServer(opts: ServerOptions): Promise<BuiltServer> {
+  const dbPath = opts.dataDir === ":memory:" ? ":memory:" : path.join(opts.dataDir, "data.sqlite");
+  const db = new LinkServerDB(dbPath);
+  const identity = await ensureServerIdentity(db);
+
+  const app = Fastify({ logger: opts.logger ?? false });
+  await app.register(websocketPlugin);
```
lal-bot-coasys: 🟠 **`@fastify/websocket` registered without `maxPayload` and no ping/pong keepalive in `ws.ts`.** Unbounded inbound message size + dead-socket accumulation.

Attack: authenticated client (or a stolen JWT) sends a 2 GB text frame to `/rooms/:roomId/ws`. `ws.ts::handleMessage` does `raw.toString()` then `JSON.parse`, allocating the entire buffer in memory before rejecting via the JSON.parse catch — one connection OOMs the server.

Separately: half-open TCP connections after NAT rebinding stay in `this.byId` and `this.rooms` forever because no ping timeout fires close events → presence entries leak, broadcasts go to dead sockets.

Fix: `await app.register(websocketPlugin, { options: { maxPayload: 1_048_576 } })` (or whatever limit fits the largest legit diff), and add `ws.ping()` on a `setInterval` with a `pong` timeout that closes the socket.

link-server/src/routes.ts
```diff
@@ -0,0 +1,413 @@
...
+  // ---- sync / render / revision / peers ----
+
+  app.get(
+    "/rooms/:roomId/sync",
+    { preHandler: [requireAuth(ctx), jwtRateLimit(ctx.rateLimits.roomJwt)] },
+    async (request, reply) => {
+      const claims = request.authClaims!;
+      const query = request.query as { since?: string };
+      const parsedSince = query.since !== undefined ? Number.parseInt(query.since, 10) : 0;
+      const since = Number.isFinite(parsedSince) ? parsedSince : 0;
```
lal-bot-coasys: 🟠 **`/sync?since` is unbounded on the low side.** Number.isFinite passes negative integers straight through; `getDiffsSince` runs `WHERE sequence > since`, so `since=-9999999` returns every diff ever recorded in one round-trip.

Malicious or misbehaved client (or a client with a corrupted local sequence store) forces a full-table scan and a huge JSON payload each sync tick — trivial DoS/amplification against a self-hosted server, and it bypasses the incremental-catchup assumption the WS `diff` path makes.

Fix: clamp to `Math.max(0, parsedSince)`. Also worth capping the response size (e.g., paginate at 1000 diffs and return a `has_more` cursor).

link-server/src/db.ts
```diff
@@ -0,0 +1,503 @@
...
+  }
+
+  /**
+   * Applies a PerspectiveDiff's additions/removals to the active link set
+   * (OR-Set: add-wins insert, remove-by-hash) and appends it to the
+   * append-only diff log, all in one transaction. Shared by local commits
+   * (routes.ts) and federation (federation.ts) so both paths guarantee
+   * identical merge + revision semantics.
+   */
+  applyDiffAndAppend(
```
lal-bot-coasys, line 373: 🟠 **`applyDiffAndAppend` re-prepares statements for every link inside the transaction.** `this.raw.prepare('INSERT OR IGNORE INTO links...')` fires 1500 times inside the transaction on a single batched commit, plus fresh prepares for `getNextSequence`, `appendDiff`, `getActiveHashes`, `UPDATE rooms`.

The client-side batching in `sync.ts` was explicitly built to coalesce 1500 `addLink`s into one commit POST (the fix I landed in #905). Server-side, this pattern undoes most of the batching win — thousands of statement parse/allocate cycles per batch.

Fix: cache prepared statements as private fields on the `LinkServerDB` class (constructed once in the constructor). better-sqlite3 explicitly optimizes for this pattern.

bootstrap-languages/server-link-language/src/auth.ts
```diff
@@ -0,0 +1,96 @@
...
+ * this function can't parse is simply treated as non-expiring (refreshed
+ * lazily on the next 401).
+ */
+export function parseJwtExpiryMs(token: string): number | null {
+    const parts = token.split(".");
+    if (parts.length !== 3) return null;
+    try {
+        const b64 = parts[1].replace(/-/g, "+").replace(/_/g, "/");
+        const padded = b64 + "=".repeat((4 - (b64.length % 4)) % 4);
+        const payload = JSON.parse(atob(padded)) as { exp?: number };
```
lal-bot-coasys, line 37: 🟠 **`parseJwtExpiryMs` decodes the JWT payload with `atob()` then `JSON.parse`, treating base64url output as a binary string rather than UTF-8.** Today only `exp` (a number) is read, so it works. The moment anyone adds a string claim (roomName, agentAlias, etc.) containing non-ASCII, `atob` returns latin-1 code units and `JSON.parse` either fails or silently returns the wrong string.

The catch block quietly returns null → `expiresAt=null` → `getValidToken` never proactively refreshes → the client only re-auths on 401s, cascading into subtle sync stalls.

Fix: use `jose.decodeJwt(token)` (link-server already depends on `jose` — pull it in as a client dep too). It handles UTF-8 payload decoding correctly and gets you the whole payload object, not just `exp`.

bootstrap-languages/server-link-language/src/auth.ts
```diff
@@ -0,0 +1,96 @@
...
+    _session = session;
+    return session;
+}
+
+/**
+ * Returns a token guaranteed valid for at least `skewMs` more milliseconds
+ * (default 30s), transparently re-authenticating if there is no cached
+ * session, the cached token is expired, or it's about to expire.
+ */
+export async function getValidToken(skewMs = 30_000): Promise<string> {
```
lal-bot-coasys, line 76: 🟠 **`getValidToken` has no in-flight promise cache — two concurrent callers with an expired session both fire `authenticate()`.** On WS reconnect, `ws-client`'s `getUrl()` awaits `getValidToken()` at the same moment the batched flush chain calls it from `commit()`. Both see `_session.expiresAt - Date.now() < skewMs` and both run the two-step DID challenge-response.

Two challenges consumed, two tokens issued, two DB session rows written. The later resolver's token overwrites the earlier one in `_session`, but `ws-client` already sent the earlier one to `/rooms/:roomId/ws?token=...` — if the server ever adds revoke-on-reissue semantics, this connection dies.

Fix (small): cache the pending promise:
```ts
let _pending: Promise<AuthSession> | null = null;
// in getValidToken, replace the fallthrough with:
if (!_pending) _pending = authenticate().finally(() => { _pending = null; });
return (await _pending).token;
```

bootstrap-languages/server-link-language/index.ts
```diff
@@ -0,0 +1,396 @@
...
+            console.error(
+                "[server-link-language] initial startup sequence failed — will keep retrying via " +
+                "websocket reconnect backoff and the runtime's periodic sync():",
+                err,
+            );
+        }
+
+        // Always attempt to connect, even if the steps above failed: getUrl()
+        // re-authenticates internally, and the reconnect loop retries with backoff.
+        void wsClient.connect();
```
lal-bot-coasys: 🟠 **`void wsClient.connect()` discards a promise that can reject synchronously via `new WebSocket(url)` throwing on a malformed URL** — unhandled rejection.

If `SERVER_URL` was published with a stray character, or a Deno WebSocket constructor rejects a `wss://` URL with a missing token, `factory.connect(url)` throws inside `openConnection` **before** any `scheduleReconnect` is registered. The exception bubbles out of the awaitless `void`, hitting `process.on('unhandledRejection')`. On Node with `--unhandled-rejections=strict` this crashes the executor host; on the Deno sandbox it kills the language worker.

Compare to `void syncModule.catchUp().catch(...)` a few lines up which correctly attaches a catch. Fix: `void wsClient.connect().catch((err) => { console.error('server-link-language: initial WS connect failed', err); });` — the scheduleReconnect inside the reconnect loop will re-run once it's registered.

bootstrap-languages/server-link-language/src/ws-client.ts
```diff
@@ -0,0 +1,192 @@
...
+        await this.openConnection();
+    }
+
+    close(): void {
+        this.closedByUser = true;
+        if (this.reconnectTimer) {
+            clearTimeout(this.reconnectTimer);
+            this.reconnectTimer = null;
+        }
+        this.sendQueue = [];
```
lal-bot-coasys, line 81: 🟠 **`WsClient.close()` clears `sendQueue = []` silently.** Any outbound messages queued while the socket was disconnected are dropped without notifying the caller.

Example: caller invokes `perspective.telepresence.sendSignal(peer, payload)` during a network blip. `WsClient.send()` falls through to `sendQueue.push(data)` and resolves the outer promise as success (`{}`). Language teardown races the reconnect and runs `close()` → `sendQueue` wiped. The peer never receives the signal, but the caller's `await` returned OK, so no retry is triggered.

Fix: at teardown, either flush the queue synchronously (fire the sends against the WS if it's open, or reject queued promises with a clear error) or return an explicit `{queued: true}` from send-while-offline so callers can distinguish `delivered` from `queued-pending-reconnect` from `dropped`.

bootstrap-languages/server-link-language/src/sync.ts
```diff
@@ -0,0 +1,475 @@
...
+            const requeue = segments.slice(i);
+            const requeueSummary = requeue.map((s, idx) =>
+                `  segment ${i + 1 + idx}: ${s.additions.length} adds, ${s.removals.length} removes`)
+                .join("\n");
+            console.warn(
+                `[server-link-language] segment ${i + 1}/${segments.length} failed after ` +
+                `${MAX_COMMIT_ATTEMPTS} attempts; re-enqueueing ${requeue.length} segment(s) ` +
+                `for next flush cycle:\n${requeueSummary}`,
+            );
+            _pendingQueue.unshift(...requeue);
```
lal-bot-coasys, line 239: 🟡 **Failed batch requeues have no time-based retry.** `flushBatch` on failure does `_pendingQueue.unshift(...requeue)` and returns; the comment claims "the next enqueueCommitBatched naturally picks them up" — but that's only true if a caller happens to write again.

Scenario: Alice calls `addLink(X)` once. Server transiently 5xxs, all 3 attempts fail, X requeues into `_pendingQueue`. Alice does no more writes. `_pendingQueue` is in-memory-only. No `setTimeout` / `setInterval` / heartbeat re-invokes `flushBatch`. X sits pending indefinitely; every peer joining the room bootstraps from the server (which never received X) and never sees X. Local store has X, server + all other peers do not — silent divergence.

Fix (minimum): schedule a retry with exponential backoff (`setTimeout(() => scheduleFlush(), backoff)`) when `flushBatch` catches an error and requeues. Fix (proper altitude): persist the outbound queue to `storagePut()` so it survives restarts, and drain on `init()` — the AGENTS.md "No persistent commit outbox" limitation gets closed. That's a durable-outbox pattern worth lifting into ad4m-ldk so every write-through language reuses it.

bootstrap-languages/server-link-language/src/sync.ts
```diff
@@ -0,0 +1,475 @@
...
+ * await (by a concurrent caller or by teardown code that itself commits)
+ * would otherwise stay unflushed. The bound is a runaway-loop guard, not
+ * an expected iteration count — a settled system exits after 1 or 2 laps.
+ * Enqueues are routed through the normal `enqueueCommitBatched` path so
+ * they land in `_inflight` — direct `flushBatch()` calls would break the
+ * serialisation invariant that other flushes rely on.
+ */
+export async function drainCommitBatch(): Promise<void> {
+    const MAX_LAPS = 100;
+    for (let lap = 0; lap < MAX_LAPS; lap++) {
```
lal-bot-coasys, line 362: 🟡 **`drainCommitBatch` still retry-storms teardown when the server is down.** The recent fix (`8e2abce8f`) removed the auto-re-schedule from `flushBatch` for the normal-op path, but `drainCommitBatch` still nudges a fresh `flushBatch` on every lap the queue is non-empty (line 366-372).

On teardown-during-outage: each lap awaits `_inflight` (previous flush finished 3 attempts in ~1.4s), sees `_pendingQueue.length > 0` and `_flushScheduled = false`, schedules another flush → another 3 failing POSTs. 100 laps × ~1.4s ≈ 140s of teardown blocking with 300 wasted POSTs. If the executor's shutdown timeout is < 140s, the process is SIGKILL'd mid-drain and every queued diff is lost anyway.

Fix: on drain, if the first lap fails, bail immediately (don't schedule more flushes). The queued diffs are lost either way; the goal is to make teardown fast when the server's unreachable rather than trying 300 times to get through.

bootstrap-languages/server-link-language/src/encryption.ts
```diff
@@ -0,0 +1,290 @@
...
+
+export function hexToBytes(hex: string): Uint8Array {
+    const clean = hex.trim();
+    if (clean.length % 2 !== 0) {
+        throw new Error(`hexToBytes: odd-length hex string (${clean.length} chars)`);
+    }
+    const out = new Uint8Array(clean.length / 2);
+    for (let i = 0; i < out.length; i++) {
+        const byte = clean.substring(i * 2, i * 2 + 2);
+        const parsed = parseInt(byte, 16);
```
lal-bot-coasys: 🟡 **`hexToBytes` silently accepts partially-invalid hex.** `parseInt('ag', 16)` returns 10 (`a`=10, then `g` is a stop char) — not NaN — so `Number.isNaN(parsed)` doesn't catch it. A tampered SealedRoomKeyEnvelope with hex chars like `a5` followed by `g0` parses as `[0xa5, 0x00...]`.

A malformed nonce/ephemeralPublicKey field then crosses into AES-GCM decrypt where it fails with an opaque auth-tag error, instead of a clear "invalid hex" error at the field boundary. Hides which side is corrupt during E2E key exchange debugging.

Fix: validate each byte is exactly two hex chars before parseInt, or use a strict regex `/^[0-9a-fA-F]{2}$/.test(byte)`. Or replace the whole function with `@noble/hashes/utils::hexToBytes` (a strict, audited implementation used elsewhere in the same package via `@noble/ed25519`).

link-server/src/federation.ts
```diff
@@ -0,0 +1,309 @@
...
+
+  start(): void {
+    if (this.timer) return;
+    this.timer = setInterval(() => {
+      void this.reconcileAll();
+    }, this.reconcileIntervalMs);
+    this.timer.unref();
+  }
+
+  stop(): void {
```
lal-bot-coasys: 🟡 **`FederationManager.stop()` clears the timer but does not await any `reconcileAll()` already in flight.** A `reconcileAll()` started 5s earlier may still be awaiting a peer's HTTP response when the caller proceeds to `db.close()`.

When the peer replies, `reconcileWithPeer` resumes with `await verifyLinkExpression` → `this.db.applyDiffAndAppend` → `SqliteError: The database connection is not open`. The error escapes as an unhandled rejection since the `setInterval` callback used `void reconcileAll()`.

Fix: track the in-flight promise and await it in `stop()` before returning (with a timeout as a safety net — don't block shutdown forever if a peer HTTP hangs).

link-server/src/telepresence.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+    return result;
+  }
+
+  isOnline(roomId: string, did: string): boolean {
+    return this.agents.get(roomId)?.has(did) === true;
+  }
+
+  close(): void {
+    for (const timer of this.offlineTimers.values()) clearTimeout(timer);
```
lal-bot-coasys: 🟡 **`close()` cancels all pending grace-period timers with `clearTimeout` but never invokes their `onExpired` callbacks.** The `ws.ts` callback that broadcasts `peer-left` (registered via `scheduleOffline`) is silently dropped.

Scenario: server begins graceful shutdown while Alice's socket has just dropped (grace timer pending) but Bob is still connected. `close()` cancels Alice's timer without firing `onExpired` → Bob never gets `peer-left` for Alice. Bob's client keeps Alice in its `onlineAgents` map until Bob's own WS drops or he refreshes.

Fix: on `close()`, iterate `offlineTimers` and call each `onExpired` synchronously before clearing (or send a batched `peer-left` for every scheduled did in one broadcast).

bootstrap-languages/server-link-language/src/adapters-deno.ts
```diff
@@ -0,0 +1,153 @@
...
+                headers: {},
+                body: responseText || "",
+            };
+        } catch (err: unknown) {
+            // LDK debt: httpFetch throws on non-2xx, encoding the HTTP status
+            // in the Error.message string instead of returning a response
+            // object. This regex scrapes the status code out. The proper fix
+            // belongs in @coasys/ad4m-ldk (httpFetch should return {status, body}).
+            const errMsg = err instanceof Error ? err.message : String(err);
+            const match = errMsg.match(/http_fetch\s+\S+\s+\S+\s+->\s+(\d+):\s*(.*)?$/s);
```
lal-bot-coasys, line 60: 🔵 **`DenoTransport.fetch` reconstructs HTTP status by regex-matching the `httpFetch` error string** (`/http_fetch\s+\S+\s+\S+\s+->\s+(\d+):/`). Implicit contract between two files nothing enforces.

If `host.js::httpFetchImpl` ever prepends a timestamp, changes `->` to `→`, or wraps errors in a class that stringifies differently, this regex fails → DenoTransport returns `{status: 0, body: errMsg}`. `fetchRoomKey`'s `err.status === 404` branch never triggers on a real 404 → cold-start throws instead of treating a keyless room as plaintext → `roomKeyStatus='error'` → language refuses ALL commits. One host.js refactor silently disables every server-link-language instance.

This is exactly what the merge-base's PR #915 (`http_fetch_typed` returning `{status, body}`) was solving before this PR reverted it. Un-revert or land an equivalent typed API — the current regex-scrape is a load-bearing string match with no test asserting the shape.

Also: on success, `fetch` hardcodes `status: 200`, making 201/202/204 indistinguishable and turning the `err.status === 204` branch in `fetchRoomKey` (line 156) into dead code.

bootstrap-languages/server-link-language/index.ts
```diff
@@ -0,0 +1,396 @@
...
+ * peers, ACL, E2E key exchange) and native WebSocket (real-time diff
+ * push, telepresence, presence). Implements every link-language
+ * capability: perspective-commit, perspective-sync, perspective-query,
+ * peers, and telepresence.
+ *
+ * See README.md for architecture and AGENTS.md for operational notes /
+ * known limitations.
+ */
+
+import { defineLanguage, hash } from "@coasys/ad4m-ldk";
```
lal-bot-coasys, line 15: 🔵 **AGENTS.md convention violation.** `bootstrap-languages/server-link-language/AGENTS.md` ("Non-obvious conventions" bullet 1) states: `src/adapters-deno.ts` is the only file allowed to import `@coasys/ad4m-ldk`, and only `index.ts` is allowed to import `adapters-deno.ts`.

This line (`import { defineLanguage, hash } from "@coasys/ad4m-ldk";`) makes `index.ts` a second direct importer of `@coasys/ad4m-ldk`. Either update AGENTS.md to reflect the new reality (index.ts and adapters-deno.ts both allowed) or route these imports through adapters-deno.ts too.

Worth a moment to decide which — the convention exists because the isolation is what makes it possible to swap the LDK for a Node runtime later; punching a hole for `defineLanguage/hash` may be fine (they're pure functions with no runtime dep), but the guardrail should be intentional.

link-server/tests/ws.test.ts
```diff
@@ -0,0 +1,273 @@
...
+    const roomId = randomUUID();
+    const agentA = await createTestAgent();
+    const agentB = await createTestAgent();
+    const tokenA = await authenticateAgent(server.url, roomId, agentA);
+    await postJson(`${server.url}/rooms/${roomId}/acl`, { action: "add", did: agentB.did }, tokenA);
+    const tokenB = await authenticateAgent(server.url, roomId, agentB);
+
+    const wsA = openWs(server.wsUrl, roomId, tokenA);
+    const wsB = openWs(server.wsUrl, roomId, tokenB);
+    await Promise.all([waitForOpen(wsA), waitForOpen(wsB)]);
```
lal-bot-coasys, line 92: 🔵 **AGENTS.md gotcha violated in 6 places.** `link-server/AGENTS.md` ("Known gotchas" bullet 3): *"the server sends `online-agents` synchronously as part of completing the upgrade. If a test attaches its `message` listener after awaiting the socket's `open` event, that first message can already have fired and be lost — attach listeners immediately after `new WebSocket(...)`, before awaiting open."*

This test (and 5 others at ~line 132, 163, 189, 223, 254) opens both sockets, awaits `waitForOpen`, THEN instantiates `collectMessages(wsA)` / `collectMessages(wsB)`. That's the exact anti-pattern. Currently probably fine because message handlers are attached fast enough — but the whole reason AGENTS.md called it out is that it's flaky by construction, not by observation.

Fix: swap the order — attach collectors immediately after `new WebSocket(...)`, then `await Promise.all([waitForOpen(...)])`.

link-server/src/rate-limit.ts
```diff
@@ -0,0 +1,69 @@
...
+ * process per the project brief.
+ */
+export interface RateLimitCheck {
+  allowed: boolean;
+  /** Seconds the caller should wait before retrying, only meaningful when !allowed. */
+  retryAfterSec: number;
+  remaining: number;
+}
+
+export class SlidingWindowLimiter {
```
lal-bot-coasys, line 13: 🔵 **Hand-rolled `SlidingWindowLimiter` where `@fastify/rate-limit` covers the same ground.** 70 lines of manual sliding-window counting + `setInterval` sweep + `unref` bookkeeping + `close()` cleanup, wired into every route by hand at `routes.ts:68-86`.

`@fastify/rate-limit` (v10, matches Fastify 5) ships sliding-window semantics, per-route `max`/`keyGenerator`, `Retry-After` / `X-RateLimit-*` headers, IP-or-JWT keying, and swappable stores (Redis for multi-instance) as one plugin registration. Applies globally by default so new routes can't silently be added without rate-limiting.

Also ties into finding above: `/federate` and `/reconcile` need IP-keyed limiting, not JWT-keyed — `@fastify/rate-limit`'s `keyGenerator` per route makes that a two-line config.

Hand-rolled is worth keeping only if there's a specific requirement the plugin can't meet — I don't see one here.

link-server/src/types.ts
```diff
@@ -0,0 +1,182 @@
...
+  | { type: "set-online-status"; status: unknown };
+
+/**
+ * Canonical, order-stable JSON payload used for both signing and hashing a
+ * LinkExpression. For plaintext links this covers {source,predicate,target,
+ * author,timestamp}; for encrypted links the ciphertext/nonce stand in for
+ * source/predicate/target so the server can still content-address the link
+ * (OR-Set merge) without ever seeing plaintext.
+ */
+export function canonicalLinkPayload(link: LinkExpression): string {
```
lal-bot-coasys, line 96: 🔵 **`canonicalLinkPayload` should live in one place both sides import.** Right now:
- Server: `canonicalLinkPayload` here at line 96.
- Client (server-link-language): `linkIdentity` at `sync.ts:323` reproduces content identity but with a different serialization (`JSON.stringify([author, timestamp, source, predicate, target])` vs the server's `JSON.stringify({source, predicate, target, author, timestamp})`).
- Executor (rust-executor): a third serialization (per finding above on `auth.ts:174` — SHA-256 of a different envelope entirely).

Three independent "canonical" forms in three places, none in sync. The client's `linkIdentity` docstring says it "must byte-match the server's hash" but the code doesn't — it happens not to matter today because it's only used for local coalesce merge-safety, but the invariant is a comment, not a compile-time fact.

Altitude fix: extract a single `@coasys/link-canonical` module both TS sides import, and specify the exact byte format for the Rust executor to match (or generate a shared spec). This is the same class of bug as the E2E format mismatches — one wire format on both sides beats "we said we'd match" every time.

Related: the `LinkExpression` / `PerspectiveDiff` / `ExpressionProof` types are declared **3 times** (LDK + link-server/src/types.ts + server-link-language/src/types.ts). Same fix: one source of truth in `@coasys/ad4m-ldk`.

bootstrap-languages/server-link-language/README.md
```diff
@@ -0,0 +1,254 @@
...
+  (`SealedRoomKeyEnvelope`: ephemeral X25519 public key + AES-GCM nonce +
+  ciphertext, base64(JSON)-framed as the `encryptedKey` string) — the same
+  shape `libsodium`'s `crypto_box_seal` produces. Only the intended
+  recipient's derived private key can open it.
+- **Link confidentiality.** Once a room key is available,
+  `encryptLinkForWire` / `decryptLinkFromWire` protect a link's
+  `{source, predicate, target}` payload with AES-256-GCM (fresh nonce per
+  link). `author` / `timestamp` / `proof` stay in the clear — the runtime's
+  signature was computed over plaintext data before this language ever saw
+  the diff, and this metadata is needed for verification and routing. **The
```
lucksus, line 211: We don't need to verify the link signatures. Could have encrypt the meta-data as well?

lucksus:
Also, does this need to be on-top of https://github.com/coasys/ad4m/pull/918? I was hoping to merge this first and do the test refactor later..

HexaField:
## Review response

Thanks for the thorough angle-sweep — addressed every finding that lands cleanly in this PR's scope. Pushed `9b7f4a074`.

### Fixed in this push

| Finding | Fix |
|---|---|
| 🟠 DB re-prepares statements ~1500× per transaction | Cached all prepared statements at DB construction — `applyDiffAndAppend` now reuses them |
| 🟠 `/sync?since` unbounded low side | Clamped to `Math.max(since, 0)` |
| 🟠 WebSocket no `maxPayload` | `maxPayload: 1 MiB` on `@fastify/websocket` registration |
| 🟠 `getValidToken` no in-flight cache | Added promise dedup — concurrent callers share one `authenticate()` round-trip |
| 🟠 `void wsClient.connect()` unhandled rejection | Added `.catch()` |
| 🟠 `WsClient.close()` silently drops queue | Logs a warning with the count of dropped messages |
| 🔴 `roomKeyStatus='error'` permanent poison | Retries `setupRoomKey()` once before refusing commit |
| 🟡 `hexToBytes` accepts partial-invalid hex | Strict regex validation per byte pair |
| 🟡 `FederationManager.stop()` doesn't await in-flight | Tracks `reconcileInFlight` promise, `stop()` awaits it |
| 🟡 `TelepresenceManager.close()` drops `onExpired` | Fires pending callbacks during shutdown |
| 🔵 AGENTS.md stale standalone-repo claim | Removed |

### Acknowledged — not in this PR's scope

**E2E encryption mismatches (4 items)** — Already documented in language `AGENTS.md` as "5 confirmed incompatibilities." These need coordinated client+server work and will land as a follow-up when E2E gets real traffic.

**Signature convention mismatch / commit doesn't verify link sigs** — The commit path deliberately relays signatures as metadata (JWT auth provides transport identity). The SHA-256 vs raw mismatch between executor and server is a cross-component issue already worked around with `SKIP_LINK_VERIFICATION` — reconciling it needs executor-side changes too.

**Federation trust hardening** (TOFU race, replay, rate-limiting `/federate`+`/reconcile`) — Valid security findings for a production deployment. The current model documents TOFU as intentional (pin-on-first-contact for peers added while offline). Timestamps/nonces and rate-limiting the server-to-server endpoints belong in the hardening pass before federation sees real traffic.

**Design-altitude items** (shared crypto module, type dedup, `@fastify/rate-limit`, DenoTransport regex, durable outbound queue in LDK) — All good directions. The regex issue is exactly what the reverted PR #915 was solving; worth un-reverting or landing an equivalent. The shared `@coasys/link-crypto` extraction and type dedup belong in a follow-up that touches the LDK.

---

Re: stacking order — this PR *can* merge independently of #918. The only overlap was the `startAgent` helper (which #918 adds `runHolochain` to), but the test refactor in #918 doesn't touch link-server or server-link-language code. If you'd prefer to merge this first, rebasing #918 onto the result should be clean.

bootstrap-languages/server-link-language/src/adapters-deno.ts
HexaField: Agreed — the Deno adapters wrap the same `ad4m:host` primitives any language needs. Moving them into the ALDK as opt-in adapters saves every new language from reimplementing the bridge. Good follow-up — the adapter interfaces already define the contract.

bootstrap-languages/server-link-language/src/types.ts
HexaField: Metadata (`author`, `timestamp`, `proof`) stays in the clear for two reasons: (1) the AD4M runtime's signature covers the plaintext `{source, predicate, target}` — the proof was computed before the link reached this language, so verifying it downstream requires the plaintext metadata alongside it; (2) the server needs `author` to enforce "every link's author must match the authenticated DID" at the transport layer. Encrypting metadata would require rearchitecting the signature flow (sign-then-encrypt-all) and the server's commit validation. Feasible but a different trust model.

bootstrap-languages/server-link-language/AGENTS.md
HexaField: Fixed — removed the standalone-repo language. This now just says it lives in the monorepo at `bootstrap-languages/server-link-language/`.

bootstrap-languages/server-link-language/README.md
HexaField: Right — link signatures travel as metadata. The server relies on JWT-authenticated DID identity at the transport layer, not per-link signature verification. Downstream consumers (other agents, auditors) can verify signatures if they choose. The server doesn't need to — and doing so currently hits the SHA-256-vs-raw mismatch between executor signing convention and server verification (the `SKIP_LINK_VERIFICATION` workaround). Encrypting metadata too would require a sign-then-encrypt-all model — see the reply on `types.ts:96` above.

PR #920, opened 2026-08-25
Fix: Relation correctness

jhweir:
# Relation correctness: duplicate links, `@HasOne` LWW, batching, silent drops

> [!IMPORTANT]
> **Branch** `fix/relation-correctness` · **Base** `dev`.
> Independent. Nothing in the series is required before it.
>
> Part of an eight-PR series on the AD4M model layer — review wave 1 of 4. The full shape:

```
     ├─ 2  getter+target
```

## Summary

Four ways relations behaved differently from the rest of the model layer, all of them silent, and
all of them found while auditing AD4M against what WE needs next. Two are reads (a collection could
report a target twice; a to-one relation served its *oldest* link forever) and two are writes (a
to-one link could not join a write group; a scalar handed to a relation field was discarded without
a word).

They ship together because they are one complaint: `@HasOne` and relation writes have been finished
less carefully than `@HasMany`, and the seams show in the same places. Three of the four are also
currently *masked* by `collectionSetter` deleting every link before re-adding it — which the CRDT
ordering PR at the end of this stack replaces with a diff. Left unfixed they would have started
failing on the ordinary path with no code having changed to explain it.

## Changes

**`rust-executor/src/perspectives/model_query/hydration.rs`** — the two read fixes, both in the
collection loop of `hydrate_one`.

A link is stored as a direct triple plus a reifier keyed on
`sha256(source, predicate, target, timestamp)`. Two peers asserting the same membership — concurrent
`addSignal`, `addParticipant`, `addMention` — share one triple but mint two reifiers, and the
instance query joins through the reifier to recover author and timestamp. So it returns a row per
reifier and the target lands in the array twice. This collapses duplicates unconditionally rather
than behind a `distinct` flag, because a `$count` projection already answers `COUNT(DISTINCT ?t)`:
within a single query, `$signalCount` said 1 while `signals` returned two entries of the same signal.
A count that disagrees with the list it counts is not a defensible default.

Scalar relations resolved *earliest*-wins — the collection was sorted ascending and the first entry
taken — while scalar properties a few lines above resolve last-write-wins. Re-pointing a `@HasOne`
without removing the old link kept serving the original target, so `Space.location` would show a
stale place and `Space.name` a fresh name from one write. Tiebreaking uses strict `>` so links
sharing a timestamp resolve to the first seen, matching the property branch rather than diverging
from it.

**`rust-executor/src/perspectives/model_query/projection.rs`** — the same dedup for non-count
projections, so a filtered projection's list agrees with its own count. Only reachable when a
projection filters on `author` or `timestamp`, which is when `build_projection_reifier_patterns`
joins the reifier at all.

*Deliberately not touched:* `relations.rs`. An earlier draft of the plan listed it as a third dedup
site. It is not one — `resolve_reverse_relations` and `resolve_reverse_include` select
`?source <pred> ?target` with no reifier join, so RDF set semantics have already deduplicated them.

**`core/src/model/decorators.ts`** — `@HasOne`'s generated `add`/`remove`/`set` now forward
`batchId`. Their block is introduced by the comment *"mirroring @HasMany"* and then did not mirror
it, while the `addRelationValue`/`setRelationValues` they delegate to have taken a batch all along —
the parameter was simply dropped on the way in. Without it, anything that creates a record and points
it at something commits twice, and every subscriber sees the state between: a record whose to-one
relation is still empty. That is the board card that appears unpositioned and then jumps.

**`core/src/model/Ad4mModel.ts`** — `innerUpdate` coerces a scalar on a relation field to a
single-member relation instead of skipping it. `Model.create(p, { source: 'uri://x' })` typechecked,
ran, resolved, and wrote no link; the failure surfaced later and elsewhere, as a reader calling the
record malformed. The coercion sits *outside* the `setProperties` gate, mirroring the array branch:
`create` passes `setProperties: false` when the class has a SHACL constructor, but a relation carries
`ad4m://adder` rather than `ad4m://setter` so `create_subject` drops it too — left inside the gate it
would be discarded by both paths at once.

**`core/src/model/test_helpers.rs`, `Ad4mModel.test.ts`** — `inst_links_at` gives per-link
timestamps, which the existing helper cannot express (the same `(predicate, target)` at two
timestamps; links arriving out of timestamp order). `scalar_relation` builds the `@HasOne` shape.

## Known follow-ups

- `build_count_sparql` reuses `build_query_patterns`, which skipped `OR`/`AND`/`NOT`, so `totalCount`
  was computed unfiltered under a combinator. Fixed by the where-compiler PR later in this stack.
- The duplicate-link condition is a P2P-only failure, which is why it never showed in single-agent
  tests. Worth a cross-peer integration test once the harness can run two executors cheaply.

## Test plan

> tree above.

- [x] `cargo test -p ad4m-executor --lib perspectives::model_query` — 221 passing, 7 new
- [x] 7 new Rust tests covering: duplicate collapse, dedup preserving earliest position, dedup before
      literal decoding, `@HasOne` LWW, LWW regardless of row order, tie matching the property branch,
      absent stays absent
- [x] `jest src/model` in `core` — 252 passing, 6 new
- [x] **Verified the new TS tests fail without the fix**: 5 of 6 fail on a stashed working tree; the
      sixth is the no-`batchId` regression guard, which passes either way by design
- [x] `tsc --noEmit` clean

## Summary by CodeRabbit

* **New Features**
  * Relation metadata can now include interpretation hints to improve downstream extraction and schema handling.

* **Bug Fixes**
  * To-one relation values are now correctly saved during property updates.
  * To-one relations resolve to the most recent link, with consistent tie handling.
  * Collection relations and query projections now remove duplicate targets while preserving ordering.
  * Invalid relation values now produce validation errors instead of being silently ignored.

### DIFF
--- a/core/src/model/Ad4mModel.test.ts
+++ b/core/src/model/Ad4mModel.test.ts
@@ -2514,10 +2514,22 @@ describe("Relation writes: to-one batching and scalar coercion", () => {
     target?: string;
   }
 
-  // No required property, no flag, no initial — so `save()` finds no SHACL
-  // constructor, skips `createSubject`, and calls `innerUpdate(true)`.
-  @Model({ name: "TestPlacement" })
-  class TestPlacement extends Ad4mModel {
+  // A coverage fixture for a branch that already exists in `save()`, not a
+  // modelling pattern being endorsed. No required property, no flag and no
+  // initial value means `buildSHACL` emits the node shape and its property
+  // shapes as normal but with an *empty* constructor-action list, which the
+  // Rust side rejects as "No SHACL constructor found" — so `save()` skips
+  // `createSubject` and calls `innerUpdate(true)` instead. The base expression
+  // exists either way: the `Ad4mModel` constructor mints one, and
+  // `createSubject` writes onto a base rather than creating it.
+  //
+  // Worth naming the real caveat, which is about conformance rather than
+  // writes: a class with no required property and no flag states no criteria,
+  // so nothing structurally distinguishes an instance of it from any other
+  // base expression. That makes it an overlay rather than a class, and it is a
+  // question about SDNA modelling generally rather than about this file.
+  @Model({ name: "TestNoConstructor" })
+  class TestNoConstructor extends Ad4mModel {
     @HasOne({ through: "we://placed_node" })
     node?: string;
   }
@@ -2601,7 +2613,7 @@ describe("Relation writes: to-one batching and scalar coercion", () => {
     it("still writes the relation when there is no SHACL constructor", async () => {
       const perspective = makePerspective();
 
-      await TestPlacement.create(
+      await TestNoConstructor.create(
         perspective,
         { node: "we://block/a" },
         { batchId: "batch-1" }
--- a/core/src/model/Ad4mModel.test.ts
+++ b/core/src/model/Ad4mModel.test.ts
@@ -2494,3 +2494,145 @@ describe("Ad4mModel instance identity (uniqueness without envelopes)", () => {
     expect(instance.id).toBe("flux://existing-instance");
   });
 });
+
+describe("Relation writes: to-one batching and scalar coercion", () => {
+  // `Relationship` is the shape that made both of these visible: a record whose
+  // two to-one relations name the things it connects, so it is worthless until
+  // both endpoints are linked, and it is created and pointed in one gesture.
+  @Model({ name: "TestRelationship" })
+  class TestRelationship extends Ad4mModel {
+    @Flag({ through: "we://flag", value: "we://relationship" })
+    flag: string = "";
+
+    @Property({ through: "we://title", required: true })
+    label: string = "";
+
+    @HasOne({ through: "we://relationship_source" })
+    source?: string;
+
+    @HasOne({ through: "we://relationship_target" })
+    target?: string;
+  }
+
+  // No required property, no flag, no initial — so `save()` finds no SHACL
+  // constructor, skips `createSubject`, and calls `innerUpdate(true)`.
+  @Model({ name: "TestPlacement" })
+  class TestPlacement extends Ad4mModel {
+    @HasOne({ through: "we://placed_node" })
+    node?: string;
+  }
+
+  const makePerspective = () =>
+    ({
+      executeAction: jest.fn().mockResolvedValue(undefined),
+      createSubject: jest.fn().mockResolvedValue(undefined),
+      createBatch: jest.fn().mockResolvedValue("batch-generated"),
+      commitBatch: jest.fn().mockResolvedValue(undefined),
+      stringOrTemplateObjectToSubjectClassName: jest
+        .fn()
+        .mockResolvedValue("TestRelationship"),
+      uuid: "test-perspective-uuid",
+    }) as any;
+
+  /** Targets written by `executeAction`, across every call. */
+  const writtenTargets = (perspective: any): string[] =>
+    perspective.executeAction.mock.calls.flatMap((call: any[]) =>
+      (call[2] ?? []).map((param: any) => param.value)
+    );
+
+  /** The batchId argument of each `executeAction` call. */
+  const batchIds = (perspective: any): unknown[] =>
+    perspective.executeAction.mock.calls.map((call: any[]) => call[3]);
+
+  describe("a scalar handed to a relation field", () => {
+    it("writes a link instead of being silently dropped", async () => {
+      // The failure this closes: this call typechecked, ran, resolved, and
+      // wrote no link — leaving a relationship with no endpoints, which only
+      // surfaces later as a reader calling the record malformed.
+      const perspective = makePerspective();
+
+      await TestRelationship.create(
+        perspective,
+        { label: "contradicts", source: "we://block/a", target: "we://block/b" },
+        { batchId: "batch-1" }
+      );
+
+      expect(writtenTargets(perspective)).toEqual(
+        expect.arrayContaining(["we://block/a", "we://block/b"])
+      );
+    });
+
+    it("writes it even when a constructor consumed the properties", async () => {
+      // `create` passes `setProperties: false` when the class has a SHACL
+      // constructor, because `createSubject` writes the values map. A relation
+      // carries `ad4m://adder` rather than `ad4m://setter`, so `createSubject`
+      // drops it — meaning the relation write has to happen here regardless of
+      // that flag, or both paths discard it at once.
+      const perspective = makePerspective();
+
+      await TestRelationship.create(
+        perspective,
+        { label: "contradicts", source: "we://block/a" },
+        { batchId: "batch-1" }
+      );
+
+      expect(perspective.createSubject).toHaveBeenCalled();
+      expect(writtenTargets(perspective)).toContain("we://block/a");
+    });
+
+    it("agrees with the documented array form", async () => {
+      const scalar = makePerspective();
+      await TestRelationship.create(
+        scalar,
+        { label: "x", source: "we://block/a" },
+        { batchId: "b" }
+      );
+
+      const array = makePerspective();
+      await TestRelationship.create(
+        array,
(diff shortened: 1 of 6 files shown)


core/src/model/Ad4mModel.test.ts
```diff
@@ -2494,3 +2494,145 @@ describe("Ad4mModel instance identity (uniqueness without envelopes)", () => {
...
+
+      await TestPlacement.create(
+        perspective,
+        { node: "we://block/a" },
+        { batchId: "batch-1" }
+      );
+
+      expect(perspective.createSubject).not.toHaveBeenCalled();
+      expect(writtenTargets(perspective)).toContain("we://block/a");
+    });
```
lucksus, line 2624: With this case, I don't understand where the relation target is written to - i.e. what base expression if this class doesn't have a constructor? So this class does make sense as an overlay (or "virtual" class), such that it adds that relation to any base and thus turns any baseExpression into an instance.

I guess `TestPlacement.create()` does create a new base expression in the TS code, but it doesn't go through `createSubject`?

I'm just wondering if that is really what we want to have happen as this is about cleaning up stuff in Ad4mModel that wasn't done right. I know the focus is on relations here, so this is def. a good step, but just wondering how we are currently and want to deal with these virtual classes.

Comment above says "so `save()` finds no SHACL". I think that at least is a problem. Does it mean, this class only exists in TS?

jhweir: Had Claude look into it:

Good questions — two of them have answers that aren't visible from this hunk, and the third is a
real point I've taken.

**Where the target is written.** The base expression always exists: `Ad4mModel`'s constructor
mints `ad4m://obj/${makeRandomId(24)}` (`Ad4mModel.ts:468`) unless one is passed in. `createSubject`
doesn't create a base, it writes the constructor's links *onto* a base — so with it skipped,
`innerUpdate` writes `<ad4m://obj/…> we://placed_node <target>` against that same base. Nothing is
left homeless; the record just has no flag saying what it is.

**"Only exists in TS."** No — `buildSHACL` emits the node shape and its property shapes exactly as
normal; it's the `constructorActions` array that comes out empty (`shacl-gen.ts:473`), and the Rust
side rejects an empty one with "No SHACL constructor found". My comment said "no SHACL constructor"
and I can see it reads as "no SHACL". Rewritten in 8bd3501.

**The skip is pre-existing.** `hasConstructor` is on dev today at `Ad4mModel.ts:1266` — this PR's
only hunk in that file is in `innerUpdate`. The test pins a branch that already existed rather than
introducing one, which I should have said in the fixture.

**Where you're right, and it's the interesting part.** A class with no required property and no
flag states *no conformance criteria*, so structurally nothing distinguishes an instance of it from
any other base expression. That isn't really a class in the duck-typed sense — it's the overlay you
describe. It shows up concretely in the `subjectClassOf` PR later in this stack, where
`most_specific` ranks candidates by required-triple count: a zero-requirement class loses every tie,
but still matches bare nodes, which is not obviously what anyone wants.

I don't think this PR is the place to decide it (it's a modelling question about SDNA, not about
relation writes), but it's worth an issue. The options I see are: refuse at `save()` when a class
declares no criteria, emit an implicit type flag for constructor-less classes, or make "overlay" an
explicit `@Model` option so the intent is declared rather than inferred from an absent flag. My
instinct is the third — the case is legitimate, it just shouldn't be reachable by accident.

**One thing I got wrong that's worth fixing here:** the fixture was named `TestPlacement`, which
implies WE has a class shaped like this. It doesn't — WE's actual `Placement` carries
`@Flag({ through: 'we://flag', value: 'we://placement' })` and goes through `createSubject` like
everything else. Renamed to `TestNoConstructor`, and the comment now says it's branch coverage for
pre-existing behaviour rather than a pattern being endorsed — 8bd3501, pushed as a separate commit
on top rather than an amend so your review anchors survive. dev is merged in and
`jest src/model src/shacl` is green at 297.

lal-bot-coasys (approved):
Read it twice, dug into the referenced call sites, and re-ran the tiebreak logic against the scalar-property branch. This lands well.

## What I like

The four fixes are one complaint framed correctly. The Rust dedup collapses on first sighting rather than sorting-then-dedup, which preserves `values`'s ascending order for free and keeps `createdAt` semantics stable across the collection. The LWW using strict `>` really does mirror the scalar-property branch a few lines above — verified against `hydration.rs`. And the reasoning for keeping the dedup *outside* `distinct` (a `$count` projection already answers `COUNT(DISTINCT ?t)`) is exactly right: a list disagreeing with its own count within one query is not a defensible default.

`@HasOne` accessors forwarding `batchId` is the smallest of the four but probably the most subscriber-visible one — the "record exists, points at nothing" frame is a real UX papercut in Flux today.

The `Ad4mModel.innerUpdate` coercion — pulling the relation branch *outside* the `setProperties` gate — is the subtlest and I stared at it longest. Your comment sells it: `create_subject` drops the value because relations carry `ad4m://adder` not `ad4m://setter`, and leaving it inside the gate would mean both paths silently drop. Coercion feels like a smell in isolation but is the right call given the alternative is a record whose relation is empty and nothing anywhere saying so.

## One thing worth raising

CodeRabbit flagged this and I think it's genuine: `interpretationHint` on relations gets registered by the decorators (`decorators.ts:364,565,891`) and used by `propose_link_child`, but `Ad4mModel.ts:374` never copies it into `relationsMetadata[relationName]`, so it never reaches SHACL generation on the relation side. It's exactly the same shape as the four bugs this PR is titled after — a silent drop between "user declared it" and "system honors it".

It predates this PR (it landed in `cb453a4db`), so it's not strictly your regression. But given the theme and given you're already touching those decorator hunks, it's a two-line fix in `Ad4mModel.ts` and a regression test, and it would keep the PR internally consistent with its own title. I'll leave the call to you — happy either way as long as it's tracked.

## Nitpick, non-blocking

The `RelationMetadata` interface in `core/src/model/types.ts:439-459` doesn't declare `kind`, `maxCount`, `filter`, or `where` — the projection at `Ad4mModel.ts:374` copies them anyway. Not this PR's job, but it's the reason the `interpretationHint` gap didn't get caught by a type error. Worth an issue if there isn't one already.

## Test plan

- Reviewed both Rust hunks against `hydration.rs` and `projection.rs` in full — the dedup happens before literal decoding, which is what the new `test_hydrate_collection_dedup_survives_literal_decoding` pins.
- Verified the tie test (`test_hydrate_scalar_relation_tie_matches_scalar_property`) actually exercises both branches with the same fixture.
- Verified CI: `build-and-test`, all four `integration-tests-*` jobs, `plugin-ad4m-tests`, CodeRabbit — all green on 8bd3501.

Approving. Nice write-up in the PR body — read like a spec, not a diary.

lucksus (approved):

lal-bot-coasys:
Retracting the `interpretationHint` concern from my review above. I traced the wrong path.

The runtime flow is `@Model` → `generateSHACL()` → `buildSHACL(..., getRelationsMetadata(target), ...)` at `decorators.ts:645-653`. That passes the *raw* registry (which fully preserves `interpretationHint`), not the projected `RelationMetadata` from `Ad4mModel.getModelMetadata()`. `buildSHACL` reads `relMeta.interpretationHint` at `shacl-gen.ts:323-325`, `SHACLShape.toLinks()` emits `ad4m://interpretation_hint`, and the Rust side picks it up as `ShaclProperty.interpretation_hint`. End to end intact.

Nico pointed out there are tests asserting this, and there are: `shacl-roundtrip.test.ts:237-288` exercises exactly this — a `@HasMany` relation with an `interpretationHint`, asserted to appear on the correct property node in the emitted links. Passes on dev.

`Ad4mModel.getModelMetadata()` is a separate public introspection API. Its projection dropping `interpretationHint` (and, incidentally, `kind`/`maxCount`/`filter`/`where`) is a completeness gap in that introspection surface, not a runtime bug. Nice-to-have for tooling, not something to block this PR on.

Sorry for the noise, James. Approval stands, and the rest of the review is unchanged.

PR #922, opened 2026-08-25
Feature: A getter relation can name its target class

jhweir:
# A getter relation can name its target class

> [!IMPORTANT]
> **Branch** `feat/getter-relation-target-class` · **Base** `dev`.
> Independent. Nothing in the series is required before it.
>
> Part of an eight-PR series on the AD4M model layer — review wave 2 of 4. The full shape:

```
     ├─ 2  getter+target
```

## Summary

`resolveRelationArgs` refused `getter` alongside `target`, and alongside `where`. Every layer
beneath already supported both pairings — SHACL emission, shape parsing, getter evaluation, include
resolution — and only the decorator said no. This deletes the two guards.

The pairing matters because a custom getter exists to express a traversal the link-shaped DSL
cannot, and the case WE is walking into is a **reified edge**: a connection stored as a record rather
than a predicate, so that it can carry a label, comments and signals. There is no single
predicate for `through` to name, because the edge is two hops. Without `target` such a relation has
no target class, `include` resolves a shape named `""` and the query errors — so a traversal could
only ever hand back bare URIs, and the caller had to re-fetch each one to learn anything about it.

## Changes

**`core/src/model/decorators.ts`** — removes the `getter`/`target` and `getter`/`where` throws, and
documents why the pairing is coherent.

`target` does two separable jobs: derive a conformance filter, and name the class the relation's
values hydrate into. Only the first competes with an explicit getter — and `buildSHACL` already
resolves that competition on its own. Its getter slot is a priority chain (`relMeta.getter` first,
auto-derived conformance second), while `sh:class` and `ad4m:targetClassName` are emitted from
`relMeta.target` in a separate branch that never consulted the getter. The generator was written for
this; the guard upstream meant it could not be reached.

`where` is the same story from the other end: the executor applies post-getter where filters
*specifically* to getter-backed relations (`apply_where_filter_to_relation`, called over
`getter_props`), and emitting `wherePredicates` needs `relMeta.target` to resolve the target's
metadata. The runtime is built for getter + target + where together.

`getter` and `through` stay mutually exclusive. That one is real — a getter replaces link-based
resolution, so there is no predicate to add to or remove from, and the relation is read-only.

**`core/src/model/relation-filtering.test.ts`** — the two tests that pinned the old behaviour now
pin the new: the explicit getter survives verbatim, `targetClassName` is emitted beside it, and a
`where` on a getter relation reaches `whereFilter`.

**`rust-executor/src/perspectives/model_query/integration_tests.rs`** — a round-trip test building
the shape this is for: two nodes joined by an edge record, asserting the far end arrives **hydrated**
(`text == "Node B"`) rather than as a URI. Self-referential (a node relates to nodes), which also
covers the target shape resolving to the shape being queried.

## Known follow-ups

- This is a **model-declaration** escape hatch: a getter is fixed per class, so "edges of kind X"
  still needs the query layer. The relation-quantifier PR later in this stack covers that.
- Nothing here changes `getter` + `through`, which remains refused. If a relation ever wants to be
  *read* by traversal and *written* by predicate, that is a separate design question.

## What this unblocks in WE

A `WeNode` can declare a `related` relation defined by a two-hop SPARQL traversal through
`Relationship`, and `include` it — getting typed instances rather than URIs, in one query. Not
adopted here; WE is untouched until the stack merges.

## Test plan

> tree above.

- [x] `cargo test -p ad4m-executor --lib perspectives::model_query` — 215 passing, 1 new
- [x] New round-trip test asserts the traversal's far end is a hydrated instance, not a URI
- [x] `jest src/model` in `core` — 247 passing
- [x] Confirmed the order the executor relies on: getters are evaluated at `query.rs:383`, includes
      resolve at `:395`, so a getter relation's URIs are populated before includes hydrate them
- [x] `tsc --noEmit` clean

## Summary by CodeRabbit

* **New Features**
  * Getter-backed relations can now specify a target model and apply filtering after retrieval.
  * Related records can be hydrated as typed model instances when traversing configured relations.

* **Bug Fixes**
  * Preserved custom getter behavior with target-model metadata and post-retrieval filters.
  * Continued to prevent incompatible combinations with through-relations.
  * Invalid filtering combinations are now consistently rejected for getter-backed relations.

### DIFF
--- a/core/src/model/decorators.ts
+++ b/core/src/model/decorators.ts
@@ -815,7 +815,7 @@ export interface RelationOptions {
      */
     through?: string;
     /** The target model class (use a thunk to avoid circular-dependency issues). Optional for untyped string relations.
-     *  Cannot be combined with `getter`. */
+     *  Combines with `getter`, where it names the class the traversal's values hydrate into. */
     target?: () => Ad4mModelLike;
     /**
      * Custom getter to resolve the relation values. Use this for custom graph traversals.
@@ -863,7 +863,8 @@ export interface RelationOptions {
      * auto-derived from the target shape (flags + required properties).
      * Providing `where` overrides this auto-derivation.
      *
-     * Mutually exclusive with `getter` and `filter: false`.
+     * Mutually exclusive with `filter: false`. Combines with `getter`, where it
+     * is applied as a post-getter filter against the target class.
      *
      * @example
      * ```typescript
@@ -942,7 +943,16 @@ function resolveRelationArgs(
         ? { ...(second || {}), target: first }
         : first;
 
-    // getter is mutually exclusive with through, target, and where
+    // `where` and `filter: false` contradict each other on every relation,
+    // getter-backed ones included — checked before the getter path returns.
+    if (opts.where && opts.filter === false) {
+        throw new Error(
+            'Relation decorator: `where` and `filter: false` are contradictory. ' +
+            '`where` adds filtering constraints; `filter: false` disables filtering.'
+        );
+    }
+
+    // getter is mutually exclusive with through
     if (opts.getter) {
         if (opts.through) {
             throw new Error(
@@ -989,16 +999,6 @@ function resolveRelationArgs(
     // Default predicate when not provided
     if (!opts.through) {
         opts.through = 'ad4m://has_child';
-    }
-
-    // where validation
-    if (opts.where) {
-        if (opts.filter === false) {
-            throw new Error(
-                'Relation decorator: `where` and `filter: false` are contradictory. ' +
-                '`where` adds filtering constraints; `filter: false` disables filtering.'
-            );
-        }
     }
 
     return opts;
--- a/core/src/model/decorators.ts
+++ b/core/src/model/decorators.ts
@@ -813,8 +813,27 @@ export interface RelationOptions {
      * The expression can reference 'Base' which will be replaced with the instance's base expression.
      * Example: "SELECT ?target WHERE { ?target <flux://has_reply> <Base> . }"
      *
-     * Mutually exclusive with `through` and `target`. When `getter` is provided the
-     * relation is read-only (no adder/remover actions are generated).
+     * Mutually exclusive with `through` — a getter replaces link-based resolution,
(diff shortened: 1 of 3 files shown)


lucksus (dismissed):

lal-bot-coasys (commented):
Read the change end-to-end, verified the two claims in the PR body against the runtime.

**Claim 1 — `buildSHACL` resolves the `getter` vs `target`-auto-getter competition on its own.** Confirmed at `shacl-gen.ts:356-403`: explicit `getter` wins the getter slot verbatim, and the `sh:class` / `targetClassName` emission at `:408-422` is a **separate `if`** (not `else`), so it runs independently of the getter branch. The generator was written for this pairing; the decorator was the only layer refusing it.

**Claim 2 — the executor already applies where-filters to getter-backed relations.** Confirmed at `getters.rs:349-354`: iterates `getter_props`, and for each prop with a `whereFilter` calls `apply_where_filter_to_relation`. Runtime is built for `getter + target + where` together, exactly as the removed guard's error message denied.

The integration test at `integration_tests.rs:5385-5498` is the right shape: reified edge, two-hop getter, self-referential target class, asserts the far end hydrates with `text == "Node B"` rather than as a bare URI. Self-referential also implicitly covers the target-class cache resolving to the shape currently being built.

**One nice-to-have, not blocking:** an end-to-end test for `getter + where` specifically. The TS test in `relation-filtering.test.ts` pins the SHACL emission (whereFilter appears alongside the getter), but nothing exercises the full path — a getter relation with a `where` clause that actually narrows results at runtime via `apply_where_filter_to_relation`. Given claim 2 relies on that path, an integration-level pin would be worth having before something in the executor decides to short-circuit that call for getter props. Happy to write it as a follow-up if you'd rather keep this PR small.

Nothing else. Deleting two throws that stood in the way of a coherent runtime path, with the reasoning captured directly at the deletion site — this is the shape a good "remove code" PR should have.

Nico's approval already there, CI fully green — leaving this as a COMMENT review rather than a second approval.

lucksus (approved):

PR #923, opened 2026-08-25
Feature: Relation quantifiers and reverse scope

jhweir:
# Filter by linked records with `some` / `none` quantifiers

> [!IMPORTANT]
> **Branch** `feat/relation-quantifiers-and-reverse-scope` · **Base** the where-clause compiler PR (`refactor/where-clause-compiler`), not `dev`.
> **Real dependency.** The quantifier compiles inside `compile_leaf_condition` and threads a resolver through `compile_where_clause` — both introduced by the base PR. ~50 references; this does not build without it.
>
> Part of an eight-PR series on the AD4M model layer — review wave 2 of 4. The full shape:

```
     ├─ 2  getter+target
```

## Summary

`where` could ask about a relation's *targets by id* — `{ comments: [id1, id2] }` — and nothing else
about them. `include`'s own `where` filters the children a query returns, never the parents it
returns them for. So "posts with no comments", or "nodes with an incoming Relationship of kind X",
were not expressible at all: the caller fetched everything with its children and counted in JS.

```ts
where: { comments: { none: {} } }                 // has none at all
where: { comments: { some: {} } }                 // has at least one
where: { comments: { some: { body: "spam" } } }   // has one that matches
```

This is the operator the where-compiler PR before it made cheap to add, and the one a knowledge-graph
template wants most now that `RelationshipType` gives edge kinds an id to filter on.

## Changes

**`rust-executor/src/perspectives/model_query/types.rs`** — `some` / `none` on `WhereOps`, each an
optional nested clause. `WhereOps` uses `deny_unknown_fields`, so adding the keys is what makes
`{ some: {…} }` parse as an operator object rather than falling through to a nested where clause.

**`rust-executor/src/perspectives/model_query/sparql_builder.rs`** — compiles to
`FILTER EXISTS { ?source <pred> ?t . … }`, or `FILTER NOT EXISTS` for `none`. A reverse relation puts
the linked record on the subject side, since a `@BelongsTo` target points *at* this instance rather
than away from it.

The nested clause names properties of the **target** class, so it is compiled against that class's
shape and the result rebased onto the variable the relation triple binds. That is why the compiler now
takes a `ShapeResolver`: with only the parent shape in hand there is no way to turn `body` into
`we://body`. It is optional — an empty nested clause is pure existence and needs no target shape, so
"has any" / "has none" compiles without one, which keeps every existing test call site working with
`None`.

`rebase_into_subquery` renames every generated variable before nesting. `FILTER EXISTS` shares the
enclosing variable scope, so an inner clause constraining the same property as the outer one would
otherwise reuse the same helper variable and silently correlate the two — the inner condition would
end up constraining the outer row rather than the linked one. Nesting stays safe because each pass
prefixes whatever the previous produced, so depth keeps names distinct with no counter to thread.

A quantifier that cannot compile in full is not pushed at all, for the same reason as `OR`: a partial
`EXISTS` matches rows the complete clause rejects, and a partial `NOT EXISTS` rejects rows it should
keep.

**`rust-executor/src/perspectives/model_query/filtering.rs`** — `matches_ops` fails closed on a
quantifier. Unlike `OR`, the fallback genuinely *cannot* answer this one: it sees a value already on
the instance, not the store, so it has no way to look up linked records. It now rejects the row and
logs, rather than ignoring the condition and returning rows that do not satisfy the query. This is
the one operator where the two layers cannot both implement the semantics, and saying so explicitly
is better than the silent-pass alternative the compile-once refactor exists to prevent.

**`query.rs`** — threads the resolver through to the three builder call sites.

## Reverse `scope`: considered and dropped

The plan paired this with a reverse form of `scope` (`?source <pred> <anchor>`, for drilling into a
relation that points the other way). On inspection it is redundant: that query is already
`where: { <relation>: anchorId }` on the child's own forward relation, which compiles to exactly the
same triple and has since before this stack. A second spelling would be surface area without
capability. The branch keeps its name; the scope half is intentionally empty.

## Known follow-ups

- Aggregate quantifiers (`atLeast: 3`) are not covered. `FILTER EXISTS` cannot count; that wants a
  `GROUP BY` sub-select, which is a different shape.
- The fail-closed path returns an empty result rather than an error. Correct in the sense that no row
  is wrongly included, but a caller hitting it sees "no matches" rather than "I could not answer
  that". Worth revisiting if it shows up in practice.

## Test plan

> tree above.

- [x] `cargo test -p ad4m-executor --lib perspectives::model_query` — 235 passing (227 → 235)
- [x] 7 unit tests: `some` with empty clause, `none` with empty clause, nested clause constraining the
      *linked* record (asserting the pattern is not left on `?source`), declining without a resolver,
      declining when the nested clause cannot compile, reverse relation reversing the triple, `some`
      and `none` together refused
- [x] End-to-end test against a real store with three posts, two comments and a two-class resolver:
      `none: {}` → the post with no comments, `some: {}` → the two with comments,
      `some: { body: "spam" }` → the one matching post

## Summary by CodeRabbit

- **New Features**
  - Added filtering for related records that meet conditions (“some”) or where no related records meet conditions (“none”).
  - Added nested filters on linked properties.
  - Relation filters now verify that linked records belong to the expected target type.
  - Improved support for complex and combined relation filters.

- **Bug Fixes**
  - Unsupported relation quantifiers are now safely rejected instead of silently ignored.
  - Prevented incorrect matches from conflicting conditions in nested relation filters.


rust-executor/src/perspectives/model_query/sparql_builder.rs
```diff
@@ -1389,14 +1431,303 @@ mod where_compiler_tests {
...
+/// this rebases `?source` onto does too, so `?source` must be substituted second
+/// or it would be rewritten twice. Nesting is safe because each pass prefixes
+/// whatever the previous pass produced, so depth keeps the names distinct
+/// without threading a counter.
+fn rebase_into_subquery(patterns: &[String], namespace: &str, target_var: &str) -> Vec<String> {
+    let prefixed = format!("?_{namespace}");
+    patterns
+        .iter()
+        .map(|p| p.replace("?_", &prefixed).replace("?source", target_var))
+        .collect()
```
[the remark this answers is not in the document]
jhweir: The finding was valid, and fixed.

What was actually wrong
[sparql_builder.rs:1455](vscode-webview://0dgj68cqksheqdhuu216btu6o95d3lusoog9q9r46j0obi0ti0hd/rust-executor/src/perspectives/model_query/sparql_builder.rs#L1455) renamed variables with two blind str::replace calls over the whole pattern string. The patterns it operates on are compiled where-clauses, and those embed user values as quoted literals — e.g. FILTER(STR(?_ft_x) = "spam"). So a nested clause like { comments: { some: { body: "?source" } } } had the value rewritten inside its own quotes to the linked record's variable name. The query then searched for a string nobody wrote and returned no rows, with nothing to indicate the question had been changed. Same for an IRI carrying ?source in a query component — validate_iri permits ?.

Not an injection: escape_sparql_string still neutralises quotes and backslashes, so the SPARQL stays well-formed. It's silent wrong-answers, and only reachable through the relation-quantifier path this PR adds.

The fix
rebase_pattern scans for variable tokens, copying "…" literals through verbatim (honouring the \" escapes, so an escaped quote doesn't read as the literal's end) and <…> IRIs likewise.

It also collapses two passes into one. The old doc comment warned that ?source had to be substituted second, because the target variable it becomes (?_{ns}t) itself starts with ?_ and would otherwise be rewritten twice. Matching by token name rather than by substring makes that ordering rule stop being load-bearing. Generated variable names are byte-identical to before, and nesting still namespaces by depth — the existing 11 quantifier tests pass unchanged.

Verification
Three new tests: the literal value, an IRI query component, an escaped quote inside a literal.
Non-vacuous — reverted to the old two-replace implementation and watched all three fail ("?_q0commentst=1" inside the IRI, \"a\\\"?_q0t\" inside the literal).
perspectives::model_query 259 passed · cargo fmt --check clean · clippy has no new findings on the file (the two hits at lines 678/786 predate this).
Committed as abd441f3c

jhweir:
@coderabbitai full review

rust-executor/src/perspectives/model_query/sparql_builder.rs
```diff
@@ -1389,14 +1431,438 @@ mod where_compiler_tests {
...
+            .include_relations
+            .iter()
+            .find(|r| r.name == prop_name)
+            .map(|r| r.target_class_name.as_str())
+            .filter(|n| !n.is_empty())?;
+        let target_shape = resolver?.get_shape(target_class).ok()?;
+
+        let compiled = compile_where_clause(inner, target_shape.as_ref(), resolver);
+        if !compiled.complete || compiled.patterns.is_empty() {
+            return None;
```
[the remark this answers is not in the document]
jhweir: Valid finding, fixed — and it turned out to have a second half the review didn't mention.

The bug
compile_relation_quantifier emitted FILTER EXISTS { ?source <we://comment> ?t . <nested> } — "does this record link to anything with the right properties". But hydration resolves comments by running Comment's own model query, which applies that class's conformance patterns ([relations.rs:199](vscode-webview://0dgj68cqksheqdhuu216btu6o95d3lusoog9q9r46j0obi0ti0hd/rust-executor/src/perspectives/model_query/relations.rs#L199)). So the filter and the row disagreed about the same link: a Post could match comments: { some: { body: "spam" } } and come back with comments empty. none had the mirror — excluding a record because of a node its relation would never show.

Post-hydration can't rescue it: matches_ops fails closed on a quantifier, so the pushdown is the whole answer.

What I did
Split shape_conformance_patterns out of build_query_patterns unchanged, and the quantifier now prepends it rebased onto the linked variable. Two things the split forced:

Helper variables renamed ?cf_ → ?_cf_. This is the part the finding didn't flag, and it would have been a fresh bug. Rebasing namespaces variables beginning ?_; conformance vars didn't. A target class with a required property named like the outer class's would bind ?cf_title in both scopes — silently requiring the Post and the Comment to share a title. build_timestamp_probe deliberately reuses that variable to correlate its reified triple with the conformance triple, so it's renamed in step.

The structural fallback is declined inside an EXISTS. It's a SELECT DISTINCT over every node carrying any of the class's predicates, and SPARQL evaluates a subquery independently of its enclosing group — a whole-store scan per quantifier, to exclude what the link triple already excluded. build_query_patterns keeps it, and now also declines it when a parent scope has narrowed the subject, which is what its old conformance_patterns.is_empty() guard meant.

I also extended it to some: {} / none: {} ("has any comment") when the shape resolves — same reasoning, and it keeps the two forms from meaning different things.

Verification
Four tests: an end-to-end query over a real store holding a genuine Comment and an imposter with the same body and no flag; the SPARQL asserting the flag on the linked variable; the helper namespacing; the absence of a structural scan.
Non-vacuous — disabled just the conformance block and all four failed.
The e2e test found a real trap first: literal:string:X bases come back from the store decoded when they're link targets, so my initial fixture matched nothing at all. Switched to IRI bases, which is what real relation targets are.
perspectives::model_query 263 passed. Across all of perspectives::: 24 failures, against 25 on the same tree without my change — the pre-existing LLM-backed interpretation_e2e / interpretation_harness_e2e suite. Set-diffed the two runs: no new failures.
cargo fmt --check clean, no new clippy findings.
Committed as 7556d92fa

rust-executor/src/perspectives/model_query/filtering.rs
```diff
@@ -185,6 +185,26 @@ pub(super) fn matches_condition(val: &Value, condition: &WhereCondition) -> bool
...
+    // they are compiled to `FILTER [NOT] EXISTS` and never evaluated here.
+    //
+    // Reaching this point means the compiler declined to push one down (no
+    // resolver, an unknown target class, or a nested clause that would not
+    // compile in full) and the clause fell back to post-hydration filtering.
+    // Fail closed and say so: ignoring the condition would return rows that do
+    // not satisfy the query, which is the failure mode this whole path exists
+    // to avoid.
+    if ops.some.is_some() || ops.none.is_some() {
+        log::warn!(
```
lal-bot-coasys: **Log-spam risk.** `matches_ops` is called per hydrated instance from `matches_where` via the `retain` closure at `query.rs:329-331`. A query that returns 10k rows and hits this fail-closed path will emit 10k identical warnings. The message itself is a compile-time diagnostic — the actionable info (missing resolver / unknown target class / nested clause failed) is at the *decline* site, not here.

Options:
1. Move the warning to `compile_relation_quantifier` at each of its three `return None` paths, with a more specific message per path (e.g. "no shape resolver available", "target class `Comment` not found in resolver", "nested clause could not compile").
2. Keep it here but gate on a `once_cell` / atomic to log once per process.

Option 1 is stronger — the compile site already knows *why* it declined, and the runtime just has to reject the row. If you keep it here, at minimum tag it with a stable key so log aggregators can dedupe.

rust-executor/src/perspectives/model_query/sparql_builder.rs
```diff
@@ -732,6 +775,33 @@ fn compile_leaf_condition(
                     }
                 }
             }
+            WhereCondition::Ops(ops) if ops.some.is_some() || ops.none.is_some() => {
```
lal-bot-coasys, line 790: **Scope question worth documenting.** This branch fires only inside the collection-relation `if let Some(prop) = ... && p.is_collection` guard at line 706. A scalar relation (`@HasOne` / `@BelongsToOne`) written as `where: { author: { none: {} } }` — a genuinely useful query, meaning "has no author assigned" — silently falls through to the property-based arm below, which doesn't know about `some`/`none` and returns `None` via the "no filter produced" gate at line 1058. From the user's perspective: no parse error, no diagnostic, always zero rows.

My read is this is intentional (the PR body specifically frames it around collection relations + `RelationshipType`), but there's no explicit "scalar relations out of scope" comment. Either add one, or emit a distinct decline diagnostic on that path so a future user can tell "scalar not supported" from "unknown target class". Not blocking — just a footgun waiting for someone to write `{ author: { none: {} } }` and stare at empty results.

rust-executor/src/perspectives/model_query/sparql_builder.rs
```diff
@@ -1389,14 +1459,566 @@ mod where_compiler_tests {
...
+fn compile_relation_quantifier(
+    shape: &ModelShape,
+    prop: &super::types::ShapeProperty,
+    prop_name: &str,
+    inner: &BTreeMap<String, WhereCondition>,
+    negate: bool,
+    resolver: Option<&dyn ShapeResolver>,
+    leaf_id: usize,
+) -> Option<String> {
+    let safe_pred = validate_iri(&prop.predicate).ok()?;
```
lal-bot-coasys: **Predicate-validation asymmetry, small.** This validates `prop.predicate` before emitting the link triple. The sibling String/StringArray arms in the same collection-relation branch (lines 711, 741) also validate — but only the *value*, not `prop.predicate` itself. The property-based branch at line 816 validates the predicate too. So we have a mix: predicates from SDNA are validated in the property arm and this quantifier arm, but trusted in the String/StringArray collection-relation arms. All paths are ultimately safe (SDNA is the source of predicates), but the asymmetry is worth normalising — validating `prop.predicate` once at the branch entry (line 706 area) and stashing the safe form would drop three call sites into one. Post-merge cleanup, not this PR.

rust-executor/src/perspectives/model_query/sparql_builder.rs
```diff
@@ -1389,14 +1459,566 @@ mod where_compiler_tests {
...
+    let link = if prop.direction.as_deref() == Some("reverse") {
+        format!("        {target_var} <{safe_pred}> ?source .")
+    } else {
+        format!("        ?source <{safe_pred}> {target_var} .")
+    };
+
+    let mut body = vec![link];
+
+    let target_class = shape
+        .include_relations
```
lal-bot-coasys, line 1639: **Target-class discovery via `include_relations`, worth noting.** This walks `shape.include_relations` to find the relation's declared target class. That means the quantifier only compiles for relations the caller has previously named in `include` on this query. A well-typed `@HasMany({ target: () => Comment })` that isn't in `include` on this call still has target metadata upstream (on the `RelationMetadata`), but if it's not in `include_relations` here, `target_class` is `None`, `target_shape` is `None`, and a nested `some: { body: "spam" }` declines. Empty `some: {}` still works (no target needed).

User-visible effect: `where: { comments: { some: { body: "spam" } } }` *without* `include: { comments: true }` on the same query silently returns zero rows. Correct or not depends on the design call — my instinct is the shape itself already knows the target class independent of `include`, so the compiler could look there first and fall back to `include_relations`. But if the intent is "quantifiers only work when hydration would resolve the same class", say so in a comment above this lookup — save the next reader a trace.

lal-bot-coasys (dismissed):
Read the compiler, the rebase logic, `compile_relation_quantifier`, and every new test. This is well-shaped: `some`/`none` as `FILTER [NOT] EXISTS`, all-or-nothing pushdown for the same reason as `OR`, target-class conformance rebased onto the linked record, and `matches_ops` failing closed. The two CodeRabbit findings (variable-substitution inside quoted literals/IRIs, target-class conformance in the EXISTS body) are addressed cleanly and pinned by tests (`the_linked_record_must_conform_to_the_target_class`, `conformance_helpers_are_namespaced_like_every_other_helper`). CI green, 235 passing.

Approving. Four observations inline — none blocking. The log-spam one is the most concrete follow-up.

One thing worth naming up here rather than inline: the `rebase_pattern` character-scanner is doing real work (quoted-literal-verbatim, IRI-content-verbatim, `?var` namespaced, `?source` remapped) and it composes correctly under nesting because each rebase pass prefixes what the previous pass produced. That's the elegant version of what a threaded counter would have been. Nice.

lucksus (dismissed):
LGTM and I think @lal-bot-coasys's points make sense as improvements still in this PR

lucksus (approved):

PR #925, opened 2026-08-25
Feature: Polymorphic include

jhweir:
# Polymorphic `include` — hydrate each target as the class it actually is

> [!IMPORTANT]
> **Branch** `feat/polymorphic-include` · **Base** `dev`.
> Its dependency, the `subjectClassesOf` PR (#924), has **merged**, so this now targets `dev` directly. `hydrate_polymorphic` calls `subject_classes_of` to classify a relation's targets before grouping them, taking the most specific class each URI belongs to.
>
> Part of an eight-PR series on the AD4M model layer — review wave 3 of 4. The full shape:

```
dev ─┬─ 1  relation-correctness ── 5  subjectClassesOf (merged) ─┬─ 6  polymorphic-include
     ├─ 2  getter+target
```

## Summary

A heterogeneous relation had no good reading. `CollectionBlock.children` holds text, images, tasks
and collections; a reified edge's endpoints are whatever two records somebody connected. Both
available options were wrong:

- **Declare the base class** and every subclass property vanishes. Not mislabelled — *absent*.
  `hydrate_one` builds its predicate map from the shape it is handed, so a `TextBlock` hydrated as a
  `WeNode` arrives with no `text` and no `tag` at all.
- **Declare nothing** and `include` has no shape to resolve, so the query fails outright.

This classifies the targets, groups them by concrete class, and hydrates each group against its own
shape.

## Why it could not be a TypeScript-side class swap

The original plan proposed choosing the right class in `jsonToModelInstance`, on the grounds that
class instantiation is the only part of hydration still in TypeScript. That line decides the class —
but the data is already gone by then. Hydration against the declared shape discards the subclass's
properties before the JSON is built, so constructing the right class over that JSON produces a
`TextBlock` with empty fields. The work has to happen in Rust, which is what this does.

## Changes

**`rust-executor/src/perspectives/model_query/types.rs`** — `polymorphic` on `ModelQueryInput`, so
it can be set per `include` sub-query. Putting it in the query rather than in SHACL means no
round-trip plumbing was needed to ship it.

**`rust-executor/src/perspectives/model_query/relations.rs`** — `hydrate_polymorphic` resolves the
target URIs with `subjectClassesOf`, groups them by concrete class, and runs one sub-query per group.
The cost is one query per *distinct class present*, not per instance — two classes in a thousand-item
collection is two queries. Each instance comes back carrying `__subjectClass`; the executor had to
know it in order to hydrate at all, so returning it costs nothing.

Three edges worth naming. A target matching no registered class is **skipped** rather than hydrated
as something it is not — the same outcome a non-conforming target has always had. `limit`/`offset`
are dropped inside the per-class sub-queries, because a limit applied per group would silently mean
something different from what it means on a single-class include; the parent's link order decides
what survives instead. `order` is honoured within each class but not across them, since sorting a
polymorphic set by a property the classes need not share is not a well-defined request.

The untyped-include failure also gets a message. It used to surface as a shape lookup for the empty
string — naming neither the relation nor the fix. It now names both.

**`core/src/model/decorators.ts`** — `polymorphic` and `classResolver` on `RelationOptions`.
`classResolver` is taken as a function rather than resolved through a global registry, for the same
reason `fromSHACL` does: the caller knows its own classes, and process-wide mutable state is awkward
across several perspectives and outright wrong across a hot reload.

**`core/src/model/Ad4mModel.ts`** — two halves. `prepareModelQueryParams` injects `polymorphic: true`
into the include sub-query when the relation declares it, so `include: { children: true }` reads
polymorphically without every call site repeating it — the relation being heterogeneous is a fact
about the data, not about one query. An explicit `polymorphic: false` at the call site still wins.
`jsonToModelInstance` reads `__subjectClass` through the resolver, and the `meta?.target` gate is
relaxed, since a polymorphic relation may legitimately declare no target at all — that is the case it
exists for. Where no class can be resolved the item stays plain JSON carrying its class name, rather
than being forced through a constructor that does not exist.

## Known follow-ups

- `polymorphic` is not yet declarable in SHACL, so a non-ORM caller (MCP, raw GraphQL) must pass it
  in the query. Adding it to the shape would follow the path PR #906 walked for
  `RelationOptions.datatype`.
- Cross-class `order` is unimplemented by design. If a use case appears, it would need a well-defined
  answer for what sorting by a property only some classes have should mean.

## What this unblocks in WE

`include: { children: true }` on a `CollectionBlock` stops erroring and returns correctly-typed
blocks, retiring the hand-written tree walk that resolves each child's type itself. A reified edge's
endpoints become hydratable, so
a board can fetch edges and their endpoints together. Not adopted here; WE is untouched until the
stack merges.

## Test plan

> tree above.

- [x] `cargo test -p ad4m-executor --lib perspectives::model_query` — 225 passing
- [x] End-to-end test: a collection whose `children` names **no** target class, holding a text block
      and an image block, asserting each arrives with `__subjectClass` set *and* with the property
      that exists only on its concrete class (`text` / `src`) — the thing a base-class hydration
      silently drops
- [x] End-to-end test that an untyped include without `polymorphic` errors with a message naming both
      the relation and the fix
- [x] `jest src/model` in `core` — 256 passing, 4 new: children constructed as the right classes, data
      preserved without a `classResolver`, the declaration reaching the executor without the call site
      repeating it, and an explicit `false` winning
- [x] `tsc --noEmit` clean

## Summary by CodeRabbit

- **New Features**
  - Added support for polymorphic relations containing different model types in a single collection.
  - Related items can be restored as their specific concrete types, including nested relations.
  - Added per-query control for polymorphic handling.
  - Items matching multiple model types retain complete classification information.

- **Bug Fixes**
  - Improved diagnostics when polymorphic handling is omitted.
  - Preserved concrete-type properties for heterogeneous related items.
  - Retained unrecognized related items as untyped data.
  - Polymorphic queries now clearly reject unsupported limits and offsets.


jhweir:
@coderabbitai full review

jhweir:
@coderabbitai full review

jhweir:
@coderabbitai full review

core/src/model/Ad4mModel.test.ts
```diff
@@ -2648,3 +2648,143 @@ describe("Relation writes: to-one batching and scalar coercion", () => {
...
+  };
+
+  @Model({ name: "PolyCollection" })
+  class PolyCollection extends Ad4mModel {
+    // No target class — the members are of genuinely different types, which is
+    // the situation `polymorphic` exists for.
+    @HasMany({
+      through: "we://children",
+      polymorphic: true,
+      classResolver: (name: string) => blockModels[name],
```
lucksus: First, I wondered why we need this. I believe the reason is that we want Ad4mModel to instantiate TS/JS classes and we need to provide those classes. We wont get that TS code from the executor.

But I still wonder why this is a function of the name. Could it just be a list of those types and we either get the name (if that's needed at all) from the class, or we just do property checks on the returned model query results to match those classes?

rust-executor/src/perspectives/model_query/relations.rs
```diff
@@ -146,6 +146,105 @@ pub(super) async fn resolve_includes_recursive(
...
+                    obj.insert(
+                        SUBJECT_CLASS_KEY.to_string(),
+                        Value::String(class_name.clone()),
+                    );
+                }
+                ordered_ids.push(id.clone());
+                hydrated.insert(id, inst);
+            }
+        }
+    }
```
lucksus, line 372: Hm, in the case of multiple different classes being instantiated around the same baseExpression, this would return all hydrations of all instances. I think that would be right here and should be filtered somewhere else (client-side?).

I think we should have a test with that case (multiple different class instances on the same base expression) and see that the whole pipeline works as expected.

lucksus (changes requested):
Some questions. And def. think there should be more tests for this. The mentioned case of multiple class instances on the same base expression. And what's missing completely is JS integration tests that drive the whole pipeline including Ad4mModel client-side.

rust-executor/src/perspectives/model_query/relations.rs
```diff
@@ -146,6 +146,105 @@ pub(super) async fn resolve_includes_recursive(
...
+        wc.insert("id".to_string(), WhereCondition::StringArray(ids));
+        group_query.where_clause = Some(wc);
+        // A limit is per-class here; applying the parent's limit to each group
+        // would silently mean something different from what it means on a
+        // single-class include, so it is dropped and the caller's link order
+        // decides what survives.
+        group_query.limit = None;
+        group_query.offset = None;
+
+        let result = Box::pin(execute_model_query_inner(
```
data-bot-coasys, line 340: **Concern (non-blocking):** dropping `limit` / `offset` silently on the sub-query means a caller writing `include: { children: { polymorphic: true, limit: 5 } }` gets every result with no signal. The comment explains why a per-class limit would be semantically weird, but users are going to try it. Consider one of:

- reject `Some(_)` here with `anyhow!("limit/offset are not supported on a polymorphic include; parent link order decides what survives")`, so callers hear about it once and move on;
- or `log::warn!` when they were set and got dropped.

Either is more honest than silence.

rust-executor/src/perspectives/model_query/relations.rs
```diff
@@ -146,6 +146,105 @@ pub(super) async fn resolve_includes_recursive(
...
+        if let Some(class_name) = classes.get(id).and_then(|names| names.first()) {
+            by_class
+                .entry(class_name.clone())
+                .or_default()
+                .push(id.clone());
+        }
+        // A target matching no registered class is skipped rather than
+        // hydrated as something it is not. It stays absent from the relation,
+        // which is the same outcome a non-conforming target has always had.
+    }
```
data-bot-coasys, line 320: **Nit:** a debug log here would rescue the future person debugging "why is my child missing". Something like:

```rust
log::debug!(
    "polymorphic include: '{id}' matches no registered class, skipping"
);
```

Matches the legacy silent-skip behavior for functional purposes, but leaves a trail when a type-link hasn't propagated yet in a live neighbourhood.

rust-executor/src/perspectives/model_query/relations.rs
```diff
@@ -146,6 +146,105 @@ pub(super) async fn resolve_includes_recursive(
     Ok(())
 }
 
+/// The JSON key carrying each polymorphically-hydrated instance's concrete
+/// class, so the TypeScript layer can construct the right model class for it.
+pub(crate) const SUBJECT_CLASS_KEY: &str = "__subjectClass";
```
data-bot-coasys, line 195: **Nit:** `"__subjectClass"` is a wire-format contract that appears again as an inline `const SUBJECT_CLASS_KEY` in `core/src/model/Ad4mModel.ts`. Drift here would degrade silently to "instance stays plain JSON" — which is exactly the failure mode that's hard to trace. Worth a doc-comment on both sides pointing at the counterpart, or extracting into a place both consumers can reference.

core/src/model/Ad4mModel.ts
```diff
@@ -35,6 +35,65 @@ import type {
...
+    if (!nested || subQuery.polymorphic) continue;
+
+    // The thunk is only evaluated here, at query time, so a self-referential or
+    // circularly-imported target that is not yet defined costs the nested
+    // default rather than the whole query.
+    let TargetClass: any;
+    try {
+      TargetClass = meta.target?.();
+    } catch (e) {
+      console.debug(`prepareModelQueryParams: target class unavailable for include '${relName}':`, e);
```
data-bot-coasys: **Question:** the `catch` here is a good default for circular imports (target-class not yet defined at module-eval time), but it also swallows a genuinely broken thunk — a typo in the imported class name, for instance — with only a debug log. Would you consider distinguishing? Roughly:

```ts
try {
  TargetClass = meta.target?.();
} catch (e) {
  const looksLikeCircularImport = e instanceof ReferenceError;
  const log = looksLikeCircularImport ? console.debug : console.warn;
  log(`prepareModelQueryParams: target class unavailable for include '${relName}':`, e);
  continue;
}
```

Not blocking — happy with a `// TODO` if this feels like scope creep.

data-bot-coasys (dismissed):
## Review of #925 — polymorphic include

Solid design, thorough tests, and the writeup is a pleasure to read. The invariant is stated precisely ("parent link order decides survivors") and the code follows it. Forward + reverse coverage, actionable error messages, and the depth-carrying default in `applyPolymorphicIncludeDefaults` all land the same way I would want them to.

I'm approving — the notes below are nits and a couple of "consider for a follow-up" observations rather than blockers.

**Concerns worth thinking about (non-blocking):**

1. `limit` / `offset` on a polymorphic sub-query are dropped silently. The reasoning is sound (per-class limit means something different), but a caller writing `include: { children: { polymorphic: true, limit: 5 } }` will get all results with no signal. Consider erroring or warning on that combination in a follow-up.
2. A target URI that resolves to no registered class is skipped with no log line. Matches legacy behavior, but in a live neighborhood where a type-link hasn't propagated yet, the child just vanishes from the read. A `log::debug!` at that point would rescue debuggers.
3. `__subjectClass` is a wire-format contract defined twice — once as `SUBJECT_CLASS_KEY` in `relations.rs`, once as an inline const in `Ad4mModel.ts`. A drift there degrades silently to "no class attached, plain JSON", which is exactly the case that's hard to debug.
4. `applyPolymorphicIncludeDefaults` swallows target-thunk throws with `console.debug`. Right default for circular imports, but a genuinely broken thunk also degrades silently — worth considering promoting to `console.warn` when the error doesn't look like `ReferenceError`.

Nice work.

core/src/model/Ad4mModel.test.ts
data-bot-coasys: You're right — the function is a superset (any list can be turned into `(name) => list.find(c => c.className === name)`), but the ergonomic default is worse than it needs to be.

**On the function-vs-list part.** `@Model({ name: "X" })` already writes `target.className = "X"` (decorators.ts:636–637), so the class is self-describing. The user shouldn't have to write `(name) => blockModels[name]` at every call site — the decorator can build that map from a plain list. So:

```typescript
@HasMany({
  through: "we://children",
  polymorphic: true,
  classes: [PolyTextBlock, PolyImageBlock],
})
```

Internally: `classResolver = (name) => classes.find(c => c.className === name)`. Same information, less boilerplate, and the ORM keeps the door open for a `classResolver` escape hatch when someone genuinely needs dynamic dispatch (hot-reload, plugin-loaded classes) — that's the case where James's function form actually earns its keep.

**On property checks instead of names.** I'd push back on that one. Two classes with overlapping required properties become ambiguous the moment they exist — `TextBlock` and `Annotation` both carrying `text` is not a hypothetical, it's exactly the SHACL structural-membership case `subject_classes_of` was written to handle. And it'd drift from what the executor is already doing: the classification step reads `rdf://type <ad4m://SubjectClass>` links (subject_classes_of.rs:54–63) — there IS a canonical name, and it's the same name the SHACL shape is registered under. Building a second, independent "class identity" out of property presence duplicates the source of truth and then races it.

So: yes to a plain `classes: [...]` list as the public API, no to dropping names entirely. Want me to sketch that as a follow-up PR after the stack merges?

rust-executor/src/perspectives/model_query/relations.rs
data-bot-coasys: Reading the code, the current behavior is actually the *opposite* of what you're describing — and I think you've caught a real design issue.

Looking at `hydrate_polymorphic` (relations.rs:189–197):

```rust
for id in target_ids {
    if let Some(class_name) = classes.get(id).and_then(|names| names.first()) {
        by_class.entry(class_name.clone()).or_default().push(id.clone());
    }
}
```

`classes.get(id)` returns the whole set of classes the URI conforms to — most specific first — and `.first()` **picks one**, silently discarding the rest. So a URI that structurally conforms to `[MessagePost, Post]` gets hydrated only as `MessagePost`; the `Post` view is dropped.

Which matters because `subject_classes_of.rs:22–24` is explicit that this is a policy call:

> Membership is therefore genuinely a *set*, and that set is what this module returns. Picking one member out of it is a policy, not a fact, so this module does not pick: it orders the set most specific first and leaves the choice to the caller.

`hydrate_polymorphic` is that caller — and it IS picking. "Most specific wins" is a defensible default, but it's a default the caller can't opt out of, and it doesn't match your description or your expectation.

Agree with your proposal: return all hydrations (one entry per (URI, class) pair, each carrying its `__subjectClass`), let the client filter by which classes it asked about via `classResolver`/`classes`. That composes cleanly with the point above — the ORM already knows which classes are relevant to it, so client-side filtering falls out for free.

**On the missing test.** You're right that it's not covered. `test_polymorphic_include_hydrates_each_child_as_its_own_class` has two children of two different classes but no single URI conforming to multiple. The test that would catch the drop-silent behavior: a URI whose flags satisfy `[TextBlock, Annotation]` (both registered), read via `polymorphic: true`, asserting we get both hydrations back with distinct `__subjectClass` — and that a client requesting only `TextBlock` in its `classes` list ends up with just one.

And the JS-integration gap you flagged in the top-level review body — driving the whole pipeline through `Ad4mModel.parseModelResult` on a real executor response — is where a design mismatch here would actually bite. Right now the Rust tests assert the JSON shape and the TS tests assert construction; nothing exercises both ends against the same server.

core/src/model/Ad4mModel.test.ts
jhweir: Agreed — done in 47753abd, `classResolver` is gone:

```ts
@HasMany({
  through: "we://children",
  polymorphic: true,
  instantiateAs: () => [PolyTextBlock, PolyImageBlock],
})
children: string[] = [];
```

You're right that the name was already available: `@Model({ name })` writes
className onto the constructor, so a name-keyed record beside the classes was
the same list written twice, free to disagree with itself. The map is now derived
from the classes.

It's a thunk rather than a bare array for the reason target is one — it's called
at query time, so a class from further round an import cycle still resolves. An
array literal in a decorator evaluates at module-eval and would reintroduce that.

Two things I kept, and I think both are right, but say if you disagree:

**__subjectClass still comes back on the wire.** Whoever does the property
checks, this side needs to know which class won in order to pick a constructor,
and the executor had to know it to hydrate at all — so returning it is free. The
alternative is matching structurally in TypeScript, which means a second
implementation of `subject_classes_of` that can drift from the Rust one, and it
isn't sound anyway: under a properties projection the client only sees the
fields the query asked for, so a truncated object can match the wrong class or
none.

**The list is advisory and cannot exclude.** A class it doesn't name still
arrives, as plain JSON carrying its concrete class name — same as when no list is
given. Declaring what this call site can construct seemed worth keeping distinct
from declaring what the relation may hold: a heterogeneous relation is open by
definition, and silently dropping members that no local constructor matched would
defeat reading it polymorphically. Hence the name — a verb phrase, so it can't be
read as a permission list. There's a test pinning it.

One question back. When you said property checks — did you mean executor-side? If
so I think what you're reaching for is narrowing a read to particular classes,
which I'd rather express in the query than on the model:

`include: { children: { ofClass: [TextBlock, ImageBlock] } }`

with matching on set membership (so ofClass: [Block] matches an ImageBlock,
i.e. isSubjectInstance) rather than on the specificity winner, and the
classification scope derived from the filter rather than declared separately. It
would also let limit/offset come back, since a scope of one class makes a
limit unambiguous again. That belongs in the where-compiler's vocabulary rather
than bolted onto include, so a separate PR — I've noted it under follow-ups. Say
if you'd rather it landed here.

rust-executor/src/perspectives/model_query/relations.rs
jhweir: Added in ea284b33 — though the premise is slightly off, so let me lay out what
the code does now.

It doesn't return all hydrations today: the grouping takes `names.first()`, so
each target lands in exactly one class group and is hydrated once. And I'd argue
it should stay that way. Links decide how many members a relation has; classes
decide only how each member is read. A node conforming to two classes is one
node with two readings, not two children — hydrating it per matching class would
make a three-link collection arrive with five entries, and every caller would
have to know to collapse them by id.

But the case you're pointing at is real, and more idiomatic than a corner case:
membership is structural, so nothing stops one base expression carrying two
classes' flags. The problem wasn't the cardinality, it was that the readings we
didn't pick vanished without trace. So every instance now carries the whole
ordered set as `__subjectClasses` beside the chosen `__subjectClass`. It costs
nothing — the classification had to produce the set in order to rank it — and a
caller wanting the other reading has the id and can query that class for it
directly.

I also wrote the pick down as the policy it is, since #924 deliberately declined
to make it. Within an inheritance chain it's a real answer: a subclass requires
everything its parent requires and more, so "matched more triples" and "more
derived" are the same ordering. Between two unrelated classes it ties and the
alphabetical tie-break decides — arbitrary, but the same arbitrary answer on
every peer, which seemed the property worth having when there's no `rdf:type`
triple to appeal to.

Two tests, both the pipeline end to end as you asked:

- a subclass and its parent both matching one target — asserts the subclass
  hydrates, its own property is present, and both names come back. This is the
  ranking's intended case and nothing covered it before; the existing test's two
  classes match one shape each, so the head of the set was never contested.
- two unrelated classes on one base expression — asserts one member for the one
  link, a deterministic winner, the winner's property hydrated, and the losing
  reading reported.

Two things I considered and didn't do, in case they're what you had in mind:

Breaking the tie with the client's class list. Tempting, since it doesn't exclude
anything — but the wire result would then depend on which constructors happen to
be in scope, so two peers reading identical data would disagree about what it is.

Hydrating against the union of the matched shapes, so the member carries both
classes' fields. Honest to the structural model and it keeps cardinality, but two
classes can map different predicates onto the same JSON key, and no single
constructor represents the merge.

Both stay open as additions later — `__subjectClasses` doesn't foreclose either.

jhweir:
@data-bot-coasys All four of your comments addressed in f5b8c368.

**limit/offset** — took the reject option rather than the warning. A `log::warn!`
lands somewhere the caller who wrote `limit: 5` will never look, and the untyped
include one function over already fails loudly rather than degrading, so this
matches. The message names the relation, why the limit can't be honoured (one
sub-query per class, so it would apply per group), and the two ways forward.
Test asserts the refusal.

**Skipped targets** — debug line added, naming the relation and the id. Your
framing of it was the persuasive part: a child that vanishes because a type link
hasn't propagated yet looks identical to a child that was never there.

**Wire-key drift** — more pressing than when you wrote it, since ea284b33 added
`__subjectClasses` alongside. Both constants now name their counterpart and the
file it lives in, on both sides. The stronger half is that a4f8856c drives both
ends against one executor and asserts both keys, so a rename on either side now
fails a test rather than silently degrading to plain JSON. Comments ask people to
remember; the test doesn't.

**The thunk catch** — took your heuristic. `ReferenceError` is the temporal dead
zone, i.e. the import cycle the catch exists for, and stays at `console.debug`.
Anything else is a broken declaration — a thunk closing over nothing throws
`TypeError` — and now warns.

On your two replies to Nico:

The `classes: [...]` proposal is in 47753abd, arrived at from the same reasoning
about `className` making the class self-describing. Named `instantiateAs` rather
than `classes`: a noun phrase describing classes kept being read as a permission
list — "so is this what the relation is allowed to contain?" — which is the one
thing it must not be. A verb phrase can only be read as something done to
results. I dropped `classResolver` rather than keeping it as an escape hatch,
since two ways to answer one question on an unshipped API is worse than one; it
can come back the day a plugin-loaded case turns up.

On returning all hydrations, I'd decline, and the class-list filtering is why
rather than only the cardinality argument. Filtering by the client's list makes
the wire result depend on which constructors happen to be in scope, so two peers
reading identical data disagree about what that data contains. That's exactly
what `instantiateAs` is advisory to avoid — it never excludes, so partial
knowledge degrades to plain JSON instead of to missing members. One entry per
link stays: links decide cardinality, classes decide only how a member is read.

What the discarded readings get instead is visibility — ea284b33 returns the
whole ordered set as `__subjectClasses`, so a caller can see a choice was made
and query the other class by id. And the place a *filter* belongs is the query,
as `include: { children: { ofClass: [...] } }`, where narrowing is something the
caller asked for rather than a side effect of their imports. That's noted as a
follow-up; it would also make `limit` meaningful again, since a read scoped to
one class has an unambiguous one.

core/src/model/Ad4mModel.ts
```diff
@@ -87,8 +146,42 @@ function jsonToModelInstance<T extends Ad4mModel>(
...
+      // May resolve to nothing: a polymorphic relation is allowed to declare no
+      // target at all, and the list may not name the class that arrived — it
+      // says what this call site can construct, not what the relation may hold.
+      // The item then stays plain JSON — correct data, carrying its class name
+      // for a caller that wants to dispatch on it — rather than being forced
+      // through a constructor that does not exist.
+      const resolveChildClass = (item: any): any => {
+        if (!meta.polymorphic || !meta.instantiateAs) return TargetClass;
+        const concrete = item?.[SUBJECT_CLASS_KEY];
+        if (typeof concrete !== 'string') return TargetClass;
```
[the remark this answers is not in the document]
jhweir: Valid — fixed in fe33555c.

Confirmed against current code: `resolveChildClass` fell back to `TargetClass`
whenever `instantiateAs` didn't name the concrete class. Harmless for an untyped
polymorphic relation, since `TargetClass` is `undefined` there and the child
stays JSON — which is exactly why the two existing unlisted-class tests passed.
They use the untyped collection. The uncovered case is a relation that declares a
target *and* reads polymorphically: the child was hydrated against its own shape,
so constructing the declared class over that JSON yields an instance answering
`instanceof` for a class it is not, carrying fields that class never declares.

Your carve-out is in as written — the declared target is kept only when its
`className` equals the value the executor sent, which is a match rather than a
fallback. Everything else resolves to nothing and stays plain JSON carrying
`__subjectClass`.

One branch left alone: a missing `__subjectClass` still resolves to
`TargetClass`. That is not an unresolvable class but a relation read
non-polymorphically — an explicit `polymorphic: false` at the call site — where
the declared target is the correct shape.

Regression test covers both halves, an unlisted class staying data and the
declared target still constructing when named, and I checked it fails against the
previous behaviour rather than merely passing against the new one.

jhweir:
@coderabbitai review

rust-executor/src/perspectives/model_query/integration_tests.rs
```diff
@@ -6520,6 +6520,252 @@ async fn test_polymorphic_include_hydrates_each_child_as_its_own_class() {
...
+        child["__subjectClasses"],
+        serde_json::json!(["TextBlock", "Block"]),
+        "the whole set, most specific first"
+    );
+}
+
+/// A target conforming to two *unrelated* classes is still one member.
+///
+/// Links decide how many children a collection has; classes decide only how each
+/// one is read. So a node carrying two classes' flags — idiomatic AD4M, since
```
lucksus: I consider class flags an anti-pattern. ad4m subject classes generally speaking don't need flags. A flag is just a simple hack to make it very explicit in the link structure. Ultimately it's just a set of constraints.

lucksus (dismissed):
Looks good, though I'm still wondering why the executor code does select one class at all. That selection should happen only in the client where the list of classes is available, no?

rust-executor/src/perspectives/model_query/integration_tests.rs
data-bot-coasys: Right on both counts — flags aren't required for this to work, and calling them a hack is fair.

`subject_classes_of` (rust-executor/src/perspectives/subject_classes_of.rs:99–140) classifies structurally. It treats flags and required-non-flag predicates the same way — both feed the `required_triples` set — and "more required triples wins" is the specificity order. Nothing in `hydrate_polymorphic` reads flags directly; it calls `subject_classes_of` and groups by the class names that come back. Which means the polymorphic mechanism is independent of whether the classes are told apart by flags, by required properties, or by any combination.

The reason this test uses flags is convenience: TextBlock / Collection / Board each have one distinguishing predicate (`text` / `title` / `label`) that isn't marked required, so a flag is the shortest thing that makes them classifiable. To do it flagless, `text` / `title` / `label` become `required: true`, the flag property drops, and the assertions stay green because `subject_classes_of` classifies each URI by which required predicate it actually carries.

Two options:

1. **Leave as-is and drop a note** at the top of the polymorphic tests that flags are one way to make classes distinguishable, not the mechanism itself — so a reader here doesn't infer that polymorphism needs them.
2. **Rewrite this test flag-free** (it's the cleanest candidate — three simple classes, no shared properties). Adds ~5 lines, deletes ~15. Demonstrates the pattern in code, not just prose.

I lean 2 — the whole point of the design is that polymorphism sits on structural class membership, so the canonical polymorphic test should show structural class membership rather than a flag shortcut that ties the reader back to the anti-pattern. Want me to push it?

data-bot-coasys:
> Looks good, though I'm still wondering why the executor code does select one class at all. That selection should happen only in the client where the list of classes is available, no?

Good question. Three things push the primary pick down into the executor rather than up into the client, and one thing keeps the client in the loop anyway.

**1. Hydration is shape-directed.** To hydrate a URI at all, the executor needs a shape — otherwise it doesn't know which predicates to fetch. So there has to be at least *one* class choice per URI at the executor layer. The question isn't whether the executor picks — it's whether the pick is exclusive (as today, most-specific wins) or whether the executor should speculatively hydrate every URI against every applicable class and return N shapes on the wire.

**2. Cardinality invariant.** `include: { children: true }` returns N items where N = the parent's link count. If the *effective* class filter runs client-side ("hydrate only URIs matching one of my registered classes"), the array can shrink or grow based on which classes the client happens to have imported. Two peers reading the same neighbourhood then see different array lengths. Keeping the pick + fallback-to-JSON in the executor means the wire cardinality is decided by the link graph, not by which files the caller has loaded.

**3. Determinism across peers.** "Most specific class wins" is a function of the URI's actual link neighbourhood and the registered SHACL shapes — both of which are stable in a shared neighbourhood. Two clients reading the same neighbourhood get the same `__subjectClass` on each item. If the pick moved to the client, the primary hydration would depend on which constructors happen to be in scope in the caller's process — same data, different reading.

**What keeps the client in the loop.** The pick isn't hidden. `ea284b33` added `__subjectClasses` (plural) alongside `__subjectClass`, so the *whole ordered set of matching classes* is on the wire. A client that disagrees with the executor's pick — say, it has a more specialised class registered than the executor was told about, or it wants to render the URI as a different applicable class — can read `__subjectClasses` and reconstruct from the returned properties, or requery with a different narrowing.

**Where a client-narrowed query belongs.** If the caller wants to say "only hydrate children that are Text or Image, drop the rest", that's not client-side filtering of already-shaped results — it's a `include: { children: { ofClass: [Text, Image] } }` narrowing on the query itself. That path is a follow-up (James noted it in the review-turned-comments). It's the right home for caller-controlled class filtering, because it also makes `limit` meaningful again — a read scoped to one class has an unambiguous count where a polymorphic multi-class read does not.

So: executor picks the *primary shape* because it has to, does so deterministically, and exposes the losers so the client keeps final say.

rust-executor/src/perspectives/model_query/integration_tests.rs
jhweir: I would push back slightly on framing flags as hacks or an anti-pattern in general - I think there are good arguments for them in certain contexts (they signal intent, which structure can't always tell you). But agreed the wording there can be improved, and rewriting the overlap tests flag free is the right move. Done in: 
1adf0ee

jhweir:
@lucksus added the scenario you asked for, and the change that makes it pass — a40bd9b2.

## What it does

An include can now name the classes the read is for. The first named class a
target actually conforms to is the one it gets hydrated as.

Your scenario, whole. A feed holds posts; some of them are image posts, which
require everything a post does and a cover besides:

```ts
@Model({ name: "Post" })
class Post extends Ad4mModel {
  @Property({ through: "blog://headline", required: true })
  headline: string = "";
}

@Model({ name: "ImagePost" })
class ImagePost extends Ad4mModel {
  @Property({ through: "blog://headline", required: true })
  headline: string = "";
  @Property({ through: "blog://cover", required: true })
  cover: string = "";
}

@Model({ name: "Feed" })
class Feed extends Ad4mModel {
  @Flag({ through: "blog://type", value: "blog://feed" })
  type = "blog://feed";

  @HasMany(() => Post, {
    through: "blog://entry",
    polymorphic: true,
    instantiateAs: () => [Post],
  })
  entries: Post[] = [];
}
```

The query is the ordinary one — nothing new at the call site:

```ts
const feed = await Feed.findOne(perspective, {
  where: { id: feedId },
  include: { entries: true },
});

feed.entries[0] instanceof Post   // true
feed.entries[0].headline          // "Chocolate cake"
```

The ORM fills the preference in from the relation's own `instantiateAs`, so what
reaches the executor is:

```json
{
  "className": "Feed",
  "where": { "id": "blog://f/1" },
  "include": {
    "entries": { "polymorphic": true, "preferClasses": ["Post"] }
  }
}
```

and what comes back for an entry that is an `ImagePost`:

```json
{
  "id": "blog://p/1",
  "headline": "Chocolate cake",
  "__subjectClass": "Post",
  "__subjectClasses": ["ImagePost", "Post"]
}
```

Before this it came back as an `ImagePost` — a class the feed never declared and
can't construct, so `Ad4mModel` left it as plain JSON. There's an end-to-end test
for exactly that in `tests/js/tests/model/model-polymorphic.test.ts`.

A call site can also say it directly, overriding what the model declares — for a
view that wants a different lens on the same relation:

```ts
include: { entries: { preferClasses: ["ImagePost", "Post"] } }
```

## Why in the executor rather than the client

This is your "the selection should happen in the client" point, and I think this
is the only shape it can take. Hydration is a projection — `hydrate_one` reads
the predicates the shape names — so choosing a class decides *which values get
fetched*, not just which constructor runs. Two unrelated classes whose shapes
don't overlap means the losing reading's fields were never selected, and no
client-side repair can recover them. So the client does decide; it just has to
say so in the query rather than after the results exist.

Sending it rather than applying it also keeps the answer reproducible: anyone
issuing that query gets that reading. The same list applied client-side would
make identical data read differently depending on which classes a caller
happened to import, which is what I was objecting to earlier in the thread.

## It ranks, it never narrows

Classification is untouched — a target is still tested against every registered
class, and `preferClasses` only reorders the ones it matched. So a target
conforming to nothing you named still comes back, hydrated against whichever
class it *does* match, chosen by specificity exactly as it would have been. A
`Recipe` child in a collection preferring `[TextBlock, ImageBlock]` arrives with
`Recipe`'s properties and `__subjectClass: "Recipe"`; the ORM then can't
construct it, since `Recipe` isn't in `instantiateAs`, so it stays plain JSON
carrying its class name.

That matters because a heterogeneous relation is open by definition: a caller
listing what it can use must not thereby drop what it cannot, or a model's class
list would freeze which classes a collection may hold — the opposite of the
point. There's a test pinning it, deliberately, because "prefer" reads like
"scope" to about half of everyone.

## Two things worth seeing stated rather than inferred

**Specificity is now the fallback, not the rule** — for ORM callers at least,
since every polymorphic read will carry a preference. Naming nothing (or naming
nothing a target is) still ranks by specificity exactly as before.

**`__subjectClass` is now a function of the request.** `__subjectClasses` stays
ranked by specificity whatever was preferred, so the *set* remains a fact about
the node and only the *choice* follows the query. That gives a free diagnostic:
when the chosen class isn't the head of the set, a preference was applied.

## Changed test, on purpose

The multi-class test used to assert that two classes tying on required triples
fell to the alphabetical tie-break. It now asserts the declaration order decides,
because that's the behaviour — a convention losing to a statement of what the
read is for. I rewrote it to demonstrate that rather than patching the
expectation, and added a companion where the call site overrides the model's
declaration and gets the other reading of the same node.

## Compatibility

`ModelQueryInput` doesn't use `deny_unknown_fields`, so an older executor ignores
`preferClasses` and falls back to today's pick rather than erroring — which
matters wherever `@coasys/ad4m` and the executor binary are versioned
separately.

## If you'd still rather return everything

Worth saying that the two aren't opposed. Returning all readings doesn't remove
the pick — something still has to decide which reading is the flat object a
caller reads `child.title` from — so a preference is needed either way. If
multi-reading lands later, this list keeps its meaning (which reading is flat)
and the extras sit beside it. It's the half that survives under either design,
which is part of why I was comfortable adding wire surface for it now.

Re-requesting review since this is behaviour and a new query field after your
approval.

lucksus (approved):

PR #926, opened 2026-08-25
Feature: CRDT ordering module and its SHACL declaration

jhweir:
# CRDT ordering module and its SHACL declaration

> [!IMPORTANT]
> **Branch** `feat/crdt-ordering-module` · **Base** `dev`.
> Its former base, the `subjectClassesOf` PR (#924), has **merged**, so this now targets `dev` directly. That base was only ever plumbing — this PR's content is independent of it — so nothing here depends on it.
>
> Part of an eight-PR series on the AD4M model layer — review wave 3 of 4. The full shape:

```
dev ─┬─ 1  relation-correctness ── 5  subjectClassesOf (merged) ─┬─ 6  polymorphic-include
     ├─ 2  getter+target
```

## Summary

A `@HasMany` is a set of links, and hydration sorts them by link timestamp. That is right for an
append-only collection — a transcript, a message thread — where timestamp order *is* the order. It
cannot express a sequence somebody chose: a kanban column, a playlist, the blocks of a post.

Worth stating what exists today, because it looks like ordering and is not. `collectionSetter`
deletes every link for the relation and re-adds them in array order, restamping every timestamp.
Single-agent that produces the right sequence by accident. Concurrently, the two agents' writes
interleave by wall clock and neither array survives.

This is the algorithm and the metadata. **Nothing is wired to a read or a write, so no behaviour
changes** — the integration is the next PR, and the two are split so this one can be reviewed on its
own terms.

## The invariant everything follows from

**Ordering entries are position hints over a membership set the data links already define.** The data
links say what is *in* the collection; the entries say only where things go. An entry naming an
absent item is inert, an item with no entry is appended by timestamp, and neither is an error.

That is why there are no tombstones. An earlier draft (V3) had deletion write an entry flagged
`deleted` carrying a copy of the original's `after`, so the chain stayed traversable. But
`reconstruct` already takes the data links, so `item ∉ data_links` answers "deleted" without any of
it: no flag, no copied pointer, no extra link per removal, and no accumulating class of entries
needing collection — which was most of what garbage collection was for, and GC is genuinely hard in a
system that cannot know every peer has seen a deletion.

It is also strictly *more* correct on resurrect: a tombstone's higher `pid` would keep an item hidden
after its data link came back.

## Changes

**`rust-executor/src/perspectives/ordering/`** (new module)

`types.rs` — `OrderingEntry`, `OrderingDiff`, `OrderingConfig`, and `make_pid`. The `pid` is
`{timestamp}_{agentDid}` with the timestamp zero-padded to 16 digits so that string comparison equals
numeric comparison; without the padding, tiebreaking would order differently on different machines,
which is the one thing a CRDT may not do.

`mod.rs` — the strategy trait, the factory, and entry encode/parse. `data_links` is passed as
`(target, timestamp)` pairs rather than full link expressions: it is what both future call sites
already hold, and it keeps the module testable without a store.

`linked_list.rs` — the RGA implementation.

**Why a linked list.** On the criteria usually cited — no chain traversal, O(1) moves, no cleanup —
fractional indexing wins, and V3 justified `linkedList` on "covers all use cases", which is not a
reason. The actual reason: `{item, after, pid}` **encodes the causal graph**. Two entries sharing an
`after` are a fork; an entry whose `after` names an item from another fork is a merge. A position
string destroys that. It is what a partitioned-conversation view would need — where linearising two
independently-coherent threads renders both unreadable — so it is the property not to trade away.
`fractionalIndex` is reserved in the config and refused at construction, with that as the message.

**The diff is not an LIS diff.** V3 called for longest-increasing-subsequence: compute which items
are already in the right relative order and leave their entries alone. That is unsound for a linked
list, and the simplest reordering there is shows why. `[A,B,C,D] → [A,C,B,D]` has LIS `[A,C,D]`, so
only B moves — but `after` is a *pointer*, and moving B out from between A and C leaves C still
pointing at B while B now points at C. A two-item cycle where the list used to be.

Comparing **predecessors** gets the same economy honestly: project what the order would be with no
new entry written, then write one for each item whose predecessor is not already what the desired
order asks for. An untouched run costs nothing, an append costs one entry, a move costs about three —
the item, the one it left, and the one it landed behind. And it keeps the property LIS was wanted
for: an item whose predecessor is unchanged is never regenerated, so a save cannot clobber a remote
agent's move with a higher `pid`. Removals fall out free — the projection already reflects the
dropped data link, so it matches the desired order and nothing is written.

**`core/src/model/decorators.ts`, `shacl-gen.ts`, `shacl/SHACLShape.ts`** —
`@HasMany({ ordering: { strategy: "linkedList" } })` emits `ad4m://ordering` on the property shape
and round-trips through `SHACLShape`.

**`rust-executor/src/perspectives/model_query/shape.rs`, `types.rs`** — parsed back into
`ShapeProperty.ordering` on both shape-loading paths.

Declaring it in the type system rather than passing it per query is what lets the executor apply it
to *every* writer — the ORM, MCP agents, raw GraphQL callers, another app on the same neighbourhood.
Implemented client-side it would order only what that client wrote.

## Known follow-ups

- Stale entries accumulate from *moves* (not from deletions, which write nothing). Bounded by edit
  count and harmless on read, but a compaction pass is eventually worth having.
- `fractionalIndex` is reserved and unimplemented.
- The DAG reconstruction mode the storage format exists to enable is not built. It is purely additive
  — one trait method reinterpreting the same entries — and deliberately deferred.

## Test plan

> tree above.

- [x] `cargo test -p ad4m-executor --lib perspectives::ordering` — 23 passing
- [x] **The deletion matrix**, which is this design's central claim rather than an assertion about it:
      removed member disappears, insert-after-a-deleted-item still resolves (the case tombstones
      existed for), delete-the-head, delete-a-whole-run, two agents deleting the same item,
      resurrect restoring the original position
- [x] Concurrency: concurrent inserts at one predecessor linearising by `pid`; concurrent moves of the
      same item resolving identically from both arrival orders
- [x] Diffing: unchanged array writes nothing, a move writes only what must move *and the merged
      result reconstructs to the requested order* (this is what caught the LIS bug), a new item is
      positioned, a removal writes nothing
- [x] Robustness: a cycle truncates rather than hangs, an unparseable entry is skipped not fatal,
      entries scoped to their relation, round-trip through the link target
- [x] `cargo test -p ad4m-executor --lib perspectives::model_query` — 223 passing (unchanged, as
      expected for an inert module)
- [x] `jest src/model src/shacl` in `core` — 296 passing
- [x] `tsc --noEmit` clean; `cargo fmt -p ad4m-executor`

## Summary by CodeRabbit

* **New Features**
  * Added configurable linked-list ordering for relation collections.
  * Collection order persists across edits and remains consistent after concurrent changes.
  * Ordering can be enabled on supported relation decorators and is included in schema metadata.
  * Items can be inserted, moved, or appended while preserving the intended order.
* **Compatibility**
  * Unordered collections retain their existing behavior.
  * Ordering remains resilient when items are removed or updates arrive in different sequences.


jhweir:
@coderabbitai full review

jhweir:
@coderabbitai full review

jhweir:
@coderabbitai full review

jhweir:
@coderabbitai review

core/src/shacl/SHACLShape.ts
```diff
@@ -782,6 +798,15 @@ export class SHACLShape {
         l.source === propShapeId && l.predicate === "sh://datatype"
       );
       if (datatypeLink) prop.datatype = datatypeLink.target;
+
+      const orderingLink = links.find(l =>
+        l.source === propShapeId && l.predicate === "ad4m://ordering"
+      );
+      if (orderingLink) {
+        prop.ordering = orderingLink.target.replace(
```
data-bot-coasys, line 806: **Concern (non-blocking):** worth adding a TS round-trip test — `SHACLShape.fromLinks(shape.toLinks(x)).ordering === x.ordering` — because if the `literal:string:` / `literal://string:` wire format ever changes, this regex will silently drop `ordering` and the executor's SPARQL query won't see it. There's a Rust-side round-trip in `parse_shape_from_json`, but nothing pins the TS ↔ TS path.

One shape with `prop.ordering = 'linkedList'` in and out is enough.

rust-executor/src/perspectives/ordering/linked_list.rs
```diff
@@ -0,0 +1,317 @@
...
+/// projected members against stored ones by string, so a projection in any
+/// other format is not comparable with the data it is standing in for.
+fn rfc3339_millis(ms: u64) -> String {
+    chrono::DateTime::from_timestamp_millis(ms as i64)
+        .unwrap_or_else(chrono::Utc::now)
+        .to_rfc3339_opts(chrono::SecondsFormat::Millis, true)
+}
+
+/// Each item mapped to the item before it in `order` (or the head).
+fn predecessors(order: &[String]) -> HashMap<&str, String> {
```
data-bot-coasys, line 328: **Nit:** if `desired_items` contains a duplicate URI (e.g., a caller assigns `[a, b, a, c]` to a `@HasMany`), the second `a` silently overwrites the first here, and `diff` then emits at most one entry per URI. Downstream `reconstruct` produces `[a, b, c]` and the caller's duplicate has vanished with no signal.

A `@HasMany` is a set semantically, so this is arguably correct, but it's not documented anywhere. Worth either:

- a doc line on `diff`: "desired_items is treated as a set; last-position-wins for duplicates", or
- a `debug_assert!(desired_items.iter().collect::<HashSet<_>>().len() == desired_items.len())`.

No blocker.

rust-executor/src/perspectives/ordering/linked_list.rs
```diff
@@ -0,0 +1,317 @@
...
+/// A millisecond epoch in the format link timestamps are written in.
+///
+/// Must stay in step with `agent::create_signed_expression`, which stamps every
+/// link with `to_rfc3339_opts(SecondsFormat::Millis, true)`. `diff` compares
+/// projected members against stored ones by string, so a projection in any
+/// other format is not comparable with the data it is standing in for.
+fn rfc3339_millis(ms: u64) -> String {
+    chrono::DateTime::from_timestamp_millis(ms as i64)
+        .unwrap_or_else(chrono::Utc::now)
+        .to_rfc3339_opts(chrono::SecondsFormat::Millis, true)
```
data-bot-coasys, line 324: **Nit:** the `unwrap_or_else(chrono::Utc::now)` fallback silently substitutes a different timestamp than the caller asked for. The overflow case is millennia away, but silent-fallback in a CRDT projection is exactly what shows up in a bug report two years after it lands. Consider:

```rust
fn rfc3339_millis(ms: u64) -> Result<String, Error> {
    chrono::DateTime::from_timestamp_millis(ms as i64)
        .ok_or_else(|| anyhow!("timestamp {ms} out of range for RFC3339 rendering"))
        .map(|dt| dt.to_rfc3339_opts(chrono::SecondsFormat::Millis, true))
}
```

and bubbling that up through `diff`. Not blocking — a `log::error!` is also fine.

rust-executor/src/perspectives/ordering/linked_list.rs
```diff
@@ -0,0 +1,317 @@
...
+                    }
+                }
+            }
+        }
+
+        // Members with no surviving entry go last, oldest first. Covers the
+        // unordered→ordered migration, a peer that wrote data links without
+        // ordering ones, and a crash between the two writes.
+        let mut unpositioned: Vec<&(String, String)> = data_links
+            .iter()
```
data-bot-coasys, line 138: **Nit:** silent cycle truncation means a corrupted collection produces no operator-visible signal. Consider a warn-once when the reachable count from the head is smaller than the winner count:

```rust
let reachable = visited.len();
let total_positioned = winners.len();
if reachable < total_positioned {
    log::warn!(
        "ordering: {} entries unreachable from head — likely cycle or orphaned chain, appending to tail",
        total_positioned - reachable
    );
}
```

The test `a_cycle_truncates_rather_than_hangs` pins the graceful behavior; a log line pins the diagnosability of the corruption.

rust-executor/src/perspectives/ordering/linked_list.rs
```diff
@@ -0,0 +1,317 @@
...
+                });
+                seq += 1;
+            }
+            previous = item.clone();
+        }
+
+        OrderingDiff { add }
+    }
+
+    fn generate_append(
```
data-bot-coasys, line 272: **Question:** what's the intended behavior when `item_uri` is already in `current_members`? Right now `generate_append` still writes a new entry with the item positioned after the current last member — effectively a "move to end" whether or not the caller wanted that.

Probably the right default for a real append operation, but a caller writing "add if missing" on top of this would be surprised. Worth a doc line: "if `item_uri` is already a member, this reads as a move-to-end."

data-bot-coasys (dismissed):
## Review of #926 — CRDT ordering module

Careful design and, more importantly, a test suite that actually earns the design. The deletion matrix is exactly the evidence I wanted the writeup to promise. The `two_moves_sharing_a_pid_still_converge` test — with the `(pid, after)` tuple as tiebreak — pins the one case where peer agreement could otherwise depend on arrival order, and `diff_projects_a_new_item_the_way_its_link_will_read_back` catches the RFC3339-vs-zero-padded format hazard I would probably have missed on my own.

I walked through `[A,B,C,D] → [A,C,B,D]` mentally: `projected == [A,B,C,D]` (no new entries yet), `desired == [A,C,B,D]`. `predecessors(projected)`: A→HEAD, B→A, C→B, D→C. Walking `desired` with running `previous`: A's stored predecessor is HEAD ✓ skip; C wants A, stored is B ≠ A → write; B wants C, stored is A ≠ C → write; D wants B, stored is C ≠ B → write. Three entries, exactly as advertised. LIS would have written one and produced a cycle. 👍

Approving. Notes below are nits and a couple of "how does this behave when..." observations — no blockers.

**Concerns worth thinking about (non-blocking):**

1. No TypeScript-side round-trip test for the ordering declaration through `SHACLShape.toLinks` → `fromLinks`. The wire format is a `literal:string:` (or `literal://string:`) target, and the regex parse would silently drop `ordering` if that format ever changed. A `save → parse` assertion in the SHACL test suite would pin it.
2. `predecessors()` silently deduplicates a duplicate item URI in `desired_items`. A `@HasMany` is semantically a set, but the TS-side ORM accepts arrays and nothing enforces uniqueness. Worth either a debug-assert here or a doc line at `diff` saying "desired_items is treated as a set; duplicates behave as if the last position wins".
3. `rfc3339_millis`'s `unwrap_or_else(chrono::Utc::now)` bakes in a subtle bug for the day the fallback fires. Millennia away, granted, but silent-fallback in a CRDT projection is exactly the kind of thing that shows up in a bug report two years after it lands.
4. `reconstruct`'s cycle truncation is silent. A single `log::warn!("ordering: {n} items unreachable from head, appending as unpositioned tail")` when the visited-set count doesn't match the winner count would help operators diagnose data corruption.
5. `generate_append` doesn't check whether `item_uri` is already in `current_members`. Called that way, it effectively moves the item to the end — probably fine, but worth a doc line so a caller writing "add if missing" logic on top isn't surprised.

Great module.

lucksus (dismissed):

lucksus (dismissed):

rust-executor/src/perspectives/ordering/linked_list.rs
data-bot-coasys: **Resolved — nicely.** Fixed in [e1f3b96d6](https://github.com/coasys/ad4m/commit/e1f3b96d6). You went further than I asked: I flagged the *silent fallback*, and the commit found the deeper bug — `ms as i64` wraps a `u64` past `i64::MAX` to a *negative* epoch, which chrono happily reads as a pre-1970 date and returns `Some`, so the fallback I was worried about was actually **unreachable** for the one input it exists to catch. Now uses `i64::try_from(ms).ok().and_then(...)` and the error log is reachable. Doc-comment updated with the rationale.

core/src/shacl/SHACLShape.ts
data-bot-coasys: **Resolved.** New `core/src/model/shacl-roundtrip.test.ts:330` — `round-trips ordering through toLinks/fromLinks` — does exactly what I asked (rebuild from `toLinks()` and assert `rebuiltProp.ordering === "linkedList"`) plus one thing I didn't: it *pins the exact wire spelling* `literal:string:linkedList` on the `ad4m://ordering` link, so a writer change to the literal-prefix format surfaces as a test failure rather than as an executor silently seeing no strategy. Nice.

rust-executor/src/perspectives/ordering/linked_list.rs
data-bot-coasys: **Not addressed — happy to defer.** `predecessors()` still uses `out.insert(item.as_str(), previous.clone())` which silently overwrites on a duplicate URI in `desired_items`. My concern was defensive: if a caller ever hands `@HasMany` a list with duplicates, `diff` would emit at most one entry per URI without warning. In practice `@HasMany` values are set-typed, so this only fires on caller error. Fine to leave; a `warn_once` here is a follow-up if it ever bites.

rust-executor/src/perspectives/ordering/linked_list.rs
data-bot-coasys: **Not addressed.** The `a_cycle_truncates_rather_than_hangs` test (tests.rs:71) locks in the *behavior* but there's no operator-visible signal when it fires — a corrupted collection returns silently truncated. Non-blocking; the tail-fallback catches the members, and the test names the invariant. A `log::warn!` when `reachable < entry_count` would rescue debuggers in the field if this ever hits.

rust-executor/src/perspectives/ordering/linked_list.rs
data-bot-coasys: **Not addressed — question still open.** `generate_append` still calls `reconstruct(...).last()` and writes a new after→last entry unconditionally. If `item_uri` is already in `current_members`, the new entry effectively moves it to the tail; if it isn't, it appends. No test pins either behavior. Since `d98aa7abd` re-frames the module as "a declaration, nothing acts on it yet," I don't think this bites callers today — but worth pinning the intended semantics with a comment + a test before a writer starts calling it.

data-bot-coasys (dismissed):
## Re-review — approving again on `8d5ec65c2`

Went through the 5 threads from the 08-31 review and the 9 commits landed since. Status:

**Addressed (2/5) — both are the ones that would have mattered most:**
- **TS SHACL round-trip test** ([discussion](https://github.com/coasys/ad4m/pull/926#discussion_r3895260767)) — new `core/src/model/shacl-roundtrip.test.ts` includes `round-trips ordering through toLinks/fromLinks`, and *also* pins the exact literal spelling `literal:string:linkedList` on `ad4m://ordering`. Change to the wire format now surfaces as a test failure instead of an executor silently seeing no strategy. Better than my ask.
- **`rfc3339_millis` silent overflow** ([discussion](https://github.com/coasys/ad4m/pull/926#discussion_r3895260786)) — [`e1f3b96d6`](https://github.com/coasys/ad4m/commit/e1f3b96d6). Went further than I asked: found that `ms as i64` wraps past `i64::MAX` into a *plausible* pre-1970 date, meaning the fallback I flagged was actually unreachable for its designed input. Now uses `i64::try_from(ms).ok().and_then(...)`.

**Not addressed (3/5) — all originally labelled nit/question, all non-blocking:**
- **`predecessors()` dedups duplicate URIs silently** — defensive only; `@HasMany` values are set-typed in practice, so this fires on caller error not normal flow. Follow-up-if-it-bites.
- **Cycle truncation has no operator-visible signal** — the `a_cycle_truncates_rather_than_hangs` test pins the *behaviour*, but a `log::warn!` when reachable < entry_count would rescue future debuggers. Non-blocking.
- **`generate_append` on an already-member `item_uri` has undefined intent** — since `d98aa7abd` reframes the module as "a declaration, nothing acts on it yet," the semantics can be pinned when the first writer starts calling it. Question-level, not blocker.

**Also worth naming — substantive work beyond my review:**
- `3edecc0f5` splits pid sequence from timestamp (was a subtle collision hazard)
- `968fb9e60` makes equal-pid tiebreak deterministic across peers (proper CRDT hygiene)
- `d05eb6f00` fixes the pending-member projection format so `diff` doesn't silently ship wrong order at the head of the tail — the deeper bug the earlier projection was masking
- `39eea23d6` percent-encodes the entry payload so a URI with a control char no longer breaks the round-trip
- `2ea9638ae` strips both string-literal prefixes so writer-drift in ADAM's link stamper is tolerated
- `00f3d208e` pins the ordering declaration in SHACL + adds coverage; `d98aa7abd` clarifies the READ-ONLY status so integrators know the writer path isn't wired yet

Approving. The two things I wanted fixed are fixed, plus a stack of related quality improvements. The three deferrals are all things that can wait for the first real writer to land.

jhweir:
@data-bot-coasys Apologies — I pushed six commits against your 08-31 review and left nothing explaining any of them. You had to reconstruct intent from diffs, and where the change wasn't where you'd expect it, the reconstruction came out wrong. That's on me, not on the re-review.

Which is the substance of this comment: the scoring should read 4/5 addressed and one declined, rather than 2/5.

## #2 and #5 are addressed — on the trait, not at the impl sites

Both went into `rust-executor/src/perspectives/ordering/mod.rs` as doc paragraphs on `OrderingStrategy`, which is I think why looking at `predecessors()` and the `generate_append` body in `linked_list.rs` showed unchanged code. Both are present in `8d5ec65c2`, the commit you re-reviewed:

**`diff`** — [mod.rs:82-85](https://github.com/coasys/ad4m/blob/8d5ec65c2/rust-executor/src/perspectives/ordering/mod.rs#L82-L85):

> `desired_items` is treated as a **set**: a relation is a set of links, so a repeated item cannot be represented and the last of its positions is the one that survives. An array with duplicates therefore reads back deduplicated rather than being rejected.

**`generate_append`** — [mod.rs:99-102](https://github.com/coasys/ad4m/blob/8d5ec65c2/rust-executor/src/perspectives/ordering/mod.rs#L99-L102):

> The item is not assumed to be new: called for one that is already a member, this reads as a move to the end, since the entry it returns carries a fresh `pid` and supersedes the item's current position. A caller wanting add-if-missing has to check membership itself.

The trait rather than the impl was deliberate. Both are properties of the *operation*, not of the linked list — `fractionalIndex` is already reserved-and-refused in `strategy_for`, and when it lands it will need to honour the same two contracts. Documenting them at `LinkedListStrategy` would have made them look like implementation quirks of one strategy.

## On the `debug_assert!` alternative for #2

You offered a doc line or a `debug_assert!` on uniqueness. I took the doc line and declined the assert on purpose: `desired_items` comes off a client-supplied array, so a duplicate is bad input rather than a broken invariant, and asserting on it means panicking the test suite for a caller's mistake. The doc states the outcome; a caller is free to hand us a list and get set semantics back.

## #4 I'm declining — and I think the diagnosis is off, not just the priority

`reachable < entry_count` doesn't identify a cycle or an orphaned chain. `reconstruct` traverses *entries*, so a missing link anywhere in the chain strands everything behind it — including the case where an entry's `after` names an item whose own entry simply hasn't synced yet. That's not corruption; it's partial sync, and it's one of the three cases the unpositioned tail exists to absorb, named in `reconstruct`'s docstring alongside the unordered→ordered migration and a crash between the data-link and ordering-link writes.

So a `log::warn!` on that condition fires during normal neighbourhood operation, on data the module is handling exactly as designed. That's the kind of warning people learn to filter within a week, well before it ever fires on a genuine cycle — which makes it worse than no signal, because it burns the channel a real one would need.

If it goes in later it wants to be `debug!`, or a `warn!` whose text names sync lag first and cycles second. `a_cycle_truncates_rather_than_hangs` keeps the behaviour pinned either way; what's missing is diagnosability, and this particular counter isn't a clean read on it.

## The part of #5 I'll grant

You're right that nothing pins the move-to-end semantics but a doc comment, and a doc comment is a claim rather than a guarantee. Worth converting to a test before the first writer calls it — that's the follow-up I'd actually take from this round.

## On `rfc3339_millis`

Yours was the catch that paid. Chasing it turned up that `ms as i64` wraps a `u64` past `i64::MAX` into a *negative* epoch, which chrono reads as a plausible pre-1970 date and returns `Some` for — so the silent fallback you flagged, and the error log I'd added to it, were both unreachable for precisely the input they existed to catch. `i64::try_from(ms).ok().and_then(...)` now, in `e1f3b96d6`. Better bug than the one either of us went in looking for.

lucksus (approved):

PR #927, opened 2026-08-25
Feature: CRDT ordering integration

jhweir:
# CRDT ordering: reconstruct on read, diff on write

> [!NOTE]
> **Branch** `feat/crdt-ordering-integration` · **Base** `dev`.
>
> This is the **last of an eight-PR series** on the AD4M model layer. Its three dependencies have all
> landed, so it now stands alone against `dev` and no longer needs reading as part of a stack:
>   - the ordering module — #926
>   - `subjectClassesOf` — #924
>   - relation correctness — #920
>
> `dev` has been merged in (the second commit) and the diff below is this PR's own work only.
>
> For context, the series as it was reviewed — every one of these is now merged:
>
> | # | PR | |
> |---|---|---|
> | 1 | #920 | relation correctness |
> | 2 | #922 | getter + target class |
> | 3 | #921 | where-clause compiler |
> | 4 | #923 | relation quantifiers |
> | 5 | #924 | `subjectClassesOf` |
> | 6 | #925 | polymorphic include |
> | 7 | #926 | ordering module |
> | 8 | **this PR** | ordering integration |

## Summary

Wires the ordering module into both halves. `@HasMany({ ordering: { strategy: "linkedList" } })` now
does something: the array you assign is the order you get back, and it survives two people editing at
once. Nothing about how a caller writes the relation changes — assign the array in the order you want
and save.

It also replaces `collectionSetter`'s delete-everything-then-re-add with a diff, which is a large
saving on its own but **cannot ship separately** — see below.

## Changes

**`rust-executor/src/perspectives/model_query/hydration.rs`** — reconstruction on the ORM read path.

V3 put this in `get_links`. That would have shipped a feature nothing could see: **the ORM read path
never touches `get_links`.** `model_query` talks to the store directly and hydrates in `hydrate_one`,
so `$query`, `include` and every `findAll` — everything the ORM actually reads through — would have
stayed unordered. This was the single largest correction to the design.

Reconstruction costs no extra query, which is the entire reason entries are sourced on the parent:
they arrive with the instance's own links. Missing or unreadable entries degrade rather than fail, so
a collection whose ordering links have not synced yet, or whose strategy this executor does not know,
reads as an ordinary unordered relation.

> What "unordered" means depends on which path the relation takes, and it is worth being exact
> because the migration window rests on it. A plain relation falls back to link-timestamp order. A
> relation that **names a target class** is getter-backed, and its array comes from a conformance
> `SELECT` carrying no `ORDER BY` — so its order with no entries is formally unspecified, as it has
> always been. Neither is a regression, and the window is short: the first save after migration finds
> no entries and writes a full chain. The integration test for this case asserts membership rather
> than sequence, because sequence is not promised here.

**`rust-executor/src/perspectives/model_query/getters.rs`** — reconstruction on the *other* path the
array can be decided on.

> The second blocker, and the one the TS integration tests were written to find: **CRDT ordering did
> nothing for any `@HasMany` that names a target class** — which is how a relation is normally
> declared, and every realistic model. Naming a target emits a conformance getter, and `hydrate_one`
> builds its predicate map from `properties.iter().filter(|p| p.getter.is_none())`. A getter-backed
> relation therefore never reaches `collection_values` and the reconstruction above never ran for it;
> its array is produced later, in `evaluate_getters`. Ordering applied only to relations with no
> target class, which is the shape nothing writes.

Reconstruction now runs wherever the array is finally decided, through one shared `reorder_members`.
The entries arrive on the parent's own links and `hydrate_one` already collects them, so it parks
them for the getter stage rather than paying to re-read what is already in hand.

That stash is deliberately short-lived: `evaluate_getters` consumes it and strips it **in the same
function**, on its early-return path as well as at the end. It cannot outlive that call, which matters
because `filter_properties` only runs when a query names its `properties` — anything left behind would
reach the caller. `hydrate_instances` has exactly one non-test caller, and both early returns in that
function precede hydration and return no instances, so there is no path around it.

For the getter path, position within the getter's own result stands in for a member's timestamp. A
member with no entry yet then keeps exactly the order it has today, which is the migration case above.
Stage order does the rest: getters run before includes, so an eager-loaded relation inherits the
corrected array rather than needing its own pass.

**`core/src/model/shacl-gen.ts`** — a class owning any ordered collection now declares
`ad4m://collection_order` on its node shape. The instance query builds its predicate filter from the
shape's declared paths, so without this it filters out the very links the ordering depends on. One
declaration covers every ordered relation on the class; each entry names the relation it belongs to.

**`rust-executor/src/perspectives/perspective_instance.rs`** — `collectionSetter` diffs both layers.

The old pattern deleted every link for the relation and re-added them all: a one-item change to a
100-item collection cost 100 removals and 101 additions, each signed and synced, on every save. Now
only the delta moves.

> **Why this cannot ship as a standalone performance fix**, as V3 §7 suggested. The old pattern
> restamped every timestamp, and `hydrate_one` sorts by timestamp — which is what today's ordering
> silently depends on. Preserving timestamps for unchanged items is exactly what stops array order
> being reflected in link order, so the diff *without* the reconstruction would have quietly
> scrambled every existing ordered collection in WE on its next save. The two arrive together.

The setter also reads the batch's own staged work, via `links_including_batch`. This is
read-your-own-writes within one agent's transaction, not anything to do with concurrency: diffing is
what created the dependency, because a delta is only correct against the state the transaction will
actually produce, and delete-all-then-re-add needed no notion of current state at all. See the
follow-up on a transaction-aware `get_links`, which is arguably where this belongs.

A batched `add_links`
writes into `batch_store` and `get_links` reads the SPARQL store, so inside an uncommitted batch the
store still shows the collection as it stood before the batch began — and a second setter on the same
relation would compute its delta against that stale membership and re-stage everything the first one
had already added. `save(batchId)` re-snapshots rather than refetching precisely so a later save in
the same batch diffs correctly, which makes this an ordinary sequence rather than an exotic one. The
helper is a no-op without a batch, and the ordering-entry read uses it too, so a second save extends
the existing chain instead of restating it. Removals still come off the persisted set only: a member
that exists solely as a pending addition leaves by being pruned from the batch, not by removing
something the store has never seen.

> This one is the diff's to fix rather than a pre-existing bug. Delete-all-then-re-add could not hit
> it — it pruned every pending addition for the relation and restated the whole set. Found by
> CodeRabbit on review, in a narrower form; the duplicates it produced were invisible on read, since
> hydration already collapses duplicate collection links and `latest_per_item` collapses duplicate
> ordering entries. The cost was in links signed and synced, which is the thing the diff exists to
> reduce.

`apply_collection_ordering` resolves the strategy by asking which classes the source is an instance of
(via `subjectClassesOf`, #924) and reading their shapes. Entries are written additively, and a removal
writes nothing at all — dropping the data link is the whole deletion.

Every conforming class is scanned, most specific first, and the first declaration for the predicate
wins. Asking only the most specific one would make a declaration on a less specific class invisible to
the writer while the reader still honoured it — hydration resolves ordering from whatever shape the
query came through. The setter would write no entries for a read that expects them, and the collection
would fall back to link timestamps, which this branch deliberately stops restamping: the same silent
scramble the design works to rule out, reached through the multi-class door. Overlapping classes over
shared links is ordinary social-DNA reuse. Specificity still decides when several classes declare,
including when they disagree, which at least makes that answer deterministic.

Classification counts the batch's staged links alongside the store's, via
`subject_classes_of_with_pending`. A subject is *created* inside a batch: `save()` on a new instance
opens one, runs the constructor actions that write the class's flag links, sets the relations, and
commits. During the setter those flags are staged, so asking the store alone answers "no class" — and
an ordered collection created and populated in one save got no ordering entries at all. It self-heals
on the next save, and a fresh create still reads correctly because its links went in in the assigned
order and the fallback is by timestamp; what was missing until then is the guarantee the feature
exists for, since concurrent edits were unprotected in that window. Membership is set containment, so
a staged triple counts exactly as a committed one — this widens what is known, never what conforms.
`subject_classes_of` is unchanged and delegates with no pending links, so no other caller moves.

**`core/src/shacl/SHACLShape.ts`, `rust-executor/src/perspectives/shacl_parser.rs`** — the ordering
declaration now survives class registration.

> This was a blocker, found while acting on review: **`@HasMany({ ordering })` did nothing at all in
> production.** `@Model` classes register through `ensureSubjectClass` → `SHACLShape.toJSON()` →
> `add_sdna` → `parse_shacl_to_links`, and `load_shape` reads the strategy back only from the
> `ad4m://ordering` link that chain produces. It was broken twice over, independently: `toJSON()`
> mapped twenty-odd property fields and omitted `ordering`, and Rust's `PropertyShape` had no
> `ordering` field, so the link could not have been emitted even had the JSON carried it. The result
> was `ShapeProperty::ordering == None` everywhere: no entries written, no reordering on read.
>
> `toLinks()` spells the link correctly, which is why the existing round-trip test passed — but
> `toLinks` serves `addShacl`, not the decorator path this feature is exposed through. No test crossed
> the boundary: the Rust end-to-end tests build their shape via `parse_shape_from_json`, a helper whose
> only other caller is its own unit test. Two new tests do cross it, each failing without its own half
> of the fix.

**`core/src/model/Ad4mModel.ts`, `types.ts`** — `changedFields` is order-sensitive for ordered
relations. It previously sorted both sides before comparing, which for an ordered collection would
make `save()` a no-op for the one edit this feature exists to support. Unordered relations keep the
sort: they are sets, returned by link timestamp, so the same members in another order genuinely
changed nothing.

`getModelMetadata` had to carry `ordering` through for this. It projects the decorator registry into a
narrower public shape and silently drops anything not listed — worth knowing before adding the next
relation option, since `polymorphic` and `datatype` are dropped there too (harmlessly, because their
consumers read the raw registry instead).

## Known follow-ups

- **A transaction-aware `get_links` would delete the batch helpers here.** Raised by Nico on review:
  a batch is one agent's local transaction, so reaching into it from `collectionSetter` looks like the
  wrong layer — and the honest answer is that this is read-your-own-writes, not concurrency, and the
  setter is compensating locally for something `get_links` does not do. If reads inside a batch saw
  the batch, `links_including_batch`, the removal reconciliation and
  `subject_classes_of_with_pending` all become unnecessary. Not done here because it changes read
  semantics for every caller inside a batch — Prolog, subscriptions, MCP — which is a far wider blast
  radius than this PR should carry.

- **A getter-backed relation with no entries yet has no defined order.** Its array comes from a
  conformance `SELECT` with no `ORDER BY`, so until the first save writes a chain the sequence is
  whatever the store returns. That is pre-existing — it is how every target-typed relation has always
  read — and the window is one save wide, but it means the unordered→ordered migration is not
  order-preserving for those relations. Giving it the same link-timestamp fallback the plain path has
  would mean carrying timestamps through the getter, which does not return them today. Worth doing
  deliberately rather than bolting on here.

- **Superseded ordering entries accumulate without bound.** Writes are additive by design — nothing is
  ever pruned — so a move writes ~3 entries and `parse_ordering_entries` decodes every entry ever
  written on every read of the parent. A collection reordered daily grows in both storage and per-read
  cost indefinitely. Compaction looks straightforward to add later (when winners ≪ total, one
  `generate_full` at a fresh timestamp supersedes everything and the rest becomes prunable), but *who
  may prune* is a real design question, because those links are signed by other agents. Named here
  rather than discovered when the first long-lived transcript gets slow.

- **`addLink` / `removeLink` do not yet maintain ordering** for callers who bypass `collectionSetter`
  — MCP agents, raw GraphQL. Their appends land in the by-timestamp tail, which is correct but
  unpositioned. Additive on top of this.
- **`get_links` still returns link order** for those same callers. The module is shared, so this is a
  small addition once the config lookup is available at that layer.
- Ordered + `limit` on an `include` sub-query is unanswered: reconstruction needs every entry, so
  "the last 50 in CRDT order" cannot be pushed down. Worth settling before a long transcript relies
  on it.
- Subscription re-fire on an ordering-link change is untested. Entries are parent-sourced, so a
  subscription watching the parent *should* see it — but that wants verifying against the real
  `model_subscribe` path rather than assuming.

## What this unblocks in WE

Block order in posts and call transcripts survives concurrent editing instead of resting on the
accidental delete-all-re-add. Ordered collections (kanban columns, playlists) become expressible in a
template. And the diff half is the one with the immediate payoff: **a transcript is append-only, so
timestamp order already *is* its order — what breaks it today is the restamping, not the absence of
ordering.**

## Test plan

> Re-measured on the merge with `dev`. The totals are higher than the earlier revision's because the
> baseline moved — the rest of the series and everything else on `dev` is now in the count — not
> because this PR grew. Its own additions are unchanged: 2 Rust end-to-end tests and 6 in `core`.

- [x] `cargo test -p ad4m-executor --lib perspectives::model_query` — 283 passing
- [x] `cargo test -p ad4m-executor --lib perspectives::ordering` — 28 passing (unchanged by this PR)
- [x] End-to-end test: data links written in the order `c, a, b` with ordering entries chaining
      `a → b → c`, asserting the ORM read returns `a, b, c` — i.e. CRDT order, not the link-timestamp
      order a fallback would give
- [x] End-to-end test that an ordered collection with no entries yet falls back to timestamps, which
      is the unordered→ordered migration path
- [x] `jest src/model src/shacl` in `core` — 351 passing, 7 of them this PR's: ordering declared in
      SHACL, the `ad4m://collection_order` predicate declared, that predicate *absent* when nothing is
      ordered, a reorder counting as a change, a reorder of an *unordered* relation still counting as
      no change, and ordering surviving the SHACL link round trip
- [x] `tsc --noEmit` clean; `cargo fmt -p ad4m-executor` clean
- [x] Full CI green on the merge commit, including `integration-tests-model` — the job that drives the
      ORM end to end, which is the half of this PR that unit tests reach least well
- [x] Two tests crossing the registration boundary, each failing without its own half of the fix:
      registering a class through `add_sdna` and reading `linkedList` back off the stored shape, and
      round-tripping `ordering` through `toJSON`/`fromJSON`. These are the ones that would have caught
      the feature being inert
- [x] An ordered collection created and populated in one batch gets its entries — the real create
      sequence, flag link then setter, in a single batch. Fails without the staged-link classification
- [x] Five tests for the batch-aware setter, each checked against the code that makes it pass:
      - two `collectionSetter` runs in one batch stage each member once — without the fix,
        `[a, a, b, b, c]` where `[a, b, c]` is correct
      - `[A] → [B] → [B, C]` stages A's removal once — without the fix, `[a, a]`
      - `[A] → [B] → [A, B]` cancels A's staged removal and leaves the original link alone —
        without the fix A is removed and re-added under a fresh signature
      - `[A] → [B] → [A]` nets to nothing staged. This is the shape that needs both reconciliations to
        compose: without the removal cancel it stages `additions [a], removals [a]`, a change that
        undoes itself; without the additions prune it stages `additions [b]`, a member no longer
        wanted. Neither of the other tests catches either case
      - a persisted member removed inside a batch is not reported as still present. This one passes
        either way; it guards the removal path, which reads the persisted set rather than the
        combined one
- [x] `cargo test -p ad4m-executor --lib perspectives:: --test-threads=1` — 794 passing, 0 failing.
      `sdna` and `interpretation_e2e` are skipped: the first aborts locally under a toolchain UB check
      inside a native dependency, the second needs a working local Ollama `gemma3:12b`. Both behave
      identically on `dev` with this branch's changes stashed
- [x] **`tests/js/tests/model/model-ordering.test.ts` — 11 TS integration tests**, the layer that was
      missing and the one that found the getter-backed defect. Real `@Model` classes, real
      registration, real batched saves, read back through the ORM: entries written on create, none for
      an unordered relation, a reorder surviving the round trip, order held through `findAll` and
      through `include`, an explicit sub-order overriding it, insertion and removal keeping position,
      two reorders in a row, an unordered relation staying a set, and every member still returned when
      there are no entries yet. **Five failed before the getter fix.**
      - The `include` pair was added last, on review: that path resolves a relation separately, taking
        the parent's already-populated array, so it inherits the corrected order only because getters
        run before includes. Correct by construction — but this PR has already shipped two defects
        that were correct by construction in exactly that way.
      - The no-entries case was rewritten after its first version asserted a sequence the executor
        does not promise. It passed locally and failed on CI. It now asserts membership, and the whole
        file was run three times consecutively to confirm it is deterministic rather than lucky.
- [x] **Two agents reordering the same collection converge** — `multi-user-simple.test.ts`. Until this
      the concurrency claim rested on the algorithm alone: 28 unit tests in the ordering module, two of
      them concurrent-move cases, and nothing exercising two agents over p-diff-sync. Both agents share
      a neighbourhood and read the collection *first*, so each reorders from the same starting state —
      asserted before either writes, which is what makes it a real conflict rather than two writes that
      serialised — and then both save together. It asserts that the peers agree and that no member is
      lost, not which order wins: the surviving entry is chosen by a pid carrying the author's DID, so
      pinning an order would be asserting an accident. Polls for agreement rather than sleeping a fixed
      amount, since sync latency is the one thing here that cannot be predicted. Four consecutive runs
      at ~4s. Runs in `integration-tests-multi-user-simple`, not the model job

## New Features
* Ordered collections now preserve their configured linked-list order when displayed.
* Ordering settings are preserved when model definitions are serialized and restored.
* Collections without ordering data fall back to timestamp order.

## Bug Fixes
* Reordering an ordered relation is recognized as a change, while unordered relations remain order-insensitive.
* Collection updates correctly reconcile additions, removals, and restorations within a batch.
* Array values are handled consistently for collection and scalar properties.
* Ordering metadata remains excluded from hydrated model properties.


rust-executor/src/perspectives/perspective_instance.rs
```diff
@@ -3367,6 +3367,196 @@ impl PerspectiveInstance {
...
     }
 
+    /// The ordering strategy declared for `(source, predicate)`, if any.
+    ///
+    /// Resolved by asking which class this source is an instance of and reading
+    /// that class's shape. Class membership is not exclusive, so the most
+    /// specific match is taken. Answering `None` — because the source is not a
+    /// subject instance, or its class declares nothing — is the ordinary case
+    /// and costs one classification.
+    async fn ordering_strategy_for(&self, source: &str, predicate: &str) -> Option<String> {
```
data-bot-coasys: The head-of-list choice you flagged as worth a look — I'd widen it, and I'd do it on this PR.

As written, write-side strategy resolution consults **only the most specific conforming class**, while read-side ordering is decided by **whatever shape the query came through**. If a URI conforms to `[Special, Base]` and only `Base` declares ordering on this predicate, the setter writes no entries — and a read through `Base` expects them, falls back to timestamps, and after this PR the timestamps no longer reflect the assigned array, because preserving them is the whole point of the diff. That is the same silent-scramble failure the description works hard to rule out, arriving through the multi-class door. Overlapping classes over shared links is the normal social-DNA reuse pattern, so the door isn't theoretical.

Suggest scanning the conforming list in order and taking the first class that declares an ordering for this predicate:

```rust
let classes = self.subject_classes_of(&[source.to_string()]).ok()?;
for class_name in classes.get(source)? {
    if let Ok(shape) = self.get_shape(class_name) {
        if let Some(p) = shape
            .properties
            .iter()
            .find(|p| p.predicate == predicate && p.ordering.is_some())
        {
            return p.ordering.clone();
        }
    }
}
None
```

Specificity still wins when several classes declare a strategy; a less specific declarer just stops being invisible; the single-class case is byte-for-byte the same behaviour. The mirror hazard — two conforming classes declaring *different* strategies for one predicate — also gets a deterministic answer (most specific wins), which is the best available.

rust-executor/src/perspectives/perspective_instance.rs
```diff
@@ -3367,6 +3367,196 @@ impl PerspectiveInstance {
...
+            let status = expr.status.clone().unwrap_or_default();
+            out.push(DecoratedLinkExpression::from((expr.clone(), status)));
+        }
+        out
+    }
+
+    /// Write the ordering entries that turn the collection's current order into
+    /// the desired one.
+    ///
+    /// Additive only: entries are never removed, and a removal writes nothing at
```
data-bot-coasys, line 3527: Non-blocking, for the follow-ups list rather than this PR: additive-only means **superseded entries accumulate without bound**. A move writes ~3 entries, nothing is ever pruned, and `parse_ordering_entries` decodes every entry on every read of the parent — so a hot collection (a kanban column reordered daily) grows in both storage and per-read cost indefinitely.

Compaction looks easy to bolt on later — when winners ≪ total entries, one `generate_full` at a fresh timestamp supersedes everything and the old entries become prunable — but *who may prune* is a real design question: those links are signed by other agents. That's why I'd rather see it named in the follow-ups now than discovered when the first long-lived transcript gets slow.

rust-executor/src/perspectives/model_query/hydration.rs
```diff
@@ -57,6 +60,37 @@ pub(super) fn hydrate_instances(shape: &ModelShape, grouped: &[InstanceLinks]) -
...
+    relation_predicate: &str,
+) -> Option<Vec<String>> {
+    let entries = parse_ordering_entries(ordering_targets, relation_predicate);
+    if entries.is_empty() {
+        return None;
+    }
+    let strategy = match create_strategy(strategy_name) {
+        Ok(s) => s,
+        Err(e) => {
+            log::warn!("ordering: {e}");
```
data-bot-coasys: Nit: this `warn!` fires per ordered collection per instance per query — a `findAll` over 200 instances whose shape names an unknown strategy logs 200 warnings on every read, and the situation it reports is stable (the shape won't change between rows). Suggest `debug!` here; the write side already warns once per save in `apply_collection_ordering`, which is the actionable place.

rust-executor/src/perspectives/model_query/hydration.rs
```diff
@@ -217,7 +261,28 @@ pub(super) fn hydrate_one(shape: &ModelShape, inst: &InstanceLinks) -> Option<Va
             let mut seen = std::collections::HashSet::new();
             values.retain(|&(target, _)| seen.insert(target));
 
-            let arr: Vec<Value> = values.iter().map(|&(target, _)| decode(target)).collect();
+            // A collection the shape declares ordered is reordered here, on the
+            // ORM's own read path. Doing it only in `get_links` — as an earlier
+            // draft had it — would leave `$query`, `include` and every `findAll`
+            // unordered, which is everything the ORM actually reads through.
+            let ordered = shape
```
data-bot-coasys, line 274: Nit: two identical `find`s over `shape.properties` for the same `name`. One lookup carrying both fields reads cleaner:

```rust
let ordered = shape
    .properties
    .iter()
    .find(|p| p.name == name)
    .and_then(|p| {
        let strategy = p.ordering.as_deref()?;
        reorder_collection(strategy, &values, &ordering_targets, &p.predicate)
    });
```

data-bot-coasys (dismissed):
**APPROVE** — with one recommendation I'd take on this PR before merge (inline at `ordering_strategy_for`), two nits, and one roadmap question. None of them block the approve: the single-class case, which is everything in production today, is correct.

What I verified beyond reading the diff:

- **The hydrate-one placement really does cover the ORM read paths.** Traced `resolve_forward_include`: with no sub-`order` it iterates `parent_ids` — the array `hydrate_one` already reordered — so `include` inherits CRDT order, and an explicit sub-`order` deliberately overrides it. `$query`/`findAll` go through `hydrate_one` directly. The V3 correction holds.
- **The batch reconcile is correct through the shapes that matter.** Traced `[A] → [B] → [B, C]` (one removal staged), `[A] → [B] → [A, B]` (removal cancelled, A's link untouched), and additionally `[A] → [B] → [A]` — which nets to *zero* staged operations, the right answer, and worth the trace because neither test covers it.
- **The non-batch path passing pre-removal membership into `apply_collection_ordering` is harmless**: `diff()` reads `current_members` only as a timestamp lookup for surviving items; removed members are never consulted. I chased this as a suspected inconsistency with the batch path and it dissolves.
- **Both CodeRabbit Major findings are addressed on the head** (`7063296`, `1be9b04`), and the regression tests it asked for exist.
- CI is fully green including `integration-tests-model`, which exercises the half unit tests reach least well.

The test suite deserves a word: the six core tests each pin one behavioural claim from the description, and the two Rust e2es pin exactly the two orders that distinguish CRDT reconstruction from the timestamp fallback. This is what reviewer-guiding test evidence looks like.

jhweir:
@data-bot-coasys Thanks — all four actioned. Three commits plus one follow-ups entry, and one thing your review should know about because it changes the basis of the approve.

**Comments**

- **`ordering_strategy_for` head-of-list** (`d88084591`) — taken, as suggested. Every conforming class is now scanned in specificity order and the first declaration for the predicate wins. Your reasoning held: the read side resolves ordering from whatever shape the query came through, so a declaration on a less specific class was invisible to the writer while the reader still honoured it, and the fallback it lands in no longer follows the assigned array now that the diff preserves timestamps.
- **Unbounded entry accumulation** — added to Known follow-ups rather than fixed here, as you scoped it, including the point that *who may prune* is the hard part given the links are signed by other agents.
- **`warn!` per row** (`44f6f1f72`) — now `debug!`. The write side still warns once per save.
- **Duplicate `find`** (`de1825062`) — folded into one lookup.

**What checking the first one turned up**

`@HasMany({ ordering })` did nothing at all in production. `@Model` classes register through `ensureSubjectClass` → `SHACLShape.toJSON()` → `add_sdna` → `parse_shacl_to_links`, and `load_shape` reads the strategy back only from the `ad4m://ordering` link that chain produces. It was broken twice over, independently:

- `toJSON()` maps twenty-odd property fields and omitted `ordering`;
- Rust's `PropertyShape` had no `ordering` field, so the link could not have been emitted even had the JSON carried it.

So `ShapeProperty::ordering` was `None` everywhere: no entries written, no reordering on read. Fixed in `e5bb0dddb`.

Worth flagging against your verification notes specifically: `integration-tests-model` was green throughout, and so were the two Rust e2es and the core SHACL tests. None of them cross that boundary — the e2es build their shape through `parse_shape_from_json`, whose only other caller is its own unit test, and the core tests cover `toLinks`/`fromLinks`, which serves `addShacl` rather than the decorator path. The green CI was real but didn't reach the thing that was broken. Two new tests do cross it now, each checked to fail without its own half of the fix.

Given the approve rested on the single-class path being correct in production, and the production path wasn't running at all, this is worth another look rather than merging on the existing approval.

rust-executor/src/perspectives/perspective_instance.rs
```diff
@@ -4406,34 +4679,117 @@ impl PerspectiveInstance {
...
+                    //
+                    // `already_removed` is read after that cancellation so the
+                    // loop below does not stage a second removal for a member
+                    // this batch has already let go: the store still lists it
+                    // until commit.
+                    let mut already_removed: std::collections::HashSet<String> =
+                        std::collections::HashSet::new();
                     if let Some(ref bid) = batch_id {
                         let mut batches = self.batch_store.write().await;
                         if let Some(batch) = batches.get_mut(bid) {
```
marvin-bot-coasys, line 4710: Nit, not blocking: the three batch tests cover `[A]→[B]→[B,C]` (removal once), `[A]→[B]→[A,B]` (cancel restamp), and a straight drop. They do not cover `[A]→[B]→[A]` — which should net to **zero** staged ops (B's addition pruned, A's removal cancelled). That's the shape that proves the two `retain`s compose rather than each being right in isolation. Cheap to add next to `test_collection_setter_in_batch_cancels_a_reinstated_removal`.

marvin-bot-coasys (dismissed):
## Review — approve

The V3 correction is the load-bearing one and it is in the right place: reconstruction runs in `hydrate_one`, which is what `$query` / `include` / `findAll` actually use, not `get_links`. The two Rust e2es pin that — CRDT order vs timestamp fallback — and the TS tests pin the other half of the contract: a reorder is dirty, an unordered shuffle is not, and `toJSON()` carries `ordering` so `ensureSubjectClass` cannot drop it.

Write path: `collectionSetter` diffs membership instead of delete-all, reconciles the open batch (additions pruned, reinstated removals cancelled, already-removed skipped), then `apply_collection_ordering` is additive-only against that same batch-aware membership. The later commits on this head close the holes the earlier reviews named: strategy walks every conforming class (`d8808459`), create-in-batch classifies from staged flag links (`0b55cd1b7`) with a regression test, unknown-strategy logs at `debug` on the read path.

CI at `0b55cd1b7` is fully green, including `integration-tests-js` and `integration-tests-model`, matching green on `dev` (`37ef3959b`). Follow-up, not this PR: additive ordering entries never compact — Data already named the prune-authority question. One missing batch shape inline.

jhweir:
@marvin-bot-coasys  Added in `d75498d97` — the head has moved one commit since your review at `0b55cd1b7`.

Your reasoning held exactly. `[A] → [B] → [A]` nets to zero staged operations, and it does need both reconciliations to compose — I checked it against each half in turn rather than just watching it go green:

- without the removal cancel: `additions [a], removals [a]` — A removed and re-added, a change that undoes itself
- without the additions prune: `additions [b]` — a member no longer wanted, left staged

So it discriminates on both, and neither of the other four batch tests catches either case. Worth having for that reason and not only for the shape.

No production code changed, so your approve should still stand on the substance. `perspectives::` is 794 passing / 0 failing.

On the follow-up: entry compaction is written into Known follow-ups in the description, carrying Data's prune-authority point — those links are signed by other agents, so *who* may prune is the part that needs deciding before the mechanics matter.

marvin-bot-coasys (dismissed):
## Re-review — approve

Prior approval at `0b55cd1b7` was dismissed by `d75498d97` (`test(ordering): cover the batch round trip that nets to nothing`). That commit is exactly the one nit I left: `test_collection_setter_in_batch_nets_a_round_trip_to_nothing` drives `[A]→[B]→[A]` through `staged_after_batched_sets` and asserts additions and removals are both empty — both `retain`s composing, not each being right in isolation.

No new blockers. CI at this SHA is fully green (build-and-test, plugin-ad4m-tests, all five integration jobs, CodeRabbit). My review thread on that nit is already resolved.

Ready to merge from my side.

jhweir:
@lucksus I've added more test coverage including multi-agent concurrent writes over p-diff-sync. Regarding the point you raised earlier about looking at links in batches:

> batching only pertains to one user, so it could be a conceptual mistake

That's exactly right, and it's the reason the batch reads exist — they aren't a concurrency mechanism at all. It's read-your-own-writes inside a single agent's transaction.

`get_links` reads the SPARQL store. A batched `add_links` / `remove_link` writes only into `batch_store`. So inside an open batch the store still shows the collection as it stood before the transaction began. That never mattered while `collectionSetter` deleted everything and re-added it: that pattern overwrites unconditionally and needs no notion of current state, which is precisely why it was immune. Diffing is what introduced the dependency — a delta is only correct against the state the transaction will actually produce, and mid-batch the store isn't it.

So: not a conceptual mistake about users, but I think you're onto something about **location**. If a batch is a transaction, reads inside it arguably ought to see it, and the honest framing is that `collectionSetter` is compensating locally for something `get_links` doesn't do. A transaction-aware `get_links` is the better shape and would delete all three helpers. I didn't do it here because it changes read semantics for every caller inside a batch — Prolog, subscriptions, MCP — which is a much wider blast radius than this PR should carry. Happy to write it up as a follow-up if you agree that's the real fix.

The three commits aren't equally load-bearing, and I'd rather say so than defend all three:

- `0b55cd1b7` (classify from staged links) isn't optional. A `save()` on a new instance *is* a batch: constructor actions write the class flag links, the setter runs, then it commits. Classifying against the store alone answers "no class", so **no ordering entries get written on create at all**.
- `706329691` (read staged additions) pays for a regression this PR introduces — the old delete-all path pruned every pending addition and restated the set, so it couldn't produce duplicates; the diff can.
- `1be9b04e7` (reconcile staged removals) is genuinely deferrable. It isn't a regression — the old code also issued duplicate removals and remove/re-added a reinstated member. Happy to pull it into a follow-up if you'd rather shrink the surface.

On the tests, you were right and I've added one (`dd513afd2`). Convergence was covered at the algorithm level only — 28 unit tests, two concurrent-move cases — with nothing exercising two agents over p-diff-sync. There's now a two-agent test in `multi-user-simple.test.ts`: both share a neighbourhood, both read the collection first so each reorders from the same starting state (asserted before either writes, so it's a real conflict rather than two writes that serialised), then both save together. It asserts that the peers agree and no member is lost — not which order wins, since that's decided by a pid carrying the author's DID and pinning it would be asserting an accident. Passes, four consecutive runs.

marvin-bot-coasys:
## How CRDT ordering actually works (review + diagrams)

Status comment on `dd513afd` — explanation of the algorithm, plus the review of the batch-vs-CRDT question.

This PR is the **integration**. The CRDT itself already landed in #926 (`rust-executor/src/perspectives/ordering/`). This PR wires it into SHACL, `changedFields()`, `hydrate_one`, and `collectionSetter`.

### The one idea

A `@HasMany` is still a **set of membership links**. Order is a second, additive set of **position hints** on the parent, under `ad4m://collection_order`. Entries are never deleted. `reconstruct` ignores stale ones by keeping the highest `pid` per item.

Membership says *who is in*. Entries say *where they go*. An entry for a deleted item is traversed but not emitted (so no tombstones). An item with no entry is appended by timestamp.

```mermaid
flowchart LR
  subgraph parent["parent: kanban://col/1"]
    P((column))
  end
  A[Task A]
  B[Task B]
  C[Task C]
  P -->|"kanban://has_task (membership)"| A
  P -->|"kanban://has_task"| B
  P -->|"kanban://has_task"| C
  P -->|"ad4m://collection_order after=HEAD item=A"| EA["entry A"]
  P -->|"ad4m://collection_order after=A item=B"| EB["entry B"]
  P -->|"ad4m://collection_order after=B item=C"| EC["entry C"]
```

`pid` is `{timestamp:016}_{seq:016}_{did}` so string comparison is a total order every peer shares.

---

### Write path (one agent, one save)

`save()` still assigns the array you want. The executor diffs membership (stop restamping the whole collection) and then asks the strategy for the **minimal new entries** that make `reconstruct(current entries, new membership) == desired`.

```mermaid
sequenceDiagram
  participant App
  participant ORM as Ad4mModel.save
  participant Setter as collectionSetter
  participant Ord as apply_collection_ordering
  participant Store as perspective store

  App->>ORM: column.tasks = [C, A, B]; save()
  Note over ORM: ordered relation: reorder is a change<br/>(unordered still sorts before compare)
  ORM->>Setter: collectionSetter(desired)
  Setter->>Setter: membership diff vs persisted + staged
  Setter->>Ord: current_members, desired
  alt no prior entries
    Ord->>Ord: generate_full chain HEAD→each item
  else already ordered
    Ord->>Ord: reconstruct(projected) vs desired
    Ord->>Ord: write entry only where predecessor changed
  end
  Ord->>Store: add ad4m://collection_order links (additive)
  Note over Store: same batch_id as membership writes<br/>local atomicity only — not the CRDT
```

`diff` is **not** longest-increasing-subsequence. `after` is a pointer: moving B from `[A,B,C,D]` to `[A,C,B,D]` also changes C's predecessor. Comparing predecessors costs about three entries for one move (the item, the one it left, the one it landed behind). Full-regenerate would restamp the whole list.

---

### Read path (every peer, independently)

`$query` / `include` / `findAll` never go through `get_links`. Reconstruction lives in `hydrate_one`, because that is what the ORM actually reads.

```mermaid
flowchart TD
  Q["model_query / findAll"] --> H[hydrate_one]
  H --> M["collect membership links<br/>by relation predicate"]
  H --> O["collect ad4m://collection_order targets<br/>on the same parent"]
  O --> P["parse_ordering_entries scoped to that predicate"]
  P --> R["LinkedListStrategy.reconstruct"]
  M --> R
  R --> L["latest_per_item: highest pid wins per item"]
  L --> F["after → followers map"]
  F --> D["DFS from LIST_HEAD<br/>forks: higher pid first"]
  D --> E["emit only if still a member"]
  E --> T["unpositioned members appended<br/>oldest timestamp first"]
```

Degrades instead of failing: unknown strategy or empty entries → ordinary unordered read.

---

### The actual CRDT: concurrent inserts are a fork

Two agents inserting after the same predecessor do **not** lock. They each add an entry. Reconstruction linearises the fork the same way on every peer.

Start: `HEAD → A → B → C`

Alice moves C to the front. Bob inserts X after A. Neither rewrites the other's items.

```mermaid
flowchart TD
  HEAD["LIST_HEAD"]
  A["A"]
  B["B"]
  C["C Alice: after=HEAD new pid"]
  X["X Bob: after=A new pid"]
  HEAD --> C
  HEAD --> A
  A --> B
  A --> X
```

Walk: start at HEAD. Two followers of HEAD (C and A) — higher `pid` first. Then A's followers (B and X) — higher `pid` first. Every peer gets the same array without talking.

**Why not regenerate the whole chain in one batch:** a batch is one agent, uncommitted local state. It cannot isolate you from Bob. Full-regen writes fresh entries for A, B, C even if you only moved C. Bob's concurrent entry for B then races yours on an item you never meant to touch. Diff only emits entries whose predecessor actually changed, so concurrent edits collide only on items both agents moved.

**Why batches still exist in this PR:** not the CRDT.

1. `staged_triples_for` — create path: constructor flags and the collection setter share one uncommitted batch, so "is this ordered?" must see staged class flags. Needed. Keep.
2. `links_including_batch` / second-save-in-same-batch — local ORM transaction bookkeeping (`save(batchId)` twice before commit). Not how two users merge. Keep only if we still support that API.

---

### Review

The integration is the right shape: reconstruct on the ORM read path, diff on write, additive entries, no tombstones, membership remains the source of truth. That is a real CRDT, and it is not the batch helpers.

I would merge the algorithm as designed. I would want `links_including_batch` justified as "the ORM already allows two saves in one batch," not as concurrent-agent machinery. If we do not care about that API for ordered collections, that slice can go; the CRDT does not need it.

`changedFields()` treating reorder as dirty for ordered relations (and still sorting unordered ones) is load-bearing — without it `save()` is a no-op for the one edit this feature exists to support.

Declaring `ad4m://collection_order` on the SHACL shape is also load-bearing: the instance query's predicate filter would otherwise drop the very links reconstruction needs.

lucksus (approved):

PR #929, opened 2026-08-25
flow: schema + @Model + FlowInstance wrapper + flow-aware extraction

data-bot-coasys:
# flow: schema + @Model + FlowInstance wrapper + flow-aware extraction

Foundation PR for the flow-implementation arc. Ships the schema, the on-graph shape, the object-oriented `FlowInstance` API, and the data path that lets the extraction LLM *see* live flows on a scope.

## What "flows" are, and why this PR

A **flow** is a state machine encoded in SDNA: a `SHACLFlow` declares named `states`, `transitions` between them, and (v5, this PR) per-state `requires: ModelQuery[]` guards that describe what evidence on the graph would let the flow move into that state. Concrete team examples:

- **Delivery flow** for a task: `identified → scoped → in_progress → in_review → done`. `requires` on `scoped` = "at least one Task with a scope description linked to this Task".
- **Deliberation flow** for a proposal: `proposed → perspectives_expressed → tension → overlap → resolution`. `requires` on `overlap` = "at least two `AgreementPoint` instances linked to this proposal".

A `SHACLFlow` is the schema. A **`FlowInstance`** is a running instance: a state-machine cursor anchored to some `subject` URI (a Task, a Proposal, a Message, …), tracking `currentState`. Multiple `FlowInstance`s per flow definition are the normal case (one per Task, one per Proposal).

After this PR, an AD4M perspective can:

1. **Declare v5-shape flows** — TS and Rust round-trip every new field (`inputTypes`, `outputTypes`, `interpretationHint` at flow + state scope, `requires: ModelQuery[]`, `semanticCheck`, `consensusRule`, `creationHint`, `context`).
2. **Mint `FlowInstance` nodes** on any subject URI via `FlowInstance.start(perspective, flow, subject)` — returns the new object-oriented `FlowInstance` wrapper.
3. **List running instances** filterable by `flowName` and/or by `subject` via `FlowInstance.findAll(perspective, flowName?, subject?)`.
4. **Read flow state** through the wrapper — `currentState`, `availableTransitions`, `proposals()` all resolve without extra round-trips because the wrapper carries the parsed `SHACLFlow` alongside the on-graph `FlowInstanceRecord`.
5. **See active flows in the extraction prompt** — the interpretation engine's extraction pass reads every active flow's context on the scope and injects an "## Active flows on this scope" block into the LLM prompt.

## Schema shape

```mermaid
classDiagram
    class SHACLFlow {
        +String name
        +String? interpretationHint
        +String[] inputTypes
        +String[] outputTypes
        +ConsensusRule? consensusRule
        +String? creationHint
        +Map~String,Any~? context
        +FlowState[] states
        +FlowTransition[] transitions
    }
    class FlowState {
        +String name
        +String? interpretationHint
        +ModelQuery[]? requires
        +String? semanticCheck
        +ConsensusRule? consensusRule
    }
    class FlowTransition {
        +String fromState
        +String toState
    }
    class ModelQuery {
        +String className
        +ModelQueryCount? count
        +Map~String,PropertyCondition~? where
        +String? didProperty
        +ModelQuery[]? or
    }
    class ConsensusRule {
        +int n
        +String? fromRole
    }
    SHACLFlow "1" *-- "0..*" FlowState
    SHACLFlow "1" *-- "0..*" FlowTransition
    FlowState "1" o-- "0..*" ModelQuery : requires
    SHACLFlow "1" o-- "0..1" ConsensusRule
    FlowState "1" o-- "0..1" ConsensusRule

    class FlowInstanceRecord {
        <<@Model on-graph>>
        +String id
        +String flow
        +String subject
        +String currentState
    }
    class FlowInstance {
        <<OO wrapper>>
        +SHACLFlow flow
        +FlowInstanceRecord record
        +String currentState
        +FlowState[] availableTransitions
        +Promise~FlowTransitionProposal[]~ proposals()
        +static Promise~FlowInstance~ start(...)
        +static Promise~FlowInstance[]~ findAll(...)
    }
    FlowInstance --> FlowInstanceRecord
    FlowInstance --> SHACLFlow

    class FlowTransitionProposal {
        <<@Model on-graph>>
        +String id
        +String flowInstance
        +String fromState
        +String toState
        +String proposer
        +String[] evidence
        +String evidenceHashes
        +String? rationale
    }
    FlowInstance ..> FlowTransitionProposal : proposals()
```

## Flow-aware extraction data path

```mermaid
sequenceDiagram
    autonumber
    participant Runner as interpretation::run
    participant Loader as flow_context::gather_active_flow_contexts
    participant Perspective as PerspectiveInstance
    participant Builder as build_interpretation_input
    participant LLM as extraction LLM

    Runner->>Loader: gather(perspective, scope)
    Loader->>Perspective: load_shacl_flows() → catalogue
    Loader->>Perspective: load_flow_instances(subject) → records
    Loader->>Loader: build_flow_contexts(records, flows)
    Loader-->>Runner: Vec<FlowContext>
    Runner->>Builder: build_interpretation_input(shapes, transcript, existing, flow_contexts)
    Note over Builder: appends "## Active flows on this scope"<br/>block per FlowContext, with:<br/>- flow name + interpretationHint<br/>- current state<br/>- reachable next-states rendered<br/>  with English `requires` + hints<br/>- $flow.base / $flow.instance tokens<br/>  + per-flow legend
    Builder-->>Runner: prompt (JSON)
    Runner->>LLM: prompt
    LLM-->>Runner: extracted instances (+ future: flow_proposals hints)
```

## Onion shells

| Shell                        | What it exercises                                                                                             |
|------------------------------|---------------------------------------------------------------------------------------------------------------|
| Schema parsing               | TS `SHACLFlow.toLinks/fromLinks/toJSON/fromJSON` round-trip every v5 field; Rust `parse_flow_from_links` / `parse_flow_to_links` mirror them; unit tests locked |
| @Model on-graph shape        | `FlowInstanceRecord` + `FlowTransitionProposal` @Model classes round-trip via the executor's model-query path; hardwired-SDNA parity tests lock TS shape ↔ Rust JSON field names |
| OO wrapper                   | `FlowInstance` wrapper carries parsed `SHACLFlow` + `FlowInstanceRecord`; `currentState`, `availableTransitions`, `proposals()` accessors                        |
| Runtime mint                 | `FlowInstance.start` + Rust `mint_flow_instance` primitive; integration test spins a live executor, mints an instance, `FlowInstance.findAll(subject: uri)` reads it back |
| Context gather               | Rust `load_flow_instances` + `load_shacl_flows` + `build_flow_contexts` compose into `gather_active_flow_contexts(perspective, scope)`; 42/42 unit tests in the split `flow_context` submodules |
| Prompt integration           | `build_interpretation_input` inserts the "## Active flows on this scope" block; `SYSTEM_PROMPT` teaches the LLM to bias toward advancing them; both extraction call sites now thread real `Vec<FlowContext>` |

## Breaking changes

Five methods removed from `PerspectiveProxy` (deleted in `ec9a98ced` as part of the v5 flow API migration):

- `startFlow(flowName, exprAddr)` → replaced by `FlowInstance.start(perspective, flow, subject)`
- `flowState(flowName, exprAddr)` → replaced by `FlowInstance.findAll(...)` + `instance.currentState`
- `flowActions(flowName, exprAddr)` → replaced by `instance.availableTransitions`
- `runFlowAction(flowName, exprAddr, actionName)` → replaced by `instance.proposeTransition({toState, reason})` + `instance.fireIfConsensus(consensusRule)` (the mutation surface lands in the next PR)
- `expressionsInFlowState(flowName, stateValue)` → replaced by `FlowInstance.findAll(perspective, flowName).filter(i => i.currentState === stateName)`

Two `SHACLFlow`-shape fields removed:

- `FlowState.stateCheck` — state is now stored on-graph via `FlowInstanceRecord.currentState`, not inferred from a link pattern.
- `SHACLFlow.startAction` — replaced by `FlowInstance.start`, which mints the on-graph record in `states[0]` (the ascending-`value` initial state).

Docs updated in this PR: `docs-src/pages/social-dna.mdx` Flows section + `PerspectiveProxy.addFlow` JSDoc `@example` both use the v5 surface.

## What is NOT in this PR

Each is its own testable shell → its own PR:

- **Deterministic post-processing** (`run_engine_proposal_pass` — engine catches transitions the LLM missed by re-evaluating `requires` on the committed graph after extraction) → PR #932.
- **`semanticCheck` 2nd LLM pass** (targeted small confirmation gate after `requires` matches, fail-safe on Ambiguous) → PR #932.
- **Harness-path attribution** — `{FlowName}_propose_transition` tool-call surface for tool-calling extractors → PR #940.
- **Mutation methods** on `FlowInstance` — `proposeTransition`, `accept`, `reject`, `fireIfConsensus` currently throw stubs; land with the consensus engine loop in the next stacked PR (`feature/flow-mutations`).
- **Subscription hooks** — `onStateChange`, `onProposalAdded`, `onProposalResolved` — same next PR.

## Test plan

- [x] TS unit: `pnpm --filter @coasys/ad4m-core test` — schema round-trip + FlowInstance wrapper accessors
- [x] Rust unit: `cargo test -p ad4m-executor --lib perspectives::flow_context perspectives::flow_classes perspectives::shacl_parser`
- [x] Integration: mint FlowInstance via `PerspectiveProxy.startFlowInstance`, read back via `FlowInstance.findAll` with subject filter
- [x] Integration: extraction pass injects "## Active flows on this scope" block with correct context
- [x] Real-LLM smoke (nightly): the added prompt block doesn't malform LLM output on `gemma3:12b`
- [x] Docs: `pnpm --filter @coasys/ad4m-docs build` clean
- [ ] CI green on merge target


core/src/shacl/SHACLFlow.ts
```diff
@@ -208,6 +313,15 @@ export class SHACLFlow {
...
     });
 
+    // Top-level interpretation hint (optional, drives AI state suggestion)
+    if (this.interpretationHint !== undefined) {
+      links.push({
+        source: flowUri,
+        predicate: "ad4m://interpretationHint",
+        target: Literal.from(this.interpretationHint).toUrl()
+      });
+    }
```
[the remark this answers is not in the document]
data-bot-coasys: Fixed in `b95be9b88`. All three sites (`this.interpretationHint`, `state.interpretationHint`, `state.semanticCheck`) now use a truthy check so empty strings emit zero predicates. The `toJSON` path got the same tightening. New regression test `empty-string interpretationHint / semanticCheck emit zero predicates` locks it in.

core/src/shacl/SHACLFlow.ts
```diff
@@ -350,6 +495,18 @@ export class SHACLFlow {
...
+    // Find top-level interpretation hint
+    const flowHintLink = links.find(l =>
+      l.source === flowUri && l.predicate === "ad4m://interpretationHint"
+    );
+    if (flowHintLink) {
+      try {
+        flow.interpretationHint = Literal.fromUrl(flowHintLink.target).get() as string;
+      } catch {
+        // Ignore parse errors
+      }
```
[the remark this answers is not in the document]
data-bot-coasys: Fixed in `b95be9b88`. Decoded literals now guarded with `typeof === 'string'` before assignment (both flow-level and per-state hints, plus `semanticCheck`). Added `isModelQueryShape` helper that requires each entry to have a non-empty `className: string` — rejects `[null]`, `[{}]`, `[{className:42}]`, `[{className:""}]`. `fromJSON` path picked up the same validation so an untrusted JSON blob can't sneak past. Two new regression tests: `malformed decoded literals leave the field unset (fromLinks)` (uses `literal:number:` and `literal:json:` to actually decode as non-string) + `malformed \`requires\` array entries leave the guard unset` (covers all four rejected shapes plus a positive control).

data-bot-coasys:
CodeRabbit follow-up on lines 891-907 (states not sanitized in `toJSON()`) — fixed in `976a63a0d`. `states` now goes through a per-state `.map` that strips empty `interpretationHint`, empty `semanticCheck`, and empty `requires` — same rule `toLinks()` uses, so JSON and links agree on "unset". Extended the existing `empty-string interpretationHint / semanticCheck emit zero predicates` test with `json.states[0]` assertions plus a `requires: []` case. 26/26 SHACLFlow tests green.

core/src/index.ts
```diff
@@ -35,7 +35,7 @@ export * from "./ai/Tasks"
 export * from "./runtime/RuntimeTypes"
 export { Model as AIModel, ModelApi, TokenizerSource, LocalModel, ModelType, ModelApiInput, TokenizerSourceInput, LocalModelInput, ModelInput, VoiceActivityParamsInput } from "./ai/AITypes"
 export * from './shacl/SHACLShape'
-export { SHACLFlow, FlowState, FlowTransition, LinkPattern, FlowableCondition } from './shacl/SHACLFlow'
+export { SHACLFlow, FlowState, FlowTransition, LinkPattern } from './shacl/SHACLFlow'
```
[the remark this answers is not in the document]
data-bot-coasys: Fixed in `a2dc174b0` — `ModelQuery` and `PropertyCondition` added to the `SHACLFlow` barrel re-export on line 38. tsc `--noEmit` clean.

data-bot-coasys:
CodeRabbit outside-diff review (SHACLFlow.ts 204-211, "Validate `or` branches recursively in isModelQueryShape") — fixed in `e50915800`. Guard now recurses into a non-undefined `or`: must be an array, every entry must itself satisfy `isModelQueryShape`. Regressions:

- `consensusRule.fromRole.or` with a `null` branch — rejected via fromLinks + fromJSON
- `requires[0].or` with a `null` branch — rejected via fromJSON

Same all-or-nothing sanitisation policy as the existing empty-string / typed-array guards. 36/36 SHACLFlow tests green locally.

core/src/perspectives/FlowModels.ts
```diff
@@ -0,0 +1,147 @@
...
+  /**
+   * RFC3339 timestamp when the proposal was written.
+   * Named `proposedAt` (not `createdAt`) because `Ad4mModel` reserves
+   * `createdAt` / `updatedAt` / `baseExpression` as synthetic hydration
+   * fields — a subclass property with those names is silently overwritten
+   * by the earliest / latest link timestamp on hydration, and the TS-side
+   * `jsonToModelInstance` converts any `createdAt` value to epoch millis.
+   */
+  @Property({ through: "ad4m://flow/created_at", required: true })
+  proposedAt: string = "";
```
lucksus: Why do we need this - as something different to the built-in createdAt?

core/src/perspectives/FlowModels.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// ── FlowInstance ────────────────────────────────────────────────────────────
+// On-graph node minted by the engine when a flow is started on a specific
+// base expression. Its URI is the value that `FlowTransitionProposal.flowInstance`
+// references. Mirrors `rust-executor/src/perspectives/hardwired_sdna/flow_instance.json`.
+
+@Model({ name: "FlowInstance" })
+export class FlowInstance extends Ad4mModel {
```
lucksus: Ah, I thought we wanted to introduce a new base class like `Ad4mModel` form flows. I get that we need this Ad4mModel subclass and shacl shape too. But instead of relying on functions in `PerspectiveProxy`, having TS class that encapsulates all the flow handling in an object-oriented way..

core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -1140,6 +1139,82 @@ export class PerspectiveProxy {
...
+     * `FlowInstance` wrapper class (§4.3, wrapper-side lands in the following
+     * slice).
+     *
+     * @param flowName - Name of a `SHACLFlow` already registered on the perspective
+     * @param baseExpression - URI of the subject expression the flow runs on
+     * @throws When the flow is unknown or has zero declared states (use the
+     *   zero-state action-flow path — coming with §6.3 fireAction — for that shape).
+     * @returns The freshly-minted `FlowInstance` (hydrated).
+     */
+    async startFlowInstance(flowName: string, baseExpression: string): Promise<FlowInstance> {
```
lucksus: ...so that this, for instance, would be a method on `Ad4mFlow` (or so)

rust-executor/src/perspectives/flow_context.rs
```diff
@@ -0,0 +1,1677 @@
...
+            id: "literal://string:channel-x".to_string(),
+            predicate: "ad4m://hasChild".to_string(),
+        };
+        assert_eq!(scope_subject(&r), "literal://string:channel-x");
+    }
+}
+
+// ============================================================================
+// Slice 10.3d — end-to-end integration test
+// ============================================================================
```
lucksus: This file is too big. At least the e2e tests should have their own file. But even without, 1.4k lines should be split up into submodules.

rust-executor/src/perspectives/flow_context.rs
```diff
@@ -0,0 +1,1677 @@
...
+// This test is the onion shell for the current cut — the natural stopping
+// point Nico asked for on 2026-08-26 evening ("cut based on what we can
+// test really, stack it like onions"). If it stays green, slice 10.3c's
+// `run.rs` wiring is by construction correct: the two lines it replaced
+// (`build_interpretation_input(shapes, transcript, &existing, &[])`) now
+// receive the exact `Vec<FlowContext>` this test builds, and slice 10.2's
+// unit tests already cover what happens on that value inside the prompt
+// builder.
+//
+// No LLM is spun up — `setup_perspective_no_llm` gives a real
```
lucksus: I think this PR/onion layer should have real LLM based interpretation tests already.

lucksus (changes requested):
Questions and change-requests below. The design doc is referenced but missing from the changeset. Either explain things inline - or add the design doc.

data-bot-coasys (commented):
Thanks Nico — all five change-requests read. No blockers on my side; here's the plan I'm executing:

1. **`proposedAt` vs built-in `createdAt`** (`FlowModels.ts:102`) — right, we don't need it. Same logic for `startedAt` on `FlowInstance`. The reserved-field bug I hit earlier (`e6362e5ca`) was solved by renaming, but the actual right answer is drop these fields and rely on `Ad4mModel`'s built-in link-timestamp hydration. Removing both.

2. **`Ad4mFlow`-style OO wrapper** (`FlowModels.ts:111` + `PerspectiveProxy.ts:1161`) — you're right, and the design doc §4.3 already specs it (I under-implemented). Renaming @Model class `FlowInstance` → `FlowInstanceRecord` (aligns with existing Rust `FlowInstanceRecord`), introducing new OO wrapper `FlowInstance` per design §4.3 with `currentState` / `availableTransitions` / `proposals` / `history` / `evidence` accessors + `proposeTransition` / `accept` / `reject` methods. Subscription hooks (`onStateChange` etc.) land after the state-changed event exists in slice 10.6 — noting as TODO with the design doc line reference. Renaming `startFlowInstance` → `startFlow(shape, base)` returning the wrapper.

3. **`flow_context.rs` split** — 1,677 lines is fair game. Splitting into `flow_context/mod.rs` + submodules: `types.rs` (records + FlowContext + NextStateSummary), `render.rs` (English rendering), `loader.rs` (parse_flows_from_bag + load_shacl_flows + load_flow_instances + gather_active_flow_contexts), `tests_e2e.rs` (integration).

4. **Real-LLM interpretation test at this layer** — agreed, doable via Marvin tunnel + OpenAI-compatible ModelApiInput (same pattern as #879 S8). Will add a `#[ignore]`-by-default test that hits Marvin's Ollama and asserts an instance is minted from a transcript that references an active flow. Guarded on `MARVIN_LLM_URL` env, so CI passes without it and local/manual runs verify the full stack.

5. **Design doc missing from changeset** — adding `docs/design/flow-interpretation-hints.md` (canonical copy of the v5 doc) in this PR so referenced sections are inline.

Executing now, will push as I go. Standing-orders discipline: silent unless blocked.

— Data

core/src/perspectives/FlowModels.ts
data-bot-coasys: R2/R3 landed in `feb2bfb9b`: new `FlowInstance` OO wrapper in `core/src/perspectives/FlowInstance.ts` — carries the parsed `SHACLFlow` + on-graph record and exposes `uri / subject / flowName / currentStateName / startedAtMillis / currentState / availableTransitions / proposals()`. Raw @Model class renamed `FlowInstance` → `FlowInstanceRecord` (SDNA name unchanged, parity tests still lock the same target class). `startFlowInstance` / `getFlowInstances` on `PerspectiveProxy` now return `FlowInstance[]` wrappers. Mutations (`proposeTransition / accept / reject / fireAction`) + subscription hooks (`onStateChange / onProposalAdded / onProposalResolved`) throw typed `not yet implemented (slice 10.6 — consensus engine)` errors so call-sites surface a clear "not yet" rather than silent no-op. 6 new wrapper tests (currentState-as-object, availableTransitions filter, terminal-state, empty proposals, typed stub errors, record passthrough). `tsc --noEmit` clean, `pnpm run build` from core/ clean.

rust-executor/src/perspectives/flow_context.rs
data-bot-coasys: R5 landed in `e2d651a71`: new `rust-executor/src/perspectives/flow_context/real_llm_e2e.rs` — real OpenAI-compatible-endpoint test that exercises the whole extraction pipeline with slice 10.3c's flow-aware wiring live. Seeds Delivery flow + mints FlowInstance at `identified`, sanity-asserts `gather_active_flow_contexts` returns it (else the test would measure nothing), runs `run_interpretation` against real Ollama with a scoping conversation + `Scope::Model` narrowing, asserts >=1 Task extracted (the minimum bar for "LLM output survived the injected active-flow prompt block"). Uses standard `setup_interpretation_e2e` harness — env-overridable via `INTERPRETATION_E2E_BASE_URL / _MODEL / _API_KEY`, defaults to Ollama at localhost:11434 with gemma3:12b. Retries up to 3× to soak single-sample LLM flake, matching `interpretation_e2e`'s policy. **Verified locally against Marvin's Ollama** via SSH tunnel: first attempt passed in 4.95s, extracted 3 Tasks with active-flow block in prompt. What this deliberately does NOT prove (deferred to PR #932): engine acts on extracted instance (post-processing); LLM proposes a FlowTransitionProposal on its own (`flow_proposals` field). CI runs on Marvin so the default endpoint hits Ollama locally there.

rust-executor/src/perspectives/flow_context.rs
data-bot-coasys: R4 (split flow_context.rs, 1677 lines → 4 submodules) landed in `adcdbbfa8`. R5 (real-LLM e2e) landed in `e2d651a71` as a 5th sibling — `rust-executor/src/perspectives/flow_context/real_llm_e2e.rs`. The e2e file (`e2e_tests.rs`, no-LLM composition proof) stays separate from `real_llm_e2e.rs` (real-LLM full pipeline proof) so they can run independently.

data-bot-coasys:
@lucksus all 5 change-requests now addressed (six commits, head `e2d651a71`):

| # | Ask | Fix | Commit |
|---|-----|-----|--------|
| R1 | `proposedAt` vs built-in `createdAt` (FlowModels.ts:102) | Drop `startedAt` / `proposedAt` — Ad4mModel synthesises `createdAt` on hydration | `0387ef718` |
| R2/R3 | `Ad4mFlow` OO API; method moves off `PerspectiveProxy` | New `FlowInstance` wrapper class in `core/src/perspectives/FlowInstance.ts` — carries parsed `SHACLFlow` + on-graph record; exposes `uri/subject/flowName/currentStateName/startedAtMillis/currentState (FlowState)/availableTransitions/proposals()`; mutations + subscription hooks throw typed "not yet implemented — slice 10.6" errors. Raw @Model class renamed `FlowInstance → FlowInstanceRecord` (SDNA name unchanged, parity tests still lock the same target class). `startFlowInstance` / `getFlowInstances` return `FlowInstance[]` wrappers | `feb2bfb9b` |
| R4 | `flow_context.rs` too big (1677 lines) | Split into 4 submodules: `types.rs` (data shapes) + `render.rs` (pure ModelQuery→English) + `loader.rs` (perspective I/O) + `e2e_tests.rs` (composition proof, no LLM) | `adcdbbfa8` |
| R5 | Real-LLM interpretation tests already | New `rust-executor/src/perspectives/flow_context/real_llm_e2e.rs` — seeds Delivery flow + mints FlowInstance + runs `run_interpretation` against real Ollama with scoping conversation + `Scope::Model` narrowing. **Verified against Marvin: first attempt passed in 4.95s, extracted 3 Tasks with active-flow block in prompt.** | `e2d651a71` |
| R6 | Design doc missing from changeset | Added `docs/flow-interpretation-hints-design.md` — the v5 authority the module doc comments cite | `e7536295f` |

Local sanity: `cargo fmt --all --check` clean, `cargo check --tests --release` clean, TS `pnpm run build` from `core/` clean, `tsc --noEmit` from `tests/js` — no new errors on `flow-models.test.ts`.

CI running on the pushed head — watching. Ready for re-review whenever.

rust-executor/src/perspectives/flow_context/e2e_tests.rs
```diff
@@ -0,0 +1,260 @@
...
+            source: identified_uri.to_string(),
+            predicate: Some("ad4m://interpretationHint".to_string()),
+            target: lit("The team has named a piece of work but has not yet scoped it."),
+        },
+        Link {
+            source: scoped_uri.to_string(),
+            predicate: Some("ad4m://interpretationHint".to_string()),
+            target: lit("The team has agreed on what the work is and can begin execution."),
+        },
+    ];
```
lucksus: Why are these links extra and not coming in through `parse_flow_to_links(&delivery_flow_json(), "Delivery")`?

rust-executor/src/perspectives/shacl_parser.rs
```diff
@@ -1050,36 +1422,477 @@ mod tests {
...
+            assert!(state.consensus_rule.is_none());
+        }
+    }
+
+    /// Full v5-shape read — hand-built links matching what
+    /// `core/src/shacl/SHACLFlow.ts::toLinks()` emits. Catches drift
+    /// between the TS writer and the Rust reader (which are the two
+    /// halves Model C hangs on).
+    #[test]
+    fn parse_flow_from_links_reads_v5_predicates() {
```
lucksus: Why do we have versioning of Flow types to the degree of different function names? What we're writing now is the only relevant shape going forward.

lucksus (changes requested):

rust-executor/src/perspectives/flow_context/e2e_tests.rs
data-bot-coasys: Fixed in `cd8d4f5be` (R7). `parse_flow_to_links` now emits every declared field (`interpretationHint` at flow + state scope, `inputTypes` / `outputTypes`, `requires`, `semanticCheck`, `consensusRule` at both scopes, `creationHint`, `context`) — mirrors the TS `SHACLFlow.toLinks`. The `v5_links` hand-append block is gone from both `e2e_tests.rs` and `real_llm_e2e.rs`; the Delivery flow is now seeded as one JSON blob round-tripped through the writer.

rust-executor/src/perspectives/shacl_parser.rs
data-bot-coasys: Fixed in `683f10a77` (R8). Every v4/v5 tag dropped from code + tests. Renames:

- `test_parse_flow_with_v5_fields` → `test_parse_flow_full_field_set_deserialises`
- `test_parse_flow_v4_shape_still_works` → `test_parse_flow_omitting_optional_fields_still_works`
- `parse_flow_from_links_roundtrips_v4_writer_output` → `parse_flow_from_links_roundtrips_minimal_flow`
- `parse_flow_from_links_reads_v5_predicates` → `parse_flow_from_links_reads_all_predicates_from_hand_built_links`

Also added `parse_flow_from_links_roundtrips_full_flow` — closes the writer→reader loop the "hand-built links" test only proved one side of, now that the writer emits every field.

Design doc left alone since its version history is legitimately historical (the doc itself evolved through v1–v5); code doesn't need to carry that.

lucksus (dismissed):

jhweir (commented):
@data-bot-coasys I had Claude review this branch. It seems well architectured but a few issues were raised that I think it's worth looking at before we merge.

# Review — ad4m PR #929 (flow: schema + @Model + FlowInstance wrapper + Model C data path)

Reviewed the branch locally (`pr929`, 23 files, +7.2k). The design is genuinely good — my concerns are almost all about where the wiring diverges from it.

## What's right

**Model C is the correct choice**, and the wiring is appropriately conservative: `active_flows` is omitted entirely when empty, so a perspective with no flows gets a byte-identical prompt. That's the right risk posture for touching the extraction hot path, and it's test-locked.

**`requires: ModelQuery[]` over the earlier SPARQL `guardQuery`** is the strongest call in the doc — guards talk about subject classes, reuse the ORM, and the same evaluator can back both transitions and `availableActions`. Roles-as-ontology-classes (§7.2) follows from it for free and genuinely delivers "new role types without code changes."

**Zero-state-flows-are-actions** is a real unification, not a framing trick. Like → moderated Like → Deliberation as one continuum is the kind of primitive that earns its keep.

The `flow_context` split (`types` / `render` / `loader`, render pure) is the right seam, and the silent-degradation discipline in `gather_active_flow_contexts` is correct for something on the extraction path.

## Architectural problems

**1. The scope filter is bolted to the wrong config field.** `run.rs:625` passes the dedup `scope` straight into `gather_active_flow_contexts`, and `scope_subject` filters `FlowInstance.subject == scope.id`. But `existing_scope` is documented in `config.rs:152-159` as the *parent* node of a dedup subtree — "existing items live under THIS project node." Both branches are wrong:

- `None` (the default the watcher passes unless an operator configured otherwise): every `FlowInstance` on the entire perspective goes into every extraction prompt. Unbounded, uncapped. Design §12 flagged exactly this ("prompt-budget for AutoProcessor scope union… hard cap on unique row count") and none of it is here.
- `Some`: filters on the parent, so flows running on the *children* — the actual Tasks and Proposals — are excluded. That's the entire Model C use case.

The right scoping key is already in the watcher: `item_ids` / the drained batch bases. Also, §3 principle 1 says scope is "base item + everything graph-dangling from it"; exact `subject ==` equality doesn't implement that at all.

**2. Two disjoint answers to "what state is this flow in."** `startFlow` / `flowState` / `flowActions` / `runFlowAction` / `expressionsInFlowState` still read and write link markers via `stateCheck`. `startFlowInstance` / `FlowInstance.currentState` read `FlowInstanceRecord.currentState`. Nothing reconciles them — `startFlowInstance` never runs `startAction`, `runFlowAction` never touches `currentState`. A perspective can be in `doing` by markers and `ready` by record simultaneously. This needs deciding *before* 10.6 starts firing transitions, or the consensus engine writes one representation while existing UI reads the other. Relatedly: the design says `stateCheck` is removed, but `core/src/shacl/SHACLFlow.ts` keeps it **required** on `FlowState`, so every v5-shape flow has to supply a dummy link pattern.

**3. `availableFlows` now inverts its own semantics.** Post-`flowable` retirement it returns a flow iff `inputTypes` is empty or contains `"any"` — so flows that declare real typed inputs, the whole point of v4/v5, are *never* available. Only untyped and wildcard flows are. The doc comment calls concrete matching "a follow-up," but this is a live public API returning a confidently wrong set rather than a partial one. The expression's class is knowable; either implement it here or make it throw.

**4. `$flow.base` and `$did` reach the LLM verbatim.** `render_model_query` in `rust-executor/src/perspectives/flow_context/render.rs` stringifies `where` values as-is, so the prompt reads `at least 1 match of coasys://Scope where forTask = "$flow.base"`. Every worked example in §6 uses that token. Substitution is trivial at this layer — `FlowContext` already carries `subject` and `instance_uri`. Since *this* PR ships the prompt, it belongs here rather than in #932.

**5. Flow identity is the bare name.** `FlowInstanceRecord.flow` is `"Delivery"`, and the join in both `build_flow_contexts` and `getFlowInstances` is keyed on that name. The design's own `AvailableAction.flow` is a URI, and `SHACLFlow` already has `namespace` + `flowUri`. For shareable social-DNA modules composed from different communities — the stated goal — name collision is when-not-if, and it silently binds an instance to the wrong shape. Store `flowUri`, keep `name` as the label.

**6. Two mint paths, one dead.** `mint_flow_instance` and `ensure_flow_model_classes` are called only from tests — no WS-RPC exposure, despite `flow_classes.rs:16` claiming it. The live path is TS `FlowInstanceRecord.create`. Fine as forward-staging, but the module-level `#![allow(dead_code)]` will hide drift between them, and the header overclaims today's state.

## Smaller wrinkles

- `getFlowInstances` calls `Ad4mModel.registerAll` on a **read** path — a query mutates the perspective, which in a neighborhood is a synced write to every peer. The Rust loader already handles "Shape not found" gracefully; the TS read could too.
- `FlowInstance` exports seven methods that unconditionally throw while typed as returning values. The design doc is the record of intent; a `Promise<FlowTransitionProposal>` that always throws is a type-level lie the caller finds at runtime. I'd leave them off until 10.6.
- `startFlowInstance` seeds `flow.states[0]`, relying on the new sort-by-value in `fromLinks`. That convention isn't declared anywhere on `SHACLFlow` — no `initialState` field, no normative doc. A flow whose lowest-value state isn't its entry point starts silently wrong.
- `render_model_query` ignores `linkedTo`, so the LLM never learns "linked to the flow" vs "linked to the base" — which is precisely what separates evidence from noise in the §6.1 Deliberation example.
- On R5: I'd take your own offer and delete the real-LLM smoke test. It burns an LLM call in CI for coverage `interpretation_e2e.rs` already provides, and the flow-specific assertion lands in #932 anyway.

## Net

#1 and #2 are the ones I'd block on — #1 because the headline feature is either unbounded or filtering on the wrong node depending on config, and #2 because the dual state representation gets much more expensive to unwind once the consensus engine starts writing. #3 and #4 are small fixes with outsized effect on whether this is usable end-to-end. Everything below that is fine to carry forward.

jhweir:
@data-bot-coasys Went back through the branch at `f9df951b0` and checked each point from my Aug 31 review against the code as it stands. The two I flagged as blockers are properly fixed, and #3 / #4 are done too — nice work. A few things are still open, one of them new since the dev merge.

One process note first: there's no reply from you on the thread, so the PR reads as unanswered even though all the work is there. Everything below I had to reconstruct from commit messages. A summary comment on the PR would help whoever reviews next.

## Confirmed fixed

**#1 — scope filter** (`4dd5d2dec`). `gather_active_flow_contexts(perspective, subjects: &[String])` now derives its subjects from `cursor.sources`, with empty → `Ok(vec![])`. That's the right key and the right bounded default. I checked it survived the `run.rs` conflict in the `f9df951b0` merge — both the strategy path (`run.rs:643-658`) and the harness path (`run.rs:889-900`) still carry it.

Your commit body defers two sub-items — widening to the §3 principle-1 graph closure, and the §12 prompt budget cap. Agreed, both are separate slices. The unbounded-default problem is genuinely gone, which was the blocking part.

**#2 — dual state representation** (`d1198b1ca`, then `ec9a98ced`). Going all the way to deleting the legacy path rather than keeping both behind the disjointness guard is the better outcome. One representation of flow state, no drift to unwind when 10.6 starts firing transitions.

**#3 — `availableFlows`** (`f0012788b`, `a87cbf68b`). Resolving through `subjectClassesOf([exprAddr])` and intersecting with `inputTypes` is exactly it, and keeping the empty / `"any"` branch preserves the old wildcard behaviour.

**#4 — `$flow.base` reaching the LLM** (`b099f4fae`). The token-preserving shape with a per-instance `tokens` legend is better than the substitution I asked for — cheaper prompts and less URI hallucination, and it composes with the post-LLM substitution pass in 10.6. Your reasoning for leaving `$did` alone holds: the acting DID isn't a property of the flow instance. Worth noting it now reaches the prompt as the one sentinel with no legend entry, so if the LLM ever sees a `$did` guard it has nothing to resolve it against — fine to leave, just flagging it as a known edge.

**Wrinkle: throw-stubs** (`8411ca8ba`). Removed, and moving instance lifecycle onto `FlowInstance.start` / `.findAll` statics is a cleaner surface than what I was arguing about.

## Still open

**1. Real-LLM test is ungated — new drift from the merge.** #943 landed on dev while this branch was open: all 24 real-LLM Rust tests now carry `#[ignore = "llm-e2e"]` and run in the nightly `llm-e2e` workflow instead. `real_llm_e2e.rs::model_c_real_llm_extraction_with_active_flow_in_prompt` doesn't have the attribute, so it still hits Ollama on every push to this branch.

This is more than convention drift. Nico's reason for #943 wasn't only CI duration — it's that a queue of heavy LLM jobs on Marvin monopolises the GPU that Flux and WE need for live transcription and interpretation. An ungated per-push LLM test on this branch reintroduces exactly the contention he just removed, and it does so on the branch most likely to see rapid pushes as 10.6 lands. Please add `#[ignore = "llm-e2e"]` before merge.

Related, and worth a look while you're in there: Nico says the nightly didn't fire last night. Two candidates, both in `.circleci/config.yml` on dev:

- `nightly-llm-e2e` uses the legacy `workflows.<name>.triggers.schedule` form. CircleCI has moved scheduling to scheduled pipelines, and legacy scheduled workflows have stopped firing on a lot of orgs. If that's it, the workflow simply never gets created and there's nothing to find in the UI.
- The skip-if-unchanged preflight from `7a61ae4e3` calls `circleci-agent step halt` when the Insights lookup returns a `commit_id` matching `CIRCLE_SHA1`. If `CIRCLECI_API_TOKEN` from the `nightly-llm-e2e` context isn't resolving, or the Insights response shape isn't what the parser expects, a bad `LAST_SHA` comparison could halt the job. That failure mode reports green in seconds, which reads as "didn't run" rather than "failed".

Checking whether a pipeline was created at all separates the two — no pipeline points at the trigger, a green-in-seconds job points at the preflight.

**2. Public docs describe deleted methods.** `docs-src/pages/social-dna.mdx:419-440` still documents `startFlow`, `flowState`, `flowActions`, `runFlowAction` and `expressionsInFlowState`, all removed in `ec9a98ced`. Also: removing five public methods from `core` is a breaking change and the PR description doesn't say so. Both worth fixing here rather than leaving for someone to trip over.

**3. `registerAll` still on the read path.** `FlowInstance.ts:167` — `findAll` registers before querying, so a read mutates the perspective, which in a neighbourhood is a synced write to every peer. The Rust loader already degrades gracefully on "Shape not found"; the TS read can too.

**4. `#5` flow identity is still the bare name.** `FlowInstanceRecord.flow` is still `ad4m://flow/flow_name` holding `"Delivery"`, and both `build_flow_contexts` and `FlowInstance.findAll` still join on it. No commit touches this. I'd rather do it here than later: it's a wire-format change to `FlowInstanceRecord`, and the cost goes up as soon as instances exist in real perspectives. Store `flowUri`, keep `name` as the label — `SHACLFlow` already carries both.

**5. `#6` dead mint paths.** `mint_flow_instance` / `ensure_flow_model_classes` are still test-only and `#![allow(dead_code)]` is still at `flow_classes.rs:18`. Fine as forward-staging. But the module header has drifted further: it still cites "the WS-RPC exposure of `startFlowInstance` (client mirror)" as a live caller — that exposure never existed, and after `8411ca8ba` the TS method doesn't either. Please correct the header to say what's actually true today.

**6. `linkedTo` is dropped on the way to the prompt.** It's parsed into the Rust struct (`shacl_parser.rs:189`) and round-trips through storage, but there are no references to it in `render.rs`, so the LLM never learns "linked to the flow" vs "linked to the base". That's the distinction that separates evidence from noise in the §6.1 Deliberation example.

**7. `states[0]` as the initial state.** Partially addressed — the comment at `SHACLFlow.ts:904` now explains the sort-by-value. Still no `initialState` field or normative statement on `SHACLFlow` itself, so a flow whose lowest-value state isn't its entry point still starts silently wrong. A declared field would be better than a convention documented at the parse site.

## Where that leaves it

CI is green across all seven jobs on the head, and the branch is mergeable — just needs an approving review.

Items 1–3 are mechanical and I'd like them in this PR — the `#[ignore]` attribute especially, since without it merging this puts the GPU contention straight back. The nightly investigation is separate and shouldn't hold the PR up. Item 4 is the one with real cost-of-delay. Items 5–7 I'm happy to carry forward into #932 if you'd rather keep this one moving; say so explicitly in a comment and I'll take it.

data-bot-coasys:
@jhweir Thanks — fair point on the missing thread reply. Summary of everything since your Aug 31 review, with head now `7f54ea5d8`:

## Confirmed fixed (your Sept 1 list)

| # | Item | Commit(s) |
|---|------|-----------|
| 1 | Scope filter — `gather_active_flow_contexts(subjects: &[String])` | `4dd5d2dec` |
| 2 | Docs describe deleted methods + breaking-change note | `7f54ea5d8` (this push) |
| 3 | `registerAll` on `FlowInstance.findAll` read path | `5b018d5c9` (J#7) |
| 4 | Flow identity → `flowUri` (not bare name) | `5b018d5c9` (J#5) |
| 5 | `mint_flow_instance` module-level `allow(dead_code)` + header overclaim | `5b018d5c9` (J#6) |
| 6 | `linkedTo` rendering in prompt sentences | `5b018d5c9` (J#10) |
| 7 | `states[0]` as initial state — normative doc | `5b018d5c9` (J#9) |

## Sept 1 item 1 detail — LLM gate

`f0074baf8`: added `#[ignore = "llm-e2e"]` to `real_llm_e2e::model_c_real_llm_extraction_with_active_flow_in_prompt` and wired it into `.circleci/config.yml`'s nightly `llm-e2e` workflow alongside the other 24 tests gated in #943. Per-push Ollama contention on this branch is gone.

## Sept 1 item 2 detail — docs (this push)

- `docs-src/pages/social-dna.mdx` Flows section rewritten to the v5 API: `new SHACLFlow(...)` + `addState`/`addTransition` Delivery example, `FlowInstance.start` / `.findAll` / `.proposeTransition` / `.aggregateVotes` / `.fireIfConsensus` / `.selectFireCandidate` for lifecycle, `perspective.addFlow(name, flow)` for registration. Callout points at the v5 design doc.
- `PerspectiveProxy.ts` `addFlow` JSDoc @example: was still using `stateCheck` on states + `startAction` on the flow + `action: 'addLink'` effects — all removed from `SHACLFlow` in `ec9a98ced`. Now uses only fields that exist on `FlowState` / `FlowTransition` / `SHACLFlow` today.
- **Breaking change in PR description**: I'll update the description in a moment to call out the five deleted `core` methods (`startFlow`, `flowState`, `flowActions`, `runFlowAction`, `expressionsInFlowState`) and the `stateCheck`/`startAction` deprecation.

## Nightly investigation — deferring per your suggestion

Agreed both candidates you flagged (legacy `workflows.<name>.triggers.schedule` vs preflight-halt) are separate work. I'll open a separate issue with the diagnostic split (pipeline-created? vs green-in-seconds?) so this PR doesn't hold on it.

## Items 5–7 status

Per your offer to carry forward: items 5 (dead mint paths) / 6 (`linkedTo` in prompt) / 7 (`states[0]` convention) are all in `5b018d5c9` on this PR — no carryover needed. `linkedTo` in particular has 5 new unit tests covering `"flow"` / `"base"` / `{ via, to }` shapes + unknown-shape silent-fallback.

CI green across all 7 jobs on `7f54ea5d8`. Ready for re-review.

/cc @lucksus

jhweir (approved):

Issue #21, opened 2021-08-27
Add proper port checking for ad4m-executor, such that holochain, gql & dapp server will not collide with other processes running on ports


jdeepee:
- Port for above processes should be returned somehow after calling `.init()`

PR #930, opened 2026-08-25
feat: link server + server-link-language

HexaField:
### What

This PR adds a self-hosted **link server** and its matching **server-link-language** — a complete alternative sync backend to Holochain for AD4M perspectives. The link server stores and relays perspective diffs. The language handles commits, sync, queries, peers, telepresence, and mandatory end-to-end encryption.

### Why

Holochain-based sync works but carries operational overhead (conductor process, DHT bootstrap, DNA compilation). A lightweight HTTP + WebSocket server gives operators a simpler deployment path for private or small-group perspectives. The link server runs as a single process, stores data in SQLite, and needs no external dependencies beyond Node.js.

E2E encryption closes a trust gap in the original design: a malicious server operator can no longer read link data or retain plaintext room keys.

#### New packages

| Package | Path | Role |
|---|---|---|
| **link-server** | `link-server/` | Fastify HTTP + WebSocket server. SQLite (better-sqlite3) storage. DID challenge-response auth → JWT sessions. OR-Set link store with XOR-based incremental revision. Diff log with configurable retention. Rate limiting (sliding-window, in-memory). Telepresence routing. Docker deployment. |
| **server-link-language** | `bootstrap-languages/server-link-language/` | AD4M link language (Deno sandbox). Implements perspective-commit, perspective-sync, perspective-query, peers, and telepresence. Talks to one link-server instance per room. |

#### E2E encryption flow

```mermaid
sequenceDiagram
    participant Admin as Admin Client
    participant Server as Link Server
    participant Member as Member Client

    Admin->>Admin: Generate AES-256-GCM room key locally
    Admin->>Server: GET /acl → member X25519 public keys + ownership signatures
    Admin->>Admin: Verify each member's X25519 signature before sealing
    Admin->>Admin: Seal room key to each verified member (ECIES: ephemeral X25519 + HKDF-SHA256 + AES-256-GCM)
    Admin->>Server: POST /keys/rotate {keys: [{did, encryptedKey}]}
    Server-->>Server: Store sealed envelopes (never sees plaintext)
    Member->>Server: GET /keys → sealed envelope for this DID
    Member->>Member: Open envelope with own X25519 private key
    Member->>Member: Encrypt link data with room key before commit
    Member->>Server: POST /commit {additions: [{data: {ciphertext, nonce}, link_hash}]}
```

**X25519 key derivation:** The Deno sandbox never exposes raw Ed25519 private keys. The language derives a deterministic X25519 keypair by signing a fixed domain-separated message (`SHA-256(sign(FIXED_MESSAGE))` → X25519 private scalar). The public key piggybacks on the DID auth flow along with a signature proving ownership.

**X25519 ownership verification:** During DID auth step 2, the client signs its X25519 public key hex via `agentSignStringHex(x25519PublicKeyHex)` and sends the signature to the server. The server verifies via `verifyHex(ed25519PubKey, hashMessageForVerify(x25519PublicKeyHex), x25519Signature)` and stores both key and signature. Before sealing room keys, the admin client retrieves the ACL (which now includes X25519 signatures) and verifies each member's key ownership — preventing a malicious server from substituting its own X25519 key to intercept sealed room keys.

**Key versions:** Each rotation increments a version counter. Links carry their `key_version`. The admin auto-grants historical versions to late-joining members via `POST /keys/grant`.

**Admin transfer:** `POST /rooms/:roomId/admin/transfer` transfers admin privileges to another room member. The endpoint validates that the target holds room membership, prevents self-transfer, and updates the room's `admin_did` atomically. The former admin retains membership but loses admin privileges (key rotation, ACL management, grants).

#### Sync architecture

```mermaid
flowchart LR
    subgraph Client["Language Client"]
        C[commit] --> TW[toWireDiff<br/>encrypt if E2E]
        TW --> POST[POST /commit]
        WS_IN[WS onDiff] --> AIW[applyInboundWireDiff<br/>single choke point]
        HTTP_IN[catchUp] --> AIW
        AIW --> FWD[fromWireDiff<br/>decrypt] --> Store[local store]
        AIW --> Emit[emitPerspectiveDiff]
    end
    subgraph Server["Link Server"]
        POST --> DB[applyDiffAndAppend<br/>SQLite transaction]
        DB --> WS_OUT[broadcast to room]
        SYNC[GET /sync?since=N] --> DIFFS[diff log]
    end
```

**Dual inbound paths:** WebSocket pushes and HTTP polling both funnel through `applyInboundWireDiff` — the single function that applies diffs, updates the store, advances the sequence cursor, and emits. No second code path.

**Batched commits:** Burst `addLink` calls (e.g., 1500 in a stress test) coalesce into minimal POST segments. Adjacent diffs merge when safe (no add/remove conflict on the same link identity). Failed segments re-enqueue with exponential backoff.

**Diff ordering:** `applyDiffAndAppend` processes additions before removals within a single transaction. The XOR revision self-inverse property guarantees that simultaneous add+remove of the same link leaves it removed and the revision returns to `EMPTY_REVISION`.

**Diff retention overflow:** The server caps stored diffs per room via `maxDiffsPerRoom`. Clients that fall behind the pruned window recover via `GET /render` (a full bootstrap snapshot of all active links) rather than replaying the missing diff history.

#### Trust model — `link_hash` for encrypted links

Encrypted links carry a `link_hash` computed from the plaintext before encryption. The server stores and compares these hashes without accessing the plaintext. This design relies on two properties documented in `types.ts`:
1. A client that encrypts a link and a client that decrypts it will agree on the hash — because the hash derives from the canonical plaintext fields, not the ciphertext.
2. The server cannot forge a valid `link_hash` without knowing the plaintext — because SHA-256 preimage resistance prevents deriving the plaintext from the hash alone.

#### Files changed

**link-server/src/:**
- `auth.ts` — DID challenge-response, JWT session issuance/verification, ACL enforcement
- `routes.ts` — REST API: auth, commit, sync, render, revision, peers, ACL, key rotation, key grant, missing-keys query, admin transfer
- `db.ts` — SQLite schema + prepared statements. OR-Set link store, append-only diff log, session management, room key storage, X25519 signature column, admin transfer. WAL mode.
- `encryption.ts` — ECIES key wrapping primitives (ephemeral X25519 + HKDF-SHA256 + AES-256-GCM). Test-use exports only.
- `ws.ts` — WebSocket connection registry, first-message auth, room broadcast, telepresence signal routing, ping/pong keepalive
- `rate-limit.ts` — Sliding-window rate limiter (in-memory, single-process)
- `telepresence.ts` — Online agent tracking with grace period for reconnects. `scheduleOffline` timer callback wrapped in try/catch.
- `types.ts` — Shared types (LinkExpression, PerspectiveDiff), canonical link hashing (with trust model documentation), XOR revision
- `server.ts` — Server factory (`buildServer`), wires all components, session sweep timer
- `index.ts` — CLI entry point

**link-server/tests/:**
- `e2e.test.ts` — end-to-end auth, ACL, key rotation, telepresence. Includes admin-loss scenario (non-admin gets 403 on admin operations), X25519 signature validation (invalid signature rejected at auth), and ACL response X25519 signature inclusion.
- `commit-sync.test.ts` — commit, sync, revision convergence, removal. Includes `applyDiffAndAppend` ordering invariant test and diff retention overflow + bootstrap recovery test.
- `admin-transfer.test.ts` — admin transfer: success, old admin loses privileges, non-member rejected, self-transfer rejected, non-admin rejected, new admin can rotate keys.
- `link-hash-invariant.test.ts` — cross-package hash consistency (server `linkHash` matches client `hashLink`), canonical link payload for encrypted links, null/undefined predicate equivalence.

**bootstrap-languages/server-link-language/src/:**
- `sync.ts` — HTTP commit/catch-up, batched commit coalescing, `applyInboundWireDiff` choke point, bootstrap, `_pendingMissingVersions` retry set
- `encryption.ts` — X25519 key derivation from signing capability, room key sealing/opening, key ring, link encrypt/decrypt (AES-256-GCM). Includes `verifyX25519Ownership(did, x25519PublicKeyHex, signatureHex)` with inline base58 decoder and Ed25519 verification.
- `api.ts` — Typed HTTP client for every server endpoint. Auth step 2 sends `x25519Signature`.
- `auth.ts` — DID challenge-response client, token caching, re-authentication. Computes `x25519Signature = agent.signStringHex(x25519Hex)` during auth.
- `store.ts` — Local KV-backed link store with source/target/predicate indexes, sync cursor (revision + sequence)
- `ws-client.ts` — WebSocket client with reconnect backoff, first-message auth
- `telepresence.ts` — Client-side online agent tracking, signal/broadcast dispatch
- `adapters.ts` / `adapters-deno.ts` — Dependency injection for Deno sandbox (storage, transport, agent, runtime, WebSocket factory)
- `types.ts` — Wire types, encrypted link detection. `AclMember` and `MemberMissingKeys` carry `x25519Signature`.

**bootstrap-languages/server-link-language/:**
- `index.ts` — `performAdminKeyGrants()` and `performRotation()` verify X25519 ownership before sealing room keys for each member.
- `tests/encryption.test.ts` — `verifyX25519Ownership` test suite: valid signature accepted, wrong signature rejected, substituted key rejected, malformed DID rejected, empty signature rejected.

**Other:**
- `.circleci/config.yml` — `integration-tests-js` job builds + publishes the server-link-language, starts a link-server, and runs neighbourhood tests against both holochain and server-link backends
- `rust-executor/src/perspectives/perspective_instance.rs` — Diagnostic logging in `commit()` demoted from `info!` to `debug!`
- `tests/js/` — Integration test harness: `linkServer.ts` (start/stop), `linkLangConfig.ts` (server-link variant config), `neighbourhood.ts` (parameterised tests run against both backends)

#### Test coverage

| Suite | Count |
|---|---|
| **link-server** (server-side) | 108 tests |
| **server-link-language** (client-side) | 117 tests |
| **Total** | 225 tests, 0 failures |

#### References

- Supersedes [#919](https://github.com/coasys/ad4m/pull/919) (wrong base branch)
- E2E client-side key generation: [#984](https://github.com/coasys/ad4m/pull/984) (merged into this branch)
- Prior review: [#919 comments](https://github.com/coasys/ad4m/pull/919)

#### Known limitations

- **Single-process rate limiting** — In-memory sliding window. Acceptable per the self-hosted single-process brief; would need shared state (Redis, SQLite) for horizontal scaling.


lal-bot-coasys (commented):
Full audit of #930 against my 25 findings on #919. Head SHA is `23c83802986f6a6b29842adaad95854447f0451b` — same commit both PRs point at, so what merges from #930 is exactly what was in #919's final state.

## What Josh addressed — 10 fixes, well-executed

Verified each against current HEAD. All five server-side fixes and four of five client-side fixes are correct.

**Server:**
- ✅ `db.ts` — all 30+ prepared statements now cached in `stmts`, all inline `.prepare()` calls replaced. Correct pattern for better-sqlite3.
- ✅ `/sync?since` clamped to `>= 0` at `routes.ts:233` (`Math.max(parsedSince, 0)`). No upper-bound cap or pagination, but the negative-integer full-scan is closed.
- ✅ `@fastify/websocket` registered with `maxPayload: 1 MiB` at `server.ts:65-73`. (See caveat below — half-open sockets / no ping-pong is a separate part of #8 that's still open.)
- ✅ `FederationManager.stop()` now `async`, tracks `reconcileInFlight`, awaits before returning. `server.ts:117` updated to `await federation.stop()`. Correct.
- ✅ `TelepresenceManager.close()` fires pending `onExpired` callbacks before clearing. Structure changed to store `{timer, onExpired}` tuples. Correct.

**Client:**
- ✅ `hexToBytes` — strict `/^[0-9a-fA-F]{2}$/` regex per byte pair, `parseInt('ag', 16)` case rejected.
- ✅ `getValidToken` — in-flight `_authInFlight` promise cache, second concurrent caller awaits the first's `authenticate()`. `resetAuth()` clears it. Correct.
- ✅ `void wsClient.connect()` — `.catch()` attached with a log line.
- ⚠️ `roomKeyStatus === "error"` — retries `setupRoomKey()` **once** inline at commit time before refusing. My finding asked for (a) retry on each WS reconnect, (b) retry on each `sync()` with backoff, (c) surface a `LinkLanguageEvent`. What's shipped is the minimum to break the permanent-poison — every subsequent commit attempts a fresh setup. Acceptable, but rapid-fire commits during outage each pay the retry cost, and the sync loop still doesn't retry independently.
- ⚠️ `WsClient.close()` — logs a warning when dropping queued messages. My finding asked for either synchronous flush or explicit rejection of the queued promises. The caller still doesn't know its `send()` returned successfully-but-lost. This makes the loss observable to operators, not to code.

And the previous commit `5282b9b2f` addressed the `flushBatch` auto-reschedule half of finding #17 (retry storm). The `drainCommitBatch` half is still open — see below.

## What is NOT addressed — 13 findings

Categorized by severity, with current line references.

### 🔴 Critical — the E2E stack is still incompatible with itself

`bootstrap-languages/server-link-language/AGENTS.md:121-138` **still declares in Josh's own words**: *"E2E encryption has 5 confirmed incompatibilities with the server."* Enumerated in the file:

1. **Encrypted link wire shape mismatch** (my finding #3, `routes.ts:198-200` still uses `isEncryptedLinkData({ciphertext, nonce})` while client sends `{encrypted: hex(nonce||ciphertext)}`).
2. **KDF mismatch** — server `SHA-256(shared)`, client `HKDF-SHA256(shared, salt, info, 32)`.
3. **Key envelope framing** — server `JSON.parse(row.encrypted_key)` at `routes.ts:400` sends an object; client `decodeSealedEnvelope` at `encryption.ts:219` calls `JSON.parse(atob(encoded))` expecting a base64 string. `atob(object)` coerces to `"[object Object]"` and throws (my finding #2).
4. **X25519 derivation still mismatched.** Josh partially addressed this by adding `ctx.db.setX25519PublicKey` at `routes.ts:161-163` — the client's derived pubkey now reaches the server's ACL. But `encryption.ts:47` and `encryptRoomKeyForDid` at `:56-72` **still use `edwardsToMontgomeryPub(didToPublicKey(did))` instead of the stored ACL value**. The wire path is complete; the read at seal time isn't. So the ACL row is populated-and-ignored, and the mismatch stands (my finding #1, half-fixed).
5. **`encryptedKey` object vs string** — same as #3 above.

Every commit into an E2E-enabled room fails with one of these mismatches. AGENTS.md's own advice: *"Fixes localize to `src/api.ts` (response shaping) and `src/encryption.ts` (envelope framing) — core sync logic stays unchanged."*

**Plaintext rooms probably ship; E2E rooms do not.** If launch means "E2E is a checkbox in the room-create dialog", the checkbox is a trap.

### 🔴 Critical — federation security posture shift, undocumented

Finding #7 (`link-server/src/auth.ts:174` — `verifyLinkExpression` verifying raw payload while AD4M signs `SHA-256(payload)`) has an interesting resolution: **`verifyLinkExpression` no longer exists**. Josh added `hashMessageForVerify` (`auth.ts:72`) which correctly hashes-then-verifies for the auth challenge (`routes.ts:137-138`), but the per-link verification was **removed entirely**. `routes.ts:189-198` now says:

> *"Link signatures travel as metadata — the server stores and relays them as-is. JWT auth (bound to the agent's DID) proves identity at the transport layer. Downstream consumers can verify signatures if they choose; the server does not need to."*

This is a defensible design choice — receivers verify, senders are JWT-authenticated. But three things are load-bearing and none of them are true today:

1. **`AGENTS.md:66` still says** *"Every individual link inside a federated diff is still independently verified against its own author's signature (`verifyLinkExpression`)"* — this is now false. The doc lies about the security model.
2. **Federation compounds the issue.** `handleIncomingFederate` (`federation.ts:173-197`) verifies the peer server's signature over the *envelope* but does not verify individual link signatures inside `body.diff`. A compromised or malicious peer server can inject links with arbitrary `author` fields — recipients get them via WS broadcast tagged as if from those authors.
3. **Receivers who trust the server aren't warned.** No known client-side documentation says "you must verify link signatures yourself" — a caller reading a link from the language reasonably assumes signature validity.

Either (a) put per-link verification back in with the right hashing, or (b) commit fully to the JWT-plus-peer-signature model, update AGENTS.md to say so, and add a client-side verification hook or an explicit `signaturesUnverified: true` field so downstream doesn't get to be surprised.

### 🔴 Critical — federation replay & TOFU race

- **#5 TOFU-before-verify race** (`federation.ts:133-136`) — UNFIXED. `resolvePeerTrust` still pins `serverPublicKey` into the row when `!byUrl.peer_public_key`. Admin `addPeer(roomId, https://peerB)` while peerB is briefly down → row inserted with null pubkey → attacker who reaches the endpoint first pins their key.
- **#6 replay attack surface** (`types.ts:96-112` `canonicalFederationPayload`) — UNFIXED. Signed payload still contains only `{kind, roomId, diff, sequence, revision}`. No timestamp, no nonce. `remove(X)` from six months ago can be replayed today.

### 🟠 High

- **#9 `/federate` and `/reconcile` skip rate limiting** (`routes.ts:349, 372`) — UNFIXED. Every other room endpoint has `jwtRateLimit(ctx.rateLimits.roomJwt)`; these two have no preHandler at all. Unauthenticated attacker sends 10k links per second, each triggering `verifyPeerSignature` (ed25519).
- **#12 `parseJwtExpiryMs` binary-string decode** (`auth.ts:37-42`) — UNFIXED. `atob(padded)` still treated as binary. Not triggered by numeric `exp` but any future field with UTF-8 content silently corrupts.
- **#8 websocket half-open TCP + ping-pong** (`server.ts`) — HALF-FIXED. `maxPayload` shipped; ping/pong keepalive still absent. Half-open sockets after NAT rebind accumulate in `this.byId`/`this.rooms`, presence entries leak, broadcasts amplify.

### 🟡 Medium

- **#16 `sync.ts:239` failed-batch requeue has no time-based retry** — UNFIXED. Requeued diffs sit in `_pendingQueue` forever if the caller stops calling.
- **#17 `drainCommitBatch` retry storm on teardown-during-outage** (`sync.ts:360-378`) — HALF-FIXED. `flushBatch` no longer auto-schedules (5282b9b2f). But `drainCommitBatch` still loops 100 laps, and each lap re-schedules a `flushBatch` when `_pendingQueue > 0 && !_flushScheduled`. During outage that's 100 × 3-attempt bursts = ~300 POSTs.

### 🔵 Style / cleanup

- **#21 `adapters-deno.ts:60` regex-parses `httpFetch` error string** — UNFIXED. Fragile implicit contract; PR #915's typed API was the right fix.
- **#22 `index.ts:15` violates AGENTS.md convention** — UNFIXED. `AGENTS.md:72-74` still says *"src/adapters-deno.ts is the only file allowed to import @coasys/ad4m-ldk, and only index.ts is allowed to import adapters-deno.ts"* — and `index.ts:15` still has `import { defineLanguage, hash } from "@coasys/ad4m-ldk";`. Either amend the rule or move the import.
- **#23 `ws.test.ts` collectMessages-after-open in 6 places** — UNFIXED per my finding at lines 92, 132, 163, 189, 223, 254.
- **#24 hand-rolled `SlidingWindowLimiter`** — UNFIXED. Still 70 lines of manual bookkeeping where `@fastify/rate-limit` covers the ground.
- **#25 `canonicalLinkPayload` scattered** (`link-server/src/types.ts:96` vs client-side `linkIdentity` at `bootstrap-languages/server-link-language/src/sync.ts:323`) — UNFIXED. Two divergent serializations of the same identity.

## Verdict

I don't think this is ready for launch as-is. Two paths depending on the actual launch scope:

- **Plaintext-only launch (E2E behind a feature flag defaulted off):** most of the 🔴 E2E items go dormant. The security-posture-shift (removed `verifyLinkExpression`) still needs an AGENTS.md update and a client-side documentation pass; the TOFU race and replay window still need patching; the federation endpoints still need rate limiting. Doable in a day.
- **E2E-enabled launch:** blocked by the five AGENTS.md-documented incompatibilities. The one Josh half-fixed (X25519 wire path in place, but seal-time read still uses the wrong derivation) is small and self-contained. The other four are the "localize to `src/api.ts` and `src/encryption.ts`" pass Josh's own doc calls out. Not launched-yet-open questions — deferred bugs someone already characterised.

Josh's 10 fixes are clean. But "all your review comments addressed" is optimistic — 13 remain, including all six of my 🔴 critical items. If Nico wants an approval to unblock a plaintext-only launch, I'm willing on the strength of the fixes shipped, but the AGENTS.md-declared E2E gaps need to be either (a) fixed, (b) hidden behind an off-by-default feature flag with a clear "E2E is preview" label, or (c) explicitly acknowledged in this PR body as known unshipped.

Happy to draft any of the remaining fixes — the small ones (`/federate` rate-limit, `resolvePeerTrust` TOFU tightening, `canonicalFederationPayload` timestamp+nonce, AGENTS.md docs) are 20-line PRs each. Ping and I'll cut them.

bootstrap-languages/server-link-language/src/adapters-deno.ts
```diff
@@ -0,0 +1,156 @@
...
+            },
+            onClose(cb: (code: number, reason: string) => void): void {
+                ws.addEventListener("close", (event: CloseEvent) => cb(event.code, event.reason));
+            },
+            onError(cb: (err: unknown) => void): void {
+                ws.addEventListener("error", (event: Event) => cb(event));
+            },
+        };
+    }
+}
```
lucksus, line 156: Should we include this in the ALDK?

bootstrap-languages/server-link-language/src/auth.ts
```diff
@@ -0,0 +1,110 @@
...
+ * this function can't parse is simply treated as non-expiring (refreshed
+ * lazily on the next 401).
+ */
+export function parseJwtExpiryMs(token: string): number | null {
+    const parts = token.split(".");
+    if (parts.length !== 3) return null;
+    try {
+        const b64 = parts[1].replace(/-/g, "+").replace(/_/g, "/");
+        const padded = b64 + "=".repeat((4 - (b64.length % 4)) % 4);
+        const payload = JSON.parse(atob(padded)) as { exp?: number };
```
lal-bot-coasys: **#12 still open (latent, not a bug today).** `atob(padded)` decodes base64url-payload bytes as a **binary string**, not UTF-8. Fine today because only `exp` (numeric) is read at line 38 — numbers are ASCII-safe. The moment anyone adds another JWT claim that could contain non-ASCII (e.g. a DID that isn't purely ASCII, a user-supplied `sub`), the parse silently corrupts the field and returns `null`, which cascades into `expiresAt=null` → `getValidToken` never proactively refreshes → subtle sync stalls on 401 cascades.

Fix is one line: `import { decodeJwt } from 'jose'` (already a dep on the server side, pull into the client) → `const payload = decodeJwt(token) as { exp?: number };`. Handles UTF-8 payload decoding correctly.

bootstrap-languages/server-link-language/src/sync.ts
```diff
@@ -0,0 +1,485 @@
...
+            const requeue = segments.slice(i);
+            const requeueSummary = requeue.map((s, idx) =>
+                `  segment ${i + 1 + idx}: ${s.additions.length} adds, ${s.removals.length} removes`)
+                .join("\n");
+            console.warn(
+                `[server-link-language] segment ${i + 1}/${segments.length} failed after ` +
+                `${MAX_COMMIT_ATTEMPTS} attempts; re-enqueueing ${requeue.length} segment(s) ` +
+                `for next flush cycle:\n${requeueSummary}`,
+            );
+            _pendingQueue.unshift(...requeue);
```
lal-bot-coasys, line 319: **#16 still open (non-blocking).** Failed segments get requeued (`_pendingQueue.unshift(...requeue)`) but no timer is scheduled. The comment thread above documents this is intentional — the next `enqueueCommitBatched` naturally picks them up. But if a caller commits once and then stops (Alice makes one edit during an outage, tab stays idle for hours), the requeued diff sits in memory until either (a) another commit happens, (b) the app tears down and loses the diff, or (c) a fresh page load with a full snapshot supersedes it. `(b)` and `(c)` mean the change silently doesn't ship.

Cheap fix: on requeue, `setTimeout(() => scheduleFlush(), backoff)` with exponential backoff (start at 5s, cap at 5min). Doesn't need persistence to close the common case (short outage recovers before the tab closes).

bootstrap-languages/server-link-language/index.ts
```diff
@@ -0,0 +1,404 @@
...
+ * peers, ACL, E2E key exchange) and native WebSocket (real-time diff
+ * push, telepresence, presence). Implements every link-language
+ * capability: perspective-commit, perspective-sync, perspective-query,
+ * peers, and telepresence.
+ *
+ * See README.md for architecture and AGENTS.md for operational notes /
+ * known limitations.
+ */
+
+import { defineLanguage, hash } from "@coasys/ad4m-ldk";
```
lal-bot-coasys, line 15: **#22 still open (convention violation).** `AGENTS.md:68-70` says *"`src/adapters-deno.ts` is the only file allowed to import `@coasys/ad4m-ldk`, and only `index.ts` is allowed to import `adapters-deno.ts`."* This line imports `@coasys/ad4m-ldk` directly, in `index.ts`, so index becomes a **second** direct importer of the LDK — which is what the rule exists to prevent (the rule is what makes swapping the LDK for a Node runtime later possible).

Either move `defineLanguage` / `hash` re-exports into `adapters-deno.ts` and import from there, or update AGENTS.md to reflect the exception (e.g. "`index.ts` imports the module surface — `defineLanguage`, `hash` — directly; only I/O adapters go through `adapters-deno.ts`"). Whichever way, the convention and the code should say the same thing.

link-server/src/rate-limit.ts
```diff
@@ -0,0 +1,69 @@
...
+ * process per the project brief.
+ */
+export interface RateLimitCheck {
+  allowed: boolean;
+  /** Seconds the caller should wait before retrying, only meaningful when !allowed. */
+  retryAfterSec: number;
+  remaining: number;
+}
+
+export class SlidingWindowLimiter {
```
lal-bot-coasys, line 13: **#24 still open (nice-to-have).** 69 lines of manual sliding-window counting + `setInterval` sweep where `@fastify/rate-limit` (v10, matches Fastify 5) does the same thing with a config object. Not urgent, but the hand-rolled version has grown its own bookkeeping (`.close()`, `.unref()`, the sweep timer) that's now maintenance surface. Post-merge — worth a separate PR if this stays as a hot code path.

bootstrap-languages/server-link-language/src/sync.ts
```diff
@@ -0,0 +1,485 @@
...
+            segments.push({ additions: [...next.additions], removals: [...next.removals] });
+        } else {
+            current.additions.push(...next.additions);
+            current.removals.push(...next.removals);
+        }
+    }
+    return segments;
+}
+
+function linkIdentity(link: LinkExpression): string {
```
lal-bot-coasys, line 415: **#25 still open (coupling documented, not eliminated).** `linkIdentity` reproduces `canonicalLinkPayload` (`link-server/src/types.ts:99`) on the client side. The comment at line 327 now honestly says *"server's canonicalLinkPayload ... exactly so that two peers ..."* — the coupling is visible, which is an improvement. But two implementations of the same serialization contract, hand-maintained, will drift the first time anyone touches either.

Proper fix eventually: extract the canonical-payload spec into a shared package both sides depend on. Not this PR (would balloon the diff), but worth an issue so the next person touching either function notices before adding a field only one side knows about.

lal-bot-coasys (dismissed):
Re-review vs my 13 open findings from 2026-08-26. **11 addressed, 5 latent/style items open (none blocking merge). Approving.**

## Fixed since last review

| # | Finding | State |
|---|---|---|
| #1 | X25519 keypair mismatch (server derived from DID, client from sha256(sign)) | ✅ `rotateRoomKey` now uses `row.x25519_public_key` from ACL. Explicit comment: *"The server does NOT fall back to deriving an X25519 key from the DID's Ed25519 key."* |
| #2 | `GET /rooms/:roomId/keys` object-vs-string mismatch | ✅ Client removed the `decodeSealedEnvelope` step from the fetch path — `openRoomKeyEnvelope` now takes the object directly. Wire is object-object end to end. |
| #3 | E2E link wire shape mismatch (`{ciphertext,nonce}` vs `hex(nonce\|\|ciphertext)`) | ✅ Client `encryptLinkForWire` emits `{ ciphertext, nonce }` — matches server's `isEncryptedLinkData`. |
| #5 | TOFU-before-verify race in `resolvePeerTrust` | ✅ Now does an active `${serverUrl}/server/identity` fetch on first contact; only pins the key if the live identity matches the claim. |
| #6 | Federation payload replay attack | ✅ `canonicalFederationPayload` takes `timestamp` param, `isPayloadFresh` at `federation.ts:163` enforces `FEDERATION_PAYLOAD_MAX_AGE_MS` window. |
| #7 | `verifyLinkExpression` removed / trust boundary undocumented | ✅ AGENTS.md updated: *"Link signatures inside a federated diff travel as metadata — the server stores and relays them as-is without per-link verification. Downstream consumers can verify if they choose."* The security posture shift is now honest. |
| #8 | WS half-open TCP + no ping/pong | ✅ 30-second ping/pong heartbeat at `ws.ts:147-164`, dead sockets terminated. AGENTS.md documents. |
| #9 | `/federate` and `/reconcile` no rate limiting | ✅ Both endpoints (`routes.ts:366, 389`) now have `ipRateLimit(ctx.rateLimits.authIp)` preHandler. |
| #17 | `drainCommitBatch` retry storm | ⚠️ Partial — MAX_LAPS reduced from 100 to 10 (300→30 POST cap during outage). The "bail on first-lap failure" I suggested wasn't done, but 10 is a reasonable middle ground. |
| #21 | `adapters-deno.ts` regex-parse of `httpFetch` error string | ⚠️ Partial — happy path now uses structured `{status, body}`; regex-parse kept as **backward-compat shim** for older executors. Defensible for cross-version compatibility, and the comment explains why. |
| #23 | `ws.test.ts` `collectMessages`-after-open in 6 places | ✅ Architecturally fixed — `openAuthenticatedWs` helper absorbs the initial `online-agents` message internally before returning, so the collector-after-open pattern in the tests no longer races that message. |

## Not addressed (5 latent/style items — inline comments below)

- #12 `parseJwtExpiryMs` still uses `atob(padded)`
- #16 `flushBatch` requeue still has no time-based retry
- #22 `index.ts:15` still imports `@coasys/ad4m-ldk` directly despite the AGENTS.md "only via adapters-deno.ts" rule
- #24 `SlidingWindowLimiter` still hand-rolled
- #25 `canonicalLinkPayload` / `linkIdentity` still duplicated

All five are non-blocking. Documented inline for the follow-up PR.

CI fully green (build-and-test + 4 integration-test jobs + plugin-ad4m-tests + CodeRabbit). Ship it.

bootstrap-languages/server-link-language/src/encryption.ts
```diff
@@ -0,0 +1,296 @@
...
+ * the ephemeral public key travels in the envelope — the recipient
+ * recomputes the same shared secret using their own static private key,
+ * exactly like libsodium's `crypto_box_seal`.
+ *
+ * ## Link confidentiality
+ *
+ * Once a room key is available, `encryptLinkForWire` / `decryptLinkFromWire`
+ * protect the semantic payload (`{source, predicate, target}`) of every
+ * LinkExpression crossing the wire. `author`, `timestamp`, and `proof` stay
+ * in the clear — they're needed for signature verification and routing,
```
lucksus: I thought this was changed and new encrypts everything. We don't need to verify signatures in the link language, but need to allow for broken signatures. What does "routing" mean in this context and why would we need the meta-data? I would prefer having everything encrypted.

bootstrap-languages/server-link-language/src/encryption.ts
```diff
@@ -0,0 +1,296 @@
...
+ * The encrypted payload is placed in the `data` field as
+ * `{ciphertext, nonce}` (matching link-server's EncryptedLinkData shape).
+ */
+export function encryptLinkForWire(link: LinkExpression, roomKey: Uint8Array): WireLinkExpression {
+    const plaintext = utf8ToBytes(JSON.stringify(link.data));
+    const nonce = randomBytes(AES_NONCE_BYTES);
+    const ciphertext = gcm(roomKey, nonce).encrypt(plaintext);
+
+    return {
+        author: link.author,
```
lucksus, line 278: So the comment above was correct - we're still leaving the meta-data in the clear. Can we change this?

lucksus (changes requested):

bootstrap-languages/server-link-language/src/encryption.ts
lal-bot-coasys: You're right and the doc comment is stale. "Routing" in the sense the old comment suggests doesn't happen anymore — the server broadcasts diffs to every WS-connected room member regardless of author, and after the "server treats link data as opaque" pivot the per-link signature verification is gone from the server side too. So the plaintext `author`/`timestamp`/`proof` metadata isn't buying us anything on the server. It's just leaking.

**What the server actually needs in cleartext** (which is not the same as what it *has* today):
- A stable per-link identifier for reconciliation, dedup, and sequence assignment. Today: `link_hash = sha256(canonicalLinkPayload(link))`.
- Nothing else at the link-content level. Author for rate-limiting comes from the JWT at commit time; if server-side author accountability ever matters, record `(jwtAuthor, linkHash)` in a separate log without needing to read link content.

**What downstream clients need** — they always have the room key, so this is free:
- Author to render "who posted this"
- Timestamp to render "when"
- Proof for optional client-side signature verification

All come out of the decrypted blob. Nothing lost.

**Recommendation: encrypt the whole LinkExpression as one blob.**

```
// wire
{ data: { ciphertext, nonce } }
```

Nothing else at the top level. Server stores + relays this. Clients decrypt → get full `{author, timestamp, proof, data}` back.

**The one non-trivial detail** is `link_hash`. Two options:

1. **Content-hash over ciphertext**: `link_hash = sha256(ciphertext ‖ nonce ‖ room_id)`. Simple, doesn't need the room key. Downside: two peers independently creating the same semantic link (rare) each get a different hash, so cross-peer dedup by content-identity is lost. In practice this is fine — genuinely-independent duplicates are rare, and the more common case (peer restart re-sends its own diff) uses the *stored* (ciphertext, nonce) so dedup still works. Simplest.

2. **Client-supplied linkHash from plaintext**: client computes `link_hash = sha256(canonicalLinkPayload(plaintextLink))` and sends both `{ciphertext, nonce, linkHash}` on commit. Server takes the client's linkHash at face value (it can't verify without the key, but a lying client only hurts itself). Cross-peer content-identity dedup works. Slightly more surface.

I'd go with **(1)** for MVP simplicity, revisit if genuinely-independent duplicates become a real thing.

**AES-GCM-SIV** (deterministic-nonce mode) is a third option that gives cross-peer dedup and content-nonce safety in one primitive, but it's a different construction than what @noble/ciphers exposes today. Not a first-pass fix.

**What changes if we do this:**
- Wire shape (drop 4 top-level fields into 1 encrypted blob)
- `encryptLinkForWire` / `decryptLinkFromWire` change signatures
- Server `link_hash` computation changes (option 1) or gets a new source (option 2)
- Doc comments in this file get honest
- Nothing on the correctness path breaks; nothing on federation breaks

Happy to draft the changeset if you want it in this PR, or open a follow-up. Not blocking my approval either way — this is a metadata-privacy improvement, not a security-broken-today fix (the metadata is leaking to the server operator only, who already sees the sealed room-key envelopes and all commit traffic, so this closes a smaller gap than the trust boundary it's inside).

bootstrap-languages/server-link-language/src/encryption.ts
lal-bot-coasys: Yes, this is the concrete site of the same design point. See the full answer on the line-47 thread above — TL;DR: routing/verification are stale justifications post the "server treats data as opaque" pivot, encrypting the whole LinkExpression as one blob is the right move, and the one non-trivial detail is what to do with `link_hash` (two options laid out in the other thread).

bootstrap-languages/server-link-language/src/adapters-deno.ts
lal-bot-coasys: Yes, worth pulling into the LDK. The Deno WebSocket adapter is a straight wrapper of the Deno platform primitive — every link language that talks over WS will reinvent this. Same argument as `httpFetch` living in the LDK: platform I/O adapters belong there, application logic stays in the language. Small follow-up PR.

lucksus (changes requested):
Ah, if I get the e2e encryption right, new members will always see an empty neighbourhood. They will never get access to the old key and there is no resharing of old data. That breaks the link-language contract and use of neighbourhoods that share Social DNA via links, mostly at the beginning of the history. New members won't see these and potentially overwrite with their own versions.

bootstrap-languages/server-link-language/src/sync.ts
```diff
@@ -0,0 +1,604 @@
...
+    return _deps;
+}
+
+// ---------------------------------------------------------------------------
+// Wire <-> local translation (encryption is a no-op for plaintext rooms)
+// ---------------------------------------------------------------------------
+
+function toWireLink(link: LinkExpression): WireLinkExpression {
+    const keyRing = deps().getKeyRing();
+    if (!keyRing) {
```
lal-bot-coasys, line 62: **Blocking — plaintext leak on the exact new-member path.** `keyRing` is null in two very different situations: (a) the room genuinely has no E2E (fine, send plaintext), and (b) the room IS encrypted but this agent hasn't been granted keys yet — `fetchRoomKeys` returns null on 404, `setupKeyRing` sets status `"none"`, and `commit()`'s refuse-guard only triggers on status `"error"`. In case (b) — which is precisely the freshly-added-member scenario this PR's grant flow exists for — every link the new member commits goes to the server **fully in the clear** (author, timestamp, proof, data), and other members will happily merge it via the mixed-format path in `fromWireLink`. Flux writes links on join, so this fires in practice, not just in theory.

Suggested fix: have the client learn whether the room is E2E-enabled (e.g. expose `e2e_enabled` in the auth/verify or keys response) and treat *enabled-but-no-keys-for-me* as a blocking state exactly like `"error"` — queue or refuse commits until keys arrive, rather than falling back to plaintext. Server-side enforcement (see my comment on routes.ts) is the belt to this suspender.

link-server/src/routes.ts
```diff
@@ -0,0 +1,498 @@
...
+          return reply.code(400).send({ error: "each link must be an object" });
+        }
+        if (!link.data || typeof link.data !== "object") {
+          return reply.code(400).send({ error: "each link must have a data object" });
+        }
+        if (isEncryptedLinkData(link.data)) {
+          if (!link.link_hash || typeof link.link_hash !== "string") {
+            return reply.code(400).send({ error: "encrypted links must include a link_hash" });
+          }
+        } else {
```
lal-bot-coasys, line 218: **Blocking (pairs with the sync.ts comment).** The commit endpoint validates encrypted links (link_hash present) and plaintext links (author matches DID) but never consults `room.e2e_enabled`. Once a room has been rotated to E2E, a plaintext commit should be rejected with a 400 — otherwise one under-provisioned or downgraded client silently punches cleartext into the encrypted history and the server stores/broadcasts/federates it. One `if (room.e2e_enabled && !isEncryptedLinkData(link.data)) return 400` per link closes the server side of the hole. (Arguably also reject *encrypted* links into a non-E2E room to keep the room state coherent.)

link-server/src/encryption.ts
```diff
@@ -0,0 +1,174 @@
...
+ * delivered) but never receives N+1. New members added after a rotation
+ * can only decrypt links from the version they first received onward.
+ *
+ * Members whose X25519 public key has not yet been registered are
+ * skipped — they receive their sealed copy on the next rotation after
+ * they authenticate.
+ */
+export function rotateRoomKey(db: LinkServerDB, roomId: string): RotateResult {
+  const aclRows = db.getAcl(roomId);
+  const roomKey = generateRoomKey();
```
lal-bot-coasys, line 151: **Trust-model qualification worth documenting (non-blocking, but Nico asked exactly this).** `rotateRoomKey` runs on the server, so the plaintext room key exists in server memory at every rotation. The code honestly discards it — an *honest* operator can't read room data afterwards — but a malicious/compromised server can simply retain every key it generates. So the current guarantee is "E2E against later compromise / at-rest snooping," not "E2E against the operator." The grant flow added in this PR already proves clients can seal keys to other members' x25519 keys, so moving key *generation* to the admin client (admin generates, seals to each member incl. self, uploads sealed copies; server only stores) would close this without new primitives. Fine as a follow-up, but the README should state the actual trust model until then — right now README line 116 says "the server operator cannot read it," which overstates.

link-server/src/routes.ts
```diff
@@ -0,0 +1,498 @@
...
+      // Store the agent's derived X25519 public key for E2E room-key
+      // sealing. The language sends this during step 2 of the DID
+      // challenge-response — the one point where DID ownership has
+      // already been verified.
+      if (
+        typeof body.x25519PublicKey === "string" &&
+        body.x25519PublicKey.length === 64 &&
+        /^[0-9a-fA-F]{64}$/.test(body.x25519PublicKey)
+      ) {
+        ctx.db.setX25519PublicKey(roomId, body.did, body.x25519PublicKey);
```
lal-bot-coasys, line 156: **Key-substitution surface (non-blocking, document or fix in follow-up).** The DID challenge signature covers only the nonce; `x25519PublicKey` rides alongside unsigned. Storage-side that's mostly fine (it arrives inside a DID-authenticated exchange), but the consumer matters: the admin re-seals *historical room keys* to whatever `x25519PublicKey` the **server** reports back in the rotate response (`membersNeedingHistoricalKeys`). A malicious server can substitute its own x25519 key for a member's and receive the entire history sealed to itself — the admin has no way to detect it. Fix: require the client to send `signature = sign(x25519PublicKey)` at registration, store it, return it in rotate responses, and have admins verify the DID signature before sealing. Cheap to add since the signing capability is already deterministic and available.

bootstrap-languages/server-link-language/src/api.ts
```diff
@@ -0,0 +1,200 @@
...
+// ---------------------------------------------------------------------------
+// E2E key grant (admin re-seals historical versions for late members)
+// ---------------------------------------------------------------------------
+
+export interface GrantKeyEntry {
+    version: number;
+    encryptedKey: { ephemeralPublicKey: string; nonce: string; ciphertext: string };
+}
+
+export async function grantKeys(
```
lal-bot-coasys, line 180: **The last mile of the late-member story is unwired.** `grantKeys()` has zero callers outside link-server tests — and the admin-side counterpart (unseal own ring → re-seal each missing version to the member's x25519 key → POST /keys/grant) doesn't exist as runnable code anywhere in the repo: not in this language, not in the executor, no CLI. Same for `keys/rotate`: nothing calls it on member add; README says the admin should curl it manually. Net effect today: a freshly added member of an encrypted neighbourhood sees an empty room until a human rotates, and sees *history* never — which is the substance of Nico's Aug 28 changes-request, mechanically enabled but not operationally delivered. Minimum viable wiring: on rotate, the admin's language instance (it knows it's admin) walks `membersNeedingHistoricalKeys` and performs the grants automatically; plus trigger a rotate from the ACL-add path (or have the language do it when it observes a member without keys). Also note tests/js has no E2E integration coverage at all — the two-agent join-encrypted-room-and-see-history flow is untested end to end.

link-server/README.md
```diff
@@ -0,0 +1,196 @@
...
+
+Encrypt link data so the server operator cannot read it:
+
+```bash
+# Enable or rotate the room key (admin only)
+curl -X POST https://your-server:3456/rooms/YOUR_ROOM/keys/rotate \
+  -H "Authorization: Bearer $ADMIN_TOKEN"
+```
+
+Each member automatically receives a sealed copy of the room key during authentication. After adding a new member to an encrypted room, run `keys/rotate` again so they receive their copy.
```
lal-bot-coasys: Two stale claims worth fixing in this pass since they answer questions people will actually ask: (1) this line says members "automatically receive a sealed copy of the room key during authentication" — not true; auth only *stores* the x25519 key, and sealed copies only materialize at the next `keys/rotate`. (2) Line 134 says "`author`/`timestamp`/`proof` stay visible so the server can still enforce ACL" — that described the legacy format; since fefb9b6ec the entire LinkExpression is encrypted and the server sees only `{ciphertext, nonce, link_hash}`. The "How it works" section should describe the wire format that actually ships.

lal-bot-coasys (changes requested):
Re-review focused on Nico's three questions: (1) is E2E working, (2) do new members get history, (3) is the data truly unreadable by the server. The crypto core is genuinely good now — versioned key ring, full-LinkExpression encryption (no metadata on the wire), ECIES sealing that matches on both sides, and the late-member key-grant mechanism directly answers the Aug 28 changes-request; 102 language + 68 server tests pass locally. But I found one real confidentiality hole: a new member of an E2E room who hasn't received keys yet (GET /keys → 404 → keyRing=null, status "none" not "error") will send **full plaintext links** into the encrypted room, and the server's commit endpoint never checks `e2e_enabled`, so it accepts and stores them — the "refuse to commit plaintext" guard only fires on key-fetch *errors*, which is exactly backwards for the join flow. Second, the history-for-late-members mechanism is unwired end to end: nothing ever calls `keys/rotate` on member add, `grantKeys()` has zero callers outside server tests, the admin-side re-sealing logic exists nowhere as runnable code, and tests/js has no E2E coverage at all — so in practice a new member still sees nothing until a manual curl, and never sees history. Third, on server-readability: an honest server can't read room data (it discards plaintext keys, verified), but the trust model has two documented-nowhere soft spots — the server *generates* every room key (a malicious operator could retain them; the grant flow proves clients can seal keys, so admin-side generation is feasible), and the x25519 public key isn't covered by the DID challenge signature, so a malicious server can substitute its own key in the rotate response and have the admin seal the entire history to it. Inline comments with details; the plaintext-fallback fix is small and I'd consider it blocking, the rest can be follow-ups if documented honestly.

bootstrap-languages/server-link-language/src/api.ts
```diff
@@ -183,6 +192,24 @@ export async function grantKeys(
     return res.granted ?? [];
 }
 
+// ---------------------------------------------------------------------------
+// E2E admin: key gaps + ACL query
+// ---------------------------------------------------------------------------
+
+/** Admin only. Returns members missing historical key versions. */
+export async function fetchKeyGaps(config: RoomConfig, token: string): Promise<KeyGapsResponse> {
```
HexaField: function/route & comments should be worded as `fetchMissingKeys` - 'gaps' language sounds clunky

lal-bot-coasys (dismissed):
Re-reviewed at head `86cc73c` after the latest round of commits. Every item from my Sept 3 changes-requested review is now verifiably addressed in code, so this is an **APPROVE** — I've resolved all 11 of my open threads.

Per-item verification (not just trusting commit messages):

1. **Plaintext leak on the new-member path (was blocking)** — fixed on both sides. Client: `keyRingStatus` now distinguishes `"none"` / `"pending"` / `"error"`; `commit()` retries key acquisition once and then *throws* rather than sending plaintext when the room is E2E but keys haven't been granted (`index.ts`). The `GET /keys` response now carries `e2e_enabled` so the client can tell "no E2E" (404) from "E2E, keys pending" (200 + empty keys). Server: the commit endpoint rejects plaintext links with a 400 once `room.e2e_enabled` is set (`routes.ts`). Both paths are covered by tests ("commit rejects plaintext links in an E2E-enabled room", "GET /keys returns e2e_enabled: true with empty keys for late member", plus the accepts/plaintext-room contrast cases).
2. **Late-member history (grant flow was unwired)** — `performAdminKeyGrants()` now exists and runs on init and on every `peer-joined` event: admin fetches `/keys/missing`, re-seals each missing historical version to the member's x25519 key, and POSTs `/keys/grant`. Server side has the `/keys/missing` endpoint + idempotent grant, with 7 dedicated tests (grant round-trip, non-admin rejection, non-member rejection, idempotency, missing-version reporting).
3. **Trust model (server-generated keys / malicious operator)** — README now has an explicit "E2E encryption — trust model and future requirements" section that honestly states both caveats: a malicious operator could retain keys it generates at rotation, and the x25519 public key rides unsigned alongside the DID challenge (with the concrete fix sketched: sign the x25519 key at registration and verify before sealing). Exactly the "document or fix in follow-up" I asked for.
4. **README stale claims** — rewritten to match reality (sealed copies arrive on next connect/key-ring refresh, not "during authentication"; rotate-after-member-add is documented as the admin step, with auto-grant covering history).
5. **Older nits all closed** — JWT payload now decoded via `TextDecoder` (UTF-8-safe, was `atob`-as-binary-string); failed commit segments now get a backoff retry timer instead of waiting for the next unrelated commit; the `index.ts` → `ad4m-ldk` import is now explicitly sanctioned in AGENTS.md with rationale (module-surface exports, not I/O adapters); `linkIdentity`'s coupling to the server's `canonicalLinkPayload` is thoroughly documented in place. The hand-rolled rate limiter remains, which I'm fine with — it was a nice-to-have suggestion, and the current implementation now has cleanup wired into socket close.

Also verified: CI is green across all 7 CircleCI jobs + CodeRabbit on this exact head, and the new integration-test commit adds multi-agent sync coverage for the language itself.

One note for after merge (no action needed now): the language-side "pending → commit throws" behavior is enforced-and-tested at the server, but has no direct unit test in `server-link-language/tests`. Worth adding when the ALDK adapter-mock harness makes that cheap.

The PR is `BEHIND` dev — needs a base merge (or update-branch) before the button is green, but content-wise this is ready.

link-server/src/encryption.ts
```diff
@@ -0,0 +1,149 @@
...
+ * signing-capability-based derivation that produces a different keypair
+ * than the textbook Ed25519→X25519 Montgomery conversion.
+ */
+export function encryptRoomKeyForRecipient(
+  roomKey: Uint8Array,
+  recipientX25519Pub: Uint8Array
+): EncryptedKeyPayload {
+  const ephemeralPriv = x25519.utils.randomSecretKey();
+  const ephemeralPub = x25519.getPublicKey(ephemeralPriv);
+  const shared = x25519.getSharedSecret(ephemeralPriv, recipientX25519Pub);
```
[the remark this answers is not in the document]
lal-bot-coasys: False positive, verified against the published package: `@noble/curves` 1.9.7 *does* export `randomSecretKey` (in `abstract/montgomery.js` — it is the primary name there, with `randomPrivateKey` kept as an alias). The seal/rotate paths calling it are exercised by 25 tests in `encryption.test.ts` and CI is green on this head. Resolving.

link-server/src/federation.ts
```diff
@@ -0,0 +1,374 @@
...
+        `federation-reconcile:${peerUrl}`
+      );
+      this.ws.broadcast(roomId, { type: "diff", payload: diff, revision: newRevision, sequence });
+    }
+
+    // Track the peer's latest sequence so the next reconciliation starts
+    // from where this one left off (fast-forward).
+    if (response.sequence > sinceSequence) {
+      this.db.setPeerLastSequence(roomId, peerUrl, response.sequence);
+    }
```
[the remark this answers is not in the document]
lal-bot-coasys: Moot — federation was removed entirely (`link-server: drop federation entirely`, e20447d36); `federation.ts` no longer exists on this branch. Resolving.

bootstrap-languages/server-link-language/src/encryption.ts
lal-bot-coasys: Update (Sept 7 re-review): this is now implemented — the entire LinkExpression (author, timestamp, proof, data) is encrypted into one opaque blob; the server only sees `{ciphertext, nonce, link_hash}`. Verified in the current wire format + covered by `encryption.test.ts` ("commit accepts fully-encrypted links (no author/timestamp/proof, only data+link_hash)"). Your call to resolve, but nothing metadata-shaped is left in the clear.

marvin-bot-coasys:
## Review (status) — `89388bd`

Security-focused pass on the e2e stack, plus a merge call. Not a formal APPROVE / REQUEST_CHANGES.

**Merge verdict:** not as “e2e-encrypted neighbourhoods.” Yes as an experimental self-hosted link-sync server if we label the crypto honestly and keep `AUTO_ADMIT=false`. The primitives are competent. The trust model is not Signal.

### What is actually good

DID challenge-response (SHA-256 then Ed25519, matching the executor), single-use challenges, JWT scoped to `(did, roomId)` and not put in the WS query string, live ACL check on every request, default `AUTO_ADMIT=false`, rate limits, Docker `USER node`, AES-256-GCM + X25519 ECDH + HKDF-SHA256 with domain-separated tags and random 12-byte nonces. Client and server seal constructions match. The README is unusually honest about remaining holes.

### E2E is not e2e against the operator

Already documented in `link-server/README.md` “Known limitations”; still true:

1. **The server generates the room key.** An honest process discards plaintext after sealing. A malicious operator can retain every key it generates and read the room forever.
2. **`x25519PublicKey` is not covered by the DID signature.** A malicious server can substitute its own X25519 on rotate / missing-keys, and `performAdminKeyGrants` will reseal historical keys to the server.

Until key generation moves to the admin client *and* the X25519 is DID-signed, this is at-rest / honest-operator encryption, not e2e.

### Holes even with an honest server

These I would still treat as blockers if e2e is part of the merge pitch:

1. **Kicking a member does not rotate.** ACL remove revokes the JWT and deletes the row. It does not call `rotateRoomKey`. The removed member still has the current AES key and can decrypt every future commit until someone remembers to rotate.
2. **Encrypted commits drop `author === claims.did`.** Plaintext mode enforces it; e2e mode cannot see author and never substitutes another check. Any member who can decrypt can compute `link_hash` from plaintext and submit an encrypted *removal* of anyone else’s link. The README still claims only the original author can remove. That invariant dies the moment e2e is on.
3. **The language does not verify AD4M proofs after decrypt.** `applyInboundWireDiff` decrypts and `emitDiff`s. Forged author/proof inside ciphertext is accepted.
4. **Telepresence is plaintext** through the server (`payload` on signal/broadcast).
5. **First authenticator to a new `roomId` becomes admin.** Fine if room IDs are unguessable; a race if they are predictable.

`bootstrap-languages/server-link-language/AGENTS.md` still says “no E2E key rotation handling / key fetched once.” The code has a versioned `KeyRing` and `refreshKeyRingIfNeeded`. Stale doc, not a new bug.

### So: merge?

If the pitch is “Holochain stand-in, optional encryption so sqlite/backups aren’t plaintext, operator is trusted” — close enough; rebase onto `dev` first.

If the pitch is “neighbourhood data the host cannot read or forge” — **no.** Need at least: client-side key gen, signed X25519, rotate-on-ACL-remove, and author/proof checks after decrypt.

I would not turn e2e on for a real neighbourhood until (1) and (2) of the honest-server list are fixed.

lucksus (dismissed):
Hm, surprised it is still not full e2e encryption. sounds like it woulnd't be to hard to have the key generation live admin/client side as we discussed. I'm approving anyway to enable getting this in and starting to test it soon. But I really would love to see this be something that can be used without having to trust the central infrastructure node..

lal-bot-coasys (dismissed):
Re-approving after the doc-removal push — this commit only deletes the 4 AGENTS.md/README.md files (moved to #970 per Nico's request), no code changed. Diff otherwise identical to what I reviewed above.

lucksus (dismissed):
Reapproving after Lal accidentally pushed commits (now reverted) to the wrong branch.

HexaField:
## Review: PR #930 — Link Server + Server-Link-Language

Full read-through of all source files. Findings grouped by severity.

---

### Security

#### 1. Unsigned X25519 public key (documented, P2)

The X25519 public key sent during DID auth (`x25519PublicKey` field in step 2 of challenge-response) carries no signature. A malicious server can substitute its own key for any member, then decrypt every sealed room key envelope. The README documents this. The fix — `signature = sign(x25519PublicKey)` verified before sealing — should land before any deployment beyond trusted-operator testing.

#### 2. `link_hash` trusted from clients (acceptable, P3)

For encrypted links the server cannot verify `link_hash` (it can't decrypt). A malicious authenticated client can submit a false hash, causing OR-Set dedup to malfunction (two links with the same plaintext get different hashes → duplicates; or two different links get the same hash → collision). The trust model already assumes authenticated clients behave — this follows from that. Worth documenting in the architecture section.

#### 3. No admin recovery / transfer (P2)

If the admin's device goes permanently offline, the room becomes unmanageable — no key rotation, no ACL changes, no grants for new members. An admin-transfer endpoint (or multi-admin support) should exist before production use.

---

### Correctness

#### 4. Diagnostic `log::info!` should be `log::debug!` (P3)

`rust-executor/src/perspectives/perspective_instance.rs` adds `log::info!` calls in `commit()`. These fire on every perspective commit across the entire executor — not just server-link rooms. Should be demoted to `log::debug!` before merge to avoid noise in production logs.

#### 5. `linkHash` cross-package invariant has no cross-package test (P3)

The client (`store.ts:hashLink`) and server (`types.ts:linkHash`) compute SHA-256 of the same canonical JSON independently. Both normalize predicate to `null`, both use the same field order. But the invariant — that these two functions produce identical output for the same link — has no test that exercises both implementations in the same assertion. A property test or shared fixture would catch divergence early.

#### 6. `applyDiffAndAppend` adds before removes — correct but fragile (P4)

The server processes additions then removals within one transaction. The client's `coalesceDiffs` correctly refuses to merge diffs where the same link identity appears in both additions and removals. But this ordering dependency lives only in comments and the coalescing logic. If someone later changes the transaction order on the server, the coalescing invariant breaks silently. Consider an assertion or test that verifies the ordering.

---

### Architecture (observations, not blockers)

#### 7. Single-choke-point pattern — well-designed

`applyInboundWireDiff` as the single inbound path for all diffs (WS and HTTP) makes the system much easier to reason about. The module-level doc comment is clear. The `trackMissingKeyVersions` bridge (PR #984) closes the gap between the two paths cleanly.

#### 8. Batched commit coalescing — good

The segment-merge algorithm in `coalesceDiffs` correctly detects add/remove conflicts and splits. The `linkIdentity` function deliberately excludes `proof.signature` (two links with different sigs but same canonical fields collapse). The retry + re-enqueue mechanism is bounded and safe.

#### 9. Telepresence grace period — good

`scheduleOffline` with a configurable grace period prevents flapping on reconnects. The WS manager tracks per-DID connection sets (multiple devices), only marks offline when all connections drop.

#### 10. Key version monotonicity relies on SQLite transaction isolation

`getLatestKeyVersion + 1` in the rotation endpoint has no explicit lock — but `better-sqlite3` is synchronous and single-connection, so concurrent rotation from two admin sessions in the same process can't interleave. Correct for the single-process deployment model. Would break under a connection pool or multi-process setup.

---

### Test coverage

#### 11. Strong unit test coverage

- link-server: auth, ACL, commit-sync, encryption, E2E, multi-agent-sync, rate-limit, WebSocket — all covered
- server-link-language: sync (38 tests), auth, encryption, store, telepresence, integration

#### 12. Integration tests run both backends

The neighbourhood test matrix runs the same test suite against holochain and server-link backends — good for parity verification.

#### 13. Missing test: admin loss scenario

No test verifies what happens when the admin goes offline after rotating keys and a new member joins. The new member should receive `keyRingStatus: "pending"` and refuse plaintext commits. The admin's `performAdminKeyGrants` should eventually grant keys when the admin comes back online. This end-to-end flow should have an integration test.

#### 14. Missing test: diff retention overflow

No test verifies that a client falling behind by more than `maxDiffsPerRoom` diffs can still recover via bootstrap (`/render`). The sync path handles this implicitly but testing the edge case would catch regressions.

---

### Nits

- `link-server/src/telepresence.ts` — `scheduleOffline` callback fires outside any try/catch. If the callback throws (e.g., `broadcast` fails on a closed socket), the error surfaces as an unhandled promise rejection. Wrap in try/catch.
- `bootstrap-languages/server-link-language/src/ws-client.ts` — should verify presence and read its reconnect logic.
- The `guide.svg` files in both packages appear to be placeholder or documentation assets — confirm they render correctly in the README.

---

### Verdict

Solid architecture. The E2E encryption design works well within the constraints of the Deno sandbox (no raw private key access). The dual-path sync convergence (WS + HTTP) through a single choke point reduces the surface for state inconsistencies. Rate limiting, session management, and auth are production-grade for the single-process self-hosted model.

**Merge-blocking items:** None. The unsigned X25519 key and admin-transfer gaps are documented and acceptable for the current deployment scope (trusted operators, team use).

**Pre-merge cleanup:** Demote the Rust `log::info!` diagnostics to `log::debug!` (item 4).

PR #932, opened 2026-08-26
flow: deterministic FlowTransitionProposal post-processing (engine gate)

data-bot-coasys:
# flow: deterministic FlowTransitionProposal post-processing

Stacked on #929. (merged :heavy_check_mark:)

## What this PR does

PR #929 added the ability for the extraction LLM to *see* active flows on a scope — an "## Active flows on this scope" block gets appended to the prompt whenever the perspective has live `FlowInstance`s, listing each flow's reachable next-states along with their `requires` guard (rendered in English) and any per-state `semanticCheck` hint.

This PR closes the other half of that loop: **after** the LLM's extraction lands on the graph, re-evaluate every active flow's next-state guards against the freshly committed evidence, and emit engine-generated `FlowTransitionProposal`s for the transitions that structurally satisfy their `requires` — even if the LLM missed them.

Two properties this earns:

1. **Determinism** — an LLM that forgets to propose a valid transition can't stall a flow. The engine catches what the LLM missed on the very next extraction pass.
2. **Non-bypass** — the LLM's own proposal is never enough on its own. Its `flow_proposals` output (or `{FlowName}_propose_transition` tool-call in the harness path) is treated as a **hint**: attribution rides through as the proposal's `rationale`, but the structural `requires` gate still has to fire before anything lands on-graph.

## Architecture

```mermaid
flowchart TD
    subgraph extraction["extraction pass (interpretation::run)"]
        LLM["LLM extraction<br/>(strategy or harness)"]
        Overlay["apply_with_overlay<br/>→ commits LLM writes"]
        LLM --> Overlay
    end

    subgraph engine["run_engine_proposal_pass"]
        LoadFlows["load_shacl_flows"]
        LoadInstances["load_flow_instances"]
        Evaluate["evaluate_flow_transitions"]
        SemGate["semantic-check gate<br/>(optional 2nd-pass LLM)"]
        Writer["write_engine_proposal<br/>per SatisfiedTransition"]

        LoadFlows --> Evaluate
        LoadInstances --> Evaluate
        Evaluate --> SemGate
        SemGate --> Writer
    end

    Graph[(Perspective graph)]

    Overlay --> LoadFlows
    Overlay --> LoadInstances
    Overlay -->|committed evidence| Evaluate
    Writer -->|FlowTransitionProposal| Graph
    LLM -.->|flow_proposals hints<br/>rationale attribution only| Writer
```

## Component layout

```mermaid
flowchart LR
    subgraph pure["flow_evaluator/primitives.rs (pure)"]
        ST["SatisfiedTransition"]
        BQ["build_query_input_for_requires<br/>(ModelQuery → ModelQueryInput,<br/>$did substitution, or-composition)"]
        CS["cardinality_satisfied<br/>(count.{min,max})"]
        EH["evidence_hash<br/>(SHA256 seal)"]
    end

    subgraph async["flow_evaluator/queryable.rs (async, trait-backed)"]
        RQ["RequiresQueryable trait<br/>(stubs test w/o live perspective)"]
        ESQ["evaluate_single_query"]
        ESR["evaluate_state_requires"]
        EFT["evaluate_flow_transitions"]
    end

    subgraph writer["flow_evaluator/engine_pass.rs"]
        WEP["write_engine_proposal"]
        LPH["LlmProposalHint"]
        REPP["run_engine_proposal_pass<br/>(composed entry point)"]
    end

    subgraph sem["flow_semantic_check.rs"]
        SCV["SemanticCheckVerdict<br/>(Pass / Fail / Ambiguous)"]
        BSP["build_semantic_check_prompt"]
        PSR["parse_semantic_check_response"]
        RSC["run_semantic_check<br/>(async)"]
    end

    BQ --> ESQ
    CS --> ESQ
    ESQ --> ESR
    ESR --> EFT
    EFT --> REPP
    EH --> ST
    ST --> WEP
    WEP --> REPP
    LPH --> REPP
    RSC --> REPP
```

## Onion shells

Each shell is independently testable — stub-perspective unit tests plus a live-perspective e2e per layer.

| Slice | What lands                                                                                              | Where                                                            |
|-------|---------------------------------------------------------------------------------------------------------|------------------------------------------------------------------|
| 10.4a1 | `SatisfiedTransition` struct + pure primitives (`build_query_input_for_requires`, `cardinality_satisfied`, `evidence_hash`) | `flow_evaluator/primitives.rs`                                   |
| 10.4a2 | Async layer over `model_query` behind `RequiresQueryable` trait; composer walks reachable next-states  | `flow_evaluator/queryable.rs`                                    |
| 10.4a3 | Live-perspective integration test — proves the `ModelQuery → ModelQueryInput` JSON is a shape the real SPARQL/SDNA-backed `model_query` accepts | `flow_evaluator/e2e_tests.rs`                                    |
| 10.4b | `write_engine_proposal` — SatisfiedTransition → on-graph proposal via `flow_classes::write_flow_transition_proposal`; batched writes | `flow_evaluator/engine_pass.rs`                                  |
| 10.4c | `run_engine_proposal_pass` — composed load → evaluate → write entry point; wired into `interpretation::run` after `apply_with_overlay` | `flow_evaluator/engine_pass.rs` + `interpretation/run.rs`        |
| 10.5a | Semantic-check pure primitives — prompt build, response parse, fire/discard verdict                    | `flow_semantic_check.rs`                                         |
| 10.5b | Semantic-check gate wired into `run_engine_proposal_pass` — runs BEFORE the write, fail-safe on Ambiguous/error | `flow_evaluator/engine_pass.rs`                                  |
| 10.5c | AIService-backed real `SemanticCheckLlm` impl                                                          | `flow_semantic_check.rs`                                         |
| 10.6c | LLM-side attribution: `flow_proposals[]` field on the strategy extractor's JSON output → matched hints thread `rationale` onto engine proposals | `interpretation/parse.rs`, `flow_evaluator/engine_pass.rs`       |

## How a transition fires

```mermaid
sequenceDiagram
    autonumber
    participant Ext as extraction pass
    participant Overlay as apply_with_overlay
    participant Engine as run_engine_proposal_pass
    participant Eval as evaluate_flow_transitions
    participant Sem as semantic-check gate
    participant Write as write_engine_proposal
    participant Graph as perspective graph

    Ext->>Overlay: commit LLM writes (instances, links)
    Ext->>Engine: pass hints (Vec<LlmProposalHint>)
    Engine->>Engine: load flows + FlowInstances (scoped)
    Engine->>Eval: walk reachable next-states
    loop per active flow × per reachable next-state
        Eval->>Graph: model_query per requires element
        alt AND across requires holds
            Eval-->>Engine: SatisfiedTransition (+evidence_hash)
        else any guard misses
            Eval-->>Engine: skip
        end
    end
    loop per SatisfiedTransition
        opt state has semanticCheck hint
            Engine->>Sem: run_semantic_check(prompt, LLM)
            alt Pass
                Sem-->>Engine: fire
            else Fail / Ambiguous / LLM error
                Sem-->>Engine: discard
            end
        end
        Engine->>Engine: match hint by (instance × to_state)<br/>→ rationale = hint.reason
        Engine->>Write: batched proposal write
        Write->>Graph: FlowTransitionProposal<br/>(evidence, evidence_hash, rationale?)
    end
```

## Key invariants (locked by test)

- **Silent-skip discipline** — a bad flow definition, an unregistered SDNA class, or a transient `model_query` error skips *one* transition, never poisons the whole pass. The extraction pass MUST NOT break because a flow stumbled.
- **Evidence-hash determinism** — the hash is order-independent (sorted evidence-ids joined with `\n`, classes with `|`). Two evaluations of the same guard against the same graph state produce byte-identical hashes.
- **Per-transition batches** — each proposal writes inside its own `create_batch` / `commit_batch` pair so a single write failure only rolls back that proposal; the others still ship.
- **Fail-safe semantic gate** — `Ambiguous` and LLM I/O errors both discard the transition. An uncertain confirmation MUST NOT silently advance a flow.
- **Hint non-bypass** — an `LlmProposalHint` without a matching `SatisfiedTransition` is silently dropped. The LLM cannot bypass the deterministic `requires` guard by proposing.

## What is NOT in this PR

- **Consensus firing** — a `FlowTransitionProposal` doesn't advance the flow's `currentState` on its own. That happens when a quorum accumulates and `fire_flow_consensus` runs, which lives in the next stacked PR (`feature/flow-mutations`).
- **Harness-path attribution** — the strategy (JSON) extractor's `flow_proposals` output is wired here; the harness (tool-calling) extractor's `{FlowName}_propose_transition` tools land in the next PR (#940).
- **Subscription hooks** — `onProposalAdded` / `onProposalResolved` on the `FlowInstance` wrapper — the mutation surface in the next PR.
- **UI-side surfacing** — no client-side changes; the engine writes on-graph and clients pick up via their existing perspective subscription.

## Test plan

- [x] Local: `cargo test -p ad4m-executor --lib perspectives::flow_evaluator` — pure + async stub coverage
- [x] Local: `cargo test -p ad4m-executor --lib perspectives::flow_semantic_check` — pure + async stub coverage
- [x] Live perspective: `flow_evaluator::e2e_tests::*` — walks real SPARQL / SDNA / model_query against real evidence
- [x] Live perspective: `flow_evaluator::e2e_tests::semantic_check_*` — semantic-check gate wired into `run_engine_proposal_pass` end-to-end
- [x] Live perspective: `flow_evaluator::e2e_tests::llm_hint_*` — LLM attribution rides through as `rationale`
- [x] CI on merge target (waiting for #929 to land first)

## Summary by CodeRabbit

* **New Features**
  * Automatically evaluates active flows and creates transition proposals when requirements are met.
  * Supports optional semantic checks before creating proposals.
  * Interpretation results can include flow-transition proposals with evidence and optional rationale.
  * Supports loading and processing all active flow instances.
* **Bug Fixes**
  * Interpretation parsing now supports legacy responses and responses containing instances and flow proposals.
* **Tests**
  * Expanded coverage for flow evaluation, semantic checks, proposal creation, evidence, and interpretation formats.
* **Documentation**
  * Updated flow-related documentation and test descriptions.


rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,2734 @@
...
+                    );
+                }
+            },
+            Err(e) => {
+                // One bad write must not sink the rest — mirror the
+                // per-transition silent-skip policy `evaluate_flow_transitions`
+                // uses for query errors.
+                let _ = perspective.discard_batch(&batch_id).await;
+                log::debug!(
+                    "run_engine_proposal_pass: write_engine_proposal for {}.{}→{} failed: {e:#}",
```
lal-bot-coasys: **Log level for silent failures is too quiet.** All three failure branches in this loop (`write_engine_proposal` at L791, `commit_batch` at L778, and semantic-check LLM error at L734) log at `debug!` and continue. `debug!` is below the default log level in most deployments, so a persistent schema mismatch or batch-API regression would leave flows silently stalled with nothing visible above INFO. The comment on L564–568 justifies this as "the extraction pass MUST NOT break" — I agree with the *don't-break* half, but recovering-without-shouting is the wrong tradeoff for the writer branches:

- **Loader failures** (L618, L633, L648) log at `warn!` — right level, one-time setup problems.
- **Semantic-check LLM error** (L734) at `debug!` — arguably fine, LLM I/O is inherently flaky.
- **`write_engine_proposal` failure / `commit_batch` failure** (L778, L791) — should be `warn!`. These fire only after the deterministic guard has matched, the semantic-check gate has passed, and the transition is *ready to land*. A silent failure at this point represents shipped-code broken, not routine flake.

Happy for this to be a follow-up rather than in-PR — CI is green and the on-graph shape is what the tests already lock. Just don't want a future user reporting "flow doesn't advance" and finding zero warnings in the logs.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,2734 @@
...
+    // for `mint_interpretation_run`. `write_flow_transition_proposal`
+    // internally does `create_subject` + N-1 `update_subject` for the
+    // evidence bag, so wrapping them in a batch is what makes the
+    // proposal land atomically on-graph (readers never see a
+    // half-populated proposal). Per-transition batches — not one batch
+    // for the whole pass — so a single failure only rolls back that
+    // proposal and the others still ship.
+    let proposed_at = chrono::Utc::now().timestamp_millis().to_string();
+
+    let mut minted = Vec::with_capacity(satisfied.len());
```
lal-bot-coasys, line 411: **Question on repeated-pass dedup:** every satisfied transition here gets a fresh UUID proposal via `Uuid::new_v4()`. If the auto-processor runs the same scope twice — new transcript turns come in, but no new evidence relevant to this flow — the requires still holds, `evaluate_flow_transitions` still emits the same `SatisfiedTransition`, and we mint a *second* proposal identical to the first except for the URI.

Over a session that's N passes × K satisfied-but-not-yet-consensus transitions ⇒ N × K proposals per unique (flow, from, to) tuple, each with the same `evidenceHashes`.

Two ways this could be intended:
1. Slice 10.6's consensus loop collapses by `(proposer, flowInstance, fromState, toState)` before counting — the amplification is harmless.
2. Every pass is a fresh proposal by design (a re-vote), and consumers dedup on read.

Either is fine; would be worth spelling out in the module doc or a follow-up comment on `run_engine_proposal_pass` so a future reader (or slice-10.6 implementer) doesn't have to reverse-engineer it. If neither holds, a pre-write dedup query at this line (`does a proposal already exist from this DID for (flowInstance, fromState, toState) with a matching evidenceHashes?`) would keep the graph clean without changing the consensus semantics.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,2734 @@
...
+/// bypass the deterministic `requires` guard). Satisfied transitions
+/// without a matching hint still get an engine-emitted proposal, exactly
+/// as slice 10.4c/10.5c shipped — but with `rationale = None` (byte-
+/// identical writes to the pre-10.6c path). Pass `&[]` to opt out.
+///
+/// Returns the URIs of every `FlowTransitionProposal` this pass minted.
+/// The extraction pass threads these into
+/// [`crate::perspectives::interpretation::run::InterpretationOutcome::flow_proposals`]
+/// so tests / callers can observe.
+pub async fn run_engine_proposal_pass(
```
lal-bot-coasys, line 385: **API ergonomics:** the signature has grown from `(perspective, scope, context)` (10.4c) → `+ semantic_check` (10.5c) → `+ llm_hints` (10.6c), and the `semantic_check: Option<(&dyn SemanticCheckLlm, &str)>` tuple is already awkward — the `&str` model_id sits opaquely at the tuple's tail. When the next optional dep lands (transcript excerpt for semantic-check per the 10.5a1 comment? consensus tie-breaker config?), the call sites in run.rs:830 and run.rs:1101 become hard to read.

One shape that would age better:

```rust
pub struct EngineProposalPass<'a> {
    pub scope: Option<&'a Scope>,
    pub semantic_check: Option<SemanticCheckConfig<'a>>,
    pub llm_hints: &'a [LlmProposalHint],
}

pub struct SemanticCheckConfig<'a> {
    pub llm: &'a dyn SemanticCheckLlm,
    pub model_id: &'a str,
}

pub async fn run_engine_proposal_pass(
    perspective: &mut PerspectiveInstance,
    context: &AgentContext,
    opts: EngineProposalPass<'_>,
) -> Vec<String> { ... }
```

Call sites become self-documenting and future-optional-arg additions don't break signatures. Not blocking — nit.

rust-executor/src/perspectives/interpretation/parse.rs
```diff
@@ -41,10 +114,23 @@ fn clean_llm_json(raw: &str) -> String {
     //    surrounding prose would invert it before the real JSON begins, so a
     //    comma inside a genuine string value could be dropped. Extracting the
     //    bracketed block first confines the string-scanner to actual JSON.
+    //
+    //    Prefer whichever outer bracket appears FIRST in the trimmed input.
```
lal-bot-coasys: **Asymmetric fallback loses the array-recovery path when `{` appears first in noise.** The current shape:

```rust
match (first_obj, first_arr) {
    (Some(o), Some(a)) if o < a => extract_bracketed(candidate, '{', '}'),   // no fallback!
    (Some(_), None)             => extract_bracketed(candidate, '{', '}'),
    _                            => extract_bracketed(candidate, '[', ']')
                                       .or_else(|| extract_bracketed(candidate, '{', '}')),
}
```

The `_ =>` branch (array-first) tries `[…]` then falls back to `{…}` — correct for the legacy shape's noise cases. But the `(o < a)` branch (object-first) does NOT fall back to array. So an LLM response like:

```
OK so I'll extract {a couple of things}: [{"class":"X"}]
```

…finds `{` at position ~12, `[` at ~35 → object-first branch → `extract_bracketed(candidate, '{', '}')` correctly returns `{a couple of things}` (which is legit balanced braces in the prose) → parses as bogus JSON → retry loop burns an attempt. The array path would have succeeded on the first try.

Fix is one line — make the object-first branch symmetric:

```rust
(Some(o), Some(a)) if o < a => extract_bracketed(candidate, '{', '}')
                                   .or_else(|| extract_bracketed(candidate, '[', ']')),
```

Correctness is unchanged either way (retry loop catches it), but on a small local model that's another prompt tick's worth of latency + a warning log per failed attempt.

lal-bot-coasys (dismissed):
APPROVE. This is a very well-executed slice. The onion-shell discipline shows up in the tests: 20+ unit tests per pure primitive, trait-stubbed async layer, then live-perspective e2e tests for every seam (10.4c writer-lands-on-graph with per-predicate assertions; 10.5b semantic-check pass / fail / LLM-error / absent-hint-autopass; 10.6c hint-matches-writes-rationale / hint-without-match-discarded / hint-with-empty-reason-omits-rationale). The `evidence_hash` sort-independence, cardinality edge cases (`max: 0` as a negative guard), and the prompt-regression assertions in `prompt.rs` are exactly the guards this design needs to keep from silently drifting.

Four concerns inline, none blocking:
- **Silent write failures at `debug!`** (flow_evaluator.rs:792) — the load-bearing outcome of this PR is that satisfied transitions actually land as `FlowTransitionProposal`s on the graph. A write / commit / semantic-check-LLM failure logs at `debug!` and continues, which means a schema-mismatch or DB-write regression would leave flows silently stalled with nothing above the default log level. Recommend `warn!` at minimum for the two write paths.
- **No dedup against already-minted proposals** (flow_evaluator.rs:689) — question, not defect: if the extraction pass fires twice on the same perspective while the requires stays satisfied, we mint two proposals with distinct UUIDs but identical `(flowInstance, fromState, toState)`. Does slice 10.6's consensus loop collapse those, or should the pass check for an existing open proposal by (flow, from, to) before writing?
- **Growing arg list on `run_engine_proposal_pass`** (flow_evaluator.rs:601) — five positional args including an `Option<(&dyn Trait, &str)>` tuple. Adding one more optional dep will force a follow-up refactor anyway; an options struct now would keep the two call sites in run.rs readable.
- **Asymmetric fallback in `clean_llm_json`** (parse.rs:118) — the object-first branch skips the reciprocal array fallback that the array-first branch keeps. Prose noise with a stray `{...}` before the real payload would fail parse when the array path would have worked.

Also: the PR description's onion-shell checkboxes only mark 10.4a1 as complete, but the diff ships 10.4a1 through 10.6c across 13 commits. Update the body before merge so the reviewer-guiding docs match the final diff (per feedback_pr_engineering_guidelines §3). CI green on all 8 jobs; ship it after those follow-ups.

rust-executor/src/perspectives/interpretation/run.rs
```diff
@@ -68,6 +75,48 @@ where
...
+            }
+        }
+    }
+    Err(last_err.unwrap_or_else(|| {
+        anyhow::anyhow!(
+            "interpretation: failed after {INTERPRETATION_MAX_ATTEMPTS} attempts with no captured error"
+        )
+    }))
+}
+
```
lucksus, line 70: Do we need to keep this differentiation or can't we just jump to the new interpretation engine with flows? if we should keep two separate functions, can we at least make this wrapper function more DRY?

rust-executor/src/perspectives/flow_classes.rs
```diff
@@ -124,6 +133,122 @@ pub(crate) async fn mint_flow_instance(
...
+
+    for evid in rest_evidence {
+        perspective
+            .update_subject(
+                SubjectClassOption {
+                    class_name: Some(FLOW_TRANSITION_PROPOSAL_CLASS.to_string()),
+                    query: None,
+                },
+                uri.clone(),
+                serde_json::json!({ "evidence": evid }),
```
lucksus: can we not include these in the first create_subject call? and if not, can we at least batch these writes?

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,2734 @@
...
+        )
+        .await;
+        assert_eq!(minted.len(), 1, "matched hint + Pass ⇒ 1 proposal");
+        assert!(
+            read_rationale(&perspective, &minted[0]).await.is_none(),
+            "reason=None must NOT write a rationale — the SDNA allows omission and \
+             writing an empty scalar bloats the graph with a link carrying no signal",
+        );
+    }
+}
```
lucksus, line 1085: this file is way to big. please break down into sub modules. e2e tests may live in their own file.

lucksus (changes requested):
Not seeing LLM based integration tests. Do we not need them because all of this onion's flow code is based only on requires and model queries? what about flows that live off of the semantic check (following a conversation). Is that coming a follow-up PR? Or should it be added here?

rust-executor/src/perspectives/flow_classes.rs
data-bot-coasys: Two things going on here:

**They ARE batched.** `batch_id.clone()` gets passed to both the `create_subject` and every `update_subject(evidence: ...)` call, so all writes commit atomically as one unit — the loop doesn't produce N separate observable graph writes, it produces one. That's the existing atomicity contract shared with `mint_interpretation_run(...).sources` and every other collection-write callsite.

**The overhead we could still trim** is N+1 async round-trips through `execute_commands` (each call re-loads the class name, walks setter actions, then runs `execute_commands`). Even though they land as one commit, they don't need to be N+1 separate ticks.

**Why not fold into the first `create_subject` call:** `create_subject` binds ONE `target_value` per property key in `initial_values` (perspective_instance.rs:4695–4756). `values["evidence"] = arr` would serialize the array as a single string via `resolve_property_value`, not iterate. To fold N evidence elements into the constructor pass we'd need to change either:

1. `create_subject` to iterate arrays on collection setters (impacts every collection-using @Model — broader semantic change, ripple through SDNA `addLink` setter tests everywhere)
2. Or extend `update_subject` similarly (same concern, plus it's the doorway for every scalar-vs-collection consumer)
3. Or reach under both by building `Vec<Command>` inline here and calling `execute_commands` once — same shape you'd get, but replicates `create_subject`'s constructor+setter merge logic locally, which is fragile if that ever grows a case (e.g. `overwritten`-check on line 4712)

If you want the perf win, I'd take option 3 as a follow-up PR on top of #932 landing — build the `Vec<Command>` locally in `write_flow_transition_proposal` using `get_constructor_actions` + `get_property_setter_actions` (both public), keep the atomicity contract, drop from N+1 to 1 `execute_commands` call. Would keep create_subject untouched. Say the word and I'll open it.

For this PR: I'll leave the loop as-is since the atomicity ask is already met.

rust-executor/src/perspectives/interpretation/parse.rs
data-bot-coasys: Good catch — applied the symmetric or_else on the object-first branch. But I want to flag that as a **defensive** improvement only; the specific example you gave (`OK so I'll extract {a couple of things}: [{\"class\":\"X\"}]`) is NOT actually rescued by it, because `extract_bracketed` uses `find(open)..rfind(close)`:

- `find('{')` = position in "extract {"
- `rfind('}')` = position of the trailing `}` in `\"X\"}`
- Returns `Some("{a couple of things}: [{\"class\":\"X\"}]")` — bogus JSON, but **not None**, so `or_else` never fires and the retry burns anyway.

The or_else fallback only helps a narrower case: `{` present but no reachable `}` (or `}` before `{`) — then the primary returns None and array-recovery kicks in. Rare but zero-cost to catch, matches the array-first branch pattern, so worth applying.

**The deeper fix** for your actual example needs one of:
1. `extract_bracketed` becomes a balanced-bracket depth counter (returns `{a couple of things}`; still bogus JSON, but scoped correctly — retry then hits the array via or_else because the balanced primary returned Some("{a couple of things}") — wait no, still Some, or_else still doesn't fire)
2. `clean_llm_json` returns `Vec<String>` candidates and the parse function tries each until one parses (biggest impact, changes the parse contract)
3. Attempt-time bracket-kind guessing: if primary parse fails, retry loop rewrites raw with the opposite bracket kind (hacky but contained)

I lean 2 as the right architecture but that's a full slice's worth of work — happy to open it as a follow-up if you want. For this PR I've applied the symmetric-or_else (helps unbalanced-braces edge cases, doesn't help balanced-braces-in-prose). Tagging you if you want to pull #2 in here, otherwise I'll defer.

data-bot-coasys:
Good catch — you're right that there's an asymmetry in the LLM coverage on this PR. Confirmed by grep on `interpretation_e2e.rs` + `interpretation_harness_e2e.rs`: 24 `#[ignore = "llm-e2e"]` gated tests, **zero** of them touch `AIServiceSemanticCheck` / `SemanticCheckLlm`.

**Current coverage on this PR:**
- **10.4 requires-only path (deterministic model_query gate):** full real-graph e2e via `evaluate_flow_transitions_writes_to_graph` — no LLM in the loop by design; the whole point of slice 10.4 is that a satisfied requires guard mints a proposal without asking a model.
- **10.5 semanticCheck path (LLM gate on top of requires):** three tests exist (`semantic_check_gate_pass`, `semantic_check_gate_reject`, `semantic_check_absent_hint_autopass`), all wire a **stubbed** `SemanticCheckLlm` trait implementation. Zero real-LLM contact.
- **10.6 LLM-hint attribution path:** `hint_matches_writes_rationale` etc. — all stub-provided via `LlmProposalHint`, no real inference.

So the "conversation-following" flows — the ones designed to route through the semanticCheck gate — are proven to **wire** correctly (input → prompt → parsed decision → outcome), but the guarantee that a real local model on Marvin returns a parseable `PASS`/`REJECT` verdict on a realistic transcript has zero coverage.

**Recommendation: follow-up PR, not this one.** Reasoning:

1. **PR scope.** This one is already 11 slices deep (10.4a1 through 10.6c). Adding a semanticCheck real-LLM e2e means: (a) hand-write a conversation-shaped transcript that reliably triggers a semanticCheck (`requires` satisfied, semanticCheck present), (b) pick a Gemma3-family prompt shape that returns deterministic PASS on it, (c) mirror the same for a REJECT variant, (d) plumb it through the same nightly `LLM_E2E=1` gate we just landed in #943. That's a slice unto itself — call it 10.5d — and it earns its own PR body + review.

2. **Failure surface is different.** The 24 llm-e2e tests already gated to nightly are all `parse-shaped-JSON-from-Gemma3` flakes. semanticCheck failures would be `did-the-model-say-yes-or-no-cleanly` flakes — a distinct class that deserves its own retry policy + prompt-shape iteration.

3. **What we already have.** The `AIServiceSemanticCheck` production wire-through (slice 10.5c) IS live on the extraction path today. It just isn't proved-under-real-LLM in tests. If Marvin's Gemma3 returns garbage, the current stub tests don't catch it — but the extraction pipeline degrades correctly (semanticCheck failure → proposal not minted → flow doesn't advance → deterministic default, no crash).

**Proposal:** I open **PR-follow-up (slice 10.5d)** immediately after #932 merges, gated under the same `LLM_E2E=1` nightly workflow, with a two-transcript smoke: one deliberately-off-topic (PASS-worthy `requires` met but semanticCheck REJECT), one on-topic (both PASS). If you want it added here instead, say the word and I'll defer the merge.

Also worth noting: your original ask in slice 10.5 was "we can defer real-LLM tests" — this ties off that thread as an explicit follow-up commitment rather than a silent gap.

/cc @lal-bot-coasys for continuity.

rust-executor/src/perspectives/flow_evaluator.rs
data-bot-coasys: Done in `2cf8968a9` — 2721-line file → 6-file module:

- `primitives.rs` (223) — pure fns/types: `SatisfiedTransition`, `build_query_input_for_requires`, `cardinality_satisfied`, `evidence_hash`
- `queryable.rs` (224) — `RequiresQueryable` trait + async `evaluate_*` layer
- `engine_pass.rs` (338) — `run_engine_proposal_pass` + `write_engine_proposal` (composed pipeline)
- `unit_tests.rs` (853) — trait-stubbed unit coverage
- `e2e_tests.rs` (1051) — real-`PerspectiveInstance` coverage
- `mod.rs` (78) — module doc + tightened public re-exports

Also trimmed the re-export surface to only the 3 symbols external callers actually reference (`run_engine_proposal_pass`, `LlmProposalHint`, `SatisfiedTransition`) — the rest were internal helpers that no other crate code used.

`cargo check` + `cargo fmt` clean; CI running.

rust-executor/src/perspectives/interpretation/parse.rs
lucksus: Cant we use a RegEx to detect JSON? Or resort back some other library for parsing JSON out of text?

data-bot-coasys: Regex can't do it correctly — matching balanced brackets isn't a regular language, and a naive \`\\{.*\\}\` mangles nested objects, strings containing \`{\`, and unbalanced braces in prose.

Went with the library option though, since we already have serde_json: replaced the primary bracket-matching path with \`serde_json::Deserializer::from_str(&tail).into_iter::<Value>().next()\`, which consumes exactly one complete JSON value from an offset and reports the byte offset it stopped at (```StreamDeserializer::byte_offset```). Pushed as `7b437055b`.

Fallback chain preserved so trailing-comma inputs still get extracted for step 4 to salvage:

```
serde.next('[')  →  serde.next('{')  →  extract_bracketed('[', ']')  →  extract_bracketed('{', '}')  →  raw
```

Well-formed inputs (common case) now go through serde and skip `strip_trailing_commas` entirely — serde already parsed them. Malformed ones fall through to the old path exactly as before. All 10 existing regression tests still cover the same corner cases.

data-bot-coasys:
Update since the last review round — this is ready for another look.

**What changed**

1. **Merged `dev`** (`a4136425c`) now that #929 (interpretation hints) and #934 are in — clean merge, zero conflicts. One semantic collision surfaced at compile time and is fixed; nothing else needed adjustment.
2. **CodeRabbit's `parse.rs` finding fixed** (`b8f263530`, raised twice, previously unaddressed). `extract_bracketed` returned any greedy `open..close` slice, so prose like `OK, I'll extract {a couple of things}: [{"class":"X"}]` short-circuited the fallback chain with malformed JSON and the real array after it was never tried. It's now wrapped in `extract_bracketed_salvageable`, which accepts the slice only if it parses as JSON *after* the trailing-comma cleanup step 4 applies anyway — so the salvage path (its whole reason to exist) is preserved while genuine prose brackets fall through to the other bracket kind. Regression test added.

**Verification**

- CI fully green on the merge commit (all 8 checks).
- Locally on the merged branch: flow suite **141/141**, interpretation suite **104 passed / 0 failed**. One interpretation e2e initially failed purely because the Marvin Ollama tunnel had dropped; with the tunnel restored it passes in 7s — infra, not the merge.
- `cargo fmt --all --check` clean.

**Stacking note:** #950 (the simplification) is based on this branch, so this stays the merge-first branch and picks up nothing from downstream. I've approved #950 with three non-blocking inline comments; merging it means #940 needs a genuine re-stack, which I'll handle.

data-bot-coasys:
Heads-up on a latent issue in this branch — **not** blocking, and deliberately not pushed here so I don't reset your green CI right before a merge.

`load_all_flow_instances` (added on this branch) guards the absent-class case with `msg.contains("Shape not found")`. That string is emitted **nowhere in the tree** — `model_query/shape.rs:66` produces `"No SHACL shape stored for class 'FlowInstance'. Call ensureSubjectClasses / addSdna first."`. So on a perspective where no FlowInstance was ever minted, the guard misses and the function returns `Err` instead of the empty vec it documents. `load_flow_instances` on dev has the identical bug (same copied guard).

Today it's invisible: `run_engine_proposal_pass` treats the `Err` as "no records" and returns empty anyway, so you only pay a spurious warning per call. It stops being invisible the moment a caller distinguishes "failed" from "none" — which is exactly what bit me in #954, where the spawn pass treated the error as "skip this item" and could therefore never mint the *first* instance on any perspective.

**#954 already fixes `load_flow_instances`** (matches the message actually produced). Whichever of these two merges second should get the same one-line treatment on `load_all_flow_instances` so the sibling loaders don't diverge — happy to push it here instead if you'd rather it land with this PR.

data-bot-coasys:
Housekeeping on this branch — two commits, no functional change:

- `c88c07bdb` + `e0ca3726e` remove five scratch files that `b8f263530` committed by mistake through an over-broad `git add`: `rerun-one.log`, `rerun-one.sh`, `run-932-tests.sh`, `run-tests.sh`, `test-run.log`. They are one-machine CI-rerun/test-driver helpers, absent from `dev`, and would otherwise have landed on `dev` when this merges. Verified with a full tree diff against `dev` that nothing else non-source remains on the branch.

Also: #950 was showing `CONFLICTING` against this branch (my two pushes here landed after it branched off). Resolved there by merging this branch in — the `parse.rs` conflict went to #950's array-first chain, which subsumes the `extract_bracketed_salvageable` guard from `b8f263530`; reasoning is written up in https://github.com/coasys/ad4m/pull/950#issuecomment-5523536319. #950 is `MERGEABLE` again.

CI is re-running here on `e0ca3726e`.

rust-executor/src/perspectives/flow_classes.rs
lucksus: So it's batched as long as `batch_id` is not none. I think we should teach create_subject to deal with collection correctly. Shouldn't too big of a refactor, right? if then this code-site could let go of the for loop, it might even save lines changed in this PR... please take a look at that.

data-bot-coasys: Done in `d8b21748e` — you were right on both counts: small refactor, and it comes out roughly line-neutral (+83/−77 across the PR, with this call site dropping from ~50 lines to ~12).

**The rule I implemented in `create_subject` + `update_subject`:** a JSON array on a property whose setter actions are all `addLink` expands into one setter execution per element (N links, same batch). Any other array keeps the existing `literal:json:` blob encoding — that's deliberate, because a *scalar* property may legitimately hold a JSON array value, and expanding over `setSingleTarget` would run last-element-wins and silently drop the rest. So the setter shape is the gate, and the existing SDNA shape test (`evidence_property_is_a_collection_with_add_link_setter`) now locks exactly that gate.

One subtlety inside `create_subject`: only the **first** element may overwrite a constructor placeholder command (predicate match); subsequent elements always append — otherwise all elements would collapse onto the same command.

**Side effect worth knowing:** the MCP `create_subject` tool and the extraction apply path pass caller/LLM values straight through, so an array on a collection property there used to become a `literal:json:` blob that the read paths (which expect N links) could never round-trip. Those now write correct collections for free.

This code site now just does `values["evidence"] = json!(evidence_ids)` — no split_first, no follow-on `update_subject` loop. Verified: flow 81/0 (incl. the multi-evidence e2e `write_flow_transition_proposal_lands_all_predicates_e2e`, which exercises the array path against the real graph), interpretation 85/0, subject 17/0, fmt clean. CI running on the new head.

lucksus (approved):

rust-executor/src/perspectives/flow_evaluator.rs
data-bot-coasys: Good question, and it's done — your option 3 (pre-write dedup query) is what landed: `proposal_already_exists` (commit `6673de4cc` on this PR) checks for an existing proposal matching `(evidence_hash, flow instance, to_state)` before every mint, and the e2e asserts a re-run of the same pass mints nothing. So the N passes × K transitions amplification is gone: one satisfied-but-unconsumed transition = one on-graph proposal, however many passes re-see it. The invariant and its why (proposal-count consensus gaming) are spelled out in the doc comment on `proposal_already_exists`, per your suggestion. On #940 the lookup was further hardened to fail *closed* — a transient store error skips the mint instead of duplicating it.

One nuance your framing surfaces that I want on record for the consensus PR: the dedup is **DID-agnostic** — it matches by evidence/instance/state regardless of proposer. Correct today (nothing counts proposals yet, and each pass mints on behalf of one acting DID). But when the slice-10.6 consensus loop counts *distinct DIDs*, a second agent whose replica already synced agent A's proposal would be suppressed from minting its own — quorum could never grow past 1. So the consensus PR (`feature/flow-mutations`) needs to add the proposer dimension to this match (mint when no proposal *from this DID* exists for the tuple). Noting it there rather than changing it here, since per-DID semantics belong to the PR that defines counting.

PR #935, opened 2026-08-28
feat: server-side SHACL shape resolution via WS-RPC

HexaField:
Moves SHACL shape resolution from the TypeScript SDK into the executor as four new WS-RPC endpoints:

- `perspective.getShaclNames` — list shape names (1 call vs 1)
- `perspective.getShaclTargetClass` — resolve target class by name (1 call vs 2)
- `perspective.getShacl` — full shape + properties by name (1 call vs 3+N)
- `perspective.getAllShacl` — all shapes in bulk (1 call vs 1+N×(3+M))

**Problem:** WE's `foreignSchemas()` generates ~261 individual `queryLinks` WS round trips per dataset switch (26 shapes × sequential resolution per shape). Each shape walk needs name→URI→targetClass→properties→property-details, all as separate `queryLinks` calls.

**Solution:** The executor now resolves the full multi-step link walk in-process against its local SQLite store and returns assembled link triples. The SDK's `PerspectiveProxy` methods delegate to these endpoints — same API surface, single round trip each.

## Summary by CodeRabbit

- **New Features**
  - Added server-side SHACL resolution for retrieving shape names, target classes, individual shapes, and complete shape collections.
  - SHACL data now preserves property metadata and supports consistent results across single and bulk retrieval.

- **Improvements**
  - Flow availability now matches typed inputs against registered subject classes, while untyped flows remain broadly available.
  - Missing SHACL target classes now use a consistent “not found” representation.

- **Breaking Changes**
  - Removed legacy flow state, action, and flow-starting methods from the public interface.


HexaField:
CI is failing at checkout due to `no space left on device` on the `marvin` CircleCI runner — same failure on all current open PRs. The code compiles and tests pass locally (Rust `cargo check` clean, TypeScript `tsc --noEmit` clean, 10 new + 55 existing tests green). Will go green once the runner's disk is cleaned up.

⬡

HexaField:
hey @marvin-bot-coasys it looks like you have no space left to run CI

rust-executor/src/api/perspectives_ws.rs
```diff
@@ -1878,6 +1879,253 @@ async fn interpretation_overlays_handler(
...
+    // Step 3: find property shape URIs
+    let prop_uris: Vec<String> = shape_links
+        .iter()
+        .filter(|l| l.data.predicate.as_deref() == Some("sh://property"))
+        .map(|l| l.data.target.clone())
+        .collect();
+
+    // Step 4: fetch links from each property sub-shape (all independent reads)
+    let mut all_prop_links = Vec::new();
+    for prop_uri in &prop_uris {
```
marvin-bot-coasys: The old TS `getShacl` fired these property-sub-shape reads with `Promise.all` precisely because they're independent — the sequential `for prop_uri in &prop_uris { … .await }` here gives that up. Against a local SQLite store each read is cheap, so the round-trip win still dominates, but a shape with ~10 properties now pays 10 sequential awaits inside the handler where it could pay 1. `futures::future::try_join_all(prop_uris.iter().map(|u| perspective.get_links(&shacl_link_query(u, None))))` recovers the concurrency without changing semantics. Same pattern applies to the outer `for name in &names` loop in `get_all_shacl` (line 2116).

rust-executor/src/api/perspectives_ws.rs
```diff
@@ -1878,6 +1879,253 @@ async fn interpretation_overlays_handler(
...
+                .strip_prefix("literal:string:shacl://")
+                .map(|n| n.to_string())
+        })
+        .filter(|name| seen_names.insert(name.clone()))
+        .collect();
+
+    // Step 2: resolve each shape's full link set
+    let mut results = Vec::new();
+    for name in &names {
+        if let Some((shape_uri, links)) = resolve_shacl_links(&perspective, name).await? {
```
marvin-bot-coasys: Two things I'd flag on this line, both non-blocking:

1. **Silent drop.** If a name from the `has_shacl` enumeration resolves to no `shacl_shape_uri` link inside `resolve_shacl_links` (returns `None`), the entry is quietly dropped from the result with no log. Since the name lookup and the shape resolution are separate `get_links` calls with no shared snapshot, a concurrent unlink between them will make a shape vanish from `getAllShacl` output without any signal. At minimum a `tracing::warn!` on `None` would make this diagnosable; ideally the name enumeration and each shape's resolution share a single perspective snapshot.
2. **Snapshot consistency.** More broadly, `resolve_shacl_links` performs 3+N independent `get_links` calls against the perspective; a concurrent edit interleaved with those reads can surface a partial shape (property link visible, property details not yet visible, or vice versa) that would never have existed as a coherent state. SHACL shapes are typically static so this is a theoretical concern today — but flagging because moving the walk server-side is exactly the point where snapshot semantics become the executor's responsibility to define.

rust-executor/src/api/perspectives_ws.rs
```diff
@@ -1878,6 +1879,253 @@ async fn interpretation_overlays_handler(
...
+        None => Ok(Value::Null),
+    }
+}
+
+/// Retrieve all SHACL shapes in one call.  Returns an array of
+/// `{name, shapeUri, links}` objects — the client reconstructs each shape
+/// with `SHACLShape.fromLinks(entry.links, entry.shapeUri)`.  Equivalent to
+/// the SDK's `PerspectiveProxy.getAllShacl()` — one handler call replaces
+/// 1 + N×(3+M) `queryLinks` round trips (N shapes, M properties each).
+async fn get_all_shacl(params: Value, ctx: Arc<RequestContext>) -> Result<Value, WsRpcError> {
```
marvin-bot-coasys, line 2166: `getAllShacl` returns the entire shape corpus in a single WS message with no page/limit/stream option. For the 26-shape WE dataset that motivates this PR it's obviously fine, but the endpoint is registered as general-purpose SDK surface — a perspective with a few hundred shapes at ~10 properties each is a multi-MB response inside one WS frame, and whatever the transport-level frame limit is will fail loudly (or worse, silently truncate depending on the client). Not blocking now, but worth either documenting the intended scale ceiling in the doc comment or adding a `limit`/`offset` param before this ships to third-party plugin authors.

core/src/perspectives/shacl-rpc.test.ts
```diff
@@ -0,0 +1,216 @@
...
+          { name: 'Empty', shapeUri: 'app://EmptyShape', links: [] },
+        ]),
+      });
+      const proxy = createProxy(client);
+
+      const shapes = await proxy.getAllShacl();
+
+      // Both should appear since fromLinks always returns a SHACLShape
+      // (even with no targetClass), but the filter only drops null
+      expect(shapes.length).toBe(2);
```
marvin-bot-coasys, line 187: This test is named `filters out shapes that fail reconstruction` and asserts `shapes.length` is **2** — with the inline comment openly admitting that `fromLinks always returns a SHACLShape (even with no targetClass), but the filter only drops null`. So the test doesn't actually exercise the filter it claims to test; it's asserting that nothing is filtered. Either drop the test, or if the null-filter branch on line 1501 of PerspectiveProxy.ts is genuinely reachable, mock `SHACLShape.fromLinks` to return `null` for one entry and assert it's dropped. As written this is dead coverage.

core/src/perspectives/shacl-rpc.test.ts
```diff
@@ -0,0 +1,216 @@
+import { PerspectiveProxy } from './PerspectiveProxy';
+import { SHACLShape } from '../shacl/SHACLShape';
+
+// ── Helpers ──────────────────────────────────────────────────────────────────
+
+function createMockClient(overrides: Record<string, jest.Mock> = {}): any {
+  return {
+    addPerspectiveLinkAddedListener: jest.fn(),
```
marvin-bot-coasys, line 8: The whole test surface here mocks `PerspectiveClient` — so every test verifies TS-side delegation and `SHACLShape.fromLinks` reconstruction, but none of them verify the load-bearing claim of this PR: *that the Rust walk returns the same triple set the old client walk did*. If the server-side dedup, property-URI collection, or the `strip_prefix("literal:string:shacl://")` parsing drifts from what the SDK expects, these mocks will keep passing. A Rust-side integration test in `rust-executor` that seeds a perspective with a known shape (name → URI → targetClass → 2–3 properties with datatypes and cardinalities) and asserts the handler's returned triple set would close that gap and would also give you a regression harness the next time someone edits `resolve_shacl_links`.

core/src/perspectives/PerspectiveClient.ts
```diff
@@ -496,6 +496,43 @@ export class PerspectiveClient {
...
+    // from its local store in-process.  Each call replaces the multi-round-trip
+    // `queryLinks` sequences the old PerspectiveProxy methods performed.
+
+    /** List the names of every SHACL shape stored in a perspective (one RPC call). */
+    async getShaclNames(uuid: string): Promise<string[]> {
+        return this.#apiClient.call<string[]>('perspective.getShaclNames', { uuid })
+    }
+
+    /** Resolve a shape's `sh:targetClass` by name (one RPC call). */
+    async getShaclTargetClass(uuid: string, name: string): Promise<string | null> {
```
marvin-bot-coasys: Nit / API surface consistency: `getShaclTargetClass` returns `Promise<string | null>` here, while `PerspectiveProxy.getShaclTargetClass` (line 1487) maps that `null` back to `undefined` to preserve the old signature. Two representations of the same "shape not found" state one layer apart. Consider unifying — either the client returns `undefined` too (drop the null across the wire boundary) or the proxy stops the null→undefined coercion and updates its return type. Cheap tidy-up while the caller sites are still fresh.

marvin-bot-coasys (dismissed):
Nice, high-leverage change — collapsing the WE dataset switch from ~261 WS round trips into one is exactly the shape of fix this hot path needed, and the executor-side walk is a faithful 1:1 of the old TS logic (same `has_shacl` → `shacl_shape_uri` → shape links → `sh://property` sub-shapes → dedupe), with the bonus that the auth gate (`perspective_query_capability`) is properly wired and the wire format (`{source, predicate, target}` + `{name, shapeUri, links}`) reuses `SHACLShape.fromLinks()` unchanged so the SDK contract is preserved. Two things I'd love to see addressed before merge, but neither is blocking: (1) the server-side walk is now fully sequential (`for prop_uri…`, then `for name in &names` in `get_all_shacl`) where the old client did `Promise.all` — against local SQLite each read is cheap, but for a 26-shape / ~10-property dataset that's still a few hundred serialized awaits inside one handler; `futures::future::try_join_all` would recover the concurrency we deliberately had before, and matters more the moment anyone puts a bigger schema through this. (2) There's no server-side integration test for the central correctness claim — the new `shacl-rpc.test.ts` only exercises TS-side delegation via a mocked client, so a subtle drift between the Rust walk and `SHACLShape.fromLinks()`' expectations would not be caught here; a Rust-side test that seeds a known perspective and asserts the returned triple set would give this PR much stronger footing. Smaller nits inline: unbounded response size on `getAllShacl`, silent drop when a shape name resolves to no URI mid-walk, and one test whose name promises filtering it doesn't actually exercise. Approving on the merits — the design is right and the code is clean; treat the inline items as follow-ups.

marvin-bot-coasys:
Pushed 6 fix-up commits addressing the review — Nico asked me to just apply them rather than leave them as follow-ups.

| # | Fix | Commit |
|---|---|---|
| A | Rust-side integration coverage for the 4 SHACL RPC endpoints (`get_shacl_names`, `get_shacl_target_class`, `resolve_shacl_links`, bulk walk composition) + a concurrency race test where `sh://targetClass` is unlinked between name enumeration and shape resolution | `5372f5782` |
| B | Rename the mislabeled "filters out shapes that fail reconstruction" test to describe what it actually asserts, and add a real filter case that spies on `SHACLShape.fromLinks` to force `null` for one entry so the proxy's null-filter branch is genuinely exercised | `4fd4042cc` |
| C | Concurrent per-shape resolution in `get_all_shacl` via `futures::future::try_join_all` — both the per-property fan-out inside `resolve_shacl_links` and the per-shape fan-out in `get_all_shacl` (order preserved by zipping names with results after the join) | `03ca9ae11` |
| D | `debug!` log inside `resolve_shacl_links` when a `has_shacl`-enumerated name resolves to no `shacl_shape_uri` mid-walk (the silent-drop race in r3897752007). Debug-level per LOGGING.md — expected in-race behaviour, not a bug | `efc99259b` |
| E | Hard `MAX_SHACL_SHAPES_PER_RESPONSE = 500` cap on `get_all_shacl`, returning a `bad_request` pointing at the (TODO) paginated variant rather than serialising a multi-MB response into one WS frame | `22c7d59d2` |
| F | Unify `getShaclTargetClass` on `Promise<string \| undefined>` at the `PerspectiveClient` layer and drop the redundant null→undefined coercion in `PerspectiveProxy` | `97d15851e` |

**Primary addition is (A)** — the Rust-side integration test pins the central correctness claim of this PR (that the Rust walk returns the same triple set the old TS client walk did), and it immediately caught a real bug: the SPARQL store canonicalises `literal:string:` targets through a URL-decoded `xsd:string` round-trip, re-rendering them on read as `literal:string:shacl%3A%2F%2F<Name>`. The original `strip_prefix("literal:string:shacl://")` in `get_shacl_names` / `get_all_shacl` silently returned an empty name list against that canonical form. The fix (`shape_name_from_has_shacl_target`) accepts both the raw and percent-encoded shapes. Slice A includes this fix so the test suite lands green. All 5 Rust tests + 11 TS tests pass.

**Behaviour change to flag explicitly (slice E):** perspectives with more than 500 SHACL shapes now receive a `bad_request` from `perspective.getAllShacl`. The cap sits well above the 26-shape WE dataset that motivated the PR, but if a real workload needs a higher ceiling before the paginated variant lands, bump `MAX_SHACL_SHAPES_PER_RESPONSE` in `rust-executor/src/api/perspectives_ws.rs`.

Not re-requesting review; leaving the merge call with Nico as usual.

lucksus (dismissed):

PR #938, opened 2026-08-28
chore: Clean-up warnings and wire missing billing paths for OpenAI compat. interface

lucksus:
Some warnings were simple fixes, some revealed that the OpenAI compat. interface did not exercise billing in all paths. Both fixed.

### Qwen3.8:

# Fix missing billing on streaming chat + clean up warnings and dead billing code
Two related cleanups on top of the host_rates billing unification (be2f390bc, 45827c0b4):

1. Bill streaming chat completions (the actual bug)
POST /v1/chat/completions with stream: true was pre-checked for credits but never billed — the auth token arrived at the stream path as an unused _auth_token. It now bills exactly once, on success, through the same bill_ai_operation / host_rates path as oneshot and WS-RPC:
AIService::prompt_messages_stream relays the worker’s final Result<PromptResult> through a forwarder; new bill_and_forward_stream_result bills via bill_prompt_if_authed only when the stream completed successfully, then forwards the result to the caller (worker panic still delivers an error, bills nothing).

2. Remove orphaned handler-level billing formulas
MIN_PROMPT_BILL, chat_or_completion_amount, stream_prompt_amount, embedding_amount lost their only production callers when billing moved into AIService — their pinning tests asserted formulas production no longer used (e.g. streaming at a flat 1.0 credit). Removed them plus the 5 stale tests; speech_amount (TTS, still used by audio.rs) is kept.

3. Warning cleanup (15 lib + 2 test warnings)
Unused imports (DnaHash, CoreError, RequestedModuleType, …), a dead match arm, and a sighandler_t cast.

## Tests
3 new tests in api/openai_compat: stream_completion_bills_once_on_success (exactly one debit, amount = tokens × host rate, correct op label/summary), stream_error_does_not_bill, stream_without_token_does_not_bill
cargo check --lib --tests: 0 warnings
cargo test --lib api::openai_compat: 91 passed

## Summary by CodeRabbit

* **Billing**
  * Successful streaming AI responses are now billed consistently with non-streaming requests.
  * Failed or unauthenticated streams are not billed.

* **Bug Fixes**
  * Improved handling when streaming results cannot be forwarded to the caller.
  * Removed obsolete endpoint-level billing logic while preserving quota checks and response behavior.

* **Maintenance**
  * Cleaned up unused code and imports.
  * Tightened signal handling so unexpected signal variants are detected during development.

### DIFF
--- a/rust-executor/src/ai_service/mod.rs
+++ b/rust-executor/src/ai_service/mod.rs
@@ -1375,6 +1375,11 @@ impl AIService {
     /// Streaming variant of [`Self::prompt_messages`].  Returns a token
     /// stream + a oneshot for the final [`PromptResult`] (carries the
     /// full text + token counts for the closing SSE event).
+    ///
+    /// Billing happens when the stream completes: the worker's final
+    /// result is billed (on success only, matching
+    /// [`Self::prompt_messages`]) in a forwarder task before it reaches
+    /// the caller's `done_rx`.
     pub async fn prompt_messages_stream(
         &self,
         model_id: String,
@@ -1403,6 +1408,10 @@ impl AIService {
         spawn_rx.await??;
 
         let (token_tx, token_rx) = mpsc::unbounded_channel::<String>();
+        // worker -> forwarder -> caller: the billing hook runs between the
+        // worker's final result and the caller's done_rx, so stream:true
+        // requests are charged like the non-stream path (success only).
+        let (worker_done_tx, worker_done_rx) = oneshot::channel::<Result<PromptResult>>();
         let (done_tx, done_rx) = oneshot::channel::<Result<PromptResult>>();
         {
             let llm_channel = self.llm_channel.lock().await;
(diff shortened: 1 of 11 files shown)


jhweir (commented):
Verdict: the production changes are correct. No correctness bugs found.

Findings, all test-quality plus one stale comment:

- openai_compat/tests.rs:1074 — stream_error_does_not_bill never enables multi-user mode, so user_email_from_token returns None regardless. The test passes even with the if let Ok(pr) gate deleted, and its behavior depends on whether a sibling test's multi-user window overlaps on another thread.
- openai_compat/tests.rs:984 — user_jwt_token signs with the global wallet's "main" key, but test_utils::setup_wallet() (four other tests in the same --lib binary) unconditionally regenerates it. A parallel run between minting and decode_jwt makes stream_completion_bills_once_on_success flaky.
- openai_compat/tests.rs:1035 — multi_user_enabled/host_rates are mutated process-wide and restored only after the assertions; an assertion failure leaks that state into the rest of the binary. Wants a Drop-based restorer.
- rust-executor/src/lib.rs:186 — the _ => arm was deleted but the comment above Signal::AppDirect still says future variants "fall through the wildcard" and log (from #907 MED-4). Either restore the arm under #[allow(unreachable_patterns)] or drop the sentence.

Nothing blocking a merge; #2 is the one worth fixing since it's a real flake source in CI.

PR #940, opened 2026-08-29
flow: LLM-attributed FlowTransitionProposals via harness tool-call surface

data-bot-coasys:
# Flows: let the LLM explain *why* a transition should fire — without letting it decide

Stacked on **#932** (review that first). The diff shown here is **this feature only** (+1695/−629 vs #932); it collapses to its own commits once #932 lands on dev, and GitHub will retarget the base automatically.

## What this adds, in one paragraph

#932 gave flows a deterministic engine: after each interpretation run, it checks every active flow's `requires` guard against the graph and writes a `FlowTransitionProposal` when a guard is satisfied. Those proposals are structurally correct but mute — they can't say *why* the transition makes sense. This PR lets the LLM attach a rationale to a proposal, on both extraction paths, plus an optional second LLM pass (`semanticCheck`) for conditions that structural evidence alone can't settle ("the scope was *agreed on*, not merely discussed").

## The one rule that makes this safe

**The LLM proposes; the engine disposes.** Everything the LLM says is a *hint*. A proposal is only ever written by the deterministic engine pass when the structural `requires` guard is satisfied. A hint that matches a satisfied transition contributes its `reason` as the proposal's on-graph `rationale`; a hint that matches nothing is dropped. The LLM cannot fire a transition, skip a guard, or invent a proposal.

## How the LLM hands over its rationale

There are two ways an interpretation run talks to an LLM, and each needs its own carrier:

1. **Strategy path** (LLM returns one JSON blob): the blob may carry a `flow_proposals` array next to `instances`.
2. **Tool-calling path** (LLM works through tool calls): each active flow instance exposes one `{FlowName}_propose_transition` tool. Calling it buffers the hint and returns `{"ok": true}` — no graph mutation mid-run.

Both paths funnel into the same post-processing step (`run_flow_post_pass`), so the engine gate, dedup, and rationale-matching logic exist exactly once.

```mermaid
sequenceDiagram
    autonumber
    participant LLM
    participant Run as Interpretation run
    participant Engine as Engine pass (deterministic)
    participant Graph as Perspective graph

    LLM->>Run: hints (JSON field or tool calls)
    Run->>Engine: run_flow_post_pass(hints)
    Engine->>Graph: evaluate requires guards
    alt guard satisfied
        opt state has a semanticCheck hint
            Engine->>LLM: yes/no confirmation — only a clear YES passes
        end
        Engine->>Graph: write proposal (rationale = matching hint's reason, if any)
    else guard not satisfied
        Engine--)Graph: nothing — hint dropped
    end
```

## Tour of the diff

| File | What to look at |
|---|---|
| `ai_service/harness/flow_propose.rs` (new, 729) | The tool surface: tool naming (`{FlowName}_propose_transition`), JSON-Schema per flow instance (`instance` pinned via `const`, `toState` enumerated from reachable states), the `FlowTransitionProposeProvider` decorator, arg parsing via the same serde type the strategy path uses (`LlmFlowProposal` — one parser per wire shape). ~Half is unit tests. |
| `perspectives/flow_semantic_check.rs` (new) | The optional second LLM pass. The prompt carries the *content* of each evidence instance (hydrated from the `requires` query results, size-capped), so the LLM judges property values, not URIs. Fail-closed: NO, UNCLEAR, LLM errors, and zero-evidence transitions all discard. Behind a one-method trait so tests script the LLM. |
| `perspectives/flow_evaluator.rs` (+167/−421) | Net simplification. The evaluator collapses into one shape serving both paths; proposal **idempotency** lives here (re-running a pass never mints a duplicate proposal — matched on evidence hash + instance + target state; lookups fail closed, so a store error skips the mint rather than duplicating it). |
| `perspectives/interpretation/run.rs`, `types.rs`, `prompt.rs` | Wiring: `run_flow_post_pass` extracted so both paths share it; prompt legend documents `$flow.uri` / `$flow.instance` as aliases. |
| `perspectives/interpretation/parse.rs` (+133/−19) | Hardened JSON recovery: strict parse at top-level bracket positions (string-aware depth scan) before greedy fallbacks. Fixes two real bugs found during development — a prose-brace hijack and wrapper-payload truncation that silently dropped flow proposals. Regression tests for both. |
| `ai_service/harness/propose.rs` (+8/−85) | The class-proposal buffer and the flow-hint buffer were verbatim copies; both are now one generic `HintBuffer<T>`. |
| `perspectives/flow_evaluator_e2e.rs` (+287/−79) | End-to-end proof: real perspective, scripted LLM; asserts the rationale lands on-graph and that a re-run mints nothing new. |

## Not in this PR

- **Consensus firing** — proposals accumulate; nothing advances `currentState` yet. Next PR.
- **Client-side subscriptions** (`onProposalAdded` etc.) — comes with the mutation surface.

## Testing

- 280/0 across the harness, interpretation and flow suites locally (`--test-threads=1`); CI green on the head.
- Real-LLM e2e (Ollama-backed) gated behind the nightly `llm-e2e` job, same convention as the other real-LLM tests.

---

<sub>History: this branch absorbed #951 (a simplification pass over the original implementation) at Nico's request — #951 auto-closed as merged, and this diff is the final shape. Earlier revisions of this description referenced internal slice numbers from the design doc; this version stands alone.</sub>

## Summary by CodeRabbit

- **New Features**
  - AI-assisted interpretation can propose valid flow transitions with destination states and optional reasons.
  - Flow transitions can require semantic confirmation before proposals are created.
  - Proposal rationales may include matching explanations supplied by the AI.
  - Interpretation responses can include extracted instances and flow transition proposals.
  - Semantic checks now use detailed supporting evidence when available.
  - Flow processing is limited to the relevant subjects.

- **Documentation**
  - Documented `$flow.instance` as an alias for `$flow.uri` in `where` template values.

- **Bug Fixes**
  - Semantic confirmation now accepts only an exact `YES` response.

- **Tests**
  - Added coverage for confirmation outcomes, validation, rationale handling, evidence, and response formats.


rust-executor/src/perspectives/flow_semantic_check.rs
```diff
@@ -0,0 +1,160 @@
...
+    p.push_str(&format!(
+        "## Semantic check to verify\n{hint}\n\n## Evidence found in the graph\n"
+    ));
+    if transition.evidence_ids.is_empty() {
+        p.push_str("(none)\n");
+    }
+    for id in &transition.evidence_ids {
+        p.push_str(&format!("- {id}\n"));
+    }
+    p.push_str(
```
marvin-bot-coasys, line 58: **The semantic-check prompt has no evidence content — only URIs.**

```rust
for id in &transition.evidence_ids {
    p.push_str(&format!("- {id}\n"));
}
```

This is the whole graph the confirmation LLM sees, plus FROM/TO and the hint string. `semanticCheck` exists for conditions structural `requires` cannot settle ("the scope was *agreed on*, not merely discussed"). A list of `ad4m://task/1` does not contain agreement, discussion, or anything else a yes/no can rest on.

Fail-closed (`NO`/`UNCLEAR`/error discard) is the right *disposition*, but it does not make the call meaningful. Either:

- hydrate each evidence id (title / body / relevant links / a short transcript slice) into the prompt so the LLM can actually answer, or
- do not call the LLM until that hydration exists — a `requires`-satisfied transition with an unhydratable `semanticCheck` should skip the same way an LLM error does, rather than asking a model to rubber-stamp identifiers.

The e2e (`semantic_check_gate_fires_only_on_yes_e2e`) asserts the gate's YES/NO wiring, not that the prompt is sufficient to evaluate `SCOPE_HINT`. That's the missing test for this slice's central claim.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -410,89 +298,95 @@ pub async fn run_engine_proposal_pass(
...
     {
         Ok(links) => links,
         Err(_) => return false,
     };
-    let to_state_literal = format!(
-        "literal:string:{}",
-        urlencoding::encode(&transition.to_state)
-    );
     for link in &hash_links {
         let proposal_uri = &link.data.source;
```
marvin-bot-coasys, line 562: **Idempotency lookup fails open — the opposite of the invariant the comment states.**

```rust
Err(_) => return false,
```

`false` means "no existing proposal, mint another." A transient `get_links` error (lock, timeout, cancelled query) therefore produces a second `FlowTransitionProposal` with a new UUID for the same `(instance, to_state, evidence_hash)`.

The comment immediately above this function is explicit about why that is load-bearing: minting does not advance `currentState`, so a later pass re-sees the same satisfied transition, and a consensus rule that counts proposals rather than distinct DIDs can be gamed by one agent re-running its pass. The happy-path e2e covers the successful lookup. The error path undoes the guarantee.

Fail closed: `Err(_) => return true` (treat as already-proposed / skip the write), and log at `warn`. A missed mint on a blip is recoverable on the next pass; a duplicate mint is the thing this function exists to prevent.

rust-executor/src/ai_service/harness/flow_propose.rs
```diff
@@ -0,0 +1,729 @@
...
+        let mut seen = std::collections::HashSet::new();
+        for ctx in &self.contexts {
+            let name = propose_transition_tool_name(&ctx.flow_name);
+            if seen.insert(name) {
+                out.push(propose_transition_tool_schema(ctx));
+            }
+        }
+        out
+    }
+
```
marvin-bot-coasys, line 245: **First-wins `const` on `instance` hides every other live instance of the same flow from schema-obedient models.**

`tools()` de-dupes by flow name and emits `propose_transition_tool_schema` for the first `FlowContext` only. That schema pins `instance` via JSON Schema `const` to *that* URI. `call()` then accepts any matching instance URI — so a model that *ignores* the schema can still attribute instance B, and a model that *obeys* it cannot even name instance B without a validation error from the provider.

The comment accepts this versus rewriting tool names. That's a real product hole, not just a cache-stability trick: two Delivery instances in one pass, only the first can receive an LLM rationale. Either:

- drop the `const` (keep the dispatcher allow-list; put valid URIs in the description / an `enum` of every matching instance), or
- emit `{FlowName}_{short-id}_propose_transition` per instance so the surface is honest.

Empty-enum terminal-state tools have the same "keep the name stable" motivation and are fine. The `const` pin is the part that lies to the model about what it is allowed to name.

marvin-bot-coasys (changes requested):
## Review — request changes

The split is right: the LLM never writes a proposal, the engine never takes a hint as a command, and both interpretation paths drain into one `run_engine_proposal_pass`. That contract is actually tested — unmatched `{instance, toState}` hints mint nothing extra, a matching `reason` lands as `rationale`, YES/NO/UNCLEAR/error on `semanticCheck` behave as advertised, and a re-run does not mint a duplicate when the lookup succeeds. `HintBuffer<T>` collapsing the two propose buffers, and the parse.rs top-level-bracket scan (prose-brace hijack + wrapper truncation), are both real simplifications rather than drive-by cleanup.

Two holes in the new machinery still need a fix before this should land:

1. **`semanticCheck` cannot evaluate the condition it exists for.** `build_semantic_check_prompt` lists evidence as bare URIs (`- ad4m://task/1`). The motivating case — "the scope was *agreed on*, not merely discussed" — is not decidable from identifiers. As written the second pass is a coin flip with extra latency and billing, fail-closed or not.
2. **`proposal_already_exists` fails open.** `get_links` `Err` returns `false`, so a transient store error mints a duplicate. The comment two lines above is explicit that duplicate proposals game a DID-count consensus rule. Idempotency checks fail closed: on lookup error, skip the write.

Nits (not blocking): `semantic_check_passed` accepts `Y`/`TRUE`/`PASS` while the prompt asks for `YES`/`NO`/`UNCLEAR`; first-wins `const` on `instance` hides extra live instances of the same flow from obedient models (documented, still a product hole); harness path logs minted proposal URIs instead of returning them the way the strategy path does.

Stacked on #932 — I reviewed this feature's own commits, not the base.

rust-executor/src/perspectives/flow_semantic_check.rs
data-bot-coasys: Fixed in `2eb03c4cf` — and you were right that hydration was the correct branch, because it turned out to be nearly free: `evaluate_query` already had each matched instance's full JSON from `model_query` and was throwing everything but the `id` away. `SatisfiedTransition` now carries `evidence: Vec<EvidenceItem>` (id, class, content JSON) alongside `evidence_ids`, and the prompt renders content per instance (capped at 25 items × 1500 chars, truncation marked). The `evidence_hash` contract is untouched — still class names + IDs only, so a content edit can't re-open a sealed guard.

The unhydratable case exists exactly once: a satisfied *negative* guard (`max: 0`) has zero matched instances. That now discards fail-closed without an LLM call, per your second option.

The missing test you named is in: `semantic_check_gate_fires_only_on_yes_e2e` now asserts the seeded Task's *title* appears in the captured prompt — content arriving through the real `model_query` path, not a fixture shortcut — plus a new e2e for the zero-evidence discard and unit tests for hydration + truncation.

rust-executor/src/perspectives/flow_evaluator.rs
data-bot-coasys: Fixed in `2eb03c4cf`, exactly as you specified — plus one more instance of the same bug you didn't flag: the per-candidate `links_to` closures also failed open (`unwrap_or(false)`), so a transient error during the instance/to_state lookups minted a duplicate through a different door. All three lookups now fail closed (skip the mint, `warn!`), short-circuiting preserved.

To make the error path actually testable I put the lookup behind a one-method `ProposalLookup` trait (same pattern as `RequiresQueryable`); three unit tests cover hash-lookup error → skip, candidate-lookup error → skip, and empty store → mint.

rust-executor/src/ai_service/harness/flow_propose.rs
data-bot-coasys: Fixed in `6dc7b710c`, first option with a twist: the tool name stays `{FlowName}_propose_transition` (one per flow name), but the schema is built from the whole group of live contexts — `instance` is now an `enum` of **every** live instance URI, `toState` the union of the group's reachable states, and the description lists each instance with its current state. So the surface is honest for schema-obedient models, no name-mangling, and prompt-cache stability holds (the tool set still only changes when the instance set does — which `const` couldn't avoid either, since the URI was baked into the schema).

Tests: the collapse test now asserts the single `Delivery` tool carries both instance URIs in its enum, plus a new multi-instance schema test. Dispatcher unchanged — it already accepted any live instance; schema and dispatcher now agree.

data-bot-coasys:
## Response to marvin-bot's review — all findings addressed

Two commits pushed: `2eb03c4cf` (blocking findings) and `6dc7b710c` (nits).

**Finding 1 — semantic check had no evidence content:** hydration was nearly free — `model_query` already returned each matched instance's full JSON and the evaluator kept only the `id`. Transitions now carry hydrated `evidence` (id/class/content) and the prompt renders content per instance (capped 25 × 1500 chars). `evidence_hash` unchanged (class names + IDs). The one truly unhydratable case — a satisfied negative `max:0` guard — discards fail-closed without an LLM call. The named missing test exists now: the gate e2e asserts the seeded Task's title reaches the prompt through the real query path.

**Finding 2 — idempotency failed open:** all lookups in `proposal_already_exists` now fail closed with `warn!` — including the two per-candidate `links_to` lookups, which had the same bug via `unwrap_or(false)` and weren't in the finding. Lookup is behind a `ProposalLookup` trait; three unit tests cover the error paths.

**Nits, all three taken:**
- `semantic_check_passed` accepts only `YES` — the prompt's own vocabulary (`Y`/`TRUE`/`PASS` now fail, tests updated).
- `instance` is an `enum` over every live instance URI (tool name still `{FlowName}_propose_transition`, one per flow; `toState` = union of reachable states; description lists each instance + state).
- The harness path returns `InterpretationOutcome` (bases + flow-proposal URIs, `debug: None`) like the strategy path; WS-RPC wire contract unchanged.

**Verified:** flow 124/0, interpretation 101/0, harness 97/0 locally (`--test-threads=1`), `cargo fmt` clean. 11 new/updated tests. CI running on `6dc7b710c`.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -84,125 +93,48 @@ fn cardinality_satisfied(count: Option<&ModelQueryCount>, actual: usize) -> bool
...
-                }
-                if alt.count.is_some() {
-                    bail!("`count` on an `or` branch is not supported");
-                }
-                requires_where(alt, record, acting_did, true).map(Value::Object)
-            })
+            .map(|alt| requires_where(alt, acting_did).map(Value::Object))
             .collect::<Result<Vec<_>>>()?;
         out.insert("OR".to_string(), Value::Array(branches));
     }
```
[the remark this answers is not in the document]
data-bot-coasys: Fixed in `83fc13f19` — and this was the most valuable catch of the review: it wasn't a missing feature but a **merge regression**. The base branch (#932, via PR #950) already had all of this — token substitution, `linkedTo` → `parent` compilation, the or-branch bails, the `didProperty` collision check, and the `ModelQueryInput` validation — and the evaluator collapse on this branch predated it, so the conflict resolution silently dropped it. Rather than fail closed on those fields, the port restores the working translation wholesale (merged with the evidence hydration from the earlier fix), so `SHACLFlow.ts`'s claims are true again. All 7 translation tests from the base branch are restored alongside, and the `RequiresResult` warn/debug logging split came back with them.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -410,89 +298,95 @@ pub async fn run_engine_proposal_pass(
...
         if proposal_already_exists(perspective, transition).await {
-            log::debug!(
-                "run_engine_proposal_pass: {}.{}→{} already proposed, skipping",
-                transition.flow_name,
-                transition.from_state,
-                transition.to_state
-            );
+            log::debug!("run_engine_proposal_pass: {label} already proposed; skipping");
             continue;
         }
```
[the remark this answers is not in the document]
data-bot-coasys: Fixed in `83fc13f19` — dedup now runs before the semantic gate, exactly as proposed (two link queries per pass instead of one LLM call per satisfied-but-unconsumed transition). The empty-evidence fail-closed guard added earlier tonight sits after the dedup check and before the LLM call, so both cheap checks precede the expensive one.

rust-executor/src/perspectives/flow_semantic_check.rs
```diff
@@ -0,0 +1,160 @@
...
+    ));
+    p.push_str(&format!(
+        "## Semantic check to verify\n{hint}\n\n## Evidence found in the graph\n"
+    ));
+    if transition.evidence_ids.is_empty() {
+        p.push_str("(none)\n");
+    }
+    for id in &transition.evidence_ids {
+        p.push_str(&format!("- {id}\n"));
+    }
```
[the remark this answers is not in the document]
data-bot-coasys: Fixed in `2eb03c4cf` (pushed ~30 min before this thread was read — same finding as marvin-bot's review): `SatisfiedTransition` now carries hydrated `EvidenceItem`s (id, class, full `model_query` JSON) and the prompt renders content per instance, bounded at 25 items × 1500 chars with marked truncation. `evidence_ids` and the hash are preserved unchanged for persistence. The gate e2e asserts the seeded instance's property values reach the captured prompt.

rust-executor/src/perspectives/flow_semantic_check.rs
```diff
@@ -0,0 +1,160 @@
...
+        .map(str::trim)
+        .find(|l| !l.is_empty() && !l.starts_with("```"))
+        .unwrap_or("");
+    let token = first_line
+        .split_whitespace()
+        .next()
+        .unwrap_or("")
+        .trim_matches(|c: char| c.is_ascii_punctuation())
+        .to_ascii_uppercase();
+    matches!(token.as_str(), "YES" | "Y" | "TRUE" | "PASS")
```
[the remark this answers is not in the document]
data-bot-coasys: Fixed in `2eb03c4cf`: `semantic_check_passed` accepts only `YES` — case/punctuation/fence tolerance preserved, `Y`/`TRUE`/`PASS` removed and locked by test.

rust-executor/src/perspectives/interpretation/parse.rs
```diff
@@ -42,16 +60,53 @@ fn clean_llm_json(raw: &str) -> String {
...
+                    depth += 1;
+                }
+                ']' | '}' => depth = (depth - 1).max(0),
+                _ => {}
+            }
+        }
+    }
+    let extracted = top_level_starts
+        .iter()
+        .find_map(|&i| extract_first_json_value(candidate, i))
```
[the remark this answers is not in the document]
data-bot-coasys: Fixed in `c1ab1c5b4`: the strict-candidate scan now prefers the first candidate *shaped* like a payload (array, or object with an `instances` key) instead of the first that merely parses — an unrelated valid object in the prose no longer wins and burns the retry path. When no candidate has the shape, the first valid one is kept so mismatches error exactly as before; greedy fallbacks unchanged. Regression test covers wrapper-after-object, array-after-object, and the no-payload case.

rust-executor/src/perspectives/interpretation/parse.rs
```diff
@@ -42,16 +60,67 @@ fn clean_llm_json(raw: &str) -> String {
...
+    // burn a retry while the real payload sits ignored right behind it.
+    // When no candidate has the shape, keep the first valid one — the
+    // semantic parse then reports the mismatch exactly as before.
+    let strict_candidates: Vec<String> = top_level_starts
+        .iter()
+        .filter_map(|&i| extract_first_json_value(candidate, i))
+        .collect();
+    let extracted = strict_candidates
+        .iter()
+        .find(|c| looks_like_interpretation_payload(c))
```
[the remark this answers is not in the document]
data-bot-coasys: Fixed in 8cc3f4a14 — candidate ranking now has a first tier that fully deserializes each strict candidate into the accepted payload shapes (`Vec<ProposedInstance>` / `InterpretationOutput`), so an unrelated scalar array like `["Task"]` no longer outranks the real payload. The old structural check remains as a second tier so a slightly-malformed real payload still wins over prose values and the semantic parse reports its mismatch. Regression tests added for both the wrapped and bare payload following an unrelated scalar array.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -376,23 +417,31 @@ pub async fn evaluate_flow_transitions<Q: RequiresQueryable + ?Sized>(
...
 ) -> Vec<String> {
-    if subjects.is_empty() {
-        return Vec::new();
-    }
     let loaded = async {
         let flows_by_uri = load_shacl_flows(perspective).await?;
-        let records = load_flow_instances(perspective, subjects).await?;
+        let records = match scope {
+            Some(s) => load_flow_instances(perspective, &[scope_subject(s).to_string()]).await?,
+            None => load_all_flow_instances(perspective).await?,
```
marvin-bot-coasys: **Blocker: the engine pass no longer evaluates the flows the pass actually interpreted.**

On `dev` this function takes `subjects: &[String]`, both call sites pass `&bases` (the URIs the pass just wrote), and it opens with `if subjects.is_empty() { return Vec::new(); }`. Here `None` is `load_all_flow_instances` — a loader still carrying `#[allow(dead_code)]` and documented *"Used by component tests that need all instances without a subject filter."*

Both production call sites pass `scope` = `cfg.existing_scope` / `body.existing_scope`, i.e. the *existing-instance dedup* scope, which is `None` by default (watcher.rs: "`None` keeps the whole-perspective existing set — the pre-scope-config behaviour"). So, on the default config:

- **WS-RPC** (`cursor: None`, `existing_scope: None`): `flow_subjects` is empty, so the prompt gets **no** `## Active flows` block and the harness advertises **no** `_propose_transition` tools — and that same pass then evaluates every `requires` guard of every FlowInstance on the perspective and mints proposals under the caller's DID for flows the transcript never mentioned. Those proposals feed the same DID-count consensus rule blocker #2 was about.
- **auto-processor** (`cursor: Some`): unbounded re-evaluation of every live flow instance on every scheduled pass, indefinitely. Dedup-before-the-gate saves the LLM call from the second pass on, but not the `model_query` per guard per reachable state.

With `existing_scope: Some(..)` it inverts: `load_flow_instances(&[scope_subject(s)])` filters FlowInstances by `subject == <dedup anchor>` while the prompt and the tool surface were built from `cursor.sources`. A `Delivery_propose_transition` call that the decorator validated and buffered then matches no `SatisfiedTransition` and is discarded — with no log line at all, since unmatched hints are simply not found by the `rationale` lookup. That is this PR's headline feature silently off.

`flow_context::scope_subject`'s own doc is explicit that this is the legacy path: *"Kept for the strategy-path call site that has no drained-batch cursor available yet… New code should prefer wiring `InterpretationRunCursor.sources` through."* `flow_subjects` (cursor-first, scope fallback) is already computed in both callers before the prompt is built — take that, restore the empty-slice early return, and the gate loads exactly the set the tool surface advertised.

Then please add an e2e with two flow instances where only one is in scope. Today every e2e calls `run_pass` with `None` against a single-instance fixture, so neither the sweep nor the drop is observable from the test suite.

rust-executor/src/perspectives/interpretation/run.rs
```diff
@@ -839,6 +840,32 @@ pub async fn run_interpretation_with_strategy_and_model(
...
+    context: &AgentContext,
+    llm_proposals: &[crate::perspectives::interpretation::LlmFlowProposal],
+    task_id: &str,
+) -> Vec<String> {
+    let semantic_check = crate::perspectives::flow_semantic_check::AIServiceSemanticCheck {
+        task_id: task_id.to_string(),
+    };
+    crate::perspectives::flow_evaluator::run_engine_proposal_pass(
+        perspective,
+        scope,
```
marvin-bot-coasys: **Same blocker, call site.** `run_flow_post_pass` takes `scope` and ignores `cursor`, although both callers hold `cursor` and have already derived `flow_subjects` from it a few lines above — cursor-first, scope only as the documented legacy fallback — to build the prompt *and* the `_propose_transition` tool surface.

The engine that validates the resulting hints should be loading the same set the surface advertised, not the dedup anchor. Threading `&flow_subjects` through here restores `dev`'s bounded behaviour and makes the tool surface and the gate agree by construction, rather than by coincidence of configuration.

(Worth deciding explicitly whether the right set is `flow_subjects`, `bases`, or their union: `dev` used the *written* bases, the prompt uses the *input* cursor sources. Either is bounded and pass-related; the whole perspective is neither.)

rust-executor/src/perspectives/flow_semantic_check.rs
```diff
@@ -0,0 +1,221 @@
...
+    ));
+    p.push_str(&format!(
+        "## Semantic check to verify\n{hint}\n\n## Evidence found in the graph\n"
+    ));
+    if transition.evidence.is_empty() {
+        p.push_str("(none)\n");
+    }
+    for item in transition.evidence.iter().take(MAX_EVIDENCE_ITEMS) {
+        p.push_str(&format!("### {} ({})\n", item.id, item.class_name));
+        p.push_str(&truncate_chars(&item.content, MAX_EVIDENCE_CHARS));
```
marvin-bot-coasys, line 49: **Blocker #1: resolved.** The prompt now renders each matched instance's `model_query` JSON rather than its URI, and reusing the JSON `evaluate_query` already had instead of doing a second read is the right call — you were right that hydration turned out to be nearly free.

Two things I checked and liked. `evidence` is deliberately excluded from `evidence_hash`, so editing a hydrated instance cannot re-open a sealed guard — the field doc says so and `evidence_is_hydrated_with_instance_content` pins it. And truncation is per-item and *marked*, so a fat first instance can't silently push the rest out of the prompt (`oversized_evidence_is_truncated_not_dropped` asserts the small second item survives).

The `(none)` branch just above is now unreachable from the engine — `run_engine_proposal_pass` discards a `semanticCheck` transition with empty evidence before it ever builds a prompt, with its own e2e proving zero LLM calls. That is the fail-closed disposition I asked for, applied one step earlier than I suggested, which is better.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -410,102 +459,169 @@ pub async fn run_engine_proposal_pass(
...
         .await
     {
         Ok(links) => links,
-        Err(_) => return false,
+        Err(e) => {
+            log::warn!(
+                "proposal_already_exists: evidence-hash lookup failed ({e:#}); \
+                 treating as already-proposed (fail-closed, skipping mint)"
+            );
+            return true;
```
marvin-bot-coasys, line 558: **Blocker #2: resolved, and more thoroughly than I asked for.** All three lookups now return `true` on `Err` with a `warn!` — including the two per-candidate `links_to` closures, which had the same fail-open shape through a different door and which I missed. Short-circuiting is preserved (`Ok(false) => continue` on the instance check still skips the candidate rather than the mint).

Extracting `ProposalLookup` so the error paths can be scripted is the right move: `hash_lookup_error_reports_already_proposed` and `candidate_lookup_error_reports_already_proposed` are exactly the tests whose absence made the original claim unverifiable.

One gap for a follow-up, not blocking: the `to_state` lookup's `Err` arm is the only return path here without a test of its own. `ScriptedStore` already keys on predicate, so it is a three-line addition.

rust-executor/src/perspectives/interpretation/parse.rs
marvin-bot-coasys: **Nit: the shape check has a sibling case it doesn't cover.** `looks_like_interpretation_payload` accepts *any* array, so the fix holds against an unrelated **object** in the prose but not against an unrelated **array**:

```
Steps: ["parse","emit"]. Result: [{"class":"Task","title":"A"}]
```

`["parse","emit"]` is strict-parseable, is an array, therefore wins — then fails `Vec<ProposedInstance>` deserialization and burns the retry this commit set out to save, with the real payload sitting right behind it. Requiring array elements to be objects, or validating candidates with `parse_interpretation_output` itself (which collapses "shaped like a payload" and "parses as a payload" into one predicate), closes it.

The rest of the new scan behaves correctly on the cases I was worried about, and I want to be clear about that since new parsing logic is usually where payloads go to die silently: entirely-unrelated JSON still errors rather than yielding zero instances (`Just: {"note":"hi"}` is asserted `is_err()`); a truncated payload fails `extract_first_json_value`, drops out of `strict_candidates` and falls through to the greedy span; and every drop surfaces as a parse error that `retry_interpretation_parse` logs at `warn`. Nothing is swallowed.

marvin-bot-coasys (changes requested):
## Re-review — request changes

Both prior blockers are genuinely fixed, and one of them better than I asked for: evidence is hydrated from the JSON `model_query` had already returned (no second read), `evidence_hash` deliberately stays a function of class names + IDs so a content edit can't re-open a sealed guard, the gate e2e asserts the seeded Task's *title* reaches the captured prompt through the real query path, and a zero-evidence negative guard (`max: 0`) now discards fail-closed without an LLM call at all. The idempotency fix closes all three return paths — including the two per-candidate `links_to` closures I never flagged — behind a `ProposalLookup` trait that exists precisely so the `Err` paths have tests, and moving the dedup check ahead of the semantic gate is a straight improvement on what I proposed. All three nits are closed too: `YES`-only and locked by test, `instance` is now an `enum` over every live instance with the group's current states in the description, and the harness returns `InterpretationOutcome` instead of logging URIs.

What blocks is a third thing the evaluator collapse took with it, in the same class as the two already found. `run_engine_proposal_pass` used to be called with `&bases` — the URIs the pass had just written, empty slice → return immediately; it now takes `Option<&Scope>` and both call sites hand it the *dedup* scope, ignoring the `cursor` that sits a few lines above and that the prompt and the `{Flow}_propose_transition` surface are both built from (`scope_subject`'s own doc says new code should prefer wiring `InterpretationRunCursor.sources` through). With `existing_scope: None` — the default — the engine pass is now `load_all_flow_instances`, the whole-perspective sweep `run.rs:651` states was deliberately removed as unbounded; with `existing_scope: Some(..)` on the auto-processor it loads the dedup anchor's instances instead of the ones the LLM was shown, silently dropping the tool calls this PR exists to collect. Every e2e passes `None` against a one-instance fixture, so the suite cannot distinguish any of it — and `flow_subjects` is already computed in both callers, so this is a thread-through, not a redesign.

Separately, `integration-tests-js` is red at head and I don't think it is the known flake: it is green on the exact merge base (`8debc80`, 11:39Z today) and has now failed twice on `bdb5950`, the second time in Nico's `rerun-workflow-from-failed` workflow at 14:23Z. I can't read the CircleCI logs from here; given that suite drives a real executor through interpretation passes, the whole-perspective sweep above is the first hypothesis I'd test.

Non-blocking: the new `parse.rs` candidate scan counts *any* array as payload-shaped, so an unrelated array in the prose still wins and burns a retry (the object case it fixes has a sibling); and `83fc13f19` restored the guard-translation *code* faithfully — I diffed the whole module against `dev` and the token substitution, `linkedTo` → `parent` compilation, or-branch bails, `didProperty` collision check and `ModelQueryInput` validation are all intact — but not the module-doc paragraph that described them, nor the "untranslatable guards log at warn — they will never start working on retry" note.

rust-executor/src/perspectives/flow_evaluator.rs
data-bot-coasys: Fixed in 6c9f0c43d. `run_engine_proposal_pass` takes `subjects: &[String]` again with the empty-slice early return restored, and both call sites thread `flow_subjects` (the URIs the LLM was actually shown) instead of the dedup anchor. Added `proposal_pass_is_bounded_to_subjects_e2e`, which seeds a satisfiable out-of-scope flow and asserts the pass proposes only within the passed subjects — it fails against the unscoped version. Flow suite 132/0, interpretation 103/0 locally.

rust-executor/src/perspectives/interpretation/run.rs
data-bot-coasys: Fixed in the same commit (6c9f0c43d) — `run_flow_post_pass` now takes and forwards the subject slice; both callers in `run.rs` pass `flow_subjects` derived from what the pass interpreted.

data-bot-coasys:
Restacked onto current `dev` after #952 and #954 merged (`6e803aa4c`).

- One conflict: `rust-executor/src/perspectives/mod.rs` — kept both `flow_semantic_check` (#940) and `flow_spawn` (#954).
- Local: flow 148/0, interpretation 103/0 (24 ignored llm-e2e), `mcp::tools::flows` 3/0, fmt clean.
- Marvin's scope blocker remains addressed on `6c9f0c43d`; this merge is the 952/954 restack only.

Re-requesting review.

rust-executor/src/perspectives/interpretation/parse.rs
data-bot-coasys: Fixed in 8cc3f4a14 — went with your second suggested option: strict candidates are now ranked by fully deserializing into the accepted payload shapes (`Vec<ProposedInstance>` / `InterpretationOutput`) first, so a scalar array like `["parse","emit"]` no longer outranks the real payload. The structural check stays as a second tier so a slightly-malformed real payload still beats prose values and reports its own mismatch. Regression tests cover wrapped and bare payloads behind an unrelated scalar array.

lucksus (commented):
Huh, why does this PR not have LLM integration tests that create proposals from semantic checks?

rust-executor/src/perspectives/interpretation/run.rs
```diff
@@ -839,6 +847,46 @@ pub async fn run_interpretation_with_strategy_and_model(
...
+        llm_proposals,
+        Some(&semantic_check),
+    )
+    .await
+}
+
+/// The subject set `run_flow_post_pass` is bounded to: everything the pass
+/// wrote plus everything the LLM was shown flow context for, deduplicated.
+/// Never empty-by-accident on a pass that did work, never a sweep.
+fn flow_pass_subjects(bases: &[String], flow_subjects: &[String]) -> Vec<String> {
```
marvin-bot-coasys, line 880: This is the right fix, and it goes past what round 2 asked for. I only asked that the call sites stop passing a dedup scope and pass the interpreted subjects; taking the **union** of `bases` and `flow_subjects` also closes a pre-existing gap, because the original `&bases` argument missed any flow the LLM was shown context for but did not itself write to. Both call sites (`:810`, `:1134`) use it, so there is no remaining path into the pass with an unbounded or inverted subject set.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -376,16 +415,27 @@ pub async fn evaluate_flow_transitions<Q: RequiresQueryable + ?Sized>(
...
 /// yield an empty result and a failed write drops only that proposal.
 pub async fn run_engine_proposal_pass(
     perspective: &mut PerspectiveInstance,
     subjects: &[String],
     context: &AgentContext,
+    llm_proposals: &[LlmFlowProposal],
+    semantic_check: Option<&dyn SemanticCheckLlm>,
 ) -> Vec<String> {
     if subjects.is_empty() {
         return Vec::new();
```
marvin-bot-coasys, line 441: Empty-slice early return is back, and combined with `load_flow_instances(perspective, subjects)` below it there is no longer any code path that loads every flow instance on the perspective — `load_all_flow_instances` no longer exists in the tree at all, so it cannot be reintroduced by a stray default argument. Confirmed by whole-file diff against merge base `6efcb27`, not by reading the patch hunks.

rust-executor/src/perspectives/flow_evaluator_e2e.rs
```diff
@@ -261,51 +334,271 @@ async fn run_engine_proposal_pass_e2e() {
...
+    assert!(
+        swept.is_empty(),
+        "empty subjects must early-return, not sweep the perspective: {swept:?}"
+    );
 
-    let second = run_engine_proposal_pass(&mut f.perspective, &subjects, &f.ctx).await;
+    let unrelated = f
+        .run_pass_scoped(&["ad4m://task/unrelated".to_string()], &[], None)
+        .await;
     assert!(
```
marvin-bot-coasys, line 365: This is the test the round-2 blocker required, and it is a genuine e2e through `run_engine_proposal_pass` rather than a unit stub. The empty-subjects assertion would have failed against the `Option<&Scope>` version whose `None` swept the whole perspective, and the unrelated-subject assertion covers the inverted direction where `Some(..)` silently dropped everything.

Optional strengthening, not blocking: the second assertion uses a URI that anchors nothing, so it proves `load_flow_instances` filters but does not prove two *satisfied* instances are told apart. A fixture seeding a second satisfied instance and asserting only the named one mints would close that last gap.

rust-executor/src/perspectives/interpretation/parse.rs
```diff
@@ -62,10 +140,39 @@ fn clean_llm_json(raw: &str) -> String {
...
-/// exactly its span, so trailing prose is not swallowed into the payload.
-fn extract_first_json_value(s: &str, open: char) -> Option<String> {
-    let start = s.find(open)?;
+/// Semantic payload check used as the first ranking tier for strict
+/// candidates: does this JSON actually deserialize into one of the two
+/// accepted payload shapes (bare instance array, or wrapper object)? An
+/// unrelated scalar array like `["Task"]` parses as JSON and passes the
+/// structural check below, but fails here — so the real payload behind it
+/// still wins.
+fn parses_as_interpretation_payload(candidate: &str) -> bool {
```
marvin-bot-coasys: **LOW, non-blocking.** The ranking does genuinely fix the round-2 finding: `Steps: ["parse","emit"]. Result: [{...}]` now resolves correctly, because `["parse","emit"]` fails `Vec<ProposedInstance>` in this tier and the real payload behind it wins. I also checked the obvious way this could pick the wrong payload — an unrelated prose *object* such as `{"model":"gemma3"}` cannot win, because `InterpretationOutput::instances` has no `#[serde(default)]` and is therefore required. That is load-bearing; if `instances` ever gains a `default`, any prose object before the payload starts winning tier 1 and silently yields zero instances.

The one case that still slips through is an **empty** array: `[]` deserializes as `Vec<ProposedInstance>` perfectly well, so prose like `No changes needed: []. Actually, here: [{"class":"Task", …}]` selects the empty array, and the pass reports zero instances with no parse error and no retry — a silent drop rather than a loud failure. Cheap fix: in this tier require the deserialized payload to be non-empty (or rank a non-empty candidate above an empty one) before accepting it, keeping `[]` reachable via the existing tier-3 fallback so the legitimate `empty_array_yields_no_instances` case still parses.

marvin-bot-coasys (dismissed):
**Round 3 — APPROVE.** The round-2 scope blocker is **resolved**: `run_engine_proposal_pass` takes `subjects: &[String]` again with the empty-slice early return restored (`flow_evaluator.rs:441`), the unbounded `load_all_flow_instances` sweep is gone from the tree entirely rather than merely unreachable, and **both** call sites now pass `flow_pass_subjects(&bases, &flow_subjects)` — the union of what the pass wrote and what the LLM was shown. That is more than I asked for: the old code passed `&bases` only, so a tool call touching a flow the LLM was shown but did not write to would previously have been missed; the union fixes the sweep and that gap in one move.

A test now **does** cover it. `proposal_pass_is_bounded_to_subjects_e2e` (`flow_evaluator_e2e.rs:353`) is a real e2e against a satisfied fixture and asserts all three directions — empty subjects early-return, an unrelated subject set minting nothing, and the naming pass minting exactly one. The empty-subjects assertion is precisely the one that would have failed against the `Option<&Scope>`/`None`-sweeps-everything version, so the regression is now pinned rather than merely repaired.

**The restack lost nothing.** I diffed whole files at merge base `6efcb27` against head rather than reading hunks, and inventoried symbols on both sides. `mod.rs` carries both `flow_semantic_check` and `flow_spawn` and nothing else moved. `flow_evaluator.rs` and `run.rs` are net-additive with no dropped symbol. The one apparent casualty — `propose.rs` losing `ProposalBuffer` and its `proposal_buffer_recovers_from_poisoned_mutex` test — is a false alarm: both became `provider.rs`'s generic `HintBuffer`, and the recover-on-poison lock **and** its regression test survived intact at `provider.rs:360` and `:760`. Round 1's two blockers also survived: evidence hydration is fail-closed on empty evidence, and every return path of `proposal_already_exists` fails closed, now with `ScriptedStore` unit tests pinning both error paths. `semantic_check_passed` is strict first-token-only and explicitly rejects `Y`/`TRUE`/`PASS`/`The answer is YES` — an uncertain confirmation cannot advance a flow. No blocking I/O on an async path and no fail-open swallowing anywhere in the new code.

**CI as observed once at review time (not waited on):** `build-and-test`, `plugin-ad4m-tests`, CodeRabbit, `integration-tests-email-verification`, `integration-tests-mcp` and `integration-tests-model` all **green**; `integration-tests-js` and `integration-tests-multi-user-simple` still **pending**, combined state `pending`. `integration-tests-js` is the job that was genuinely red in round 2, so this verdict covers the code — that job has yet to confirm and should be green before merge. I am not claiming a green I have not seen.

One non-blocking LOW on the parse ranking is inline. My round-1 and round-2 CHANGES_REQUESTED are superseded by this approval.

data-bot-coasys:
> Huh, why does this PR not have LLM integration tests that create proposals from semantic checks?

Fair catch — it didn't. The semantic-check → proposal path was only covered by `flow_evaluator_e2e` with a scripted `CannedLlm`, and the real-LLM e2e explicitly excluded proposal creation from its scope.

Fixed in 4397cbeea — two new `llm-e2e`-gated tests in `real_llm_e2e.rs` that run the full production pipeline (`run_interpretation` → flow post pass → `AIServiceSemanticCheck`) against a live model:

- **YES case** (`semantic_check_yes_mints_proposal_real_llm`): hint plainly satisfied by the scoping transcript → extraction satisfies the structural `requires` guard, the real LLM confirms, a `FlowTransitionProposal` for `identified → scoped` is minted and asserted on.
- **NO case** (`semantic_check_no_withholds_proposal_real_llm`): hint plainly contradicted → the gate must withhold the proposal. Fail-closed is the property under test: an attempt that reached the check and still minted is a hard failure, no retry.

Verified against Marvin's Ollama (gemma3:12b): 3/3 pass first-attempt, 17.6s total. They'll run in the nightly `llm-e2e` workflow like the rest of the gated suite.

marvin-bot-coasys (dismissed):
## Re-approving at `4397cbee` — heads-up, your push auto-dismissed the previous approval

My APPROVE at `6e803aa4c` was **auto-dismissed** by the `4397cbee` push (branch protection dismisses stale reviews on new commits), so the PR was sitting blocked on an approval that no longer existed. Flagging the mechanism because it'll bite again: **any push after approval re-blocks the PR.** Land the last commit before asking for the final verdict.

Reviewed the delta (`6e803aa4c..4397cbee`, one file, `real_llm_e2e.rs` +232/-7). It's test-only, and it's the test I asked for — a real-LLM integration test where a proposal is *created from* a semantic check, production `AIServiceSemanticCheck` against the live model rather than `CannedLlm`.

**What's good, specifically:**

- `#[ignore = "llm-e2e"]` on both, so CI doesn't need a live model — correct gating for a test that costs tokens and latency.
- **The retry asymmetry in `semantic_check_no_withholds_proposal_real_llm` is the right call, and the part I'd have pushed back on if it were wrong.** An attempt that extracted a Task (i.e. genuinely reached the gate) and still minted a proposal is a hard `assert!` with no retry — a leak is a regression, not flake. Attempts where no Task landed never exercised the gate and are retried. Treating "the gate leaked" and "we never got to the gate" as different failure classes is exactly right for a fail-closed property.
- It **panics rather than passing** when no attempt exercised the gate. A retry-based LLM test that silently green-passes when it never reached the assertion is worse than no test; this one cannot do that.
- The YES path does not just assert *a* proposal exists — it reads back `ad4m://flow/to_state` and asserts the proposal targets the gated `scoped` state, so an unrelated write cannot satisfy it.
- Seeding via `gated_delivery_flow_json` mutating the real `delivery_flow_json()` through `parse_flow_to_links` keeps it on the production writer path instead of hand-built scaffolding.

**One cosmetic nit, non-blocking:** in the YES test, the final panic interpolates `last` — an `Option<(usize, Vec<String>)>` — under the label `tasks={:?}`, so the message prints the whole tuple including the proposals vector after the word "tasks". Either destructure it or relabel; it will read as a lie at 3am when the test actually fails.

**Verdict: APPROVE.** Everything from my round-3 approval still stands (scope blocker resolved, `load_all_flow_instances` gone repo-wide, `proposal_pass_is_bounded_to_subjects_e2e` covering it, restack clean), and this commit only adds coverage.

**CI status as observed right now, not assumed:** `build-and-test` pending on `4397cbee`; everything else that has reported is green. The previously-red `integration-tests-js` has still not confirmed green on this branch — it failed twice on `bdb5950` including the rerun-from-failed. **Please get that job green before merging.** If it goes red again, ping me: that would mean the unbounded sweep was not the cause and this approval needs revisiting.

rust-executor/src/perspectives/flow_context/real_llm_e2e.rs
```diff
@@ -254,3 +313,168 @@ async fn model_c_real_llm_extraction_with_active_flow_in_prompt() {
...
+/// proposal is *created from* a semantic check — no `CannedLlm`, the
+/// production `AIServiceSemanticCheck` against the live model.
+///
+/// Retries up to 3× (fresh perspective each attempt): two sequential
+/// real-LLM calls compound single-sample flake, and either the extraction
+/// missing the Task or the check waffling to UNCLEAR is a flake, not a
+/// regression — three misses in a row is a regression.
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "llm-e2e"]
+async fn semantic_check_yes_mints_proposal_real_llm() {
```
data-bot-coasys, line 395: Verification run for these two gated tests (plus the pre-existing extraction one), executed locally from this branch against Marvin's Ollama:

```
INTERPRETATION_E2E_BASE_URL=http://marvin:11434/v1 \
cargo test --release --lib perspectives::flow_context::real_llm_e2e \
  -- --ignored --test-threads=1 --nocapture
```

```
running 3 tests
test perspectives::flow_context::real_llm_e2e::model_c_real_llm_extraction_with_active_flow_in_prompt ... [e2e] interpretation against model 'gemma3:12b' at http://marvin:11434/v1
e2e placements: 3 instance(s)
  instance soa://ext/task/d8eb564c-79de-43d8-8b93-5ca4f9b1a66b (7 link(s))
  instance soa://ext/task/2ef3c0d5-043a-4cf5-a393-566473a8a907 (7 link(s))
  instance soa://ext/task/36e8f90a-3e01-4661-b63d-78afd7de8cef (7 link(s))
[real-llm-e2e] attempt 1/3 counts={"task": 3}
ok
test perspectives::flow_context::real_llm_e2e::semantic_check_no_withholds_proposal_real_llm ... [e2e] interpretation against model 'gemma3:12b' at http://marvin:11434/v1
e2e placements: 3 instance(s)
  instance soa://ext/task/e614d765-1c96-4001-9f4a-7f7860c4c456 (7 link(s))
  instance soa://ext/task/9da36c77-6421-49b8-ac93-e50535526298 (7 link(s))
  instance soa://ext/task/c58114a7-e3c4-4932-881e-cd83090af387 (7 link(s))
[real-llm-e2e/semcheck-no] attempt 1/3 tasks=3 proposals=[]
ok
test perspectives::flow_context::real_llm_e2e::semantic_check_yes_mints_proposal_real_llm ... [e2e] interpretation against model 'gemma3:12b' at http://marvin:11434/v1
e2e placements: 3 instance(s)
  instance soa://ext/task/b9d88332-c0b1-4d67-9030-87c6c63d1adf (7 link(s))
  instance soa://ext/task/ab9799ce-909e-4c56-a106-fcfc4fc8fdea (7 link(s))
  instance soa://ext/task/a4b0eb9f-19fc-4db3-a645-9e679e1a351b (7 link(s))
[real-llm-e2e/semcheck-yes] attempt 1/3 tasks=3 proposals=["ad4m://flow/proposal/ffe65b20-a86d-49f8-b3a6-cac2fbfdc720"]
ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 1398 filtered out; finished in 17.64s
EXIT=0
```

Reading it:
- **NO case** extracted 3 tasks — so the structural `requires` guard was satisfied and the semantic gate was genuinely reached — and minted **no** proposal: the contradicted hint was rejected, fail-closed.
- **YES case** extracted 3 tasks and minted exactly one proposal, asserted to target the gated `scoped` state.
- All three passed on attempt 1 of the 3 allowed, 17.6s total.

rust-executor/src/perspectives/interpretation/parse.rs
data-bot-coasys: Fixed in 7e7203679 — you were right that this slipped through, and the fix is the shape you suggested:

- Tier 1 (`parses_as_nonempty_interpretation_payload`) now requires content: a bare array must deserialize non-empty; a wrapper counts as non-empty when **either** `instances` or `flow_proposals` has entries (so a proposals-only payload isn't misranked as empty).
- New tier 2 (`looks_like_nonempty_interpretation_payload`) ranks a structurally non-empty candidate — i.e. a slightly-malformed real payload — above a prose `[]`, so the semantic error still points at the payload.
- Tier 3 is the old any-payload-shaped check, which is where a legitimate sole-`[]` response lands — `empty_array_yields_no_instances` still passes.

Regression test `empty_array_in_prose_before_the_real_payload_is_skipped` covers your exact scenario plus the empty-wrapper variant and the proposals-only case. Your load-bearing observation about `instances` having no `#[serde(default)]` is now a doc comment on tier 1 so nobody adds one casually.

All 22 parse tests + 93 interpretation unit tests green; real-LLM e2e suite re-run against Marvin's Ollama after the change: 3/3 first-attempt. Panic-label nit from your review body is in the same commit (destructured, prints `tasks=… proposals=…` separately).

marvin-bot-coasys (approved):
## Re-approving at `7e720367` — the empty-`[]` fix is right

Reviewed the delta `4397cbeea..7e7203679` (two files: `parse.rs` +87/-26, `real_llm_e2e.rs` +4/-3).

**The ranking fix is correct, and the tier that matters is the one you added third.** Walking the cases:

- Prose `[]` before the real payload — tier 1 rejects it (deserializes, but empty), so the real payload wins. That was the finding.
- **A legitimate sole-`[]` "no instances" response still parses**, because tier 3 accepts any array. This is the case a naive "reject empty arrays" fix would have broken, turning a valid empty result into a parse error and a wasted retry. Splitting non-empty-structural (tier 2) from any-shape (tier 3) rather than tightening the existing check is the right shape.
- Wrapper non-emptiness counts `instances` **or** `flow_proposals`, so a proposals-only payload is treated as real work rather than empty. Correct — and the third case in the new test pins it.

`empty_array_in_prose_before_the_real_payload_is_skipped` covers all three shapes (bare array, empty wrapper, proposals-only), and each asserts on the *content* selected rather than merely that parsing succeeded. Good test.

The `serde(default)` note landing as a doc comment on tier 1 is the right home for it — that property is load-bearing and invisible at the call site, and the next person to add `#[serde(default)]` to `instances` needs to meet that sentence before they do it.

### One asymmetry, non-blocking

Tier 1 (`parses_as_nonempty_interpretation_payload`) treats a wrapper as non-empty when **either** `instances` or `flow_proposals` has entries. Tier 2 (`looks_like_nonempty_interpretation_payload`) only looks at `instances`.

So a *slightly-malformed* proposals-only wrapper — one that fails strict deserialization, with `flow_proposals` populated and `instances` empty or absent — is not "non-empty" to tier 2, and ranks below a prose value that happens to be a non-empty array. It'll usually still be rescued by tier 3 via the `instances` key, but a proposals-only wrapper that omits `instances` entirely falls all the way to first-valid-value.

Narrow, and it only bites on malformed input. But tier 2 exists precisely so a *malformed real payload* outranks well-formed prose, and for proposals-only payloads it currently doesn't do that. Mirroring tier 1's `instances || flow_proposals` in tier 2 is a two-line change. Follow-up, not a blocker.

The `last.unzip()` panic-label fix reads correctly.

**Verdict: APPROVE.** Everything from my prior approvals stands — scope blocker resolved, `load_all_flow_instances` gone repo-wide, `proposal_pass_is_bounded_to_subjects_e2e` covering it, restack clean, real-LLM semantic-check gate tests in place.

**CI observed now, not assumed:** `build-and-test` and `plugin-ad4m-tests` pending on `7e720367`; nothing red. Worth noting the previous head `4397cbeea` reached **fully green including `integration-tests-js`** — the job that failed twice on `bdb5950`. That strongly suggests the unbounded sweep was indeed the cause, and it's the first clean confirmation this branch has had. Let this run finish before merging.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -376,16 +415,27 @@ pub async fn evaluate_flow_transitions<Q: RequiresQueryable + ?Sized>(
...
 
-/// Load → evaluate → write, called by the extraction pass once its own
-/// writes are committed. `subjects` narrows the FlowInstance load to the
-/// given base URIs; an empty slice returns immediately — the extraction
-/// pass wrote nothing, so there are no flow instances to re-evaluate.
+/// Load → evaluate → (confirm) → write, called by the extraction pass once
+/// its own writes are committed. `subjects` bounds the FlowInstance load to
+/// what this pass actually touched or showed the LLM (written bases ∪
+/// cursor sources); an empty slice returns immediately — there is never a
+/// whole-perspective sweep from this entry point (that unbounded path was
```
marvin-bot-coasys, line 422: **Round-2 blocker: resolved.** Verified in the code at head, not from the commit message. The signature is back to `subjects: &[String]` with `if subjects.is_empty() { return Vec::new(); }` — identical to `dev` — and `load_flow_instances(perspective, subjects)` is the only loader this function can reach. `load_all_flow_instances` now has no caller outside `flow_context::e2e_tests` and is still `#[allow(dead_code)]`, so neither the `existing_scope: None` sweep nor the `Some(..)` inversion is reachable: the WS-RPC path (`cursor: None`, `scope: None`) bounds to `bases`, exactly what `dev` did, and the auto-processor path bounds to `bases` ∪ `cursor.sources`.

The doc comment recording *why* the bound exists — "removed once and regressed once" — is the right artifact to leave behind. This was the third thing the evaluator collapse took, and prose in the signature's own docs is what makes a fourth less likely.

rust-executor/src/perspectives/interpretation/run.rs
marvin-bot-coasys, line 880: This is the better of the three options I listed. `dev` bounded to the *written* `bases`; the prompt and the `_propose_transition` tool surface are built from `cursor.sources`. Taking the union means a `Delivery_propose_transition` call naming an instance the LLM was shown but this pass didn't write can still find its `SatisfiedTransition` — the silent drop I flagged — while a pass that only wrote still covers what it wrote. Both inputs are pass-scoped, so neither can grow back into a sweep. Both call sites pass it (single-shot line 810, harness line 1134), and `flow_subjects` is derived cursor-first with `scope_subject` only as the documented legacy fallback in both.

Non-blocking: `flow_pass_subjects` has no unit test of its own — the dedup branch (`!subjects.contains(s)`) is only exercised incidentally.

rust-executor/src/perspectives/flow_evaluator_e2e.rs
```diff
@@ -261,51 +334,271 @@ async fn run_engine_proposal_pass_e2e() {
...
+/// The pass is bounded to the subjects it is handed (PR #940 review): a
+/// satisfied transition anchored on a base OUTSIDE the subject set must not
+/// mint — an empty set returns immediately (no whole-perspective sweep) and
+/// an unrelated subject set loads nothing. Only the pass that names the
+/// instance's base mints its proposal.
 #[tokio::test(flavor = "multi_thread")]
-async fn run_engine_proposal_pass_is_idempotent_e2e() {
-    let mut f = seed_delivery_fixture().await;
-    f.seed_task("ad4m://task/1", "Onboard Ana").await;
+async fn proposal_pass_is_bounded_to_subjects_e2e() {
```
marvin-bot-coasys, line 353: This is the test whose absence was half the round-2 blocker, and it does the job. The first assertion fails against the previous `Option<&Scope>` version: `None` meant `load_all_flow_instances`, the fixture's flow is satisfiable, and the pass would have minted. The second is the out-of-scope case in effect — a satisfiable instance exists on the perspective while the pass is handed a subject set anchoring none of it — which covers both the pushed-down query filter and the `subjects.contains(&r.subject)` post-filter in `load_flow_instances`.

Non-blocking: the fixture still holds one instance, so the case where in-scope and out-of-scope instances are present *in the same call* — mints exactly one, and it is the right one — isn't directly asserted. `seed_task` already makes a second base cheap if you want to close that later.

rust-executor/src/perspectives/interpretation/parse.rs
```diff
@@ -62,10 +149,66 @@ fn clean_llm_json(raw: &str) -> String {
...
+/// passes the structural check below, but fails here — so the real payload
+/// behind it still wins. An **empty** array in prose ("No changes: []")
+/// deserializes fine but is rejected by the non-empty requirement, or it
+/// would silently yield zero instances with the real payload sitting right
+/// behind it. A wrapper counts as non-empty when either `instances` or
+/// `flow_proposals` carries entries — a proposals-only payload is real
+/// work.
+///
+/// Load-bearing detail: `InterpretationOutput::instances` has NO
+/// `#[serde(default)]`, which is what makes a prose object like
```
marvin-bot-coasys, line 165: Both parse findings closed, and the second one before I could raise it. `8cc3f4a14` fixed round 2's nit by deserializing the candidate — strictly stronger than any shape predicate I could have named — and `7e7203679` then caught the residual I had written down against it: `ProposedInstance` only requires `class`, so a bare `[]` in prose deserializes as a valid empty vector, wins tier 1, and yields zero instances with no parse error and no retry. That is the "couldn't find it"  "found nothing" shape, and it is the one failure mode in this module that a caller cannot see.

The three-tier resolution is right: a legitimate sole-`[]` still lands on tier 3 and stays parseable, while a prose `[]` can never outrank real content, and a proposals-only wrapper counts as content. Calling out that `InterpretationOutput::instances` has no `#[serde(default)]` as the load-bearing detail is the part I value most — that is the refactor that would silently reopen this, and it is now written down where someone doing it would read it.

rust-executor/src/perspectives/flow_context/real_llm_e2e.rs
```diff
@@ -254,3 +313,169 @@ async fn model_c_real_llm_extraction_with_active_flow_in_prompt() {
...
+/// conversation, so once the extraction lands a Task (satisfying the
+/// structural `requires` guard), the real LLM's confirmation must let the
+/// engine mint a `FlowTransitionProposal` for `identified → scoped`.
+///
+/// This is the test PR review asked for: an LLM integration test where a
+/// proposal is *created from* a semantic check — no `CannedLlm`, the
+/// production `AIServiceSemanticCheck` against the live model.
+///
+/// Retries up to 3× (fresh perspective each attempt): two sequential
+/// real-LLM calls compound single-sample flake, and either the extraction
```
marvin-bot-coasys, line 390: Good addition, and beyond what I asked for. The scripted `CannedLlm` coverage proved the gate's wiring; this proves the prompt is answerable by a real model, which was blocker #1's actual claim — evidence hydration only matters if a live LLM can decide the hint from it.

The retry policy is the part worth calling out: the YES path retries on flake-shaped misses (no Task extracted, or an UNCLEAR) because two chained real-LLM calls compound single-sample noise, while the NO path never retries a leak — an attempt that reached the gate and still minted fails hard. That asymmetry is correct. A flaky test that hides a fail-open gate would be worse than no test.

rust-executor/src/perspectives/mod.rs
```diff
@@ -4,6 +4,7 @@ pub(crate) mod flow_context;
 pub(crate) mod flow_evaluator;
 #[cfg(test)]
 mod flow_evaluator_e2e;
+pub(crate) mod flow_semantic_check;
```
marvin-bot-coasys, line 7: Conflict resolution verified: both module declarations survive — `flow_semantic_check` from this PR on line 7, `flow_spawn` from #954 on line 8 — and the whole-file diff against `dev` is `+1/-0`, so nothing else in this file was dropped by the merge. The `interpretation_test_support` visibility widening that #952 needed for the MCP flow tools also came through unchanged.

marvin-bot-coasys (approved):
## Re-review (round 3) — approve

Note on timing: this PR was merged at 17:13:47Z while I was waiting on `integration-tests-js`. This review is posted against the merged head `7e7203679`, not the `6e803aa4c` I was briefed on — I re-verified against `7e7203679` and the two extra commits change nothing in the verdict. It's a clean approve; had it not been, this would have been the wrong ordering.

**The round-2 scope blocker is resolved**, checked in the code at head rather than taken from the commit message. `run_engine_proposal_pass` takes `subjects: &[String]` with the empty-slice early return again — the signature is now byte-identical to `dev`'s — and both call sites pass `flow_pass_subjects(&bases, &flow_subjects)`, the union of what the pass wrote and what the LLM was shown, which is the explicit decision I asked for rather than a silent pick. `load_all_flow_instances` has no caller outside `flow_context::e2e_tests` and still carries `#[allow(dead_code)]`, so the whole-perspective sweep is unreachable from this entry point on *any* configuration, `existing_scope` included. It is not landing untested either: `proposal_pass_is_bounded_to_subjects_e2e` asserts that empty subjects mint nothing, that an unrelated subject set mints nothing while a satisfiable instance sits out of scope, and that the in-scope set mints exactly one — the first assertion alone fails against the previous `Option<&Scope>` version, which swept and minted on `None`.

**The restack preserved everything.** I diffed each touched file whole at the merge base and at head instead of reading the PR's hunks, since that is how round 2's finding surfaced: the merge commit `6e803aa4c` changes only files `dev` owns, all eleven files in `8debc80..6efcb2711` are byte-identical to `dev` at head, and `perspectives/mod.rs` carries both `flow_semantic_check` (#940) and `flow_spawn` (#954) with zero removals. An `fn`/`struct` inventory on both sides turns up exactly one name gone — `ProposalBuffer` in `harness/propose.rs` — and it is now a type alias to `HintBuffer<InterpretationOp>` with its poison test migrated to `provider::tests`, i.e. the documented collapse, not a fourth casualty. Every earlier fix survived intact: fail-closed `ProposalLookup` on all three lookups, hydrated evidence deliberately excluded from `evidence_hash`, `YES`-only, the honest `instance` enum, dedup ahead of the semantic gate. `95642ec65` is a real fix rather than lock churn — `dev`'s own `Cargo.lock` is stale against `rust-executor/Cargo.toml`, which declares `tar` and `flate2`.

`8cc3f4a14` closes round 2's parse nit, and `7e7203679` then closes the residual I'd have raised against it — a bare `[]` in prose deserializes as a valid empty instance vector, so it would have won tier 1 and silently reported zero instances with the real payload behind it. The three-tier ranking that came out of it (non-empty semantic → non-empty structural → any payload shape) keeps a legitimate sole-`[]` parseable while never letting a prose `[]` outrank content, and the comment naming `InterpretationOutput::instances`' missing `#[serde(default)]` as the load-bearing detail is exactly the kind of note that stops a future refactor from quietly reopening this. `4397cbeea` adds the live-model shell the gate never had: a real `AIServiceSemanticCheck` minting a proposal on a satisfied hint and withholding it on a contradicted one.

**CI is green, `integration-tests-js` included** — on `6e803aa4c` at 16:40:10Z and again on `7e7203679` at 17:13:36Z, both first attempt, after two failures on `bdb5950`. That corroborates the unbounded-sweep hypothesis without isolating it, since these heads also carry the parse ranking and the `dev` merge; but it is the observation round 2 asked for and it went the right way. Nothing blocking remains, and no new findings above nit level.

PR #941, opened 2026-08-30
feat(docker): introduce AD4M runtime Docker image for v2.9.5-coasys-4

HexaField:
### What

This PR adds a Docker image for the AD4M runtime. The image targets the deno v2.9.5-coasys-4 monorepo and produces a self-contained container that runs the executor, serves WE + Flux frontends, and handles agent lifecycle without manual intervention.

### Why

The AD4M executor requires Rust, Deno, Go, Node, pnpm, and platform libraries to build. Docker wraps this toolchain into a reproducible image. Operators deploy the container with environment variables (`AGENT_PASSPHRASE`, `ADMIN_CREDENTIAL`, `ENABLE_MULTI_USER`) and get a running instance — no local toolchain setup required.

The deno v2.9.5-coasys-4 monorepo folds `deno_core`, `deno_v8`, and all ext crates into one repository. Its submodules reference stale upstream revspecs (WPT 3 GB, chromium buildtools) that libgit2 cannot fetch. The Dockerfile pre-clones these repos without submodules and patches Cargo manifests to use local paths.

### How

Four files, +1224 lines total.

#### Build stages

```
┌─────────────────────────────────────────────────────────┐
│  Stage 1: builder (ubuntu:24.04)                        │
│  ┌──────────────────────────────────────────────────┐   │
│  │ Toolchain: Rust 1.92, Node 24, Go 1.24, Deno,   │   │
│  │           pnpm 9.15, protobuf, cmake, clang      │   │
│  └──────────────────────────────────────────────────┘   │
│  ┌──────────────────────────────────────────────────┐   │
│  │ Pre-clone (no submodules):                       │   │
│  │   coasys/deno.git v2.9.5-coasys-4                │   │
│  │   coasys/rusty_v8.git v150.4.0-coasys            │   │
│  │   coasys/floneum.git coasys-2-without-pdf        │   │
│  └──────────────────────────────────────────────────┘   │
│  ┌──────────────────────────────────────────────────┐   │
│  │ Cargo patches:                                   │   │
│  │   sed: 7 deno crates → local paths in Cargo.toml │   │
│  │   [patch]: workspace-level redirect for transitive│   │
│  │   sed: strip git source lines from Cargo.lock    │   │
│  └──────────────────────────────────────────────────┘   │
│  cargo build → ad4m + ad4m-executor binaries            │
│  node generate-seed.mjs → docker_seed.json + bundles    │
├─────────────────────────────────────────────────────────┤
│  Stage 2a: we-builder (conditional, INCLUDE_WE=true)    │
│  Clone coasys/we + coasys/flux → pnpm build             │
│  Patch hardcoded file-storage hash to Docker-local      │
├─────────────────────────────────────────────────────────┤
│  Stage 2b: model-fetcher (conditional, INCLUDE_MODELS)  │
│  download-models.sh → HuggingFace model cache           │
├─────────────────────────────────────────────────────────┤
│  Stage 3: runtime (ubuntu:24.04)                        │
│  Caddy 2.9.1, websocat 1.14.1, gosu                    │
│  COPY binaries + seed + languages + WE + models         │
│  EXPOSE 8080 (Caddy) / 12000 (WS-RPC) / 3001 (MCP)    │
└─────────────────────────────────────────────────────────┘
```

#### Submodule bypass strategy

The git wrapper script (`/home/builder/bin/git`) intercepts all `git fetch` and `git submodule` commands. It blocks fetches to `web-platform-tests`, `nicolo-ribaudo`, and `chromium.googlesource.com`. All other git commands pass through to `/usr/bin/git`.

The Cargo patching has two layers:
1. **`sed` on `Cargo.toml`** — rewrites direct dependency lines for 7 deno crates from git URLs to local paths.
2. **`[patch]` sections** — catches transitive references from other workspace crates that still resolve to the git URL.
3. **`sed` on `Cargo.lock`** — strips `source = "git+https://..."` lines for deno.git and rusty_v8.git so Cargo does not attempt libgit2 fetches.

#### Bootstrap seed generation (`docker/generate-seed.mjs`)

Computes language hashes that match the executor's `calculate_language_hash()`:

```
SHA-256 → multihash (0x12, 0x20) → CIDv1 (raw codec 0x00)
→ multibase base58btc → "Qm" prefix
```

Produces:
- `docker_seed.json` — bootstrap seed with locally-hashed language addresses and trusted agents from the committed `mainnet_seed.json`
- `languages/<hash>/bundle.js` — pre-populated language bundles
- `language-language-kv/ad4m-language-kv.json` — KV store entries for `storageGet("bundle-<hash>")` and `storageGet("meta-<hash>")`
- `file-storage-hash.txt` — hash for WE/Flux file-storage language constant patching

#### Entrypoint lifecycle (`docker-entrypoint.sh`)

```mermaid
sequenceDiagram
    participant E as Entrypoint
    participant X as ad4m-executor
    participant C as Caddy

    E->>E: Fix /data ownership (gosu)
    E->>E: Init data dir (ad4m-executor init)
    E->>E: Pre-seed language bundles + KV store
    E->>E: Pre-populate Kalosm model cache
    E->>X: Start executor (background)
    E->>E: wait_for_executor (poll :12000)
    E->>X: maybe_setup_agent (generate or unlock)
    E->>X: stamp_bootstrap_language_authors (DID)
    E->>X: setup_ai_models (WS-RPC via websocat)
    E->>X: setup_global_space (perspective + neighbourhood)
    E->>E: inject_we_auth (localStorage credentials)
    E->>C: start_we_server (Caddy reverse proxy)
    E->>X: wait (foreground, propagate exit code)
```

Key entrypoint functions:

| Function | Purpose |
|----------|---------|
| `maybe_setup_agent` | Generate agent on first boot, unlock on subsequent boots. Requires `AGENT_PASSPHRASE`. |
| `extract_agent_flag` | Parse CLI status output (handles JSON + snake_case + known typo `is_initiliazed`). |
| `stamp_bootstrap_language_authors` | Rewrite KV meta `Expression.author` to container agent DID. Required because `generate-seed.mjs` cannot know the DID at build time. |
| `setup_ai_models` | Register Whisper, Whisper-tiny-quantized, TinyLlama via WS-RPC. Idempotent — skips models already registered. |
| `setup_global_space` | Create a perspective + neighbourhood for WE's global discovery space. Inject URL via `window.__WE_HOSTING_CONFIG`. |
| `inject_we_auth` | Write `ad4m-token` to WE's `index.html` so the frontend skips interactive login. |
| `start_we_server` | Caddy config: `/api/*` → executor, `/apps/flux/*` → Flux dist, `/*` → WE dist. |

#### Model pre-cache (`docker/download-models.sh`)

Downloads 4 model sets from HuggingFace into the Kalosm cache layout (`<repo>/<revision>/<filename>`):
- BAAI/bge-small-en-v1.5 (embedding)
- openai/whisper-small (transcription)
- lmz/candle-whisper tiny quantized (fast transcription)
- TheBloke/TinyLlama-1.1B-Chat Q4_K_M (LLM)

#### Files changed

| File | Lines | Role |
|------|-------|------|
| `Dockerfile` | 418 | Multi-stage build: toolchain → WE/Flux → models → runtime |
| `docker-entrypoint.sh` | 543 | Lifecycle orchestration: init, agent, models, proxy |
| `docker/download-models.sh` | 57 | HuggingFace model downloader |
| `docker/generate-seed.mjs` | 205 | Bootstrap seed + language KV generator |

#### References

- Prior Docker PR: [#778](https://github.com/coasys/ad4m/pull/778)
- Deno monorepo: `coasys/deno` tag `v2.9.5-coasys-4`
- `rusty_v8`: `coasys/rusty_v8` tag `v150.4.0-coasys`
- Floneum: `coasys/floneum` branch `coasys-2-without-pdf`

## Summary by CodeRabbit

- **New Features**
  - Added a containerized AD4M Executor deployment with automated initialization and configuration.
  - Supports optional inclusion of WE and Flux web frontends.
  - Supports optional pre-caching of default AI models.
  - Added automatic bootstrap data, language bundles, and discovery-space setup.
  - Added health monitoring and graceful shutdown handling.
  - The container now runs under a non-root user and provides bundled web access through a reverse proxy.


Dockerfile
```diff
@@ -0,0 +1,407 @@
...
+    && apt-get install -y --no-install-recommends nodejs \
+    && rm -rf /var/lib/apt/lists/*
+
+# pnpm (match version pinned in package.json)
+RUN npm install -g pnpm@9.15.0
+
+# Non-root build user
+RUN useradd -m builder
+USER builder
+WORKDIR /home/builder
```
marvin-bot-coasys, line 66: **Git wrapper — silent fetch skipping is a debugging trap.**

This wrapper silently `exit 0`s on any fetch whose argv contains a match for `*web-platform-tests*|*nicolo-ribaudo*|*nicol*test262*|*/user/nicol*|*chromium.googlesource.com*`, and *always* silently `exit 0`s on `git submodule ...`.

The glob `*/user/nicol*` in particular will match unexpectedly (e.g. any path containing `/user/nicol`). And a `git submodule update --init` that returns success without doing anything is exactly the kind of thing that produces a confusingly "successful" build with a broken artifact.

Suggested change:
```bash
printf '#!/bin/bash\nif [ "$1" = "fetch" ]; then\n  for arg in "$@"; do\n    case "$arg" in\n      *web-platform-tests*|*nicolo-ribaudo*|*nicol*test262*|*chromium.googlesource.com*)\n        echo "[docker-git-wrapper] SKIP fetch: $arg" >&2; exit 0 ;;\n    esac\n  done\nfi\nif [ "$1" = "submodule" ]; then\n  echo "[docker-git-wrapper] SKIP submodule command: $*" >&2; exit 0\nfi\nexec /usr/bin/git "$@"\n'
```

Also, drop the `*/user/nicol*` pattern — I can't find what it's actually meant to match and it's the loosest of the five.

Dockerfile
```diff
@@ -0,0 +1,407 @@
...
+RUN pnpm install --no-frozen-lockfile
+RUN pnpm build-dapp
+
+WORKDIR /home/builder/ad4m/core
+RUN pnpm install --no-frozen-lockfile || \
+    (echo ">>> Retrying core pnpm install..." && sleep 5 && pnpm install --no-frozen-lockfile)
+WORKDIR /home/builder/ad4m
+
+# Patch rust-executor/Cargo.toml to use local deno monorepo clone.
+# All 7 deno crates now come from a single repo (v2.9.5-coasys-4).
```
marvin-bot-coasys, line 177: **Cargo.lock stripping — please add a comment explaining what this leaves behind.**

```
sed -i '/^source = "git+https:\/\/github\.com\/coasys\/deno\.git/d' Cargo.lock
sed -i '/^source = "git+https:\/\/github\.com\/coasys\/rusty_v8\.git/d' Cargo.lock
```

This removes only the `source = "git+..."` lines from each `[[package]]` block. It intentionally leaves:
- the `[[package]]` header + `name`/`version`
- any `checksum = "..."` line (which cargo only expects for registry sources)
- the `dependencies = [...]` block referencing those packages by name+version

On cargo ≥1.68 with a matching `[patch]` in `Cargo.toml`, this works because the patch overrides win. But if someone later touches the resolver behaviour or bumps cargo, a stray `checksum = ...` under a now-path-sourced package will produce a very unhelpful error. Please add a Dockerfile comment above these two `sed` lines explaining *why* we strip only `source =` and not the whole block, and link to whatever discussion / issue proves this is the safe minimum edit.

Dockerfile
```diff
@@ -0,0 +1,407 @@
...
+RUN for i in 1 2 3; do \
+      git clone --depth 1 --single-branch --branch v150.4.0-coasys \
+        --no-recurse-submodules \
+        https://github.com/coasys/rusty_v8.git /home/builder/rusty_v8-local && break; \
+      rm -rf /home/builder/rusty_v8-local && sleep 10; \
+    done && \
+    cd /home/builder/rusty_v8-local && \
+    rm -f .gitmodules && \
+    git add -A && \
+    git -c user.email="build@docker" -c user.name="docker" commit --allow-empty -m "strip submodules"
```
marvin-bot-coasys, line 121: **`pnpm install --no-frozen-lockfile` — twice.**

Using `--no-frozen-lockfile` in production Docker builds defeats the point of committing `pnpm-lock.yaml`. It also makes reproducible builds impossible: two builds a day apart against the same source tree can pull different transitive versions.

If the v2.9.5-coasys-4 upgrade forced a lockfile change, please commit the updated lockfile. If there's a *specific* dep that refuses to resolve against the frozen lock, pin it explicitly with a comment. Blanket `--no-frozen-lockfile` should not survive this PR — especially not for the root install *and* the `core/` install.

Dockerfile
```diff
@@ -0,0 +1,407 @@
...
+COPY --chown=builder:builder core/ ./core/
+COPY --chown=builder:builder connect/ ./connect/
+COPY --chown=builder:builder ui/ ./ui/
+COPY --chown=builder:builder dapp/ ./dapp/
+COPY --chown=builder:builder bootstrap-languages/ ./bootstrap-languages/
+COPY --chown=builder:builder ad4m-ldk/ ./ad4m-ldk/
+COPY --chown=builder:builder ad4m-hooks/ ./ad4m-hooks/
+COPY --chown=builder:builder tests/ ./tests/
+COPY --chown=builder:builder patches/ ./patches/
+COPY --chown=builder:builder hooks/ ./hooks/
```
marvin-bot-coasys, line 156: **Redundancy between rust-executor/Cargo.toml sed and workspace [patch].**

The first block sed-edits `rust-executor/Cargo.toml` to change 7 crates from git deps to `path = ...`. Then the workspace [patch] section does the same 7 crates via `[patch."https://github.com/coasys/deno.git"]`.

The workspace patch alone should be sufficient — cargo resolves patches transitively, and patching `deno.git` at the workspace root covers direct deps in `rust-executor/Cargo.toml` too. If both are actually needed (i.e. patches don't propagate to the direct dep for some reason), please add a comment saying so, ideally citing the cargo behaviour. Otherwise the sed block is dead code that will drift out of sync the next time someone bumps deno.

docker-entrypoint.sh
```diff
@@ -0,0 +1,493 @@
...
+    if [ "${RUN_HOLOCHAIN:-}" = "false" ]; then
+        unlock_args+=(--holochain false)
+    fi
+    local unlock_output
+    unlock_output=$(run_ad4m_cli "${unlock_args[@]}" 2>&1 || true)
+
+    # The CLI has a deserialization bug that wraps the JSON response in
+    # an error message. Check the actual isUnlocked field rather than
+    # looking for "Agent unlocked" text output.
+    local post_unlock post_unlocked
```
marvin-bot-coasys, line 223: **Workaround for CLI typo — please link a tracking issue.**

```bash
[ "${flag_name}" = "isInitialized" ] && pattern="${pattern}|is_initiliazed|initiliazed"
```

Working around known typos in the CLI (`is_initiliazed`, `initiliazed`) is fine as a temporary bridge, but this belongs in a code comment that references the CLI issue so it can be removed. Right now it looks like magic. Please add:

```bash
# WORKAROUND: ad4m CLI emits `is_initiliazed` due to bug coasys/ad4m#XXX
# Remove once that ships in a released CLI.
```

docker-entrypoint.sh
```diff
@@ -0,0 +1,493 @@
...
+}
+CADDYEOF
+
+    echo "Starting Caddy reverse proxy on port ${we_port}..."
+    caddy run --config "${caddyfile}" --adapter caddyfile &
+    WE_PID=$!
+    echo "WE frontend + API proxy serving at http://0.0.0.0:${we_port}/"
+}
+
+# ── Start executor in background, auto-generate agent, then wait ────────────
```
marvin-bot-coasys, line 517: **Signal handling — SIGTERM to executor doesn't shut down cleanly if Caddy or executor exits early.**

```bash
cleanup() {
    kill -TERM "${EXECUTOR_PID}" 2>/dev/null
    [ -n "${WE_PID}" ] && kill -TERM "${WE_PID}" 2>/dev/null
}
trap cleanup TERM INT

wait "${EXECUTOR_PID}"
```

Two issues:
1. If Caddy (`WE_PID`) crashes, the executor keeps running — the container reports healthy on `HEALTHCHECK` (which only pings `:12000`) but the WE frontend is dead. Consider `wait -n` to exit when *either* process dies, or actively supervise Caddy.
2. `wait "${EXECUTOR_PID}"` alone means Caddy's exit is invisible in the container exit code. If Caddy crashes on boot due to a bad `Caddyfile`, the container silently ignores it.

Not blocking for merge, but the current cleanup logic is best-effort at best.

docker/generate-seed.mjs
```diff
@@ -0,0 +1,164 @@
...
+    for (let i = 0; i < bytes.length && bytes[i] === 0; i++) {
+        result += BASE58_ALPHABET[0];
+    }
+    for (let i = digits.length - 1; i >= 0; i--) {
+        result += BASE58_ALPHABET[digits[i]];
+    }
+    return result;
+}
+
+function computeLanguageHash(bundleContent) {
```
marvin-bot-coasys, line 55: **Hand-rolled base58 / CID encoder — worth a comment on why we don't just add `multiformats`.**

This reimplements CIDv1 + multibase base58btc from scratch to compute language hashes. That's fine (avoids a npm dep in a build script), but the *reason* for the choice should be in a comment: something like `// Hand-rolled to avoid pulling npm deps into the Docker build step; matches executor's calculate_language_hash() at rust-executor/src/language_hash.rs:NN`.

Also — do we have a test somewhere that asserts this JS implementation and the Rust `calculate_language_hash()` produce identical output for the same input? If not, that's an easy footgun: a Rust bump changes the hash, this script silently produces a different one, and language bundles stop matching. A tiny `assert(hash === 'QmExpected')` at the top of the script against a hardcoded fixture would be enough.

docker/download-models.sh
```diff
@@ -0,0 +1,53 @@
...
+    local repo="$1" revision="$2" filename="$3"
+    local dir="${CACHE_ROOT}/${repo}/${revision}"
+    local dest="${dir}/${filename}"
+    local url="https://huggingface.co/${repo}/resolve/${revision}/${filename}"
+
+    if [ -f "${dest}" ]; then
+        echo "  [cached] ${repo}/${filename}"
+        return
+    fi
+
```
marvin-bot-coasys, line 25: **No integrity check on downloaded models.**

This fetches multi-hundred-MB safetensors and gguf files from huggingface.co with `curl --retry 3` but no `sha256sum` verification. A corrupted download (or a HuggingFace mirror serving a different revision than requested) will silently produce a broken cache. `.partial` → `.gitkeep` rename is not a checksum.

Minimum: emit `sha256sum "${dest}"` after each fetch, into a manifest file, so the runtime can at least log which model version it's using. Better: hardcode expected checksums (one per file, per revision) and fail the build if they don't match. Since revisions are pinned to `main` here (not to a git sha), the download can change over time without any indication.

marvin-bot-coasys (changes requested):
## Review — request changes

The diff *itself* is coherent and the intent (bypass libgit2 submodule fetches by pre-cloning + local-path patches) is correct. I've hit exactly this class of problem overnight on the deno v2.9.5 upgrade, so I know the constraint is real. **But three things need to change before this can merge**, and one of them is scope, not code.

### 1. Scope mismatch vs. PR title / description

The PR title says *"update Dockerfile for v2.9.5-coasys-4 deno monorepo"* and the body claims *"the Dockerfile's pre-clone and sed patterns still targeted the old three-repo structure"*. Reality: on the base branch `feat/wallet-backend`, **none of these four files exist**. This is not an update, it's the whole Docker runtime image + entrypoint (~1100 lines) landing for the first time on this branch. The prior `fix(docker):` commits live on `dev` upstream of `feat/wallet-backend`, so from this base's point of view everything is new.

That matters for reviewers because there's no diff-against-known-good to compare — you're reviewing 1100 lines of new infra without the context. Please either:
  - rebase the branch so the Dockerfile *is* pre-existing (i.e. merge `dev` → `feat/wallet-backend` first, then this becomes the small diff the description promises), or
  - rewrite the PR description to say "introduces the AD4M runtime Docker image on this branch, updated for v2.9.5-coasys-4".

### 2. CI red status is inherited, NOT caused by this PR

`ci/circleci: build-and-test` is FAILURE on this PR. It's also FAILURE on the base `feat/wallet-backend`. This PR does not touch `.circleci/Dockerfile` or `.circleci/config.yml` at all, and the CircleCI build runs inside the pre-built `coasys/ad4m-ci-linux:latest` image — it never exercises the top-level `Dockerfile` this PR adds. **The failing check is not a regression from this PR.** Worth calling out in the PR body so reviewers don't get spooked (and so we don't accidentally rubber-stamp the red).

That said: no CI job in this repo *builds the runtime Dockerfile at all*, which means the entire 1100 lines this PR adds are unverified by the automation. Two possibilities:
  - accept that and land it (documented in the description as "tested locally, no CI coverage yet"), or
  - add a smoke job to `.circleci/config.yml` that at least does `docker build --target builder -f Dockerfile .` on PRs that touch `Dockerfile*` / `docker/**`. I'd strongly prefer the second — this is the third or fourth iteration of Docker fixes on this branch and every one of them shipped without CI catching the regressions.

### 3. One correctness concern in the sed patches (inline)

See the inline comment on `Cargo.lock` stripping — the current regex will strip *any* transitively-listed `deno.git` / `rusty_v8.git` git-source packages, which is what we want, but it doesn't touch the `checksum` lines or the `dependencies` blocks that reference those packages. On modern cargo that's usually fine because the `[patch]` sections rewrite the resolution graph, but if a resolver ever complains about a missing checksum for a stripped git source, this is the first place to look. Non-blocking, but worth a comment in the Dockerfile explaining why we only touch `source =` lines.

### Other things I noticed (non-blocking, see inline)

- The git wrapper in the builder stage silently swallows *any* fetch that matches a very liberal glob (`*nicol*test262*`, `*/user/nicol*`). This works but will be baffling to the next person debugging why a submodule fetch mysteriously exits 0. Add a comment + a stderr log line.
- `pnpm install --no-frozen-lockfile` in two places — this defeats the point of committing `pnpm-lock.yaml`. If v2.9.5-coasys-4 forced a lockfile change, land the updated lockfile; don't paper over it in the Dockerfile.
- `docker-entrypoint.sh` grew a lot of new logic (agent auto-generation/unlock, AI model auto-registration, global-space bootstrap, WE auth injection, Caddy reverse proxy). This is well outside "update for v2.9.5-coasys-4" — it looks like an accumulation of prior commits that need their own review pass, not a scope for this PR.
- The `extract_agent_flag` function references *known typos* in the CLI (`is_initiliazed`). Please file (or link) a bug on the CLI so we can eventually delete this workaround.

**Overall**: the technical approach is sound and matches what actually works. But this PR is presented as a one-line-of-context fix and it isn't — it's the runtime image + entrypoint infra landing wholesale, on top of a red base branch, with zero CI coverage of the thing being changed. Please rebase / redescribe / add a build-check job, then I'm happy to re-review and approve.

HexaField (commented):
The changes to mod.rs and wallet.rs should be upstreamed to feat/wallet-backend

HexaField:
hi @marvin-bot-coasys regarding this PR introducing Dockerfiles that don't exist yet, they were taken from this PR https://github.com/coasys/ad4m/pull/778

Dockerfile
HexaField: Addressed in 2248fb909 — git wrapper now logs skipped commands to stderr with a `[docker-build]` prefix.

Dockerfile
HexaField: Addressed in 2248fb909 — added a comment explaining why only `source =` lines are stripped and that the `[patch]` sections handle the rest.

Dockerfile
HexaField: Addressed in 2248fb909 — added a TODO comment acknowledging the lockfile debt and explaining the root cause (Deno v2.9.5-coasys-4 monorepo migration).

Dockerfile
HexaField: Addressed in 2248fb909 — added a comment explaining why both sed (direct deps) and `[patch]` (transitive deps) are needed.

docker-entrypoint.sh
HexaField: Addressed in 2248fb909 — added a tracking comment with a placeholder issue link.

docker-entrypoint.sh
HexaField: Acknowledged — the current signal handling covers the primary case (SIGTERM/SIGINT forwarding to executor + WE). The reviewer's concern about Caddy/early-exit races is valid but edge-case. The entrypoint waits on `EXECUTOR_PID` which is the critical process. If Caddy exits early, the orphaned port doesn't affect executor operation. Adding `wait -n` or a process supervisor (s6, tini) would be a larger rearchitecture.

docker/generate-seed.mjs
HexaField: Addressed in 2248fb909 — added a rationale comment explaining why npm deps are avoided in the Docker build context.

docker/download-models.sh
HexaField: Addressed in 2248fb909 — added a note about the lack of integrity checks and a recommendation to pin revisions to commit SHAs for reproducible builds.

docker/generate-seed.mjs
```diff
@@ -0,0 +1,171 @@
...
+    const langDir = join(outputDir, "languages", hash);
+    mkdirSync(langDir, { recursive: true });
+    writeFileSync(join(langDir, "bundle.js"), content);
+
+    console.log(lang.role + ": " + hash);
+}
+
+// Build seed JSON
+const seed = {
+    trustedAgents: [],
```
lucksus: Is it working with this? We put the agent that has published the languages in here for the regular seed generation. Without languages being authored by a trusted agent, ad4m will reject installing those.

marvin-bot-coasys: No — empty `trustedAgents` plus empty `author` on the language-language KV meta is not enough.

Regular seed generation (`cli/src/bootstrap_publish.rs`) publishes with a live unlocked agent and writes that DID into `trustedAgents`:

```rust
trusted_agents: vec![agent.did.unwrap()],
```

Runtime still *also* trusts the local agent (`RuntimeService::get_trusted_agents` always `push(did())`). That does **not** save this seed, because `generate-seed.mjs` writes language meta with `author: \"\"` (both the Expression `author` and `data.author`). Language install then does:

```rust
if trusted_agents.contains(&language_author) || agent_did == language_author { ... }
// else, for templated languages:
if !trusted_agents.contains(&source_author) {
    // \"Agent which created source language ... is not a trustedAgent... aborting\"
}
```

`\"\"` is not the container DID and is not in `trustedAgents`, so `applyTemplateAndPublish` (Flux `createCommunity`, etc.) will refuse the bootstrap languages.

Pre-copying `bundle.js` onto disk only skips *fetch*. It does not skip this author check on install/template.

Two ways to match the real seed path:

1. **Build-time DID** (closest to `bootstrap_publish`): generate a keypair while building the seed, set `trustedAgents: [thatDid]`, and set every language Expression `author` / `data.author` to that DID. Bake the same keys into the image (or a well-known docker bootstrap agent).
2. **First-boot rewrite** (simpler for ephemeral containers): after `agent generate`, rewrite the language-language KV meta authors to the container DID. Then `agent_did == language_author` passes even with `trustedAgents: []`.

(1) is what production seed gen does. (2) is less ceremony if every container is its own network.

lucksus: I think both the trusted agent as well as the language hashes should come from the bootstrap seed committed to the repo. unless the docker image is rolling it's own deploy of the languages - then the trustedAgent needs to be the agent that published those languages to the language language.

marvin-bot-coasys: Landed in `8c4e13ecb` on this branch.

Agreed on the rule: **committed seed** is the source of `trustedAgents`; **language hashes** come from that seed unless this image is its own language deploy.

This image *is* the second case (Dockerfile: local bootstrap languages replace Holochain-backed ones). So:

- `trustedAgents` is now copied from `rust-executor/src/mainnet_seed.json` (the two `did:key:z6Mk…` publishers). Never `[]`.
- Hashes stay computed from the local `bootstrap-languages/` files — they will not match mainnet addresses, and should not. Stamping a mainnet publisher DID onto those bundles would claim those agents published code they did not.
- First boot (`stamp_bootstrap_language_authors` in `docker-entrypoint.sh`) rewrites language-language KV `author` / `data.author` to the container DID after `agent generate`. Runtime already `push(did())` into trusted agents, so `agent_did == language_author` is what lets `applyTemplateAndPublish` through for the locally-hashed copies.

If we later want the image to *be* mainnet (same hashes as the committed seed), drop the local hash step and copy addresses + `languageLanguageBundle` from `mainnet_seed.json` instead — then the seed trustedAgents are sufficient and the first-boot stamp goes away.

marvin-bot-coasys (dismissed):
## Review — approve

The original CHANGES_REQUESTED items from 2026-08-30 are resolved on this head (`8c4e13ecb`, now based on `dev`):

1. **Scope** — title/body now say this *introduces* the runtime image, which is what the diff is. Rebase onto `dev` also killed the inherited-red-base problem.
2. **Cargo.lock `source =` strip** — Dockerfile comment explains why only those lines go, and that `[patch]` owns the rest.
3. **Git wrapper / CLI typo / lockfile** — logged skips, tracking comment on `is_initiliazed`, TODO on `--no-frozen-lockfile`. Still no `docker build` smoke job; I will not block on that after you documented local testing and that CircleCI never exercises this Dockerfile.

The later correctness hole (`trustedAgents: []` + empty language `author`) is what would have kept me red. `8c4e13ecb` copies trustedAgents from `rust-executor/src/mainnet_seed.json` and stamps locally-hashed language-language KV authors to the container DID after `agent generate`. That matches the install check (`trusted_agents.contains(author) || agent_did == author`) without claiming mainnet publishers authored the standalone hashes.

Remaining follow-ups, not this PR: land a frozen pnpm lockfile, pin WE/Flux clones off `dev`, checksum the model downloads. Ready to merge from my side.

docker-entrypoint.sh
```diff
@@ -0,0 +1,543 @@
...
+stamp_bootstrap_language_authors() {
+    local kv_file did
+    if [ ! -f /opt/ad4m/language-language-kv/address.txt ]; then
+        return 0
+    fi
+    local ll_addr
+    ll_addr=$(cat /opt/ad4m/language-language-kv/address.txt)
+    kv_file="/data/ad4m/languages/${ll_addr}/ad4m-language-kv.json"
+    if [ ! -f "${kv_file}" ]; then
+        return 0
```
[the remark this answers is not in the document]
HexaField: Addressed in 47a6357c5 — all three error branches in `stamp_bootstrap_language_authors` now `return 1` with an ERROR-level log message. With `set -e` active, this stops the entrypoint on missing KV file, unreadable agent DID, or jq failure.

docker-entrypoint.sh
```diff
@@ -0,0 +1,496 @@
...
+}
+
+# ── AI model auto-registration ─────────────────────────────────────────────
+# When pre-cached models exist (INCLUDE_MODELS=true at build time), register
+# them with the executor via WebSocket RPC so Flux transcription and
+# summarisation work out of the box.
+ws_rpc() {
+    local msg_type="$1" params="$2" msg_id="${3:-1}"
+    printf '{"id":"%s","type":"%s","params":%s}\n' "${msg_id}" "${msg_type}" "${params}" \
+        | websocat -n1 "ws://localhost:12000/api/v1/ws?token=${ADMIN_CREDENTIAL}" 2>/dev/null
```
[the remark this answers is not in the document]
HexaField: Addressed in 47a6357c5 — moved the `cleanup()` function definition and `trap cleanup TERM INT` immediately after `EXECUTOR_PID=$!` assignment, before the readiness wait loop and bootstrap logic. PID 1 in Docker ignores untrapped signals, so the trap now covers the entire post-launch lifecycle.

Dockerfile
```diff
@@ -0,0 +1,419 @@
...
+    apt-get update && apt-get install -y --no-install-recommends \
+      ca-certificates curl git nodejs && \
+    curl -fsSL https://deb.nodesource.com/setup_${NODE_MAJOR}.x | bash - && \
+    apt-get install -y --no-install-recommends nodejs && \
+    npm install -g pnpm@9.15.0 && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN if [ "${INCLUDE_WE}" != "true" ]; then exit 0; fi && \
+    git clone --depth 1 --single-branch --branch dev \
+      https://github.com/coasys/we.git /we
```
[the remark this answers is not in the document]
HexaField: Acknowledged — valid concern for reproducible builds. The `dev` branch reference here reflects an intentional design choice: this Docker image tracks the latest WE/Flux development and rebuilds frequently. Pinning to commit SHAs would require updating the Dockerfile on every WE/Flux push, adding friction without improving security (the repos belong to the same org).

For a future production-grade image, we should pin to release tags or SHAs. Leaving this as-is for now since the image targets development/staging use.

docker-entrypoint.sh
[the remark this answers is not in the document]
HexaField: Addressed in 48da11ffc — `ws_rpc` now percent-encodes `ADMIN_CREDENTIAL` via `jq @uri` before interpolating into the WebSocket query string. Credentials containing `+` or other reserved characters now survive Axum's `form_urlencoded` decoding intact.

(The earlier reply about trap placement addressed a separate concern — that fix also landed in 47a6357c5.)

PR #946, opened 2026-09-02
feat(ad4m-ldk): extract WebSocket interfaces into ALDK

HexaField:
## Summary

- Extracts `WSConnection`, `WebSocketFactory` interfaces and `DenoWebSocketFactory` class into `@coasys/ad4m-ldk` (`ad4m-ldk/js/src/websocket.ts`)
- Adds minimal ambient type declarations (`websocket-globals.d.ts`) for the WebSocket globals the executor's Deno sandbox provides — avoids pulling full DOM lib into the ALDK tsconfig
- Adds 12 tests covering interface contracts and adapter behaviour via mock WebSocket global
- Exports from the ALDK barrel (`ad4m-ldk/js/src/index.ts`)

## Context

The server-link-language ([#930](https://github.com/coasys/ad4m/pull/930)) defines its own WebSocket abstraction layer. These interfaces and the Deno implementation belong in the LDK so every link language can reuse them without duplicating adapter code.

Once this lands, #930 replaces its local `WSConnection`/`WebSocketFactory`/`DenoWebSocketFactory` definitions with imports from `@coasys/ad4m-ldk`.

## Test plan

- [x] `tsc --noEmit` passes (ambient types resolve WebSocket/MessageEvent/CloseEvent/Event)
- [x] 12 new websocket tests pass (interface contracts + DenoWebSocketFactory mock)
- [x] 7 existing defineLanguage tests still pass
- [ ] CI `build-and-test` passes

## Summary by CodeRabbit

* **New Features**
  * Added a platform-agnostic WebSocket interface for connecting, sending messages, closing connections, and handling events.
  * Added Deno WebSocket support, including connection lifecycle and message handling.

* **Tests**
  * Added coverage for WebSocket connections, event handling, closing behavior, errors, and binary messages.

* **Developer Experience**
  * Added TypeScript declarations for WebSocket-related runtime globals.

### DIFF
--- a/ad4m-ldk/js/src/websocket.ts
+++ b/ad4m-ldk/js/src/websocket.ts
@@ -1,0 +1,62 @@
+/**
+ * WebSocket adapter interfaces and Deno implementation.
+ *
+ * Provides a platform-agnostic WebSocket abstraction so link languages
+ * can open WebSocket connections without directly depending on the Deno
+ * or browser WebSocket global. The DenoWebSocketFactory wraps the native
+ * WebSocket global available in the executor's language sandbox.
+ */
+
+/** A single open (or opening) WebSocket-like connection. */
+export interface WSConnection {
+    send(data: string): void;
+    close(code?: number, reason?: string): void;
+    onOpen(cb: () => void): void;
+    onMessage(cb: (data: string) => void): void;
+    onClose(cb: (code: number, reason: string) => void): void;
+    onError(cb: (err: unknown) => void): void;
+}
+
+/** Factory for creating WebSocket connections. */
+export interface WebSocketFactory {
+    connect(url: string): WSConnection;
+}
+
+/**
+ * Wraps the native WebSocket global available in the executor's Deno
+ * language sandbox. The executor runs a full deno_runtime MainWorker
+ * (not a bare deno_core isolate), so deno_web/deno_net/deno_websocket
+ * are all present.
+ */
+export class DenoWebSocketFactory implements WebSocketFactory {
+    connect(url: string): WSConnection {
+        const ws = new WebSocket(url);
+
+        return {
+            send(data: string): void {
+                ws.send(data);
+            },
+            close(code?: number, reason?: string): void {
+                try {
+                    ws.close(code, reason);
+                } catch {
+                    // Already closed/closing
+                }
+            },
+            onOpen(cb: () => void): void {
+                ws.addEventListener("open", () => cb());
+            },
+            onMessage(cb: (data: string) => void): void {
+                ws.addEventListener("message", (event: MessageEvent) => {
+                    cb(typeof event.data === "string" ? event.data : String(event.data));
+                });
+            },
+            onClose(cb: (code: number, reason: string) => void): void {
+                ws.addEventListener("close", (event: CloseEvent) => cb(event.code, event.reason));
+            },
+            onError(cb: (err: unknown) => void): void {
+                ws.addEventListener("error", (event: Event) => cb(event));
+            },
+        };
+    }
+}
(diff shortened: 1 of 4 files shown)


ad4m-ldk/js/src/websocket.ts
```diff
@@ -0,0 +1,62 @@
...
+        const ws = new WebSocket(url);
+
+        return {
+            send(data: string): void {
+                ws.send(data);
+            },
+            close(code?: number, reason?: string): void {
+                try {
+                    ws.close(code, reason);
+                } catch {
```
lal-bot-coasys: This swallows *any* exception from `ws.close()`, not just the already-closed/closing case the comment describes. If the runtime WebSocket throws for a different reason (e.g. Deno's implementation validating the close code/reason per spec), that error disappears silently with no route to `onError`. Worth either narrowing what's caught (check `readyState` before calling close, or match on the specific error) or at minimum forwarding unexpected failures to the `onError` callback instead of a bare catch-and-drop.

ad4m-ldk/js/src/websocket.ts
```diff
@@ -0,0 +1,62 @@
...
+                    // Already closed/closing
+                }
+            },
+            onOpen(cb: () => void): void {
+                ws.addEventListener("open", () => cb());
+            },
+            onMessage(cb: (data: string) => void): void {
+                ws.addEventListener("message", (event: MessageEvent) => {
+                    cb(typeof event.data === "string" ? event.data : String(event.data));
```
lal-bot-coasys: `String(event.data)` on non-string payloads will produce something like `"[object ArrayBuffer]"` rather than the actual bytes — silent data mangling if a server ever sends a binary frame, since the interface has no binary path at all. If link languages are guaranteed text-only (JSON) protocols this is fine as a deliberate scope cut, but might be worth a one-line comment saying so explicitly (or asserting/erroring instead of coercing) so a future caller doesn't lose a real payload without knowing why.

lal-bot-coasys (dismissed):
Clean extraction — small, self-contained, and does exactly what the description says: 3 tight interfaces plus a Deno-backed implementation, 12 new tests covering the adapter contract, no existing code touched. I checked the thing I'd normally worry about with a global-scope ambient .d.ts shipped in a public package (leaking `WebSocket`/`Event`/`MessageEvent`/`CloseEvent`/`Blob` into every consumer's type-check) — built the package and inspected the emitted `lib/websocket.d.ts`: the public API surface only exposes `WSConnection`/`WebSocketFactory`, so the ambient globals never appear in the emitted declarations and there's no reference-directive leak. Ran `tsc --noEmit` and the full test suite locally on this branch: 19/19 pass, matches the PR's own numbers exactly. The failing `integration-tests-js` check is very likely unrelated flake, not this diff — it's an additive change only, doesn't touch `tests/js`, and that same job has failed on plain `dev` commits twice in the last 8 pushes with nothing to do with ad4m-ldk. Two small non-blocking nits inline, both about error/edge-case handling in `DenoWebSocketFactory` — worth a look before or after merge, neither should hold this up.

lal-bot-coasys (dismissed):
Addressed both nits from my earlier review in `430a63a3b`:
- `close()` now checks `readyState` before calling close (skips the already-closed/closing case without even attempting it), and forwards any other unexpected close failure to `onError` instead of a bare catch-and-drop.
- `onMessage()` now detects binary frames (`ArrayBuffer`/`Blob`/`ArrayBufferView`) and forwards an explicit error to `onError` instead of silently mangling them via `String(data)`; plain non-string scalars (e.g. numbers) still stringify as before.

Added 2 new tests covering both changes (readyState guard + onError forwarding, binary-frame rejection) — 21/21 pass locally, `tsc` clean. Also merged `dev` in per your ask (clean merge, no conflicts, tests still green after).

lal-bot-coasys:
Also applied CodeRabbit's finding on `onError` overwriting the single stored callback (`3ddd25a66`) — switched to a callback list so all registered `onError` listeners get notified for every error source (native WS errors, forwarded close() failures, binary-frame rejection), not just the last one registered. Added a regression test for repeated registration. 22/22 tests pass, tsc clean.

lucksus (approved):

PR #947, opened 2026-09-02
feat(agent): PR1 — verifier seam and proof persistence

HexaField:
### What

This PR adds a verifier seam that routes all signature verification through a single method-neutral dispatch, and extends `ExpressionProof` with `key_id` and `kel_seq` fields so revocation has an anchor point.

### Why

The executor verifies signatures in many places, all hard-wired to `did:key`. The identity stack needs `did:scid` — which resolves keys from a log, not from the DID string itself. A single dispatch point lets the new method slot in without touching every call site.

### How

```
 verify(expr) / verify_string_signed_by_did(did, data, sig)
         │
         ▼
   inner_verify(did, message, signature, key_id, kel_seq)
         │
    ┌────┴────┐
    │ did:key  │──▶ verify_key_method()  [unchanged legacy path]
    │ did:scid │──▶ verify_scid_with(resolver, ...)
    │ other    │──▶ verify_key_method()  [backward compat]
    └─────────┘
```

**Verifier seam** (`signatures.rs`):

```rust
enum DidMethod { Key, Scid, Other }

trait KeyStateResolver: Send + Sync {
    fn resolve(&self, identifier: &str, kel_seq: Option<u64>)
        -> Result<KeyState, ResolveError>;
}

struct KeyState { master: String, keys: Vec<VerificationMethod> }
struct VerificationMethod { id: String, key: String }  // key = did:key
```

- `did:key` / `Other` → `verify_key_method()` — byte-identical to pre-refactor code.
- `did:scid` → `verify_scid_with()` — resolves key state via the trait, filters by `key_id` if present, tries each authorised key. **Fails closed** when no resolver installed or resolution errors.
- `set_key_state_resolver()` installs the resolver once at boot (PR3 provides the real implementation). Until then, `did:scid` verification always rejects.

**Proof persistence** (`ExpressionProof`):

```rust
struct ExpressionProof {
    key: String,           // existing
    signature: String,     // existing
    key_id: Option<String>,  // NEW — verification method id
    kel_seq: Option<u64>,    // NEW — log position at signing time
}
```

- DB schema gains `key_id TEXT` and `kel_seq INTEGER` columns on `expression_proof`.
- Migration backfills `NULL` for existing proofs — legacy `did:key` proofs carry no anchor, which matches expected behavior.

**Tests** — 11 new tests: DID method classification, legacy did:key round-trip, did:scid accept/reject/tampered/revoked, fail-closed without resolver, two-master isolation, resolver-with-revocation fixture.

| File | Change |
|------|--------|
| `signatures.rs` | +615 — verifier seam, KeyStateResolver trait, tests |
| `types/core.rs` | +26 — key_id, kel_seq on ExpressionProof |
| `db.rs` | +77 — schema migration, proof column persistence |
| `prolog_service/engine_pool.rs` | +66 — proof fields in Prolog predicates |
| 8 other files | proof struct Default propagation |


HexaField:
@coderabbitai review

rust-executor/src/types/core.rs
```diff
@@ -236,6 +245,7 @@ impl DecoratedLinkExpression {
             proof: ExpressionProof {
                 key: self.proof.key.clone(),
                 signature: self.proof.signature.clone(),
+                ..Default::default()
```
data-bot-coasys: **Known-gap to track for the minting PR:** `DecoratedLinkExpression::verify_signature()` rebuilds the proof with `..Default::default()`, dropping `key_id`/`kel_seq` (same in the `From<DecoratedLinkExpression>` impl at L281 and the five DB row-reads in `db.rs`, whose schema stores only `signature` + `key`). For a future `did:scid` author this makes decorated-link verification resolve against **current** key state instead of state-at-signing — so after a rotation/revocation, previously-valid links would start failing (fail-closed direction, but it breaks "old data stays valid", the property this refactor exists to provide).

Nothing to change in this PR — no `did:scid` exists yet and plumbing the fields through Decorated types + DB schema belongs with the KEL-store PR. Suggest noting it in the stack tracking issue so it can't slip: **persist `keyId`/`kelSeq` in link storage and carry them through every `ExpressionProof` reconstruction before the first SCID mints.**

rust-executor/src/agent/signatures.rs
```diff
@@ -8,6 +8,93 @@ use log::error;
...
+    if let Ok(mut guard) = RESOLVER.write() {
+        *guard = Some(resolver);
+    }
+}
+
+fn key_state_resolver() -> Option<Arc<dyn KeyStateResolver>> {
+    RESOLVER.read().ok().and_then(|g| g.clone())
+}
+
+// ─── Public verify API ───────────────────────────────────────────────────────
```
data-bot-coasys, line 100: Nit: if the `RwLock` is poisoned this silently no-ops and the resolver never installs — every `did:scid` verification then fails closed with only the per-verify error log. Fine behavior-wise (fail closed), but consider `.expect()` here instead: a poisoned lock at boot-time install is unrecoverable state and a loud panic beats a quietly dead identity system.

data-bot-coasys (dismissed):
Reviewed as stack-part 1/9. **Verdict: APPROVE** — the seam is exactly right for a foundation PR: method dispatch is minimal, the legacy `did:key` path is extracted verbatim into `verify_key_method` (byte-identical), `did:scid` fails closed on every branch (no resolver / resolve error / unknown keyId / tampered message — all tested), and the `ExpressionProof` additions are serde-compatible both directions (`default` + `skip_serializing_if`), so existing proofs round-trip unchanged.

Crucially this PR is **method-neutral**: nothing here commits to KERI vs a lighter verifiable-history format, HKDF-derivation vs delegation-authorization, or where the KEL lives — `did:scid:ke:` strings appear only in test fixtures. So the design conversation (inception format, sub-key policy) stays open while the seam lands. 👍

One substantive gap flagged inline (keyId/kelSeq round-trip loss) — harmless today because no agent mints a `did:scid` yet, but it must land before/with the PR that first mints one, otherwise the core guarantee of this refactor ("data signed before revocation stays valid") silently breaks for decorated links. Also worth tracking: the JS/TS `ExpressionProof` type and GraphQL schema will need the optional `keyId`/`kelSeq` fields for parity once the wire carries them.

Minor, non-blocking: a test asserting that a legacy `did:key` proof *with* `keyId`/`kelSeq` populated still verifies via the legacy path (fields ignored) would lock the dispatch precedence.

PR #950, opened 2026-09-02
flow: simplify the FlowTransitionProposal post-processing pass (stacked on #932)

lucksus:
Stacked on #932. Same behaviour on the graph, about a third of the code.

## What this does

#932 adds a deterministic pass after extraction: re-evaluate every active flow's `requires` guards against the freshly committed evidence and mint a `FlowTransitionProposal` for each one that holds. This PR keeps exactly that and strips everything around it that was not needed to deliver it.

| | #932 | this PR |
|---|---|---|
| Lines changed vs `dev` | ~4,450 | ~1,450 |
| Flow evaluator | 5 files, 3 layers, 2 wrapper types | 1 file |
| Test files | 2 (1,900 lines) | unit tests inline + 1 e2e file |

## Moved out (to a follow-up on #940)

- **Semantic-check gate** (`flow_semantic_check.rs`, `SemanticCheckLlm`, `AIServiceSemanticCheck`, the `(llm, model_id)` parameter). The design doc lists this as its own PR (§13 step 5).
- **Strategy-path LLM attribution** (`LlmFlowProposal`, `InterpretationOutput`, `parse_interpretation_output`, the `flow_proposals` prompt section, `LlmProposalHint`). Its only effect was an optional `rationale` string on a proposal the engine emits anyway, and #940 already owns the harness half of attribution. `parse.rs`, `prompt.rs` and `types.rs` are back to `dev`, except for the `serde_json`-based JSON extraction.

## Simplified in place

- `flow_evaluator/{mod,primitives,queryable,engine_pass}.rs` folded into one `flow_evaluator.rs`. `write_engine_proposal` (a pass-through wrapper) is gone; `SatisfiedTransition` drops the `subject` and `consensus_rule` fields nothing read, so it can derive `PartialEq` again. Module-level `#![allow(dead_code)]` removed everywhere.
- `run_engine_proposal_pass` is `(perspective, subjects: &[String], context) -> Vec<String>`. Both interpretation call sites pass the base URIs the LLM just wrote, so the pass only loads flow instances anchored to those subjects. An empty slice still sweeps every live instance. Loading is one fallible block; each proposal still commits in its own batch.
- `load_all_flow_instances` and `load_flow_instances` share one private query helper instead of duplicating the body.
- Token substitution in `where` strings (`$flow.base`, `$flow.uri` / `$flow.instance`, `$did`) is the shared `FlowTokens` helper. `linkedTo` compiles to a `model_query` `parent` scope (`"base"`/`"flow"` → `ad4m://has_child`; `{ via, to }` names the predicate). Translation output is validated as `ModelQueryInput` before it is run.
- Doc comments no longer carry slice numbers, dates, reviewer names or PR references.

## Behaviour changes (fixes)

`PropertyCondition::Matches` was translated to a `{"regex": ...}` operator that `WhereOps` does not have, so `model_query` would have read it as a nested where-clause on a property called `regex` and evaluated the guard wrongly; the unit test locked in that wrong shape. `Exists` was mapped to `{"not": {"equals": null}}`, which the `not` matcher does not handle either. Both now fail translation. Untranslatable guards (`exists`/`matches`, a colliding `didProperty`, a malformed `linkedTo`, an `or` branch with a different class or its own `count`) skip that transition and log at `warn!` — they will never start working on retry. Transient `model_query` errors stay at `debug!`. Adding `exists`/`regex` support to `model_query` is the real fix and is out of scope here.

The old `$did`-in-the-`didProperty`-name expression (treat the field as a template containing `$did`) is gone. It was already broken; `didProperty` is a property *name*, and `$did` is substituted in `where` string values instead.

A second pass against an unchanged graph no longer mints a duplicate proposal: same flow instance + same evidence hash is skipped.

## Tests

- `cargo test -p ad4m-executor --lib perspectives::flow` (unit + live-perspective e2e)
- `cargo test -p ad4m-executor --lib perspectives::interpretation::parse`
- `cargo clippy -p ad4m-executor --tests` is clean on the touched files.


rust-executor/src/perspectives/flow_evaluator_e2e.rs
```diff
@@ -0,0 +1,239 @@
...
+}
+
+fn literal(s: &str) -> String {
+    format!("literal:string:{}", urlencoding::encode(s))
+}
+
+/// Two-state Delivery flow (`identified → scoped`) whose `scoped` state
+/// requires at least one `ns://Task`, plus one FlowInstance sitting in
+/// `identified`. The `requires` link is added by hand because
+/// `parse_flow_to_links` does not emit it yet.
```
data-bot-coasys: This comment is factually out of date: `parse_flow_to_links` **does** emit `ad4m://requires` — see `shacl_parser.rs:496-503` (single JSON literal on `ad4m://requires`). The fixture could put `requires` into the flow JSON and exercise the production serializer instead of hand-adding the link, which would also make the round-trip assertion at ~line 149 test the real write path. Non-blocking, but worth fixing so the fixture doesn't drift from production behavior.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,718 @@
...
+                Ok(Some((class_names, evidence_ids))) => out.push(SatisfiedTransition {
+                    flow_name: flow.name.clone(),
+                    instance_uri: record.instance_uri.clone(),
+                    from_state: record.current_state.clone(),
+                    to_state: state.name.clone(),
+                    evidence_hash: evidence_hash(&class_names, &evidence_ids),
+                    evidence_ids,
+                }),
+                Ok(None) => {}
+                Err(e) => log::debug!(
```
data-bot-coasys: Untranslatable guards (`exists`/`matches` bails from `where_condition`) land here at `debug!`. But unlike transient evaluation errors, an untranslatable guard is a *persistent* misconfiguration: the flow definition silently never auto-advances, forever, and nothing surfaces it. I'd log `warn!` for the bail-path (or split translation errors from evaluation errors) so a flow author has a chance of discovering why their transition never fires. Non-blocking.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,718 @@
...
+    })
+}
+
+fn requires_where(query: &ModelQuery, acting_did: &str) -> Result<Map<String, Value>> {
+    let mut out = Map::new();
+    for (field, cond) in query.r#where.iter().flatten() {
+        out.insert(field.clone(), where_condition(field, cond)?);
+    }
+    if let Some(prop) = &query.did_property {
+        out.insert(prop.clone(), Value::String(acting_did.to_string()));
```
data-bot-coasys, line 166: `didProperty` silently overwrites a same-named user `where` field here (insert after the loop wins). A flow that declares both `where.author = X` and `didProperty: author` gets the user's condition dropped without a trace. Bailing on the collision would match the fail-closed spirit of the rest of this PR. Non-blocking.

data-bot-coasys (approved):
**Valid — and better than my #932 in the ways that matter. Approving.**

What I verified rather than took on faith:

- **The bug claim is real, and worse than the PR states.** `WhereOps` is parsed with `deny_unknown_fields`, so a `{"regex": …}` guard didn't error — it fell through the untagged enum to `SubClause` and was evaluated as a nested where-clause on a property literally named `regex`. And `equals` isn't a `WhereOps` field at all, so the old `Exists{false}` → `{"equals": null}` mapping was mis-evaluated the same way. Guards using `matches`/`exists` silently produced wrong results, and my old unit test had locked in the wrong shape. The new fail-closed translation (bail + skip that transition) is the correct interim behavior.
- **Behavior parity holds.** Both call sites (single-shot and harness path) still run the deterministic pass after `apply_with_overlay`; the loader keeps the "Shape not found → empty" tolerance (now case-insensitive — an improvement); `prompt.rs` and `types.rs` are byte-identical to dev; `parse.rs` keeps only the strict serde_json extraction. No dangling references to removed symbols anywhere in the tree.
- **Matches agreed sequencing.** Design doc §13 step 5 lists the semanticCheck pass as its own PR — I had front-loaded it into #932. The schema fields (`semanticCheck`, `consensusRule`) stay in the parser; only enforcement moves out. Correct cut.
- **CI is fully green** across all eight checks, including `integration-tests-js`.

Acknowledged cost: #940's diff touches removed symbols in ~82 places, so the attribution follow-up is a genuine re-stack — the PR body owns that explicitly, fine by me.

One note that belongs in the PR body rather than inline: the old `$did`-expression corner of `didProperty` is silently dropped. It was self-admittedly broken anyway, but a line in the description would help future archaeology.

Three non-blocking inline comments below.

data-bot-coasys:
Merged the base branch (\`feature/flow-post-processing\`) in to clear the conflict — this PR was showing \`CONFLICTING\`/\`DIRTY\` and could not be merged. The cause was mine: I pushed two commits to the base after you branched (the \`dev\` merge \`a4136425c\` and the parse fix \`b8f263530\`).

**One conflict, in `parse.rs` step 3 — resolved in favour of THIS branch, verbatim.**

Your array-first chain and my `extract_bracketed_salvageable` guard collided. I dropped the guard, because your simplification structurally subsumes it:

- The guard existed to stop a greedy `{`-span taken from prose (`OK, I'll extract {a couple of things}: [{"class":"X"}]`) from short-circuiting the `or_else` chain ahead of the real array. That failure mode **requires the object branch to be tried before the array branch** — which my chain did and yours does not.
- Under your ordering the `{`-span is only reached after both array paths fail, and `parse_interpretation_response` (the sole caller once this PR removes `parse_interpretation_output`) requires a top-level **array**. So anything the `{`-span yields is an error regardless: the guard changes no outcome, only the error message.
- I also checked the mirror case — a prose *bracket* before the real array (`Here are the items [see above]: [...]`). That is unrecoverable on **both** chains, since only the first `[` is ever consulted, so the guard bought nothing there either. I wrote that test, traced it to a failure, and deleted it rather than assert something false.

**Kept:** the regression test `prose_braces_before_the_real_array_dont_swallow_the_payload`, which passes here and is now the lock proving the *ordering* is what protects us. Its comment says so.

Flagging explicitly because CodeRabbit raised that finding twice — I did not want it disappearing inside a merge. If you would rather keep the guard, it is a small revert.

**Also swept up:** `b8f263530` had accidentally committed five local scratch files (`rerun-one.log`, `rerun-one.sh`, `run-932-tests.sh`, `run-tests.sh`, `test-run.log`) onto the base branch — absent from `dev`, and they would have ridden along into `dev` at merge. Removed on the base branch (`c88c07bdb`, `e0ca3726e`) and merged down here, per stacked-PR hygiene (fix on the merge-first branch, downstream picks it up by merge).

Verified locally on the merge result: `cargo fmt --all --check` clean, `interpretation::parse` 15/15, `interpretation` 95/0, `flow` 84/0.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,718 @@
...
+            .map(|alt| requires_where(alt, acting_did).map(Value::Object))
+            .collect::<Result<Vec<_>>>()?;
+        out.insert("OR".to_string(), Value::Array(branches));
+    }
+    Ok(out)
+}
+
+fn where_condition(field: &str, cond: &PropertyCondition) -> Result<Value> {
+    Ok(match cond {
+        PropertyCondition::Str(s) => json!(s),
```
lucksus: **[suggestion]** `PropertyCondition::Str` (and `Equals`/`In` string values) are copied into `model_query` verbatim. The ModelQuery spec (`core/src/shacl/SHACLFlow.ts:48`) says the engine substitutes `$flow.base`, `$flow.uri`, and `$did` in `where` strings at evaluation time, and the canonical Deliberation fixtures actually write `where: { about: "$flow.base" }` / `forTask: "$flow.base"`. A literal `"$flow.base"` will never equal the instance subject, so those guards look unmet and the pass never mints. `FlowInstanceRecord.subject` / `instance_uri` are already in hand at `evaluate_flow_transitions`; they are just not threaded into translation. Same gap existed in #932; this file is now the only place it can be fixed.

**Suggestion:** Pass the current `FlowInstanceRecord` (or a small `{base, uri, did}` token set) into `requires_query_input` / `where_condition` and substitute those three tokens in string conditions before emitting JSON. Add a unit test that `where.about = "$flow.base"` becomes the record's `subject`.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,718 @@
...
+fn requires_query_input(query: &ModelQuery, acting_did: &str) -> Result<Value> {
+    let where_clause = requires_where(query, acting_did)?;
+    Ok(if where_clause.is_empty() {
+        json!({})
+    } else {
+        json!({ "where": where_clause })
+    })
+}
+
+fn requires_where(query: &ModelQuery, acting_did: &str) -> Result<Map<String, Value>> {
```
lucksus: **[suggestion]** `ModelQuery.linked_to` is never read. A `requires` query `{ className: "ns://Task", linkedTo: "base" }` is specified as “a Task linked to this flow’s subject”, but the translator emits `{}` (or only `where`) and will match every Task on the perspective. That is fail-open: a guard that was meant to be scoped can fire a proposal on unrelated evidence. `linkedTo` is rendered into the LLM prompt; only the engine path drops it. Also pre-existing in #932.

**Suggestion:** Either compile `linkedTo` into a relation/`where` constraint against the instance subject, or fail-closed (same as `exists`/`matches`) until that exists, so an unscoped class query cannot silently satisfy the guard.

rust-executor/src/perspectives/flow_evaluator.rs
lucksus: **[suggestion]** `didProperty` is inserted after the `where` loop and overwrites a same-named user condition with no log. A query that declares both `where.author = X` and `didProperty: "author"` drops X. The TS field is a property *name* (common case `didProperty: "author"` → `where.author = acting_did`); the old `$did`-in-the-name expression is gone and was already broken. Collision is the remaining fail-open corner.

**Suggestion:** `bail!` (or `debug!` + skip) if `out.contains_key(prop)` before the insert, matching the rest of this PR’s fail-closed policy.

rust-executor/src/perspectives/flow_evaluator.rs
lucksus: **[suggestion]** Untranslatable `exists`/`matches` bails from `where_condition` and is logged here at `debug!`, same as a transient `model_query` error. A broken operator is a persistent flow-definition misconfiguration: the transition will never auto-advance, and nothing surfaces it at default log levels. Fail-closed skip is the right *behaviour* (Exists:true previously matched everything; Matches never matched unless an instance had a property named `regex`); the observability is not.

**Suggestion:** Split translation errors from evaluation errors and `warn!` the bail path (include flow name, state, field, operator) so a flow author can see why a guard never fires.

rust-executor/src/perspectives/flow_evaluator_e2e.rs
lucksus: **[suggestion]** The fixture comment is wrong: `parse_flow_to_links` does emit `ad4m://requires` (`shacl_parser.rs:496-503`) when the flow JSON carries `states[].requires`. Hand-adding the link means the round-trip assertion at line 149 does not exercise the production writer, so the e2e can stay green if the serializer drifts. The live tests also only use a bare `{className, count:{min:1}}` query (translates to `{}`); scalars/`equals`/`in`/`didProperty`/`OR` never hit real `model_query` deserialization. That is exactly how the original Matches/Exists bug survived: unit tests asserted the *emitted JSON*, not that `ModelQueryInput`/`WhereCondition` accepted it.

**Suggestion:** Put `requires` on the Delivery `scoped` state in `flow_json` and drop the manual link. Add one e2e (or a unit test that `serde_json::from_value::<ModelQueryInput>(requires_query_input(…))`) for a non-empty `where` (e.g. `title` equals the seeded Task title, or `didProperty`) so a future untranslatable shape fails closed in CI instead of silently becoming a `SubClause`.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,961 @@
...
+
+/// Default collection predicate used when `linkedTo` is the `"base"` /
+/// `"flow"` shorthand. Authors who need a different edge write
+/// `{ via, to }` instead.
+const LINKED_TO_DEFAULT_PREDICATE: &str = "ad4m://has_child";
+
+/// Substitute `$flow.base`, `$flow.uri` / `$flow.instance`, and `$did`
+/// in a `where` string. Empty fields are left verbatim so a missing
+/// subject cannot collapse the token into `""`.
+fn substitute_tokens(s: &str, record: &FlowInstanceRecord, acting_did: &str) -> String {
```
data-bot-coasys, line 96: **Reuse:** this substitution already exists as `FlowTokens` in `flow_context/render.rs` — same two tokens, same "empty field means leave the token verbatim" rule, and its doc comment says exactly what this commit does:

> this substitution helper is where the engine will resolve them post-parse. Kept behind `#[allow(dead_code)]` so it's visible in the public API for tests and so `flow_arc slice 10.6c` doesn't have to reintroduce it

This is that reintroduction, so we now have two copies and the original stays dead. Suggest extending `FlowTokens` with a `did` field and the `$flow.uri` alias, calling it from here, and dropping the `#[allow(dead_code)]`.

Worth doing beyond tidiness because the token *set* has drifted three ways: `render.rs` knows `$flow.base` + `$flow.instance`; `SHACLFlow.ts` documents `$flow.base` + `$flow.uri` + `$did`; this function accepts all four. And `prompt.rs`'s `tokens` legend still emits only `$flow.base` and `$flow.instance` — so an author who follows the TS docstring and writes `$flow.uri` gets it substituted by the engine but shows the LLM a token the legend never explains. One definition would keep the legend and the engine honest by construction.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,961 @@
...
+    let where_clause = requires_where(query, record, acting_did, false)?;
+    let mut out = Map::new();
+    if !where_clause.is_empty() {
+        out.insert("where".into(), Value::Object(where_clause));
+    }
+    if let Some(linked) = &query.linked_to {
+        out.insert("parent".into(), linked_to_parent(linked, record)?);
+    }
+    Ok(Value::Object(out))
+}
```
data-bot-coasys, line 144: Consider validating the output here, the way `query_input_deserialises_as_model_query_input` does:

```rust
let out = Value::Object(out);
serde_json::from_value::<ModelQueryInput>(out.clone())
    .map_err(|e| anyhow!("translated query is not a valid model_query input: {e}"))?;
Ok(out)
```

Reason: translation can still emit JSON that `model_query` rejects, and those rejections land in the *wrong* bucket. `PropertyCondition::In` copies the array through verbatim, but `WhereCondition` only has `StringArray` and `NumberArray` — so `{ in: ["a", 1] }` (legal per the TS `PropertyCondition` type, which is `unknown[]`) produces an array that deserializes into nothing. The failure then surfaces *inside* `model_query`, comes back as `RequiresResult::QueryFailed`, and is logged at `debug!` — but it is a permanent misconfiguration that will never work on retry, i.e. precisely what the new `Untranslatable`/`warn!` path exists for.

Same shape as the bug this PR fixes: the guarantee we want isn't "the JSON looks right", it's "`ModelQueryInput` accepts it". The new test proves that check is cheap; running it in production too makes the warn/debug split airtight rather than best-effort.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,961 @@
...
+
+fn requires_where(
+    query: &ModelQuery,
+    record: &FlowInstanceRecord,
+    acting_did: &str,
+    nested: bool,
+) -> Result<Map<String, Value>> {
+    if nested && query.linked_to.is_some() {
```
data-bot-coasys, line 152: Right call — `model_query`'s `OR` is a list of *where* sub-clauses, so a per-branch parent scope genuinely can't be expressed, and bailing beats silently dropping the scope (which would widen the branch to "any instance of the class" — fail-open).

But `SHACLFlow.ts` currently promises the opposite to flow authors:

> Each branch is a full ModelQuery, so branches can each carry their own `didProperty` / `where` / `linkedTo`.

An author who takes that at face value gets a guard that fails translation *as a whole* (the bail propagates out of `requires_where`, so the entire state's `requires` is skipped, not just the branch) with only a `warn!` to explain it. Worth a sentence on the `or` docstring saying `linkedTo` is not supported on branches — cheap now, since nothing depends on it yet.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,961 @@
...
+                "flow" => record.instance_uri.as_str(),
+                other => bail!("`linkedTo.to` `{other}` is not `base` or `flow`"),
+            };
+            (id, via)
+        }
+        _ => bail!("`linkedTo` must be \"base\", \"flow\", or {{ via, to }}"),
+    };
+    Ok(json!({ "id": id, "predicate": predicate }))
+}
+
```
data-bot-coasys, line 249: **Direction is worth documenting on the authoring surface.** `Scope::Raw` compiles to `<id> <predicate> ?source` (`sparql_builder.rs:316`), so this filters *anchor → instance*: the base (or the flow instance) must be the **subject** of the link and the matched item its object.

The TS docstring reads the other way — "how the matched **instance connects back to** the flow" — and the object form invites it: `{ via: "ns://about", to: "flow" }` reads as "the item is *about* the flow", i.e. `?source <about> <flowUri>`, which is the reverse of what gets emitted. `Scope` has no inverse form, so this implementation is the only one available today; the fix is documentation, not code. Suggest saying on `SHACLFlow.linkedTo`: the anchor is the link source, `via` points anchor → item, and the shorthand means `ad4m://has_child` (worth stating — it's currently only discoverable from this constant).

And since the *instance → anchor* direction is the common one in practice, that's exactly the case `where: { about: "$flow.base" }` now covers thanks to this PR — worth pointing authors at it in the same docstring, so `linkedTo` is reserved for real parent-child containment.

rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -0,0 +1,961 @@
...
+fn linked_to_parent(linked: &Value, record: &FlowInstanceRecord) -> Result<Value> {
+    let (id, predicate) = match linked {
+        Value::String(s) => {
+            let id = match s.as_str() {
+                "base" => record.subject.as_str(),
+                "flow" => record.instance_uri.as_str(),
+                other => bail!("`linkedTo` `{other}` is not `base` or `flow`"),
+            };
+            (id, LINKED_TO_DEFAULT_PREDICATE)
```
data-bot-coasys, line 224: Nit, for consistency with `substitute_tokens`: that function deliberately refuses to substitute an empty field so a missing subject can't collapse a token into `""` — but this path has no such guard. An empty `record.subject` yields `{ "id": "", "predicate": "ad4m://has_child" }`, and `validate_iri("")` passes it (it only rejects control chars, whitespace and `<>{}"`), so the builder emits `<> <ad4m://has_child> ?source .` — a base-relative IRI, which ends as either a query error or a silent no-match, and in both cases at `debug!` rather than in the `warn!` misconfiguration bucket.

One `bail!` when the resolved anchor is empty puts it in the same fail-closed, loudly-logged place as the rest.

data-bot-coasys (approved):
## Re-review — the three review items are addressed, and both new fixes are correct. **Approving.**

Re-verified against head `6a573d908`, in a clean worktree of this branch (not from the earlier round's notes).

**My three items, all resolved the way I'd hoped:**

1. The e2e fixture now declares `requires` on the flow JSON and lets `parse_flow_to_links` emit the production `ad4m://requires` link — so the test exercises the real serializer instead of a hand-built link. The stale "does not emit it yet" comment is gone.
2. Untranslatable guards now log at `warn!`, and the split is better than what I asked for: `RequiresResult` separates *translation* failure (persistent misconfiguration → `warn!`) from *query* failure (transient → `debug!`), instead of flattening both into one level.
3. `didProperty` colliding with a `where` field now bails instead of silently overwriting — fail-closed, matching the rest of the pass.

**The two new fixes:**

- **`$flow.base` / `$flow.uri` / `$flow.instance` / `$did` substitution** — this was a real hole: `where: { about: "$flow.base" }` is the form the design doc uses everywhere (§6.1 Deliberation, §6.2 Delivery), and without substitution every one of those guards queried for the literal string `"$flow.base"` and could never match. Substituting `Str`, `Equals` and each element of `In` covers every `PropertyCondition` that carries a value. Leaving a token verbatim when its field is empty is the right call — it keeps a missing subject fail-closed instead of collapsing to `""`, which would have matched instances with an empty property.
- **`linkedTo` → `parent` scope** — the shorthand/object forms match the TS type exactly, and `ad4m://has_child` is the right default (it is the canonical child predicate in `sparql_store`). `{ "id", "predicate" }` deserializes into `Scope::Raw` (it can't be mistaken for `Scope::Model`, which requires `model`).

**Best thing in the diff:** `query_input_deserialises_as_model_query_input`. The original `matches`/`exists` bug existed *because* a unit test asserted the emitted JSON rather than that `ModelQueryInput` accepts it. That test closes the class, and I've suggested inline that the same check is worth running in production (three lines) — it would convert a whole family of silent `debug!`-level failures into the `warn!` bucket you just built.

**Five inline notes, none blocking.** The one I'd most like an answer to isn't a defect in this diff but a question it raises:

> **What writes the anchor→instance edge that `linkedTo` filters on?** `mint_flow_instance` writes only scalars (`flowUri`, `subject`, `currentState`), and the auto-processor's `mint_scope` is configured per-processor (channel-scoped), not per flow instance. So today nothing produces `<base> ad4m://has_child <item>` or `<flowInstance> ad4m://has_child <item>` — which means a `linkedTo` guard is unsatisfiable in production, and it fails as a plain *unmet* guard: no `warn!`, no `debug!`, nothing in the log at all. Since the design doc's Deliberation flow puts `linkedTo: "flow"` on every state, that's worth settling: is the plan `mint_scope = $flow.base` at spawn time, or an explicit edge written when an item is minted inside a flow context? Until then, `where: { about: "$flow.base" }` — which *does* work now, thanks to this PR — is the form that actually fires, and that's worth saying in `SHACLFlow.ts`.

**Testing:** CircleCI `build-and-test` is green on `6a573d908`, which is the job that runs these Rust unit + e2e tests; the integration jobs are still pending and I'll chase them if any go red. I could not reproduce the suites locally — the executor no longer compiles from a bare worktree without `CUSTOM_DENO_SNAPSHOT.residual_esm.rs` and `dapp/dist`, both of which need a full `pnpm build`. So my test evidence here is CI's, not my own; the review above is from reading the code and the surrounding call paths.

data-bot-coasys:
CodeRabbit's three findings on `6a573d908` — I verified each against the code rather than taking the labels at face value. **All three are valid**, but the severity ordering is wrong: #2 and #3 are labelled Minor and, taken together, are the most serious of the three.

### 1 · `or` branches discard `className` and `count` — valid (yours)

`ModelQuery.class_name` is a required `String` and `or` is `Option<Vec<ModelQuery>>`, so every branch carries its own class and may carry its own `count`. `requires_where(alt, …, nested = true)` returns only the where-map, so `run_query` evaluates every branch against the **outer** class and the **outer** cardinality. A `{ className: "Vote" }` branch under a `{ className: "Comment" }` query silently matches comments.

The fix is the same shape you already wrote three lines up for `linkedTo` — bail when a branch carries something the flattened form cannot represent. `§7.3 multi-role composition` is exactly the case that will carry differing classes, so this will bite in the first real multi-role guard.

### 2 + 3 · Unbounded sweep × non-idempotent writes — valid, and they compound (mine, from #932)

These are pre-existing in `run_engine_proposal_pass`, not regressions from your diff, but they meet in production:

- **Scope (#3).** The auto-processor calls the interpretation path with `scope = cfg.existing_scope` — `None` unless an operator configured one (`watcher.rs:806`/`839`). The same run derives `flow_subjects` from `cursor.sources` for the *prompt context* (the J#1 fix), then hands `run_engine_proposal_pass` a `None` scope, which falls to `load_all_flow_instances` and evaluates **every live FlowInstance in the perspective**. Context is bounded; proposal generation is not. That is the same unbounded-sweep shape James raised as J#1, surviving in the sibling path.
- **Idempotency (#2).** `write_proposal` mints a fresh `Uuid::new_v4()` every call and nothing reads existing proposals — there is no dedup anywhere in the evaluator. Minting does not advance `currentState` (consensus does, separately), so a still-satisfied transition is re-proposed on every pass.

Together: the auto-processor runs indefinitely, and every run re-proposes every satisfied-but-unconsumed transition across the whole perspective. Duplicate proposals accumulate without bound. It is invisible today because nothing counts proposals yet — which is also the reason to fix it *before* the consensus writer lands, since a rule that counts proposals rather than distinct DIDs would let a single agent manufacture its own quorum.

### Merge order

Both fixes belong on `feature/flow-post-processing`, and that file is being restructured by this PR (`flow_evaluator/engine_pass.rs` → `flow_evaluator.rs`), so landing them now would hand you a conflict for no benefit. Recommendation: **merge this PR as-is** — it is approved, 8/8 green and neither finding comes from its diff — and I will land the scope + idempotency fix on the base branch immediately afterwards, before #932 goes to `dev`. Finding 1 is in your diff; happy to push it there if you'd rather not.

PR #911, opened 2026-08-22
feat: LLM tool-calling harness for interpretation passes

data-bot-coasys:
## Summary

Adds a **tool-calling harness** for interpretation passes: instead of one big single-shot LLM call that returns structured JSON, the model can iteratively (a) discover existing state via `_query`/`_list`/`_get` read tools, then (b) create new instances and relations via synthetic `<class>_create` / `<class>_propose_link_child` write tools. Writes accumulate in a per-pass buffer and are handed to the existing `apply_with_overlay` at pass end — the overlay's human-divergence gate is unchanged.

**Opt-in only.** Every existing caller of `run_interpretation*` is untouched. Set `AutoProcessorConfig.maxToolCalls > 0` (or call the new `perspective.runInterpretationWithHarness` WS-RPC method) to route through the harness. Design lives at `planning/llm-harness-design-2026-08-21-v3.md`.

## Pipeline

```mermaid
flowchart TD
    subgraph engine[run_interpretation_with_harness_and_model]
        A["prompt = system + transcript"]
        B["load SHACL classes<br/>→ Vec&lt;ClassProposeShape&gt;<br/>(includes per-relation hints)"]
        C["Ad4mMcpHandler<br/>(existing MCP tool surface)"]
        D["Ad4mToolProvider<br/>(harness_bridge)"]
        E["FilteredProvider<br/>(class-scoped reads only)"]
        F["BoundArgsProvider<br/>(auto-inject perspective_id)"]
        G["ProposeWritesProvider<br/>(+ &lt;class&gt;_create<br/>+ &lt;class&gt;_propose_link_child)"]
        H["OpenAiCompatBridge<br/>(CompletionSource → AIService)"]
        I["run_with_tools loop<br/>(≤ maxToolCalls)"]
        J["drain ProposalBuffer<br/>→ Vec&lt;InterpretationOp&gt;"]
        K["apply_with_overlay<br/>(unchanged overlay gate)"]
        L["placements returned<br/>to WS-RPC caller"]
        M{{"emit_debug_events?"}}
        N["ToolCall / ToolResult<br/>events on pubsub"]
    end

    C --> D --> E --> F --> G
    A --> I
    B --> G
    G --> I
    H <--> I
    I --> J --> K --> L
    I -.->|per tool call| M
    M -.->|yes| N
```

Per iteration of `run_with_tools`:
1. Advertise the composed provider's `tools()` to the LLM
2. LLM emits either a plain answer (loop terminates) or one or more `tool_calls[]`
[...]

PR #951, opened 2026-09-02
flow: semantic-check gate + LLM attribution, simplified (stacked on #940)

lucksus:
Stacked on #940. Companion to #950, which strips the semantic-check gate and LLM attribution out of #932. This PR brings both back here, where the harness attribution already lives, in a smaller form.

## How to review

The branch has two parts:

1. **A merge of #950** (`feature/flow-post-processing-simplify`) into #940's branch. #940 was cut from an older #932 and has not seen the `dev` merge or the module split since, so this merge is large and the "Files changed" tab is dominated by it. Once #932 (with #950) is merged into #940, this noise disappears.
2. **One commit** re-adding the moved pieces. Review that commit on its own: `rust-executor/src` only, ~730 lines including tests.

## What comes back, and how it changed

**Semantic-check gate** (`flow_semantic_check.rs`, 95 lines of code)

- `SemanticCheckLlm` is a one-method trait, `confirm(prompt) -> String`. The `model_id` parameter is gone: the AIService implementation never used it.
- `SemanticCheckVerdict` and `should_fire_proposal` collapse into `semantic_check_passed(&str) -> bool`. Fail, Ambiguous and an LLM error all behaved the same way, so there is one branch now.
- The gate lives inline in `run_engine_proposal_pass`. The `FlowContext` index that was built per pass only to fetch the flow-level hint is replaced by a lookup through the records already in hand.

**LLM attribution** (strategy path and harness path share one type)

- `LlmFlowProposal` in `interpretation/types.rs` is the single type. `LlmProposalHint` is gone; `flow_propose.rs` and `run_engine_proposal_pass` use `LlmFlowProposal` directly. The "module cycle" it was avoiding does not exist in Rust for modules within a crate.
- `parse_interpretation_output` dispatches on array vs. object and lets serde do the rest. `instances` is required in the object form, so a lone misplaced instance object is rejected without a special case. `parse_interpretation_response` is a one-line wrapper over it for the existing callers.
- `clean_llm_json` starts at the first `[` or `{` and falls back to the greedy span of that same bracket type, instead of the three-way match with eight fallbacks.
- The prompt section for `flow_proposals` is one paragraph.

**Harness wiring** is #940's, unchanged except for the type rename. #940's own `flow_propose.rs` is not reduced here; that is a separate pass.

## Tests

- `cargo test -p ad4m-executor --lib -- perspectives::flow perspectives::interpretation::parse perspectives::interpretation::prompt ai_service::harness::flow_propose`
- The e2e file gains four tests on one fixture: gate fires only on YES (YES / NO / UNCLEAR / error in one test), no hint means no LLM call, LLM proposal contributes a rationale only when it matches, and the full harness tool-call round trip ported from #940.

https://claude.ai/code/session_011SNoLB6YKTEv9dqdc3iZR1


rust-executor/src/ai_service/harness/flow_propose.rs
```diff
@@ -4,7 +4,7 @@
 //! `<class>_propose_create` / `<class>_propose_link_child` tools that
 //! accumulate SHACL-instance writes, this module gives the LLM
 //! `<flow>_propose_transition` tools that accumulate
-//! [`crate::perspectives::flow_evaluator::LlmProposalHint`]s.
+//! [`crate::perspectives::flow_evaluator::LlmFlowProposal`]s.
 //!
 //! Symmetry with `propose`:
```
data-bot-coasys, line 9: nit: The module doc references `[crate::perspectives::flow_evaluator::LlmFlowProposal]` but the type now lives in `crate::perspectives::interpretation::LlmFlowProposal` (the import at line 29 is correct). Also lines 17–18 still say `(instance_uri, to_state)` as the matching key, but the field was renamed to `instance` in this PR. Won't break anything but `cargo doc` will show a broken link.

core/src/shacl/SHACLFlow.ts
```diff
@@ -363,9 +363,18 @@ export class SHACLFlow {
    */
   public consensusRule?: ConsensusRule;
 
-  /** States in this flow */
+  /**
+   * States in this flow.
```
data-bot-coasys, line 382: nit (pre-existing, not introduced here): The `ModelQuery` JSDoc documents `$flow.uri` as a template variable, but the Rust source of truth defines `FLOW_INSTANCE_TOKEN = "$flow.instance"` and the system prompt in `prompt.rs` renders `"$flow.instance"`. No code substitutes `$flow.uri` — it would pass through unsubstituted if anyone used it. Worth aligning now that both layers are being touched.

data-bot-coasys (approved):
Reviewed both commits (`aff76ba1a` refactor + `c435486ad` re-add) and the full branch state at `c435486ad`.

**Verdict: APPROVE**

Clean simplification. The two-commit split is easy to review: commit 1 is a pure module collapse with no logic change, commit 2 re-adds the semantic-check gate and LLM attribution in a smaller form. Behaviour preservation verified: the new `SemanticCheckLlm` one-method trait + `semantic_check_passed` bool is strictly equivalent to the old verdict enum (all non-YES branches behaved identically), and both call sites thread neutral defaults (`None` / `&[]`) that reproduce the pre-feature path.

Two nits below as inline comments. Neither blocks merge.

data-bot-coasys:
Stack update: merged `feature/flow-tool-call-attribution` (post-#950 head) into this branch — merge, no cherry-picks. Two things worth eyes:

**1. The parse-chain resolution is a union, not either side** (`19c779462`). Your simplify-side chain (strict parse at the *first* bracket, fall back to that bracket type's greedy span) re-opened the prose-brace hijack that `prose_braces_before_the_real_array_dont_swallow_the_payload` locks — this branch had forked before that test landed, so its CI stayed green while carrying the regression. But the incoming array-first chain was also wrong *for this branch*: it strict-parses the first `[`, which inside a wrapper object `{"instances": [...], "flow_proposals": [...]}` is the inner instances array — silently dropping `flow_proposals`. Resolution: strict-parse at each **top-level** bracket position in order (string-aware depth scan), then the greedy fallbacks. Top-level-only is what keeps a repairable outer array (trailing commas) from being hijacked by a valid inner object. Both regression tests now pass: prose-braces and a new `wrapper_object_after_prose_keeps_flow_proposals`.

**2. The evaluator collapse had dropped proposal idempotency** (`e219f12f6`). `write_proposal` minted a fresh UUID per satisfied transition with no existence check, and minting doesn't advance `currentState` — so every later pass re-proposed everything (the CodeRabbit finding fixed on #932 as `6673de4cc`). Re-ported onto the collapsed shape: `proposal_already_exists` matches evidence_hash + instance + `to_state` (two transitions sharing identical requires guards hash identically). The e2e now asserts a re-run over unchanged evidence mints nothing — this also makes the `scope=None` sweep path (`load_all_flow_instances`) safe, so I kept your `None` arm as-is.

Everything else resolved in favour of this branch's structure. Verified: interpretation 90/0, flow 77/0, parse subset 19/19, fmt clean. CI running on `e219f12f6`.

data-bot-coasys:
**Simplification pass on top of the merge (`2708f0d08`)** — Nico asked whether more could be cut before he reviews. Net **−228 lines** on the combined diff, no behaviour change:

1. **One parser per wire shape:** the hand-rolled `_propose_transition` arg parser (`ArgError` + `parse_propose_transition_args` + 6 tests) duplicated `LlmFlowProposal`'s serde derive — the exact parser the strategy path already uses for `flow_proposals` entries. Replaced with `serde_json::from_value`. Error wording changes to serde's (still actionable; the two affected test assertions updated).
2. **One buffer implementation:** `FlowProposalBuffer` was a verbatim copy of `ProposalBuffer` including the poison-recovering `lock()`. Both are now type aliases of a generic `provider::HintBuffer<T>`; the PR #911 poison-recovery regression test moved next to the generic (`provider::tests::hint_buffer_recovers_from_poisoned_mutex`).
3. **Dead code:** `propose_transition_tool_schemas` had no production caller — `tools()` inlines its own de-dup loop — and its doc comment described de-dup semantics the real path doesn't have. Deleted; decorator doc corrected.
4. **Folded terminal-state branch** into the reachable-target check (same information in the error message, one branch fewer) and **extracted `run_flow_post_pass`** in `run.rs` so the strategy and harness paths visibly share one flow post-processing block.

**Review guidance:** the effective diff to read for #940+#951 together is `feature/flow-post-processing...feature/flow-tool-call-attribution-simplify` — 11 files, ~+1600/−540 after this pass (vs. #940's raw +4539).

Deliberately **not** cut: the dispatcher's validate-before-buffer layer (~120 lines). It re-checks what `run_engine_proposal_pass` enforces anyway, but it's what gives the LLM recoverable mid-loop error feedback instead of silent hint-dropping. Flagging in case you'd rather trade that away.

Verified: 280/0 tests green (harness + interpretation + flow filters), `cargo fmt` clean.

rust-executor/src/ai_service/harness/flow_propose.rs
data-bot-coasys: Applied in `05fd77b0e`. (The `$flow.uri` half of the second nit had already been resolved by the token unification — the engine now substitutes `$flow.uri` and `$flow.instance` as aliases, with tests; the TS doc now names both.)

core/src/shacl/SHACLFlow.ts
data-bot-coasys: Applied in `05fd77b0e`. (The `$flow.uri` half of the second nit had already been resolved by the token unification — the engine now substitutes `$flow.uri` and `$flow.instance` as aliases, with tests; the TS doc now names both.)

PR #953, opened 2026-09-03
feat: lazy perspective loading (ArcSwap + deferred init + hydration)

HexaField:
## Summary

Lazy perspective loading for multi-tenant executors. Perspectives start with in-memory SPARQL stores and hydrate on demand (first API access), so the executor reports `/health` ready before downloading large stores.

The executor binary stays identical in hosted and self-hosted multi-user deployments — the multi-tenancy detection (`db_backend=shared`) controls the deferred path.

**Stacked on** feat/wallet-backend — review that PR first.

### Per-perspective snapshot format

Replaces the monolithic `perspectives.tar.gz` with per-perspective archives + JSON manifest:

- `snapshots/{did}/manifest.json` — lists perspective UUIDs, sizes, timestamps
- `snapshots/{did}/{uuid}.tar.gz` — one archive per perspective
- Backup creates/uploads individual archives + manifest
- Restore downloads manifest first, falls back to legacy monolithic format

### ArcSwap store refactor

`PerspectiveInstance.sparql_store` changes from `Arc<SparqlStore>` to `Arc<ArcSwap<SparqlStore>>`:

- Lock-free reads via `store()` → `load_full()`
- Atomic swap on hydration
- Manual `Clone` impl — all clones share the same swap target through the outer `Arc`

### Deferred initialisation

In multi-tenant mode (`db_backend=shared`), `initialize_from_db()` creates perspectives via `new_deferred()`:
- In-memory SPARQL store (no RocksDB, no disk I/O)
- `is_hydrated() == false`
- Background tasks do NOT start

### On-demand hydration

First API access triggers `hydrate()`:
1. Download perspective archive from the remote snapshot backend
2. Open persistent OxiGraph store
3. Run migrations (rusqlite→SPARQL + named-graph→reifier)
4. Swap store via ArcSwap (lock-free, readers in flight keep old store alive)
5. Start background tasks (link language sync, neighbourhood sync, pending diffs)

### Hydration wiring

All API access paths trigger hydration before use:
- `perspectives_ws`: `get_perspective_with_access()` (~40 handlers)
- `neighbourhoods_ws`: 6 handlers
- `runtime_ws`: perspective export
- MCP tools: both auth helpers + neighbourhood publish
- MCP discovery: **skips** unhydrated perspectives (avoids hydrating all just for SHACL enumeration)

### Lazy startup

`restore_perspectives_lazy()` downloads only the manifest and caches it. Individual archives download during hydration. Executor reports healthy before any perspective fully hydrates.

### Tests

Unit tests for ArcSwap mechanics:
- `new_deferred` creates unhydrated instance
- `new()` backward compat (hydrated)
- `swap_store` atomicity + hydration flag
- Clone sharing via `Arc<ArcSwap>`
- Shape resolver Arc survival across swaps

### New dependency

`arc-swap` 1.x — lock-free atomic swap of `Arc<T>`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -61,6 +61,7 @@ version = "0.13.0-test-interpretation-2"
 dependencies = [
  "ad4m-client",
  "anyhow",
+ "arc-swap",
  "argon2 0.5.3",
  "async-trait",
  "aws-lc-sys 0.35.0",
@@ -87,6 +88,7 @@ dependencies = [
  "dirs 5.0.1",
  "env_logger 0.10.2",
  "fake",
+ "flate2",
  "fs_extra",
  "futures",
  "futures-channel",
@@ -142,6 +144,7 @@ dependencies = [
  "signal-hook",
  "sodoken",
  "sys_traits",
+ "tar",
  "tempfile",
  "thiserror 2.0.20",
  "tikv-jemalloc-ctl",
--- a/rust-executor/Cargo.toml
+++ b/rust-executor/Cargo.toml
@@ -85,6 +85,7 @@ tokio = { version = "1.25.0", features = ["full"] }
 url = "2.3.1"
 urlencoding = "2.1"
 percent-encoding = "2"
+arc-swap = "1"
 futures = "0.3.28"
 tokio-stream = { version = "0.1.12", features = ["sync"] }
 lazy_static = "1.4.0"
--- a/rust-executor/src/api/neighbourhoods_ws.rs
+++ b/rust-executor/src/api/neighbourhoods_ws.rs
@@ -95,6 +95,7 @@ async fn send_broadcast(params: Value, ctx: Arc<RequestContext>) -> Result<Value
         .map_err(|e| WsRpcError::bad_request(format!("Invalid params: {}", e)))?;
 
     let agent_context = AgentContext::from_auth_token(ctx.auth_token.clone());
+    crate::perspectives::ensure_hydrated(&uuid).await;
     let perspective_instance = get_perspective(&uuid)
         .ok_or_else(|| WsRpcError::not_found(format!("No perspective found with uuid {}", uuid)))?;
 
@@ -118,6 +119,7 @@ async fn send_signal(params: Value, ctx: Arc<RequestContext>) -> Result<Value, W
         .map_err(|e| WsRpcError::bad_request(format!("Invalid params: {}", e)))?;
 
     let agent_context = AgentContext::from_auth_token(ctx.auth_token.clone());
+    crate::perspectives::ensure_hydrated(&uuid).await;
     let perspective_instance = get_perspective(&uuid)
         .ok_or_else(|| WsRpcError::not_found(format!("No perspective found with uuid {}", uuid)))?;
 
@@ -141,6 +143,7 @@ async fn set_online_status(params: Value, ctx: Arc<RequestContext>) -> Result<Va
         .map_err(|e| WsRpcError::bad_request(format!("Invalid params: {}", e)))?;
 
     let agent_context = AgentContext::from_auth_token(ctx.auth_token.clone());
+    crate::perspectives::ensure_hydrated(&uuid).await;
     let perspective_instance = get_perspective(&uuid)
         .ok_or_else(|| WsRpcError::not_found(format!("No perspective found with uuid {}", uuid)))?;
 
@@ -160,6 +163,7 @@ async fn has_telepresence(params: Value, ctx: Arc<RequestContext>) -> Result<Val
         .map_err(|e| WsRpcError::forbidden(e))?;
 
     let uuid = params.require_str("uuid")?;
+    crate::perspectives::ensure_hydrated(&uuid).await;
     let perspective = get_perspective(&uuid)
         .ok_or_else(|| WsRpcError::not_found(format!("No perspective found with uuid {}", uuid)))?;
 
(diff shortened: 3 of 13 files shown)


data-bot-coasys (changes requested):
Reviewed head `fd9bf5278` against base `feat/wallet-backend`. All line numbers below are on that commit, and I re-read every one of these myself rather than taking a summary's word for it.

The idea is right and the deferred-store seam is clean. But lazy loading removes an invariant the backup path silently depends on, and that combination loses user data. Requesting changes on **finding 1** alone; 2 and 3 are the ones I'd want fixed before this runs multi-tenant.

---

### 1. 🔴 The periodic backup drops every unhydrated perspective from the manifest — permanently

`backup_perspectives` (`rust-executor/src/perspective_snapshot.rs:420`) builds its manifest **only** from directories that exist on disk:

```rust
let mut entries = Vec::new();                      // :433
for dir_entry in std::fs::read_dir(&persp_dir)? {  // on-disk dirs only
    ...
    entries.push(ManifestEntry { uuid, size_bytes, last_modified });
}
let manifest = SnapshotManifest { created_at: ..., perspectives: entries };  // :468
upload_json(&presign_resp.url, manifest_json)?;    // overwrites the remote manifest
```

Before this PR that was safe: an eager restore put every perspective on disk, so enumerating disk == enumerating perspectives. `restore_perspectives_lazy` (`:337`) breaks that — it downloads **no archives at all**, only caches the manifest (`CACHED_MANIFEST`, `:361`). Directories now appear one at a time, on hydration.

Failure, entirely in normal operation:

1. Hosted executor restarts. 10 perspectives, 0 directories, manifest cached with 10 entries.
2. One user opens theirs → one directory exists.
3. The periodic backup fires (`lib.rs` → `spawn_blocking(backup_perspectives)`) and uploads a manifest with **1** entry, overwriting the 10-entry one.
4. Next restart: `restore_perspective_archive` (`:372`) consults the cached manifest, finds no entry for the other 9, returns `Ok(false)`.
5. `hydrate()` logs *"No remote archive — hydrating with fresh store"* and opens an empty store. Nine users' perspectives are now empty, and step 3 has already overwritten the manifest that knew better.

The remote archives themselves survive, so this is recoverable by hand — but nothing in the running system will ever find them again.

**What makes me confident this is unintended rather than a design choice:** the PR's own test documents the opposite invariant.

```rust
// perspective_snapshot.rs:834-836
/// ... deferred perspectives have no directory yet, so they produce
/// no archive. The manifest carries forward their entry from the cached
/// manifest (the remote archive remains untouched).
```

`backup_perspectives` never reads `CACHED_MANIFEST`. The documented carry-forward is simply not implemented, and `test_backup_flush_does_not_hydrate_deferred` (`:841`) doesn't catch it because it only asserts that flush doesn't hydrate — it never calls `backup_perspectives` or inspects a manifest.

Fix is small and matches the doc: merge `CACHED_MANIFEST` entries for UUIDs with no local directory into `entries` before upload. Worth a test that asserts manifest length is preserved across a backup with one hydrated and one deferred perspective — that test would have failed here.

### 2. 🟠 `hydrate()` has no mutual exclusion — concurrent callers can corrupt the store

`perspective_instance.rs:678`:

```rust
pub async fn hydrate(&self) {
    if self.is_hydrated() { return; }   // :679 — check, then act, with nothing in between
```

`ensure_hydrated` (`perspectives/mod.rs:341`) adds no guard either, and it is reachable from ~12 handler entry points (`neighbourhoods_ws.rs` ×6, `runtime_ws.rs:133`, `mcp/tools/mod.rs:427,472`, `mcp/tools/neighbourhoods.rs:262`), plus `perspectives_ws.rs:117` calling `hydrate()` directly. Two requests for a cold perspective arriving together is ordinary traffic, not a corner case.

Both observe `hydrated == false` and proceed. Each calls `restore_perspective_archive`, which extracts a tar into `{app_data}/perspectives/{uuid}/sparql_store/`. The second extraction lands **on top of** the RocksDB directory the first has already opened, rewriting `CURRENT`/`MANIFEST-*`/SST files underneath a live store.

A per-uuid async mutex (or a `hydrating` flag other callers await on) is the smallest thing that closes this. It also removes the duplicate download, which is the expensive half.

### 3. 🟠 A failed hydration returns normally and the caller reads an empty store as success

`perspective_instance.rs:715-723`:

```rust
Err(e) => {
    log::error!("Failed to open persistent SPARQL store for {}: {} — staying in-memory", ...);
    return;      // hydrate() -> (), so no caller can know
}
```

`hydrate()` returns `()`. `ensure_hydrated` returns `()`. So after a disk-full, a permissions error, or the RocksDB lock contention that finding 2 makes reachable, the handler proceeds against the empty `new_deferred` in-memory store and `perspective.queryLinks` returns `[]` **with a success status**.

The client cannot distinguish "this perspective is empty" from "loading it failed", which is the worse of the two failure modes — a UI will render an empty perspective, and a client that writes against that view can propagate the emptiness back.

`hydrate()` should return `Result<(), AnyError>` and handlers should surface the error. The `Arc::try_unwrap` fallback below it has the same shape: it opens a *second* store at the same path (which is exactly what the RocksDB lock forbids) and then returns unhydrated.

### 4. 🟡 Blocking network + disk I/O on a Tokio worker thread

`hydrate()` is `async`, but calls the synchronous `restore_perspective_archive` inline, and that path uses `reqwest::blocking::Client` (`perspective_snapshot.rs:220,272`) plus tar/gzip extraction plus a RocksDB open.

With N cold perspectives opened concurrently where N ≥ worker-thread count, every worker parks in a blocking `send()` and the executor stops serving anything — including health checks. That is the stall this PR exists to prevent, relocated from startup to first-access.

`spawn_periodic_backup` already does this correctly for the same class of work — `spawn_blocking`. Same treatment here.

---

### What I checked and found clean

- `ArcSwap` store swap itself: the load/store discipline is fine, and `new_deferred` → swap is a single publish with no read-modify-write window.
- `flush_all_stores` correctly skips unhydrated perspectives; the test for that is real and passes on its own terms.

### Note on my method

Findings 1-4 are verified by reading `fd9bf5278` directly — the manifest code path, the doc-comment contradiction, the missing guard, the `()` return, and the `reqwest::blocking` call sites. I did **not** run a multi-tenant executor to observe the data loss in finding 1; the mechanism is read from source and the failure sequence is reasoned, not measured. If you have a hosted run that contradicts it, that beats my reading.

PR #961, opened 2026-09-03
feat(agent): PR8 — encryption keyring with versioned DEKs

HexaField:
### What

This PR adds an encryption keyring with versioned data-encryption keys (DEKs) wrapped to X25519 recipients resolved from their KELs.

### Why

Three consumers need encrypted key material: hosted-executor at-rest encryption, server-link-language room keys, and DMs. All three share the same pattern — wrap a symmetric key to a set of public keys, re-wrap on membership changes, and let leavers keep history but not future messages.

### How

**Mechanism:**

```
 mint_version(recipients)
   │
   ▼ generate DEK (32 bytes, ChaCha20Poly1305)
   │
   ▼ for each recipient:
   │   ephemeral X25519 secret → DH(ephemeral, recipient_pubkey)
   │   sym_key = SHA-256(shared_secret)
   │   ciphertext = ChaCha20Poly1305(sym_key, zero_nonce, DEK)
   │   → WrappedKey { recipient, ephemeral_pubkey, ciphertext }
   │
   ▼ DekVersion { version, wrapped[], created_at_seq }

 unwrap(version, my_secret)
   │ DH(my_secret, ephemeral_pubkey) → same sym_key → decrypt DEK
```

**Contracts:**

```rust
struct WrappedKey {
    recipient: String,           // DID or key_id
    encryption_key: String,      // X25519 pubkey targeted
    ciphertext: Vec<u8>,         // encrypted DEK
    ephemeral_pubkey: [u8; 32],  // one per wrap
}

struct DekVersion { version: u32, wrapped: Vec<WrappedKey>, created_at_seq: u64 }
struct Dek { key: [u8; 32] }   // zeroed on drop

struct VersionedKeyring {
    fn mint_version(recipients, kel_seq) -> Result<(u32, Dek)>
    fn unwrap(version, recipient_secret) -> Result<Dek>
    fn re_wrap(version, old_secret, new_recipients) -> Result<u32>
}
```

**Zero-nonce safety:** Each ephemeral key produces a unique shared secret, so nonce reuse across wraps never happens.

**Forward secrecy on leave:** A departing member retains versions they held — they decrypt history up to exit. New versions wrap only to current members. No re-encryption of old data needed.

| File | Lines |
|------|-------|
| `keyring.rs` | +551 — DEK generation, wrap/unwrap, VersionedKeyring, tests |
| `Cargo.toml` | +2 deps (x25519-dalek, chacha20poly1305) |


rust-executor/src/agent/keyring.rs
```diff
@@ -0,0 +1,551 @@
...
+    recipient_pubkey: &[u8; 32],
+) -> Result<(Vec<u8>, [u8; 32]), KeyringError> {
+    let ephemeral_secret = StaticSecret::random_from_rng(rand::thread_rng());
+    let ephemeral_pubkey = PublicKey::from(&ephemeral_secret);
+
+    let recipient_pk = PublicKey::from(*recipient_pubkey);
+    let shared_secret = ephemeral_secret.diffie_hellman(&recipient_pk);
+
+    // Derive a symmetric key from the shared secret via SHA-256.
+    let sym_key = Sha256::digest(shared_secret.as_bytes());
```
data-bot-coasys, line 133: **Protocol drift vs #930 — wire-incompatible KDF.**

This uses `SHA-256(shared_secret)` directly as the symmetric key. The server-link-language keyring in #930 uses `HKDF-SHA256(shared_secret, salt=eph||recip, info="adam-server-link-language:room-key-seal:v1")` — a proper domain-separated KDF.

These produce different symmetric keys from the same DH output, so wrapped keys from one implementation can't be unwrapped by the other. The design doc D3 asks for "one keyring implementation, three consumers" — if the Rust crate is that implementation, it should match #930's KDF so the JS side can be replaced.

Suggest: switch to HKDF (the `hkdf` crate is already in the dep tree via other paths) with domain-separated info, e.g. `"ad4m:keyring:dek-wrap:v1"`. Also consider matching #930's random nonce + AES-256-GCM to avoid a second divergence, or document why ChaCha20 is preferred here (performance on non-AES-NI targets is the usual argument — fine, but state it).

rust-executor/src/agent/keyring.rs
```diff
@@ -0,0 +1,551 @@
...
+    let shared_secret = recipient_secret.diffie_hellman(&ephemeral_pk);
+
+    let sym_key = Sha256::digest(shared_secret.as_bytes());
+    let cipher =
+        ChaCha20Poly1305::new_from_slice(&sym_key).map_err(|e| KeyringError::CryptoError(e.to_string()))?;
+
+    let nonce = Nonce::default();
+    let plaintext = cipher
+        .decrypt(&nonce, wrapped.ciphertext.as_ref())
+        .map_err(|_| KeyringError::UnwrapFailed)?;
```
data-bot-coasys, line 162: **Plaintext DEK bytes not zeroized.**

`cipher.decrypt(...)` returns a `Vec<u8>` holding the raw DEK bytes. After `key.copy_from_slice(&plaintext)`, that `Vec` is dropped without zeroization — the 32 DEK bytes remain in freed heap memory until the allocator reuses the page.

The `Dek` struct correctly zeroizes on drop, but this intermediate `Vec<u8>` undoes part of that guarantee. In a hosted multi-tenant executor (the target scenario), freed-but-not-zeroed memory is the attack surface at-rest encryption is meant to close.

Fix: wrap the decrypt result in `zeroize::Zeroizing<Vec<u8>>`:
```rust
let plaintext = zeroize::Zeroizing::new(
    cipher.decrypt(&nonce, wrapped.ciphertext.as_ref())
        .map_err(|_| KeyringError::UnwrapFailed)?
);
```
`Zeroizing<Vec<u8>>` calls `zeroize()` on the Vec's backing buffer before deallocation.

rust-executor/src/agent/keyring.rs
```diff
@@ -0,0 +1,551 @@
...
+            created_at_seq: at_seq,
+        });
+
+        Ok(self.versions.last().unwrap())
+    }
+
+    /// Wrap the current DEK to an additional recipient WITHOUT minting a new
+    /// version. Used when enrolling a new device — the DEK stays the same,
+    /// the new device just gets a copy.
+    pub fn wrap_to_additional(
```
data-bot-coasys, line 255: **History visibility on join — `wrap_to_additional` only covers the latest version.**

This method wraps the current DEK to a new recipient (device enrollment). But for the room-key consumer, a new *member* joining an existing space needs old DEK versions to decrypt message history.

In #930's link-server, `rotateRoomKey` has a `MemberKeyGap` mechanism that detects which members are missing historical versions and returns that as structured data for the admin/caller to act on. This keyring has no equivalent — old versions are stored but there's no API to wrap a specific historical version to a new recipient.

This is also the implicit "history visibility on join" decision point from the design doc. Two options:
1. **Deliberate omission (document it):** New members only decrypt from their join version onward. Add a doc comment stating this is the intended policy.
2. **Add `wrap_version_to_additional(version: u32, ...)`:** Let callers selectively grant old versions, enabling per-space "share history" vs "forward-only" policy at a higher layer.

Either way, pin down the decision explicitly rather than leaving it implicit in the API shape.

rust-executor/src/agent/keyring.rs
```diff
@@ -0,0 +1,551 @@
...
+        unwrap_secret: &StaticSecret,
+        at_seq: u64,
+    ) -> Result<&DekVersion, KeyringError> {
+        if recipients.is_empty() {
+            return Err(KeyringError::EmptyRecipients);
+        }
+
+        // Unwrap the current DEK.
+        let current = self.versions.last().ok_or(KeyringError::VersionNotFound(0))?;
+        let _old_dek = unwrap_from_version(current, unwrap_secret)?;
```
data-bot-coasys, line 293: **`rewrap` mints a fresh DEK but the name suggests re-wrapping the same key.**

The function name `rewrap` reads as "wrap the existing key to new recipients" — but it actually generates a *new* DEK (line 298 calls `self.mint()`), giving forward secrecy on revocation. The `_old_dek` unwrap on line 293 is purely an authorization check.

This is the right behavior — but the name should match. Consider `rotate` or `mint_excluding` to signal "new key material." The `_old_dek` variable naming helps, but callers reading the public API won't see it.

Minor: the authorization-via-unwrap pattern means `rewrap` requires the caller to be a current member. This is correct for the revocation case, but worth a doc note — it means only a remaining member can trigger rotation, not an external admin.

rust-executor/src/agent/keyring.rs
```diff
@@ -0,0 +1,551 @@
...
+        assert_eq!(recipients[0].encryption_key, hex::encode(pk.to_bytes()));
+    }
+
+    #[test]
+    fn dek_zeroed_on_drop() {
+        let dek = generate_dek();
+        // Verify the DEK has content (not all zeros).
+        assert!(dek.key.iter().any(|&b| b != 0));
+        // Drop happens automatically; zeroize ensures the key gets cleared.
+        // We can't test post-drop memory, but the Drop impl calls zeroize.
```
data-bot-coasys, line 549: **Test has no effective assertion about zeroization.**

This test generates a DEK, confirms it's non-zero, then drops it. The comment correctly says "we can't test post-drop memory" — but then the test passes identically with or without the `Zeroize` impl on `Dek`. An always-passing test is worse than no test (it implies coverage that doesn't exist).

Two options:
1. **Delete the test** and rely on the `Drop` impl being visually obvious (it's 3 lines).
2. **Use unsafe to verify** — allocate the `Dek` on the heap, grab a raw pointer to the key bytes, drop it, then read through the pointer. This is UB in the strict sense but is the standard pattern for zeroize verification in crypto crates (e.g., `zeroize`'s own test suite does this). If you go this route, gate it behind `#[cfg(not(miri))]`.

data-bot-coasys (changes requested):
## Review: PR8 — encryption keyring with versioned DEKs

**Verdict: REQUEST CHANGES** — the keyring design is solid and the API shape is right for the three-consumer vision (at-rest, room keys, DMs). But the wrapping protocol diverges from #930's link-server keyring in ways that make "one implementation, three consumers" impossible without a follow-up alignment pass. Two items need resolving before merge.

### Blocking

1. **Wire-incompatible KDF with #930.** This uses raw `SHA-256(DH_shared_secret)` + ChaCha20Poly1305 + zero nonce. The server-link-language in #930 uses `HKDF-SHA256(shared_secret, salt=eph||recip, info=domain_tag)` + AES-256-GCM + random nonce. Different KDF + different AEAD = two implementations that can never interop. The design doc D3 says "extract the keyring versioning from the link-language PR into a shared crate" — this PR creates a parallel, incompatible implementation instead. **Fix: converge on HKDF with domain separation.** AEAD choice (ChaCha vs AES-GCM) is less critical but should be documented.

2. **Intermediate plaintext not zeroized.** `unwrap_with_secret` decrypts into a `Vec<u8>`, copies to the `Dek` struct, and drops the Vec without clearing it. In the hosted multi-tenant scenario, freed-but-not-zeroed memory is exactly what at-rest encryption should prevent. **Fix: `zeroize::Zeroizing<Vec<u8>>`** — one-line change.

### Non-blocking (address or document)

3. **History visibility on join is implicit.** `wrap_to_additional` only wraps the *latest* version. No API grants old versions to new members. #930 has `MemberKeyGap` detection for this. Either add `wrap_version_to_additional(version, ...)` or document that forward-only is the deliberate policy.

4. **`rewrap` naming.** It mints a *fresh* DEK (forward secrecy), not re-wrapping the same key. Consider `rotate` to match #930's `rotateRoomKey`.

5. **`dek_zeroed_on_drop` test is assertion-free.** Passes with or without the `Zeroize` impl. Either verify via unsafe post-drop read or remove the test.

### What's good

- Clean, generic API — `VersionedKeyring` is genuinely consumer-agnostic. It doesn't assume perspective storage or room semantics.
- `Dek` zeroization via `Drop` is correct.
- Ephemeral-per-wrap (one ephemeral ECDH per recipient) gives forward secrecy on key compromise.
- `recipients_from_keys` bridges cleanly to the KEL's `KeyEntry` struct.
- Test coverage of the core flows (round-trip, enrol, revoke, leaver-history, error paths) is thorough.
- Module-level "honest limit" documentation is appreciated — the design doc's transparency ethic is carried through.

### Drift summary vs #930

| Aspect | This PR (#961, Rust) | #930 (TS, link-server) |
|---|---|---|
| AEAD | ChaCha20Poly1305 | AES-256-GCM |
| KDF | SHA-256(shared_secret) | HKDF-SHA256 + domain tag |
| Nonce | Zero (ephemeral key ensures uniqueness) | Random 12-byte |
| History grant | No API (implicit forward-only) | `MemberKeyGap` detection |
| Rotation name | `rewrap` | `rotateRoomKey` |
| Key source | `rand::thread_rng()` | `crypto.getRandomValues()` |

The versioning model (monotonic version numbers, per-member wrapping, old versions retained) is consistent between both — the structural design is aligned. The divergence is in the wrapping-layer crypto.

### Testing statement

- **CI:** `build-and-test` reports FAILURE, but the same check also fails on the base branch (`feat/agent-identity-07-social-recovery`) — pre-existing, not introduced by this PR. `plugin-ad4m-tests` passes.
- **Local:** Attempted `cargo test --release -p ad4m-executor --lib agent::keyring` in a fresh worktree. The keyring module compiled successfully, but the test binary could not link due to missing `CUSTOM_DENO_SNAPSHOT.bin` (requires a full `pnpm build` that's impractical in a review worktree). I did **not** run the 9 unit tests — I reviewed them by reading. They cover the central claims (round-trip, enrol-without-version, revoke-excludes, leaver-history, error paths).
- I did **not** verify CI green — `build-and-test` is red (pre-existing).

PR #970, opened 2026-09-04
docs: rust-executor refactoring spec and AGENTS.md module guides

lucksus:
## Summary

Docs-only. Adds the rust-executor refactoring spec and per-directory agent
guides so humans and coding agents can find the right file, and the
invariants, without reading 7k-line sources.

No runtime or test changes.

## Spec

`planning/rust-executor-refactoring-spec-2026-09-04.md`

Target layout and ordered work items for splitting the oversized modules
(`perspective_instance`, `db`, languages, AI, WS/MCP, Prolog isolation,
`agentic/` cycle break, `AppContext`). Includes conventions, open decisions
with defaults, and crate-level done criteria.

This PR does not start those items.

## Agent docs

Canonical instructions live in `AGENTS.md`. Sibling `CLAUDE.md` is only
`@AGENTS.md`, so Claude Code, Codex, Copilot, Cursor, and Grok load the
same text.

- Root `AGENTS.md`: package map, repo rules, Holochain/test gotchas
- `rust-executor/AGENTS.md`: crate map, build/test, do/don't
- Per-module guides: `api`, `mcp`, `perspectives`, `ai_service`,
  `prolog_service`, `languages`, `js_core`, `holochain_service`


data-bot-coasys:
# Refactoring spec — three-pass synthesis

2026-09-04, late. Three independent passes over `planning/rust-executor-refactoring-spec-2026-09-04.md` (PR #970): Marvin (Grok) on architecture, Lal on the English layer, Data (Fable 5) on both plus this synthesis. Each pass was run against the real tree at the branch head, not the spec's description of it. Where we disagree with each other I say so; where all three converge I say that too, because convergence from three different models reading independently is itself evidence.

## TL;DR

1. **The spec's mechanical programme is right and should run.** All three passes independently protected the same items: §0 ground rules, item 1 (Prolog — a bugfix wearing a refactor ticket), item 2 (MCP dedup fixes two live bugs), 4a (schema versioning), 6 (CapSpec), 11 (AppContext with the `--test-threads=1` success metric), §4 "do not freelance" defaults.
2. **Item 7 (`agentic/`) is the one wrong call.** It groups harness + interpretation + flows + auto-processor by *"uses an LLM"* — but flows are a graph state machine, interpretation is a text→graph pipeline, auto-processor is a scheduler. They share a **host**, not a domain. Keep item 7's cycle-break graph (forbidden edges, `ToolProvider` in `ai_service/tools.rs` — that analysis is correct); drop the folder.
3. **The substantial architecture shift Nico asked for exists and is cheap to start: *Perspective as host, aspects as plugins*.** The internal analogue of Languages. Flows are the worked example, and the seam is already 90% proven by tonight's own PRs. Details below, including the part I think Marvin's version undersells: aspects are **SDNA-activated**, which is what makes them social-DNA modules rather than just Rust plugins.
4. **The English layer gets one merged contract template** (Lal's structure × Marvin's contract fields), and the drift Lal found tonight (`languages/AGENTS.md` vs `js_core/AGENTS.md` disagreeing about op2 counts, *on an unmerged branch*) is the proof the docs need the same de-duplication discipline as the code.
5. **First slice pays this week without touching the flow stack:** contract-template AGENTS.md rewrite, item 1, `FlowStore`/`SemanticCheck` traits in place with no directory moves.

---

## 1. Host + Aspects — endorsed, with three modifications

Marvin's core proposal (his review, §1–2): `PerspectiveInstance` *hosts registered aspects* instead of containing them. Each aspect is a holon — inbound trait (what the host calls: `on_created`, `on_interpreted`, `accept`, `tick`), outbound trait (what the aspect needs: `FlowStore`, `SemanticCheck`), an English contract, and a test surface that mocks the outbound trait. The `interpretation/run.rs` hook (today: a direct call to `flow_evaluator::run_engine_proposal_pass` at `run.rs:867`) becomes a typed inbound call, which turns the #940 round-2 bug class — threading the wrong subject list from another module — into a compile error.

I verified his load-bearing claims against the tree and they hold. Three modifications:

### 1a. Aspects are SDNA-activated — this is the AD4M-native half of the pattern

Marvin says aspects are "the internal analogue of Languages" and moves on. The analogy is deeper and it changes the design: **a Language is code addressed by data** (a language address in a perspective/neighbourhood handle activates it). An aspect should be the same: *code activated by the presence of its social DNA in the graph*. A perspective with no `SHACLFlow` definitions never runs a spawn pass, never pays flow's cost, never needs its invariants. `flow_classes.rs` (hard-wired SDNA) is already the activation key — we just don't treat it as one.

Concretely: the registry holds all compiled-in aspects; the host consults `aspect.activated_by()` (a set of class/predicate URIs) against the perspective's SDNA before dispatching events. This buys:

- **Flows + ontologies become literally shareable social-DNA modules** — join a neighbourhood whose SDNA declares flows and the flow aspect wakes up. That's the sentence from the flows vision doc, made mechanical.
- The composability Nico asked for ("modules that are almost holons that we can easily extend by just changing in the scope of that module") gets a *runtime* meaning, not just a source-layout meaning.
- A horizon (not this quarter): aspects behind a stable trait pair are the shape you'd need if some ever became WASM plugins, the way Josh's Living Web specs treat sync modules. Nothing in the first slice commits to this; nothing forecloses it.

### 1b. Rust privacy is the primary enforcement; the dependency test is the backstop

The spec (item 7.6) and Marvin both lean on a `#[cfg(test)]` module-dependency test for the forbidden edges. Good backstop — but Rust already has a stronger, always-on mechanism: **visibility**. If an aspect module exports *only* (a) its inbound trait impl registration and (b) a `types.rs` of wire-shape types, and everything else is `pub(self)`/private, then `mcp` physically cannot import `flow_consensus` internals; the compiler is the CI.

This also resolves a real gap in Marvin's design that tonight's #968 exposes: WS/MCP/TS surfaces need `FireOutcome` and proposal types. If *all* imports of the aspect are forbidden, the types have nowhere to live except a global types crate (churn). Rule instead: **types-only imports are allowed edges; behavior imports are not** — and privacy enforces the split mechanically because behavior simply isn't `pub`. The dependency test then only needs to check the coarse crate-level edges privacy can't express (e.g. `ai_service` must not depend on any aspect).

### 1c. Trait home and registry sequencing

`GraphHost` + `PerspectiveAspect` need a home neither side depends on: a small `perspectives/host.rs` (traits only, no logic) now; the registry lives on `AppContext` **once item 11 exists**, and until then a `Vec<Arc<dyn PerspectiveAspect>>` on the instance is fine. Marvin's warning stands and I'll sharpen it: **`GraphHost` must be narrower than `PerspectiveInstance` or nothing was encapsulated — you've renamed `self`.** The forcing function: define `GraphHost` as exactly the surface `flow_evaluator`/`flow_spawn` already use today (get_links / add_link / model_query / did / uuid), nothing more, and let it grow only when a second aspect demands it. Inbound events stay coarse (`on_created`, `on_interpreted`, `accept`, `tick`) — never per-link, or the hot write path pays for every registered aspect.

### The holon acceptance test (for review, and for agents)

A one-line rule that makes "is this a holon" checkable in review, which is what the agentic-coding angle needs: **a PR that adds a flow feature touches only `flow/` files; touching a second aspect requires a stated reason in the PR description.** Same rule per aspect. When that rule holds for a month, the directory move (wherever the team wants the folders) is mechanical and boring — which is exactly when moves should happen.

## 2. The English layer — one merged template

Lal and Marvin converged on "AGENTS.md files are file inventories, not contracts" from opposite directions (she measured reader search-cost across six docs with five different section vocabularies; he asked what an agent loading ONE subtree needs). Their templates overlap ~70%. The merge, ordered by what a reader must *decide*:

```markdown
# <module>/ — contract

<One sentence: what this IS. The existing docs nail this — keep the bar.>

## Boundary
Owns / does NOT own (and where that lives instead). Cross-links to the
docs that must be read together. A Boundary line that reaches into a
sibling's internals is only legal if the coupling is permanent and mutual —
if a spec item plans to remove it, the fact belongs in Status.

## Map
| File/dir | Role |   ← table only, one present-tense line each

## Contract
Inbound: who calls us, with what (names, not vibes).
Outbound: what we call, with what. "Does not import X, Y, Z."
One mermaid: inbound → this → outbound. Not a second essay.

## Invariants
Numbered. Fail-closed behaviors named. Lock order if any.

## What breaks if you change X
The three most expensive mistakes, from real history where possible.

## Tests
Central-claim test names. A missing one of these is a regression.

## Status
One glossed line per active spec item ("item 3 — splits X into … see
planning/<spec>.md#item-3"), or "Stable — no split plan."
```

Both of their filled-in examples (Marvin's `flow/`, Lal's `perspectives/`) already fit this merged shape with minor reslotting, and both are worth landing nearly as-is. Lal's riders travel with it: over-40-line docs self-declare their tier in Status; earned code samples live under Map as a "Shape" block.

Two of Lal's findings deserve to be spec text, not review commentary:

- **The drift is already real.** `languages/AGENTS.md` carries a full duplicate `## js_core/` section that has *already diverged* from `js_core/AGENTS.md` (op2 counts present in one copy, absent in the twin) — on a branch that hasn't merged. The English layer has the identical two-places disease item 8 is fixing in the Rust. Fix: one-line cross-link, delete the duplicate section.
- **Item numbers are permanent.** Append, never renumber; supersede rather than reuse. Otherwise every "spec item 7" citation in a module doc silently dangles the day the spec is revised.

On the name: Lal pushed back on `agentic/` as industry-trend vocabulary ("every other directory name describes a mechanism or a data shape; this one describes a category of hype"), Marvin independently said "if they need a parent later, name it `aspects/` after the host relationship." Since the folder move is deferred anyway, the naming decision is free to defer with it — recorded as D6 below.

## 3. Verifiability — what CI enforces vs what review owns

Merged from Marvin's table and Lal's three buildable lints. The principle (Marvin's, and it matches the retro's mental-model-tiers framing): **CI checks structure; review checks prose. Never ask CI to judge English.**

| Claim | Enforced by |
|---|---|
| Every `src/<dir>/` has `AGENTS.md` + `CLAUDE.md`, and `CLAUDE.md` == `@AGENTS.md` byte-exact | walk + compare |
| Template shape (headings present, in order) | heading-schema lint — prerequisite for the rest to parse |
| No duplicated doc content across siblings | shingled-overlap lint above a threshold → "X% shared with Y, check for drift" — *would have caught tonight's languages/js_core drift mechanically, before merge* |
| Every "item N" citation resolves to a heading in the spec | dangling-ref grep |
| Central-claim tests named in AGENTS.md exist | `cargo test -- --list` contains them |
| File caps (800/300) | `wc`-class check, tests excluded |
| Forbidden coarse edges | privacy first (§1b), `#[cfg(test)]` dependency test as backstop |
| Aspect registry ↔ tree ↔ GraphCoder node list | registry introspection vs concept-map hash — this is the Josh tie-in from the retro: GraphCoder hashes the nodes, AGENTS.md is the English snapshot of one node, CI checks they agree |
| `#[ts(export)]` types regenerated | fail CI if `core/src/generated/api/` dirty |

All the doc lints fit one `scripts/lint-agent-docs` script — small enough for week 1, and per Lal it's a fair sales pitch on its own ("this would have caught the bug we just found by hand").

## 4. Sequencing — first slice that pays, and what to defer

All three passes flagged the same hazard: week 4's "move everything into `agentic/`" collides with the in-flight flow stack, and combining it with the `perspective_instance.rs` split (3b) in the same fortnight is the same class of collapse that ate three behaviors in #940 this afternoon. Revised schedule:

| Week | Items |
|---|---|
| 1 | Contract-template AGENTS.md rewrite (in place, no moves) · item 1 Prolog · item 2 MCP · `FlowStore` + `SemanticCheck` traits in place · `scripts/lint-agent-docs` |
| 2 | 5 types · 10 holochain · 4a schema · 3a prep |
| 3 | 3b split · 4b db · 6 CapSpec |
| 4+ | 8 query-stack · 9 services · 11 AppContext · aspect registry once 11 exists |
| later | directory moves, which the traits will have made mechanical |

Why the traits cost so little now: `SemanticCheck` already effectively exists (`CannedLlm` vs `AIServiceSemanticCheck` is tonight's #940 test seam), and `FlowStore` is just naming the surface flow code already uses. "Traits in place, no moves" doesn't conflict with #967/#968 as long as they're thin.

## 5. Design docs vs module docs — the tense rule

Lal's rule, worth adopting verbatim as spec text: **a present-tense structural fact may live only in the module doc; the design doc indexes and explains why, and never restates a landed shape.** A design doc is allowed to be wrong about the future without being wrong about anything — it's a plan. A module doc that describes something not in the tree today is a bug. Enforcement starts free (PR-checklist line: "module AGENTS.md updated to match landed structure" in every item's acceptance criteria) and graduates to the dup-content lint.

This is also the answer to "we should use more design documents": yes — one living design doc per aspect in `planning/`, dated, holding the *why* and the decision history, cross-linked from the aspect's Status section. The flow work already proved the pattern; the rule above is what keeps those docs true after the code lands instead of rotting into `PHASE1_*`.

## 6. Concrete edits requested on #970

1. Item 7: keep steps 1–2 and the edge/trait analysis (steps 5–6); **replace the `agentic/` directory move (steps 3–4, 7) with "Host + Aspects"** as sketched in §1 — traits in place first, moves deferred and renamed D6.
2. §5.5 + ground rule 5: replace the "≤40 lines: purpose, files, entry points…" description with the merged contract template (§2); add the item-numbers-are-permanent rule and the tense rule (§5).
3. §2 target layout: mark the tree explicitly as **target, not current** (Lal: it's a trap for an agent that jumps to §2 for "where does X live" and gets tomorrow's answer), and replace the `agentic/` subtree with `flow/`, `interpretation/`, `auto_processor/` as sibling aspects plus `perspectives/host.rs`.
4. §6 schedule: swap in the revised table (§4).
5. New CI item (S): `scripts/lint-agent-docs` with the checks in §3.
6. Fix the `languages/AGENTS.md` duplicate `## js_core/` section on this PR (it's this PR's own files).
7. §4 new decisions: **D6** — parent folder name if aspects ever get one (`aspects/`? nothing?): defer until after traits exist. **D8** — types-only import edges: adopt the §1b rule (privacy-enforced, types `pub`, behavior not). *(D7 — registry location — resolves itself: `AppContext`, per item 11.)*

## Attribution

Marvin: the Host+Aspects inversion, the holon test worked on flows, the interpretation-hook removal proof, verifiability table, sequencing hazard. Lal: the drift finding, the reader-cost analysis, the template's decision-ordering, the tense rule, the three buildable lints, the naming pushback. Data: SDNA activation (§1a), privacy-as-enforcement + types-only edges (§1b), trait home/narrowness forcing function (§1c), the holon acceptance test, the merge itself. Full source docs: Marvin's `ad4m-refactor-spec-review-2026-09-04.md`, Lal's `ad4m-refactor-english-layer-2026-09-04.md` — both worth reading whole; this synthesis compresses but doesn't replace them.

data-bot-coasys:
Per Nico's ask, the aspect design has moved out of comments into the PR itself (`7c8992021`):

- **New:** [`planning/aspect-holons-refactor-2026-09-04.md`](https://github.com/coasys/ad4m/blob/refactor/rust-executor-spec-and-agent-docs/planning/aspect-holons-refactor-2026-09-04.md) — phase-2 design: three rings (host / aspects / capability services), the `GraphHost`/`PerspectiveAspect` traits, wire surfaces (`WireContribution` — transport stays in `api/`+`mcp/`, surface definitions belong to aspects), Rust↔TS spanning via the `generate:api-types` codegen, SDNA activation, and two-phase sequencing with explicit gates. Answers the open questions: API service in/out, WebSocket handlers, client-side spanning, whether `Ad4mModel`/`ModelQuery` is an aspect (it isn't — it's host).
- **Spec edits (minimal):** item 7's `agentic/` directory move superseded (cycle break steps 1–2/5–6 stand), §2 marked target-not-current, D6–D10 pointer.
- **Drift fix:** the duplicated `## js_core/` section between `languages/AGENTS.md` and `js_core/AGENTS.md` is de-duplicated (single source now `js_core/`, with the op counts).

The synthesis comment above remains the review record; the doc is the design of record.

data-bot-coasys:
**Aspect holons doc rev 2** pushed (`04c240ea8`), following Nico's feedback that the plugin system was too narrow with `model_query` declared host-by-fiat.

Core change: **being an aspect** (module shape, contract, dependency discipline) is now separate from **being optional** (an activation policy: `always` / `sdna` / `config`). The host shrinks to a kernel (links, commit pipeline, sparql index, SDNA storage, registry), and **model, subscriptions, and notifications join flow/interpretation/auto_processor as aspects** — six total, in a declared dependency DAG (`requires()`, topo-ordered registry init, `ModelView` as model's exported trait).

Also new: §6 on the Languages symmetry (executor level = same pattern one level up; extending it there is deferred, D11), wave-based sequencing (flows as template → notifications/subscriptions → model last), D9 resolved **yes**, D12 keeps `sparql_store` in the kernel since the commit path writes it.

Doc: [`planning/aspect-holons-refactor-2026-09-04.md`](https://github.com/coasys/ad4m/blob/refactor/rust-executor-spec-and-agent-docs/planning/aspect-holons-refactor-2026-09-04.md)

PR #971, opened 2026-09-05
fix(plugins/ad4m): declare contracts.tools for static tools

data-bot-coasys:
## Summary

- Declares `contracts.tools` for **all 53** tools the AD4M plugin registers with `api.registerTool()`:
  - **8 statically-named** plugin-managed tools (`ad4m_get_sample_config`, `ad4m_refresh_ad4m_tools`, `ad4m_subscribe_to_mentions`, etc.)
  - **45 executor-provided** tools discovered via MCP `tools/list` and re-registered with an `ad4m_` prefix (verified against the live launcher on `marvin.fritz.box:3001` on 2026-09-05: `ad4m_signup`, `ad4m_login_email`, `ad4m_neighbourhood_join_from_url`, `ad4m_add_link`, …).
- Without this, OpenClaw's plugin loader silently no-ops every `api.registerTool()` call with `"plugin must declare contracts.tools for: <name>"` and the plugin exposes zero actionable tools to the agent — despite looking healthy in logs.

## Symptom in the wild

Three bots on the openclaw AD4M plugin (Data, Lal, Marvin) all pointing at the shared launcher on `marvin.fritz.box:12100` — all three see no `ad4m_*` tools in their toolsets. All plugin logs clean, no auth failures on the executor side, no MCP session activity from any of them. Manifest gate is the shared cause.

## Reproduction & verification

- `openclaw plugins doctor` reports `"plugin must declare contracts.tools before registering agent tool <name>"` for each tool.
- After this patch: doctor clean; `openclaw plugins enable ad4m --accept-capabilities` succeeds; capability re-consent is the standard one-time step when the manifest surface widens.
- Root cause identified by Lal in independent investigation; initial 8-static-tool patch drafted by her, widened here to include the 45 executor-provided tools after verifying they follow the same registration path (`plugins/ad4m/index.ts:291`).

## Follow-up (not in this PR — separate design discussion)

Per-perspective SHACL-derived tools (`ad4m_Message_create`, `ad4m_Task_add_child`, etc.) can't be exact-string declared because their names depend on user-added subject classes. Loader does exact-string set membership only (no glob/prefix); the matcher API used for `tools.allow` supports patterns but `contracts.tools` does not. This is an architectural gap worth discussing with OpenClaw core — probably wants a `contracts.toolPrefixes` or similar. For now the MVP (signup + neighbourhood + message exchange) lives entirely in the 53 tools declared here, so it unblocks today's three-bot chat while leaving SHACL-tool dynamic registration as a separate follow-up.

## Test plan

- [x] `openclaw plugins doctor` clean on this checkout
- [x] Plugin re-enable with `--accept-capabilities` succeeds
- [ ] After gateway restart: `ad4m_*` tools present in agent toolset (blocked on restart; verifying in fresh session next)
- [ ] End-to-end: three bots sign up via `ad4m_signup` on shared launcher and exchange messages in a neighbourhood via `ad4m_neighbourhood_join_from_url` + `ad4m_add_link`


HexaField (approved):
Nice work, feel free to also put up a PR and run this in the ad4m-wind-tunnel repo's agent integration test scenarios.

data-bot-coasys:
Closing as superseded — not abandoned.

`fix/plugin-static-instance-tools` (#972) is built directly on this branch's commit (`9dc2eff61` is its first commit), and then **rewrites** the very list this PR adds: the 54 dynamic-era entries here (`refresh_ad4m_tools`, `get_children_body_parsed`, `create_subject`, …) describe the tool surface as it existed *before* the static-tools work, and several of those tools no longer exist on the executor.

Merging this first would land a stale manifest that #972 immediately deletes. The content that stays correct ships in #972.

Review queue is now: #973 (Rust executor side) · #972 (plugin side: static surface + waker fixes + skill/docs) · #975 · #976.

PR #977, opened 2026-09-06
docs(plugins/ad4m): rewrite AD4M skill + references for the static tool surface

data-bot-coasys:
Stacked on #972 (plugin static tool surface). Docs only — no code.

## Why

The shipped AD4M skill is from the dynamic-tools era: it instructs bots to call `channel_query`, `message_create`, `get_children_body_parsed` and `set_agent_profile`, several of which no longer exist on the static surface. A bot following it today fails at step one. Tonight three agents (Data, Marvin, Lal) onboarded onto a multi-user executor and hit every gap in sequence, so this rewrite is grounded in a real cold-start rather than a read-through.

## What changed

`SKILL.md`
- Rule 0: explicit native surface inventory. Phantoms (`agent_status`, `remove_link` — declared in `contracts.tools`, absent on the executor) and non-bridged-but-real tools (`signup`, `verify_email_code`, `add_model`, `get_children_body_parsed`) are each called out with the right caveat instead of silently omitted.
- Rules 4–6: `instance_*` as the default vocabulary; the `expression_address` vs `base_uri` trap.
- Rule 8: Flux channel creation rewritten for `instance_create`.
- Rule 3c: the `mcporter` fallback now covers **authenticated** calls — a bare call carries no identity and fails with `Perspective not found` on a perspective you just created. Documents the bare-JWT `Authorization` header via `$env:` indirection (no token in argv). Verified live against mcporter 0.13.10.
- Rules 12–13: the waker. A profile is necessary but not sufficient, a success string is not proof, `list_waker_subscriptions` verification is mandatory, and the pending/30s-auto-retry behaviour from #972 is described.
- Password hygiene: `key=@path`, with `$(cat …)` explicitly named as unsafe rather than offered as a workaround.

`references/architecture.md`
- **The highest-value fix:** the PropertyShape table only had `node_kind: "IRI"` for references, which produces a plain **string** property, not a relation — `describe_perspective` reports `"type": "string"` and nothing is ever hydrated. Adds `relation_kind` (`hasOne` / `hasMany` / `belongsToOne` / `belongsToMany`) and `target_class_name`, a worked `hasOne` example, and the `describe_perspective` check that confirms it.
- The setter asymmetry: a `hasOne` without a `setter` is rejected at write time, but `describe_perspective` renders it identically to a working relation — you cannot tell it is dead until you try.
- The `class_name` must equal the local name of `target_class` rule (enforced by #976).
- Generated `{class}_*` tools reframed as the `dynamicClassTools=true` opt-in path, not the default.

`references/setup.md`
- Leads with the one-command `openclaw ad4m-setup` path — it already fully automates multi-user onboarding, and nothing documented it, so we did all of it by hand tonight.
- Step 4 (unlock) names the two misleading errors a locked wallet produces (`User key not found`, `main key not found`) and frames the lockout as designed behaviour whose complaint is the late, unclear message (addressed by #975), not the lockout itself.
- Running without `--admin-credential` documented as an intentional test-only convenience with a "never off loopback" warning.
- New troubleshooting rows: locked wallet, pending subscription.

## Provenance

Drafted by Lal after an audit of the current skill, then **cold-bot tested**: she built an Author/Book ontology from these drafts and the live tools only, without reading anyone else's schema. Five defects came out of that run; the two that made the docs unusable (relation fields, authenticated fallback) are fixed here, and the other three are reproductions of executor-side findings tracked separately (#976 covers the `class_name` one).


data-bot-coasys:
Folded into #972 rather than reviewed separately.

`fix/plugin-static-instance-tools` has been fast-forwarded to this branch's head (`8fdf1f97b`), so every commit here now lives on #972 — no rebase, no cherry-pick, no duplicate SHAs. The diff of this PR against its base is now empty.

Rationale: these docs describe exactly the tool surface #972 introduces, and they moved four times in twenty minutes today as that surface changed (`add_model`/`signup`/`list_link_language_templates`, then `verify_email_code`). Keeping them in a separate PR guarantees one of the two is wrong at any given moment. As one PR, code and docs land together or not at all.

Closing. Docs review happens in #972.

PR #988, opened 2026-09-08
flow: accept/reject wire surfaces on the consensus engine (replaces #968)

data-bot-coasys:
**Retargeted to `dev`, and it now carries the whole stack.** #987 is merged (`18f3c813`); #1013 → #1012 → #1011 were merged down into this branch in that order, so this one PR is the tail that lands all four. Each kept its own merge record, and nothing was rebased or cherry-picked.

The diff below is therefore larger than the accept/reject work this PR was opened for. The three that rode in were reviewed and approved separately — this section is a map, not a re-review request:

| rode in | what it is | already approved as |
|---|---|---|
| `flow_classes.rs`, `PerspectiveProxy.ts` | flow runtime classes register on `addFlow`; `perspective.remove` accepts a bare `Link` | #1011 (#1007, #1008) |
| `auto_processor/claim.rs`, `watcher.rs` | the processing claim is renewed while a pass runs, so a long pass no longer loses its lease mid-flight | #1012 (#1009, #1010) |
| `interpretation/prompt.rs` | the LLM is told a relation's cardinality | #1013 (#1005) |

**Verified on the merged tree before the cascade** (`d68885fe4`, top of stack with `dev` merged in): `cargo check --release -p ad4m-executor --all-targets`, `cargo fmt --all --check`, the `generate_snapshot` build, and the crate suite at **1654 passed / 0 failed**. Rust only — the JS suites are CI's to run, and CI on this head is the run that gates the merge.

---

This is #968's accept/reject wire surfaces, **re-implemented** against the clean-slate consensus engine rather than rebased onto it. #968 was written against `flow_consensus.rs`, which #987 deletes; a rebase would have produced a surface wired to a design that no longer exists. #968 should be closed in favour of this.

## Start here

| # | read | lines | why |
|---|---|---|---|
| 1 | `flow_instance/accept.rs` — `reject_flow_proposal` + its doc comment | 45 | the only new engine behaviour in this PR |
| 2 | `api/perspectives_ws.rs`, `mcp/tools/flows.rs` | 127 | the two wire surfaces; both are thin |
| 3 | `core/src/perspectives/FlowInstance.ts` | 27 | the SDK shape a client actually touches |
| — | `mcp/tools/mod.rs`, `side_effects.rs`, `PerspectiveClient.ts`, `PerspectiveProxy.ts` | ~42 | registration and pass-through. Skip unless reviewing line by line |

## What reject means here

Reject is **withdraw your own contribution**, not "cancel the proposal".

[...]

PR #987, opened 2026-09-08
flow: consensus engine — state is a fold over the links present now (clean slate)

data-bot-coasys:
Clean-slate flow consensus engine, **standalone against `dev`** — not stacked on #967 or #985. Those two do not merge; this replaces both.

Written to Nico's brief of 2026-09-08: *the code design is the deliverable*. Legibility was the acceptance criterion, not line count.

> **Since the last description — this PR is now the whole of Track B.** Two things happened.
>
> 1. The **Local-cache series** (Sep 15 evening, `673108bd6..1cc88a195`): the `currentState` cache and fired marks became executor-private `Local` links, sync triggers re-derivation, two readers became cache-first. Section [Local cache + cache-first reads](#local-cache--cache-first-reads-sep-15-series).
> 2. **The four stacked follow-ups were merged into this branch** (Sep 16, Nico's call, so he reviews one final state instead of five): **#1030** roles as-of vote time · **#1034** timestamps as instants · **#1035** contested surfaced to consumers · **#1039** foreclosure requires atom evidence. Each merged with its own merge commit and kept its approval; section [Merged in](#merged-in-the-four-follow-ups-1030-1034-1035-1039). `dev` is also merged in (through #996/#1031/#1036/#1038).
>
> Now **67 commits / 32 files / +7,211 −155** against `dev`. Track A (#988 → #1011 → #1012 → #1013) is a sibling chain off this branch and is *not* included here.

Closes #1027. Closes #1000. Closes #998. Closes #999.

---

## Start here (reviewer's route)

| # | Open | Lines | What you get |
|---|---|---:|---|
| 1 | `flow_instance/mod.rs` — the module doc at the top | 38 | the model in one paragraph, before any code |
| 2 | `flow_instance/fold.rs` — `fold`, `settle`, `settle_edge` | **76** | **the engine's entire belief system.** `fold` is 10 lines, `settle` 18, `settle_edge` 48 |
| 3 | `flow_instance/atom.rs` — `signed_by`, `unique_field`, `valid_votes` | 322 | the trust story: what counts as a proposal and a vote, and why a foreign link is invisible |
| — | `roles.rs` (116), `pass.rs` (203), `accept.rs` (158), the rest of `mod.rs` | ~700 | plumbing. Each module doc states its contract; **nothing in them can move a state the fold did not derive.** Skip unless reviewing line by line |
[...]

PR #985, opened 2026-09-08
flow: derived state — the fold is the authority, not the cached link

data-bot-coasys:
**Stacked on #967 — base is `feature/flow-firing-engine`, do not merge before it.**

Slice 1 of the derived-state design in `planning/flow-derived-state-implementation-plan.md`. It makes a forged state link stop mattering to every decision the engine makes.

## The problem

#967 writes a flow's state to a `currentState` link and then trusts that link when deciding what a flow's state is. Any member of a neighbourhood can write that link. So the whole ceremony — vouching signatures, evidence hashes, quorum — is bypassed by forging one triple.

Two things found while reading the code, both worse than a forgeable cache:

- **Identity was an author-string match.** `flow_consensus.rs` never mentioned `proof`, while every synced link carries a persisted `proof.valid` verdict from signature verification on receive. A vote counted if its `author` field said the right DID.
- **Atom fields hydrated last-write-wins.** A peer could append a later `to_state` to someone else's honest proposal; the engine read the peer's value while still attributing the proposal to the original author. With two target states sharing `requires`, that hijacks a vote; otherwise it gets the proposal hard-deleted as superseded.

The old `accept_api_n2_quorum_fires_e2e` passed a "second agent" vote signed with the literal string `"test-signature"`. It now fails without a real second signer, which is the point.

## The design

```
        BEFORE                                  AFTER

  currentState link  ──► state            marked-fired atoms
   (any peer writes)                              │
                                          rebuilt from raw links
                                          per read, each checked:
                                            proposer-authored
                                            signature valid
                                            fields unique
                                            edge declared
                                            quorum met
                                                  │
                                                  ▼
                                        fold(genesis, flow, atoms)
                                                  │
                                                  ▼
                                                state
                                      (currentState = write-through
                                       cache; only reader is a warn
                                       when it disagrees)
```

State is a fold from the flow definition's initial state over transition atoms the pass marked fired, each rebuilt from raw links and re-verified on every read. History is re-verified; evidence hashes are recomputed only at the frontier before a new fire, never over history.

## Where it lives, and why it should be readable

New `perspectives/flow_instance.rs`, written as small named functions with one contract each so the API can be judged from the signatures:

| Function | Contract |
| --- | --- |
| `signed_by(link, did)` | author matches **and** `proof.valid == Some(true)` — the one identity primitive |
| `self_authored_proposer(links)` | the proposer, only if they authored their own proposer link |
[...]

PR #973, opened 2026-09-06
feat(mcp): static instance_* tool surface + describe_perspective (dynamic class tools behind flag)

data-bot-coasys:
## Static MCP tool surface: `describe_perspective` + `instance_*`, dynamic per-class tools behind a flag

External MCP clients now get one stable, DNA-independent tool set. The class name is a **parameter**, the schema is **data** (`describe_perspective`), and every write is validated against the class SHACL before it touches the graph. The old `*_subject` family and `children.rs` are gone; the per-class `{class}_*` tools stay in the codebase but are hidden over MCP unless `dynamicClassTools` is on (the in-process harness is unaffected). #972 (plugin side) is folded in and closed.

Design note: `planning/mcp-tool-surface-redesign-2026-09-06.md` (naming settled with Nico 2026-09-06). Review aids on the branch root: `CONSOLIDATION_NOTES.md`, `PLUGIN_ALIGN_NOTES.md`, `DOCS_RESTRUCTURE_NOTES.md` (each has a "decisions not covered by the brief" section, see the review guide below).

### Size and shape

```
71 files, +12,209 / -3,395
│
├─ rust-executor/src/
│  ├─ mcp/tools/instances/   +4,627        NEW static instance_* suite
│  │  ├─ tests.rs            +2,108        unit tests
│  │  └─ 10 tool files       +2,519        create/query/update/…
│  ├─ mcp/ (rest)            +1,150/−1,500 docs tool + rework
│  │  ├─ docs/*.md           +1,179        agent-facing docs (5 files)
│  │  ├─ tools/*             ~+650/−300    mod, perspectives, auth, …
│  │  └─ subjects+children   −1,196        old per-class code deleted
│  └─ non-MCP + cli/         ~+230         decode_jwt guard, config, flag
│
├─ plugins/ad4m/             +2,900/−1,000
│  ├─ staticToolDefs.ts      +763          NEW verbatim tools/list capture
│  ├─ index/setup/config     +770/−210     plugin rework
│  ├─ tests (vitest)         +1,071        index + bundle
│  ├─ wakerSubscriptionMgr   +18/−287      logic moved into core →
│  └─ skills/ad4m/           +455/−1,097   docs rewritten for static surface
│
├─ core/src/perspectives/    +396          WakerSubscriptionManager + NEW tests
│
├─ tests/js/                 +980          mcp-static-tools NEW, mcp-http rework
│
└─ docs-src/                 +340          running-an-executor NEW, mcp.mdx
```

[...]

PR #996, opened 2026-09-09
Feature: Anthropic provider and streaming

jhweir:
# feat(ai): Anthropic provider, streaming, native tool calling, model discovery

## What

This PR lets the executor talk to Anthropic. It adds a `RemoteChat` trait, moves the existing
OpenAI client behind it, and adds a second implementation for the Messages API. The Anthropic
client supports prompt caching, token streaming, and native tool calling. The PR also adds
`ai.discoverModels`, which asks an endpoint which models it serves, and fixes a lost
`tool_call_id` on the OpenAI-compatible surface.

This PR is independent. It is not stacked on anything and it replaces nothing. It overlaps one
open PR, [#969](https://github.com/coasys/ad4m/pull/969), in three files. The
["Conflicts"](#conflicts-with-open-prs) section covers that.

## Why

The executor could reach one kind of remote model: an OpenAI-shaped HTTP endpoint.
`LlmModel::Remote` held a concrete `ChatGPTClient`. A different vendor's wire format had nowhere
to live.

Anthropic marks its prompt cache with explicit `cache_control` breakpoints. An OpenAI request
struct has no field for one. An interpretation pass makes up to `max_tool_calls` completions
against a system prompt that does not change between them. Without a breakpoint the node pays
full input price on every turn, and this executor bills for each of them.

### The reviewer's route

Read four files in this order. Skip the rest. Line counts are implementation only, and exclude
tests.

| # | File | Read | Lines | What you get |
|---|---|---|---|---|
| 1 | `ai_service/providers/mod.rs` | the whole file | 267 | The contract. `RemoteChat`, the neutral turn and tool types, and the one predicate that decides how tools travel. Everything else follows from this file. |
| 2 | `ai_service/providers/anthropic.rs` | `build_body`, `split_system`, `read_content`, `chat_stream` | 445 | How Claude is spoken to. The cache breakpoint, the content blocks, the SSE reader. |
| 3 | `ai_service/mod.rs` | `build_remote_client`, `turns_for_task`, `prompt_with_tools`, the `PromptTurns` arm | 201 | Where a provider is chosen, and where the worker thread calls one. |
| 4 | `api/openai_compat/harness_bridge.rs` | `complete`, `complete_natively`, `structured_turn` | 134 | The one branch that can change behaviour for an existing caller. |
| — | **Skip:** `ai_service/providers/openai.rs` | — | 133 | Moved code. It sends the same bytes it sent before. The only new part is `list_models`. |
| — | **Skip:** `types/core.rs`, `api/ai_ws.rs`, `openai_compat/chat.rs`, `core/src/ai/AIClient.ts` | — | 70 | One enum variant, one RPC handler, a two-line fix, and one SDK method. Each is self-contained. |

The contract that makes the skips safe: **no file outside `providers/` names a wire format.**
`ai_service/mod.rs` holds a `Box<dyn RemoteChat>` and never learns which one answered.

### The model

A provider is a wire protocol, and the protocol is now a seam. `RemoteChat` has three members: a
completion, a streamed completion, and a question about whether the protocol carries tools. Each
implementation owns its own request shape, its own auth, and its own parsing. Callers hold the
trait. Adding a third vendor means adding a file and one match arm. It does not mean touching
`LlmModel`, which answers a different question: local weights, or a remote endpoint.

### Files changed

Ten files across two packages. The `rust-executor` crate holds the provider module, the service
that drives it, and the two API surfaces that reach it. The `core` package gains one SDK method,
because the new RPC handler was otherwise unreachable from TypeScript.

### Structure

```mermaid
graph TD
    M["<b>providers/mod.rs</b> 👈 read this first<br/>RemoteChat + neutral types<br/>267 lines · decides everything"]

    A["providers/anthropic.rs<br/>Messages API · 445 lines"]
    O["providers/openai.rs<br/>moved, unchanged · 133 lines"]

    S["ai_service/mod.rs<br/>picks and calls a provider<br/>201 lines"]
    H["openai_compat/harness_bridge.rs<br/>native or injected · 134 lines"]

    A -->|implements| M
    O -->|implements| M
    S -->|"holds Box&lt;dyn RemoteChat&gt;"| M
    S -->|"constructs, once per model"| A
    S -->|"constructs, once per model"| O
    H -->|"calls prompt_with_tools"| S
    H -->|"uses the neutral types"| M

    classDef core fill:#1f6feb,stroke:#0b3d91,color:#fff
    classDef skip fill:#30363d,stroke:#8b949e,color:#c9d1d9
    class M core
    class O skip
```

Dependencies point one way, toward `providers/mod.rs`. No caller imports a provider
implementation. `ai_service/mod.rs` constructs one and immediately forgets its type.

### Flows

A tool-calling turn takes one of two paths. The branch is the only new decision that an existing
caller can notice.

```mermaid
flowchart TD
    H["harness loop"] --> B{"does this provider<br/>carry tools?"}

    B -->|"no — every local model,<br/>every OpenAI-shaped endpoint"| I["render tools into<br/>a system prompt"]
    I --> P["prompt_messages"]
    P --> W1["worker thread"]
    W1 --> OA["OpenAI client"]
    OA --> X["read calls out of the reply text"]
    X --> D1["drop: the tool_call blocks<br/>are stripped from the content"]
    D1 --> H

    B -->|"yes — Anthropic"| N["prompt_with_tools"]
    N --> W2["worker thread"]
    W2 --> AN["Anthropic client"]
    AN --> R["read tool_use blocks"]
    R --> D2["drop: unknown block types,<br/>and input_json_delta while streaming"]
    D2 --> H

    classDef unchanged fill:#30363d,stroke:#8b949e,color:#c9d1d9
    class I,P,W1,OA,X,D1 unchanged
```

The greyed path is unchanged code. A local model and an OpenAI-shaped endpoint run exactly what
they ran before this PR.

### What is trusted, and where each failure stops

Nothing in this PR trusts a provider response. Every field is optional on the way in.

| Input | Reaches the model or caller? | Stops where |
|---|---|---|
| Wrong API key | No | The provider answers 401. The error carries the provider's own message, because a status alone does not separate a bad key from a bad model name. |
| Unknown model name | No | Same path as a wrong key. |
| Unreachable endpoint | No | A connection error that names the URL, that being the thing an operator got wrong. |
| Malformed SSE line | No | Skipped. The stream continues. One lost delta beats an aborted completion the caller pays for. |
| `input_json_delta` in a stream | No | Skipped. It is partial JSON for a tool call and would splice a fragment into the visible answer. |
| Unrecognised content block | No | Skipped. It adds no text and mints no call. |
| Tool call with unreadable arguments | No | Dropped at the bridge. The turn degrades to text, which is recoverable. A broken call is not. |
| A local model sent down the native path | No | Refused with a named error, in case the branch predicate is ever wrong. |
| Upstream answers with no choices | No | An error. It used to be an empty completion, reported as success and billed as one. |
| Anthropic refuses the request | No | An error naming the category. The API answers a refusal with HTTP 200, `stop_reason: "refusal"` and empty content, so reading `content` alone reports a successful empty answer and bills for the input. |
| Consumer drops the token channel | Partly | The stream stops early and returns the text written so far. |

### Honest counts

**No line budget was set for this work, so there is no overrun to report.** The design was settled
in conversation and not in a document. That is a departure from our own rules and it costs a
reviewer the design delta they would otherwise read instead of the diff.

- Implementation: **1,343 lines.** That is about a third over the ~1k guideline.
- Tests: **1,176 lines**, in 61 tests, five of which reach the real API.
- Counted by attributing each added line to a test module or not. An earlier version of this
  section said ~1,330 and ~1,150. That count was wrong: it treated everything after the first
  `cfg(test)` in a file as a test line, and one of those is a module declaration near the top of
  `providers/mod.rs`, so 300 lines of implementation were filed as tests.
- No implementation moved into a test file to make either number smaller.

The largest single file is `anthropic.rs` at 1,016 added lines. 519 of those are tests.

### What the real-API tests found

They found a bug on the first run, which is the argument for having written them.

Anthropic answers a refused request with HTTP 200, `stop_reason: "refusal"` and an empty content
array. The client read `content`, found nothing, and returned a successful empty reply. A caller
gets an empty string with no sign anything went wrong, and gets billed for the input. That is the
same failure the OpenAI arm had with `unwrap_or_default`, which this branch already fixed. The
Anthropic path shipped with its own copy of it, and no mock test would have caught it, because the
mocks assert what we send.

A refusal is now an error naming the category the API reported.

`ChatReply` also carries the usage block now. A cache assertion needs something observable, and
prompt caching fails silently by construction. The counts are exact, unlike the `chars/4` estimate
this crate bills on, so they are also a step toward fixing that.

### The simplification pass

Run over the whole branch before review. Two findings.

`messages_endpoint` and `models_endpoint` were the same function in two files, differing only in
the last path segment, and each carried its own tests. `versioned_endpoint(base_url, path)`
replaces both, and six endpoint tests become four that exercise one implementation rather than one
copy each.

`ToolCall` is field-for-field the same as `harness::HarnessToolCall` and stays that way. Merging
them means `providers` importing from `harness`, which sits above it and already carries an import
cycle its own guide asks nobody to widen. The doc comment now says so, because without it the next
reader merges them and gets the cycle.

Nothing else came out. Every public item added on this branch has a caller. Two candidates were
considered and left: collapsing `to_wire_role` and `to_wire_message` saves three lines and loses a
name that earns its place, and `ChatUsage::output_tokens` is read only by tests today but belongs
to a usage block that would be stranger partial than whole.

### Corrections against the design doc

There is no design doc. See ["Honest counts"](#honest-counts).

### Verification

Separate the two kinds of claim below. The first block is evidence a reviewer can reproduce. The
second is a verdict a reviewer has to take on trust.

**Proved:**

- **Verified against `api.anthropic.com`, three consecutive clean runs.** Five tests in
  `providers/anthropic_e2e.rs`: a completion returns text, a streamed completion arrives in more
  than one delta and reassembles into the returned text, a tool is called with parsed arguments, a
  `tool_result` block goes back and the model reads the value out of it, and a second call over an
  identical prefix reads 7,120 cached tokens after the first writes them. Run them with
  `ANTHROPIC_API_KEY=… cargo test --lib ai_service::providers::anthropic_e2e -- --ignored`.
- **CI was green on commit `523aa2e`.** All CircleCI checks reported
  SUCCESS, including `build-and-test`, `integration-tests-js`, `integration-tests-mcp`,
  `integration-tests-model`, `integration-tests-multi-user-simple`,
  `integration-tests-email-verification` and `plugin-ad4m-tests`. **Three commits have landed since,
  so CI has to run again before this claim holds for the current head.**
- 117 tests pass in `ai_service`, 281 in `api::`, run as
  `cargo test -p ad4m-executor --lib <path> -- --test-threads=1`.
- `cargo fmt --all --check` is clean from the repo root. `core` typechecks under `tsc --noEmit`.
- The wire tests observe a red state. New code has nothing to run "before the change" against, so
  the equivalent evidence is a mutation. Renaming `cache_control` to `cacheControl` and
  `input_schema` to `parameters` turns four tests red, two of them at the wire level:

  ```
  test wire_tests::the_request_body_carries_the_model_max_tokens_and_a_cached_system_prompt ... FAILED
  test wire_tests::tools_go_out_as_input_schema ... FAILED
  panicked at anthropic.rs:806: completion succeeds: Anthropic API error 501 Not Implemented
  ```

  The mock server refuses to match the body, which is the signal that the bytes changed. The
  mutation is reverted. All 31 provider tests pass again.

**Not proved, and taken on trust:**

- **The e2e tests are not in CI.** They cost money, unlike the nightly llm-e2e suite that runs
  against a local Ollama. They have to be run by hand when this client changes.
- **No bot review has happened.** CodeRabbit reports SUCCESS on the checks list, but it skipped
  the review because this PR is a draft. Read that tick as "did not run", not as "found nothing".
- **The branch predicate itself has no test.** The pieces on both sides of it are tested. Choosing
  between them reads model config from the database, and the bridge holds a concrete
  `Arc<AIService>` rather than a trait, so there is no seam to fake.
- **Interpretation has not run end to end against a real model.** The JS integration suites pass in
  CI, which covers the WS and MCP surfaces. A local model and an OpenAI-shaped model take
  byte-identical code here, so nothing should regress. The pass that would show it needs a real
  model.
- **No simplification pass has been run.** Our own rules ask for one before review.

### Deferred work

Each item names what is worse today because it is not done.

- **Streaming for OpenAI-shaped providers.** A caller streaming from one receives the whole reply
  in a single chunk. `chat_gpt_lib_rs` exposes no streaming call, so real SSE there means bypassing
  the crate.
- **A native Ollama provider.** Ollama's `/v1` endpoint ignores `options.num_ctx` and truncates
  context to roughly 16K without saying so. The native `/api/chat` endpoint honours it. The tool
  path suffers most, because injected tool definitions make the prompt larger.
- **The e2e tests in the nightly.** They exist and pass, but run by hand. The nightly llm-e2e
  workflow uses a local Ollama and needs no credentials. Wiring these in needs a funded key on that
  workflow. Until then, a change in how Anthropic accepts `cache_control` stops caching silently
  between manual runs, and the only symptom is a larger bill.
- **Anthropic-specific request features.** Adaptive thinking, effort levels and structured outputs
  are not plumbed. Nothing needs them yet.

Platform work that is not this PR's problem: the `ai_service/mod.rs` split
([spec item 9](https://github.com/coasys/ad4m/pull/970)) and the `agentic/` move (item 7). This PR
puts new code where those items send it and shortens `mod.rs` by 48 lines.

### Conflicts with open PRs

The `LlmModel` change looks disruptive and is not. Checked file by file.

| PR | Shares a file? | Cost |
|---|---|---|
| [#967](https://github.com/coasys/ad4m/pull/967), [#985](https://github.com/coasys/ad4m/pull/985), [#987](https://github.com/coasys/ad4m/pull/987), [#988](https://github.com/coasys/ad4m/pull/988) | No | Their harness work sits in `perspectives/interpretation/run.rs`, the loop's caller. |
| [#973](https://github.com/coasys/ad4m/pull/973) | No | It changes `mcp/tools/harness_bridge.rs`, a different file that shares a name. |
| [#944](https://github.com/coasys/ad4m/pull/944), [#953](https://github.com/coasys/ad4m/pull/953) | Different regions | #953 edits `Cargo.toml` at line 85. This PR edits line 174. |
| [#969](https://github.com/coasys/ad4m/pull/969) | Yes, three files | Mechanical. See below. |

#969 touches `ai_service/mod.rs` at the import block and from line 1122 down, in the tasks and
embeddings sections. Every change here sits in the worker-thread region between 700 and 1000. The
expected conflicts are the import block, the boundary near `publish_model_status`, and two
branches appending to `openai_compat/tests.rs`. None of them is semantic, in either merge order.

The alternative to the trait was one `LlmModel` variant per vendor. That is tidy at two providers
and unpleasant at three, and each of the four `match model` arms already repeated its own message
assembly. A variant now means this same refactor later, against more branches. The mechanical move
is its own commit, so a rebase across it reads cleanly.

### Commits

Eleven commits. The mechanical move is separate from every behaviour change, and the
simplification pass is separate again.

1. `refactor(ai)` — the `RemoteChat` trait, and the OpenAI client behind it
2. `feat(ai)` — the Anthropic provider and prompt caching
3. `feat(ai)` — streaming
4. `fix(openai-compat)` — `tool_call_id` on tool results
5. `feat(ai)` — `ai.discoverModels`
6. `feat(ai)` — tool types on the provider trait
7. `feat(ai)` — the native or injected branch
8. `feat(sdk)` — `AIClient.discoverModels`
9. `test(ai)` — the wire tests against a mock server
10. `fix(ai)` — refusals become errors, plus the real-API tests
11. `refactor(ai)` — one endpoint helper instead of two copies of it

One behaviour change hides inside commit 1 and deserves a reviewer's eye. The streaming arm read
`choices.first()` with `unwrap_or_default()`. An upstream answering with no choices produced an
empty completion, reported as success and billed as one. It now takes the error path that the
non-streaming arm always took. **This is not a no-behaviour-change refactor, and the commit says
so.**

## References

- Anthropic Messages API: https://platform.claude.com/docs/en/api/messages
- Anthropic prompt caching: https://platform.claude.com/docs/en/build-with-claude/prompt-caching
- OpenAI SDK compatibility, and its stated limits: https://platform.claude.com/docs/en/cli-sdks-libraries/libraries/openai-sdk
- Harness design: `planning/llm-harness-design-2026-08-21-v3.md`
- Executor refactoring spec, still in draft: https://github.com/coasys/ad4m/pull/970

## Summary by CodeRabbit

- **New Features**
  - Added Anthropic provider support, including chat, streaming, native tool use, and prompt caching.
  - Added model discovery to list available models and verify endpoint credentials.
  - Added native tool calling for compatible models.
  - Added turn-based prompts and streamed remote responses.
  - Added support for Anthropic model type configuration.

- **Bug Fixes**
  - Tool-result messages now preserve supplied call identifiers.
  - Credentials are blocked from unsafe non-local HTTP endpoints.

- **Tests**
  - Expanded coverage for discovery, streaming, tool calls, provider behavior, and error handling.


jhweir:
@coderabbitai review

jhweir:
@coderabbitai review

jhweir:
@coderabbitai review

jhweir:
@coderabbitai review

rust-executor/src/ai_service/providers/anthropic.rs
```diff
@@ -0,0 +1,1170 @@
...
+
+/// Wire version. Anthropic requires it on every request and treats it as the
+/// contract: pinning it is what stops a server-side change reshaping our
+/// responses. Bump deliberately, never to chase a new feature by accident.
+const ANTHROPIC_VERSION: &str = "2023-06-01";
+
+/// `max_tokens` is mandatory on this API — unlike the OpenAI path, where
+/// omitting it lets the server pick. 16k is large enough that a long
+/// interpretation answer is not clipped and small enough to bound a runaway.
+/// Model options override it (`ModelApi.options.max_tokens`).
```
data-bot-coasys: `ModelApi.options.max_tokens` does not exist. I grepped the whole tree for `max_tokens`: the only writers are this constant at line 324 and the two test sites, and the only other occurrences are `openai_compat/types.rs:191,308`, which are the *inbound* request fields on the compat surface — and those are dropped too (`ChatInput { ..Default::default() }` on the OpenAI side, no field at all on this one).

So a client that sends `max_tokens` to `/v1/chat/completions` has it silently ignored on both providers, and an operator who reads this comment goes looking for a model option that was never built.

Either wire it — thread the inbound `max_tokens` through `ChatRequest` and `unwrap_or(DEFAULT_MAX_TOKENS)` here — or delete the sentence. I'd take the deletion in this PR and the wiring in its own; the comment being wrong is the part that costs someone an afternoon.

(For what it's worth the 16k value itself is fine: every current Claude model tops out at 128k output, so it will not be refused.)

rust-executor/src/ai_service/providers/anthropic.rs
```diff
@@ -0,0 +1,1170 @@
...
+                            usage,
+                        });
+                    }
+                }
+            }
+        }
+
+        Ok(ChatReply {
+            text,
+            tool_calls: Vec::new(),
```
data-bot-coasys, line 531: `tool_calls: Vec::new()` unconditionally, on a provider whose `supports_native_tools()` is `true` and whose `build_body(request, **true**)` serialises `request.tools` onto the wire.

The comment on `build_body` says the shared body exists "so the streaming and non-streaming calls cannot drift in what they send". That is true and is exactly what makes this asymmetric: tools go out on the streaming path, and any `tool_use` the model answers with is discarded here. `text_delta` correctly refuses to treat `input_json_delta` as text — so the call is not corrupted into the reply, it just vanishes. The caller sees a successful completion with empty text and no calls, which reads as "the model had nothing to say".

Not reachable today: `PromptStream` in `ai_service/mod.rs:1016` builds its `ChatRequest` with `ChatRequest::new(...)` and never `.with_tools(...)`, and it is the only route to `chat_stream` outside the e2e file. So this is latent. It stops being latent as soon as something streams an agent turn — and #1001 is already stacked on this branch adding a third provider to the same trait.

Cheapest correct fix is the guard, not the feature:

```rust
if !request.tools.is_empty() {
    return Err(anyhow!(
        "streaming with tools is not supported on this provider yet — \
         tool_use blocks arrive as input_json_delta and are not accumulated"
    ));
}
```

That turns a silent drop into a loud one and costs four lines. Accumulating `input_json_delta` per `content_block_index` and flushing on `content_block_stop` is the real fix, but it deserves its own PR with its own tests. Either way the trait doc on `chat_stream` should say which one this is — right now it promises "the same" as `chat`.

rust-executor/src/ai_service/providers/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+/// are the exact numbers, carried so a caller can check them and so billing
+/// can move onto them later.
+#[derive(Debug, Clone, Default, PartialEq)]
+pub struct ChatUsage {
+    pub input_tokens: Option<u64>,
+    pub output_tokens: Option<u64>,
+    /// Tokens served from the prompt cache. Zero means the breakpoint did not
+    /// land, which is otherwise invisible: caching fails silently and the only
+    /// symptom is a larger bill.
+    pub cache_read_tokens: Option<u64>,
```
data-bot-coasys, line 244: This field is the only evidence that the cache breakpoint landed, and nothing in the running executor ever looks at it.

Traced every reader: `ai_service/mod.rs` takes `.text` and drops the rest on all three arms (`Prompt` at :940, `PromptTurns` at :912 hands the whole `ChatReply` to a `result_sender` whose consumer `prompt_with_tools` uses `.text`/`.tool_calls` only, `PromptStream` at :1034 bills with `estimate_token_count`). `harness_bridge` maps `reply.text` and `reply.tool_calls`. The single reader in the tree is `anthropic_e2e.rs:326`, which is `#[ignore = "llm-e2e"]` and returns early without `ANTHROPIC_API_KEY` — so it does not run in CI, it does not run nightly (that suite is local Ollama), and it costs money to run by hand.

Production therefore has no way to answer "is caching working", and the comment three lines up says why that matters: it fails silently and the only symptom is a larger bill. Prompt caching is the reason given for not routing Claude through Anthropic's OpenAI shim, so this is the PR's central claim going out unobserved.

One line in `AnthropicChat::chat` (and the same in `chat_stream`) pays for itself:

```rust
debug!(
    "anthropic usage: in={:?} out={:?} cache_read={:?} cache_write={:?}",
    usage.input_tokens, usage.output_tokens,
    usage.cache_read_tokens, usage.cache_write_tokens
);
```

`rust_executor::ai_service` is already at debug in the e2e harness, so it would show up in the logs we actually read. Moving billing onto the exact counts is the bigger change and I agree it belongs later — this is just about being able to see them.

rust-executor/src/ai_service/providers/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+/// The same fact as [`RemoteChat::supports_native_tools`], asked of a model
+/// config rather than of a built client — the harness has to decide how to
+/// render tools *before* anything reaches a worker thread, so it cannot ask
+/// the instance. Kept beside the trait, and covered by a test asserting the
+/// two agree, because two answers drifting apart would route a model down a
+/// path its provider cannot serve.
+pub fn api_type_supports_native_tools(api_type: &crate::types::ModelApiType) -> bool {
+    match api_type {
+        crate::types::ModelApiType::Anthropic => true,
+        crate::types::ModelApiType::OpenAi => false,
```
data-bot-coasys, line 96: The doc above says this is "a plain fact about the wire format, never a preference". For `Anthropic => true` that is right. For `OpenAi => false` it is not: the OpenAI wire format carries `tools[]` and returns `tool_calls` — it is one of the two formats that defined the shape. The `false` is a fact about `OpenAiChat`, which does not implement it.

The combination is a trap. A reader who takes the doc at its word sees an obviously wrong value, corrects it to `true`, and every test still passes — `the_two_answers_about_native_tools_agree` only asserts the two answers match, so it would need `OpenAiChat::supports_native_tools` flipped too, and then it is green again. At that point `harness_bridge::complete` routes to `complete_natively`, `ChatRequest` carries the tools, and `OpenAiChat::chat` builds a `ChatInput` from `to_wire_message` alone and drops them: the model is handed no tools, answers in prose, `read_content` finds no calls, and the harness reads the empty `tool_calls` as "done" on turn one. Silent, and it looks like the model chose not to call anything.

Two small changes make that unreachable:

1. Reword to what it is — "whether *this executor's client* for the protocol passes tools as data" — and note that the OpenAI client could support it and does not yet.
2. Make the drop impossible rather than documented: in `OpenAiChat::chat`, `if !request.tools.is_empty() { return Err(...) }`. The `ChatRequest::tools` doc already says it is "only ever non-empty for a provider whose `supports_native_tools` is true" — this is that invariant, enforced instead of asserted.

This matters more with #1001 in flight: Ollama's native endpoint *does* carry tools, so the third arm will be a case where implementation and wire format agree, making the OpenAI arm the odd one out and the wording actively misleading.

rust-executor/src/api/openai_compat/harness_bridge.rs
```diff
@@ -153,6 +162,141 @@ impl CompletionSource for OpenAiCompatBridge {
...
+/// unbalanced conversation is not.
+///
+/// A call with no id is the one case that is skipped, because the pairing it
+/// would need does not exist either way and an empty `tool_use` id is refused.
+fn to_provider_call(raw: &Value) -> Option<ToolCall> {
+    let function = raw.get("function").unwrap_or(raw);
+    let name = function.get("name")?.as_str()?.to_string();
+
+    let id = raw.get("id").and_then(|v| v.as_str()).unwrap_or_default();
+    if id.is_empty() {
```
data-bot-coasys: Note rather than a blocker — this branch contradicts the argument in its own doc comment.

Lines 267-272 make the case for keeping a malformed call: dropping it leaves the following `tool_result` pointing at a `tool_use` that is not there, and "Anthropic rejects the whole request". Then the id-less case is dropped, which produces exactly that state — the harness has already dispatched the call and appends the result on the next turn, and if that result carries a `tool_call_id` (the normal shape; `complete` mints `call_{uuid}` for every one) `structured_turn`'s `"tool"` arm turns it into a `tool_result` referencing the block you just removed.

The stated reason — an empty `tool_use` id is refused — argues for *replacing* the id, not for dropping the call:

```rust
let id = raw.get("id").and_then(|v| v.as_str()).unwrap_or_default();
let id = if id.is_empty() { format!("call_{}", Uuid::new_v4()) } else { id.to_string() };
```

That keeps the conversation balanced for the same reason `_raw` keeps the bad arguments. I could not construct a live path that produces an id-less call — both producers always set one — so this is about the defensive branch being self-consistent, not about a bug I can demonstrate.

data-bot-coasys (changes requested):
Read it in the order the description asks for, which worked — `providers/mod.rs` really is the whole contract, and the skips were safe. The layering claim holds: I grepped for wire-format names outside `providers/` and found none. `is_transport_safe` is the best thing in the diff — https-anywhere plus loopback-only-http, with `localhost.example.com` explicitly not counting as local, is the version of that check people usually get wrong. The UTF-8 line buffer in `chat_stream` is right for the reason stated, and refusal handling on both the status path and the mid-stream path closes a real hole that the OpenAI arm had.

Four things. None is a live data-corruption bug on today's call graph — I traced every caller — but two are doc-vs-code contradictions that will mislead the next reader, and two are gaps that stop being latent the moment #1001 lands on top of this. All four are small.

**1. `DEFAULT_MAX_TOKENS` documents an override that does not exist.** The comment says model options override it via `ModelApi.options.max_tokens`. There is no such field anywhere in the repo. An operator reading that comment will go looking for a setting that was never built.

**2. The streaming path sends tools and cannot read them back.** `build_body` is deliberately shared "so the streaming and non-streaming calls cannot drift in what they send" — and it does keep the send side identical, tools included. The read side has no such parity: `chat_stream` returns `tool_calls: Vec::new()` unconditionally. Not reachable today (`PromptStream` builds its `ChatRequest` without tools), so this is a trap rather than a bug — but `a_tool_use_json_delta_is_not_treated_as_text` shows you already knew `tool_use` blocks arrive in the stream, and nothing in the trait docs says streaming drops them.

**3. Prompt caching ships with no signal that it works.** Caching is the stated reason this PR exists rather than routing Claude through the OpenAI shim. `ChatUsage.cache_read_tokens` is parsed correctly and then dropped at every call site — the only reader is `anthropic_e2e`, which is `#[ignore]`d and needs a paid key. Your own module doc says caching "fails silently and the only symptom is a larger bill". That is the state this ships in.

**4. `api_type_supports_native_tools` says it is a fact about the wire format. For OpenAI it is not.** Details inline.

One note (no change needed unless you agree): the `discover_models` error path returns the upstream response body verbatim, so an AI_CREATE holder can read internal HTTP response bodies by pointing `baseUrl` at them. I convinced myself the authority is the same as `addModel` + prompt, which also round-trips a remote's text back to the caller — flagging it only so the equivalence is a decision rather than an accident.

CI is 8/8 at `0f3abc39a`. The description is the clearest in the repo this week; the reviewer's-route table saved me real time and I'd like that to become the house style.

jhweir:
@data-bot-coasys 

Responding to all four requests, plus both notes. Nothing here I pushed back on — the two doc-vs-code contradictions were the kind I would rather find in review than in a postmortem, and the two latent gaps are worth closing now precisely because #1001 is stacked on this trait.

Six commits, `f1a47e15a`..`b4e8b7aa3`, one per finding.

---

### 1. `DEFAULT_MAX_TOKENS` — `docs(ai): delete a max_tokens override that was never built` (`f1a47e15a`)

Confirmed exactly as described. `ModelApi` has four fields — `base_url`, `api_key`, `model`, `api_type` — and no `options`. The `max_tokens` on `openai_compat/types.rs:191,308` is the inbound field on the compat surface and is dropped on both providers.

Took the deletion, as you suggested, and named the gap rather than just removing the sentence:

> Nothing overrides it. `ModelApi` carries no options, and the `max_tokens` a client sends to the compat surface is dropped before it reaches either provider. Threading that through is worth doing and is not done here.

The wiring is its own PR. A comment that sends someone looking for a setting is the part that costs an afternoon; a comment that says "this is missing" costs nothing and marks the work.

### 2. Streaming sends tools and cannot read them back — `fix(ai): refuse to stream with tools rather than dropping the calls` (`68df51e43`)

Took the guard, not the feature. Four lines at the top of `chat_stream`, before `build_body`, so nothing is posted at all:

```rust
if !request.tools.is_empty() {
    return Err(anyhow!(
        "Streaming with native tools is not supported on this provider yet: \
         tool_use blocks arrive as input_json_delta and are not accumulated. \
         Use the non-streaming call for a turn that carries tools."
    ));
}
```

You were right that the trait doc was the other half of the problem — `chat_stream` promised "the same" as `chat`, which is what would let the next implementor reproduce this. It now says tools are not promised on the streaming path even where `supports_native_tools` is true, and that an implementation which cannot read them must refuse rather than answer with none.

New test `streaming_with_tools_is_refused_rather_than_dropping_the_calls`, asserting on a mock with `.expect(0)` — the claim is that the request never leaves the client.

Accumulating `input_json_delta` per `content_block_index` and flushing on `content_block_stop` is the real fix and I agree it needs its own PR and its own tests.

### 3. Caching ships unobserved — `fix(ai): log the usage block, so caching can be seen to work` (`c3d17b626`)

Your trace matches mine: `.text` on all three arms of `ai_service/mod.rs`, `.text`/`.tool_calls` at the bridge, `estimate_token_count` for billing, and `anthropic_e2e.rs:326` the single reader in the tree behind an `#[ignore]` and a paid key.

One `log::debug!` on each arm via a shared `log_usage(model, streamed, &usage)`, so the two cannot report differently. Called at both `chat_stream` exits, including the consumer-hung-up path — the tokens were still bought.

The doc on the helper says why it exists, because a future simplification pass will otherwise read it as a stray log line:

> Caching is the reason this client exists rather than routing Claude through the OpenAI-shaped shim, and it fails silently. `cache_read_tokens` is the observation that says it worked, and until billing moves off `estimate_token_count` the log is the only place it can be seen outside the `#[ignore]`d e2e suite.

Billing on the exact counts is still later.

### 4. `api_type_supports_native_tools` — `fix(ai): refuse tools on a client that cannot pass them` (`9ed072aa9`)

This was the best of the four and I think you undersold it slightly: the trap is not only that a reader corrects the value, it is that `the_two_answers_about_native_tools_agree` is a *consistency* test wearing the costume of a *correctness* test. It will stay green through the exact mistake it looks like it guards.

Both changes taken.

Reworded, on the trait and on the free function. The trait now says it is a fact about the implementation, names `OpenAiChat` as a client that answers `false` about a protocol that could carry tools, and states the obligation: a client answering `false` must refuse a request carrying tools rather than ignore them. The free function's doc adds that flipping its arm alone is not enough to make native tools work on the OpenAI path.

Enforced. `OpenAiChat::chat` now returns an error when `request.tools` is non-empty, with the message saying whose limitation it is. The `ChatRequest::tools` doc already asserted this invariant; it is now checked. Test `a_request_carrying_tools_is_refused_rather_than_stripped`, again `.expect(0)` on the mock.

That makes the mistake loud instead of silent: someone who flips both answers now gets an error naming the missing work, rather than a model answering in prose and a harness concluding it is done on turn one. And you are right that #1001 sharpens this — with Ollama's native endpoint carrying tools, the OpenAI arm becomes the only one where client and format disagree, so the old wording would have been actively misleading rather than merely loose.

### 5. The id-less call — `fix(openai-compat): mint an id for a call that arrives without one` (`85ed475f1`)

Agreed, and this one bothered me more than its severity warrants because the argument against it was sitting five lines above it. The doc makes the case for keeping a malformed call — a dropped `tool_use` leaves the next turn's `tool_result` pointing at nothing and Anthropic rejects the request — and then the next branch does the dropping.

Minted, in the shape `complete` already uses:

```rust
let id = if id.is_empty() { format!("call_{}", Uuid::new_v4()) } else { id.to_string() };
```

`a_call_without_an_id_is_skipped` becomes `a_call_without_an_id_is_given_one_rather_than_dropped`, and I added `a_call_with_an_empty_id_is_given_one_too`, since an id present but empty is the same wire failure and was reaching the same branch.

Like you, I could not construct a live path that produces one. Recorded as consistency, not as a bug.

### 6. The note on `discover_models` — `docs(ai): record that discovery returns the upstream failure body` (`b4e8b7aa3`)

I agree with your conclusion and I did not want it to stay an accident, so I wrote the decision down where it lives. The handler doc already reasoned about the authority the endpoint carries and said nothing about the body.

Keeping the verbatim body, on two grounds now stated: AI_CREATE already lets a caller name an arbitrary URL and send it credentials, so this is reach and not authority; and the body is what makes the endpoint worth having, because a status alone does not separate a bad key from a bad model name from a host that is not an LLM.

One asymmetry I did not want to paper over, so it is in the comment as the condition that would change the answer: discovery is a cleaner read primitive than `addModel` plus a prompt, because it needs no configured model and no completion. If AI_CREATE is ever granted more widely than to the node's operator, this should be revisited rather than re-derived.

---

### Verification

- `cargo test -p ad4m-executor --lib ai_service:: -- --test-threads=1` — **134 passed, 0 failed**, 10 ignored.
- `cargo test -p ad4m-executor --lib api:: -- --test-threads=1` — **289 passed, 0 failed**.
- `cargo fmt --all --check` clean. Working tree clean.
- Three new tests, one renamed. The two `.expect(0)` mocks are the interesting ones: both guards are about a request that must not be sent, so asserting on the error text alone would pass against a client that sent it anyway and then failed.

**Not re-verified, and you should read it as unproven:** the five e2e tests against `api.anthropic.com` have not been re-run since these commits. Nothing here touches `build_body`, the SSE reader or `read_content`, and `log_usage` reads a struct that was already parsed — but that is an argument, not a run. **CI has not run on `b4e8b7aa3` yet**; it was 8/8 on `0f3abc39a` as you saw.

### What I did not do

- Accumulating `input_json_delta` into structured calls on the streaming path. Its own PR, with its own tests.
- Native `tools[]` on `OpenAiChat`. Same. The guard is what makes deferring it safe rather than silent.
- Threading `max_tokens` from the compat surface through `ChatRequest`. Same, and now marked in the code.
- Billing on `ChatUsage` rather than `estimate_token_count`. Bigger than this PR, unchanged.

Thanks for tracing every caller rather than pattern-matching — the "not reachable today, and here is the exact line that makes it so" framing on #2 and #4 is what made them quick to act on, because I could check the claim instead of re-deriving it. Noting for the record that the reviewer's-route table did its job in both directions: you said it saved you time, and it is also what made your findings land on the four files I most wanted eyes on.

rust-executor/src/ai_service/providers/anthropic.rs
```diff
@@ -0,0 +1,1250 @@
...
+    http: reqwest::Client,
+    api_key: String,
+    /// Fully-resolved messages endpoint, e.g. `https://api.anthropic.com/v1/messages`.
+    endpoint: String,
+}
+
+impl AnthropicChat {
+    pub fn new(api_key: &str, base_url: Url) -> Self {
+        Self {
+            http: reqwest::Client::builder()
```
data-bot-coasys: **Blocking: the API key follows a redirect off-host.**

This client sets a timeout and nothing else, so it takes reqwest's default redirect policy, `Policy::limited(10)`. On a cross-host redirect reqwest strips the headers it knows to be sensitive — `Authorization`, `Cookie`, `Proxy-Authorization`, `WWW-Authenticate` — and `x-api-key` is not one of them. `send` sets it per request at line 340, so it is re-attached to the redirected request and travels to whatever host answered.

The OpenAI arm happens to be safe from this by accident: `list_models` uses `bearer_auth`, which lands in `Authorization`, which reqwest does strip. The asymmetry is exactly why this is easy to miss in a PR that adds a second provider.

```rust
http: reqwest::Client::builder()
    .timeout(REQUEST_TIMEOUT)
    .redirect(reqwest::redirect::Policy::none())
    .build()
```

Nothing in the Messages API redirects, so refusing to follow one costs nothing and the failure is a visible error rather than a key on someone else's access log.

rust-executor/src/ai_service/providers/anthropic.rs
```diff
@@ -0,0 +1,1250 @@
...
+
+/// Ask Anthropic which models this key can reach — `GET {base}/v1/models`.
+///
+/// Same role as the OpenAI listing (see that module for why it is standalone),
+/// with different auth: an `x-api-key` header and the pinned wire version
+/// rather than a bearer token.
+pub async fn list_models(api_key: &str, base_url: Url) -> Result<Vec<String>> {
+    let endpoint = super::versioned_endpoint(base_url, "models");
+
+    let response = reqwest::Client::new()
```
data-bot-coasys: **Blocking, and the sharper half of the same bug: this is the call `is_transport_safe` guards, and a redirect walks around the guard.**

`discover_models` refuses to send a key unless the URL is https or loopback. Then this client, with the default redirect policy, follows wherever that https host points — including to `http://`. So:

- caller passes `https://collector.example/` with a key → `is_transport_safe` returns true
- the host answers `302 Location: http://collector.example/`
- reqwest follows, `x-api-key` is a custom header and is not stripped, the key crosses the network in the clear

That is the precise thing the check exists to prevent, achieved by the party the check assumes is hostile. `.redirect(reqwest::redirect::Policy::none())` on this builder too — `reqwest::Client::new()` has to become `Client::builder()` here, which is the only reason this is a second comment rather than the same one.

Worth a test: a mock server answering 302 to a second mock, asserting the second one never sees `x-api-key`.

rust-executor/src/api/ai_ws.rs
```diff
@@ -47,6 +47,88 @@ async fn list_models(_params: Value, ctx: Arc<RequestContext>) -> Result<Value,
...
+
+    Ok(serde_json::to_value(models)?)
+}
+
+/// Whether a credential may be sent to this URL.
+///
+/// True for https anywhere, and for http on loopback only. A hostname that
+/// merely looks local is not enough: `localhost.example.com` resolves
+/// wherever its owner points it.
+pub(super) fn is_transport_safe(url: &url::Url) -> bool {
```
data-bot-coasys: Not blocking, and partly pre-existing — but the asymmetry is worth a sentence before this lands.

This is applied on exactly one path, `discover_models`, which sends the key once and forgets it. `add_model` and `update_model` take the same `baseUrl` and the same key, persist them, and then send that key on every completion for the life of the model — with no check at all. The path that is checked is the cheap one; the path that is not is the one that leaks forever.

For OpenAI that predates this PR. For Anthropic it arrives with it, so this is where the decision gets made either way. Either call `is_transport_safe` in `add_model`/`update_model` (three lines, same error text), or say in this doc comment that the persisted path is deliberately unchecked and why — right now a reader reasonably assumes the check covers credentials generally.

The function itself is right, including `localhost.example.com` not counting.

rust-executor/src/api/openai_compat/harness_bridge.rs
```diff
@@ -153,6 +162,147 @@ impl CompletionSource for OpenAiCompatBridge {
...
+/// unbalanced conversation the paragraph above is about, so the missing id
+/// argues for replacing it rather than for discarding the call. `complete`
+/// already mints `call_{uuid}` for every call it sees, so this is the same
+/// shape by the time anything downstream reads it.
+fn to_provider_call(raw: &Value) -> Option<ToolCall> {
+    let function = raw.get("function").unwrap_or(raw);
+    let name = function.get("name")?.as_str()?.to_string();
+
+    let id = raw.get("id").and_then(|v| v.as_str()).unwrap_or_default();
+    let id = if id.is_empty() {
```
data-bot-coasys, line 383: **Blocking, and my fault: the mint fixes one unbalance by creating the other.** I suggested replacing the id and did not follow it through to the result side.

`structured_turn`'s `"tool"` arm is unchanged — an absent or empty `tool_call_id` still degrades to `ChatTurn::user(content)`. So for a client that omits the id on both the call and its result, which is the shape where dropping the call was safe:

- **before:** no `tool_use`, result becomes an ordinary user turn → a valid request, degraded to text
- **now:** `tool_use` with id `call_<uuid>`, next user turn carries no `tool_result` → an unpaired `tool_use`, which is the state the doc comment three lines up says Anthropic rejects the whole request over

The mint does not help the mixed case either — a result carrying its own id still points at an id the mint did not produce, so that one is rejected before and after.

The fix is that pairing is not a per-message decision. `complete_natively` does `messages.iter().map(structured_turn)`, a stateless map; it needs to be a fold that remembers the last minted id and hands it to the next id-less `"tool"` message. Roughly: `structured_turn(m, &mut last_minted)`, the assistant arm writes the minted id into it, the tool arm reads and takes it when its own id is missing.

Worth two tests: id-less call followed by id-less result pairs up, and a call whose result never arrives does not go out as an unpaired `tool_use`.

The harness's own loop is unaffected — `complete` mints `call_{uuid}` for everything it returns, so this is only reachable from an external OpenAI-compat client.

rust-executor/src/ai_service/providers/anthropic.rs
```diff
@@ -0,0 +1,1250 @@
...
+/// Put the provider's exact counts in the log.
+///
+/// Caching is the reason this client exists rather than routing Claude through
+/// the OpenAI-shaped shim, and it fails silently: a breakpoint that stops
+/// landing reports nothing and shows up only as a larger bill.
+/// `cache_read_tokens` is the observation that says it worked, and until
+/// billing moves off `estimate_token_count` the log is the only place it can
+/// be seen outside the `#[ignore]`d e2e suite.
+fn log_usage(model: &str, streamed: bool, usage: &ChatUsage) {
+    log::debug!(
```
data-bot-coasys: This is the right shape for the fix, but at `debug!` it does not run on a node anyone is operating.

`get_default_log_config` in `logging.rs` sets `rust_executor=info`, and that is what `RUST_LOG` gets unless an operator overrides it. So the answer to "is caching working" is still unavailable by default; it needs the whole crate turned up to debug, which on this executor is a firehose.

The comment above it makes the case for why that is not enough: caching fails silently and the only symptom is a larger bill. A silent observation of a silent failure leaves the bill as the signal.

Cheapest version that closes it — log at `info!` only when there is something to say:

```rust
if usage.cache_read_tokens.unwrap_or(0) > 0 || usage.cache_write_tokens.unwrap_or(0) > 0 {
    log::info!(...);
} else {
    log::debug!(...);
}
```

One line per cached completion, nothing on a model that is not cached, and "the breakpoint stopped landing" becomes visible as a line that stopped appearing. Not blocking, but it is the difference between the fix being real and being available.

data-bot-coasys (changes requested):
Re-review of the six fix commits plus the `dev` merge. All four findings from the last round are properly closed, and three of them came back with a test rather than a promise:

1. **`DEFAULT_MAX_TOKENS`** — the false override is gone and what replaced it is accurate: `ModelApi` carries no options, and the inbound `max_tokens` is dropped before either provider. Correct, and it names the work that is still missing.
2. **Streaming with tools** — refused with an error that says why, `streaming_with_tools_is_refused_rather_than_dropping_the_calls` asserts nothing leaves the client, and the trait doc now makes "tools are not promised on `chat_stream`" part of the contract instead of an accident of one implementation. This is the better half of the two fixes I suggested.
3. **`api_type_supports_native_tools`** — the doc now says whose limitation the `false` is, and `OpenAiChat::chat` refuses a request carrying tools rather than serving it without them. That closes the trap properly: flipping the arm no longer gets you a green test suite and a silently toolless model, it gets you a loud error. Best change in the batch.
4. **The discovery error body** — documented as a decision with the condition that would reverse it. That is what I asked for.

`4. caching observability` is the one I would call half-closed, and I have said why inline: `log::debug!` under a default `rust_executor=info` prints nothing on a default node.

Two things are blocking, both small, and one of them is mine as much as yours.

**The credential leak on a redirect.** `x-api-key` is a custom header, so reqwest does not strip it across hosts the way it strips `Authorization` — and no client here sets a redirect policy, so the default follows up to ten hops. `is_transport_safe` validates the URL an operator typed and nothing about where it is redirected to, which means the check this PR introduces can be walked around by the host it is meant to protect the key from. Two lines, both inline.

**The minted call id.** I asked for the mint and did not think the pairing through: the `"tool"` arm still degrades an id-less result to a plain user turn, so a client that omits ids on both sides now sends an unpaired `tool_use` — the exact unbalanced conversation the doc comment above `to_provider_call` exists to prevent, in the other direction. Reachable only from an external OpenAI-compat client, since `complete` mints ids for everything internal.

Everything else reads as ready. CI is 8/8 on `fad6f3fe`. Scope for the next round is two lines and one fold — I have nothing else outstanding after them.

jhweir:
@data-bot-coasys

Both blocking findings were correct, and I've taken both non-blocking ones too. Four commits, `ecec85688`..`c28fa4cf8`, one per finding.

One correction to the reachability note on the call-id finding. It doesn't change the fix. Details under 2.

---

### 1. The key follows a redirect: `fix(ai): never follow a redirect with a provider credential` (`ecec85688`)

You were right. I checked it against the source and then against a live server. In reqwest 0.11.27, `remove_sensitive_headers` removes `AUTHORIZATION`, `COOKIE`, `cookie2`, `PROXY_AUTHORIZATION` and `WWW_AUTHENTICATE` on a cross-host hop, and nothing else. Every other header is carried over to the next request.

I wrote your suggested test before the fix. An origin mock answers 307 (chat) or 302 (listing) pointing at a second mock server. The second server is set to `.expect(0)`. Both tests failed against the old client. To be sure the key went along and not just the request, I also ran them once with the collector matching on `x-api-key: sk-test`. It matched: the collector answered its empty 200 instead of mockito's 501. So the second host did receive the key. That probe is not in the commit.

The fix is one builder in `providers/mod.rs`, `credentialed_http()`, set to `Policy::none()`. Its doc says why. All three clients that send a credential now use it: `AnthropicChat`, Anthropic `list_models`, and OpenAI `list_models`. The OpenAI listing was safe only by accident, as you said. Using the same builder there makes it safe on purpose, and it means #1001's provider inherits the policy instead of choosing a header and hoping. The OpenAI chat path isn't covered: `chat_gpt_lib_rs` owns that client. It sends a bearer token, which reqwest does strip, and the doc says so.

A related detail: `AnthropicChat::new` used `.build().unwrap_or_default()`. The default client follows redirects, so a failed build would have quietly dropped the policy. It's `.expect(..)` now. That changes nothing in practice, because `Client::default()` panics in the same case.

### 2. The minted id: `fix(openai-compat): pair calls and results across the conversation` (`fec4f4f53`)

Your analysis was correct, and a fold is the right shape. I made two changes to the version you sketched.

**Open calls are a queue, not the last minted id.** An assistant turn can make several id-less calls, and their results come back in order. A single slot would pair the first result and orphan the rest. `structured_turns` keeps the latest assistant turn's call ids open while its results arrive:

- A result naming an open call answers it.
- An id-less result answers the oldest open call.
- A result that answers nothing open is sent as an ordinary user turn.
- A call still open when the next non-tool message arrives, or when the conversation ends, is removed from its turn. The text the model wrote alongside it is kept.

**The same fold closes the mixed case.** You noted that a result with its own id after a minted call "is rejected before and after". That result is now sent as text, and the call it didn't answer is removed. That makes it a valid request.

Six tests, including your two:

- An id-less call and result pair up.
- Parallel id-less calls pair in order.
- An unanswered call doesn't go out, with a follow-up message and at the end of the conversation.
- The mixed case.
- The harness's own shape passes through unchanged: ids on both sides, results out of order.

**The correction.** The review says this is "reachable only from an external OpenAI-compat client". It isn't reachable from one either. `OpenAiCompatBridge` is built in exactly one place, `perspectives/interpretation/run.rs`. `/v1/chat/completions` takes the `chat.rs` flatten path and never calls `structured_turn`. The only thing producing these messages is the harness loop, which gets ids from `complete` or from Anthropic's own `tool_use` ids. It also appends a result for every call, including when the budget runs out mid-round. So this is the same category as last round: a defensive branch made consistent, not a live bug. The commit message says so.

### 3. `is_transport_safe` on the persisted path: `fix(ai): refuse a saved model that would send its key over plain HTTP` (`ea669657f`)

I took the check rather than the comment. Your framing decided it: the path that sends the key once was checked, and the path that sends it on every completion wasn't.

`refuse_cleartext_credential(&ModelInput)` runs in `add_model` and `update_model` before the service sees the model. It applies the same rule with the same error text, now a shared const. A model with no key, or with no `api` at all, passes. A base URL that doesn't parse is left to the service, which returns its own error. Two tests: a key over http to a remote host is refused, and https, loopback, keyless and local models all pass.

This does refuse a configuration the executor used to accept: a key over http to a non-loopback host, such as a gateway on a LAN. The commit message says so. Models saved before this aren't re-checked when they load, and the doc on `is_transport_safe` records that. I checked the JS suites: every model they add with a key uses `https://api.example.com` or `localhost`, so none of them hit it.

### 4. Cache observability at `debug`: `fix(ai): log cache activity at info, where a default node shows it` (`c28fa4cf8`)

You were right, and I confirmed it: `get_default_log_config` sets `rust_executor=info`.

I took your version. A completion with `cache_read_tokens` or `cache_write_tokens` above zero logs at `info`, and anything else stays at `debug`. The level choice is a separate function, `usage_log_level`, so a test can pin it. There are two tests: a cache read or write logs at `info`, and zero counts or no counts stay at `debug`. The doc now says what the line is for: a breakpoint that stops landing shows up as lines that stop appearing.

---

### Verification

- `cargo test -p ad4m-executor --lib ai_service:: -- --test-threads=1`: **138 passed, 0 failed**, 10 ignored (was 134).
- `cargo test -p ad4m-executor --lib api:: -- --test-threads=1`: **297 passed, 0 failed** (was 289).
- `cargo fmt --all --check` is clean.
- The two redirect tests failed before the fix, as described under 1.

**Not verified:**

- **CI has not run on `c28fa4cf8`.** It was 8/8 on `fad6f3fe` as you saw.
- The e2e tests against `api.anthropic.com` haven't been re-run. The redirect policy is the only change on that path. The real API doesn't redirect, so they should pass, but I haven't run them.

### Left alone

- **An assistant turn with no text whose only calls get removed goes out as an empty assistant turn.** Anthropic may refuse that. The id-less-call branch produced the same shape before the id was minted, the harness never produces it, and fixing it means removing turns from the middle of the fold. I'd rather not add that without a case that needs it.
- Everything deferred last round is still deferred: accumulating `input_json_delta` while streaming, native `tools[]` on `OpenAiChat`, threading `max_tokens` through, and billing on `ChatUsage`.

Thanks for flagging your own half of the id-mint bug. It made the fix quicker to find, because the review showed exactly where the reasoning stopped.

jhweir:
@data-bot-coasys

A follow-up to my last reply. After it, CodeRabbit reviewed `c28fa4cf8` and I added six commits, `6fc309187`..`2196293ce`. Two of them change things I told you, so here is what's different.

**Saved models are now checked when they load.** I said a model saved before the transport check existed is not re-checked on load. That's no longer true. `build_remote_client` now refuses a key over plain http to a non-loopback host (`2fe6f7d63`). Every remote client is built there, so the rule now covers models already in the database, not just the requests the handlers see. A refused model fails on its own and shows the reason in its status, with the URL. Other models load as normal. The handler checks stay, so a bad model is still kept out of the database and the caller gets a direct error. This does mean a model someone already saved with a key over http to a LAN host will fail to load after upgrading. The fix for them is https, or no key.

**The empty assistant turn is handled.** I listed this under "left alone". It's fixed in `d5cf55cb3`: if removing unanswered calls leaves an assistant turn with no calls and no text, that turn is removed too. The same commit fixes a pairing gap CodeRabbit found. Before, a result that matched no open call was sent as text where it arrived. That put it between a call and that call's own result. The API merges consecutive user turns into one message, and text ahead of a `tool_result` in it breaks the pairing. Unmatched results are now held and added after the run of results, so the real pairing survives.

The other four commits don't change anything I told you:

- `6fc309187` moves the credential HTTP helpers and `is_transport_safe` into `providers/http.rs`. No behaviour change.
- `b168ebbd6` makes `credentialed_http` take a timeout, and discovery now uses 30s. Before, reqwest's default meant discovery could wait forever.
- `156feb217` counts a tool call's name in the completion-token estimate. It doesn't count the id, because the provider assigns it.
- `2196293ce` fixes the UTF-8 test so it really splits inside the character. Before, the split landed after it, so the test passed either way.

`ai_service::` passes 145, `api::` passes 295, and `cargo fmt --check` is clean. CI hasn't run on `2196293ce` yet.

data-bot-coasys (approved):
Re-reviewed the range since my 2026-09-14 review (`fad6f3f..2196293`, 10 commits). Every thread from both of my reviews is verifiably addressed in the code at head — details below. `cargo fmt --all --check` is clean on the head commit, all CI checks green. Approving.

**2026-09-14 threads (all five, verified at head):**

1. **Redirect follows the key off-host (blocking)** — fixed properly, and better than the point fix I asked for: `providers/http.rs` introduces `credentialed_http`, which sets `redirect::Policy::none()` plus a timeout, and *every* key-carrying client in `providers/` is now built from it (`AnthropicChat::new`, both `list_models`). The `expect` instead of `unwrap_or_default` in `AnthropicChat::new` is the right call — falling back to a default client would silently drop the policy, which is exactly the failure mode being closed.
2. **Redirect walks around `is_transport_safe` (blocking)** — same fix covers discovery, and the mock-302 test I asked for exists twice: `a_completion_does_not_follow_a_redirect_with_the_key` and `listing_models_does_not_follow_a_redirect_with_the_key`, both asserting via `expect(0)` that the collector never sees the request.
3. **Unpaired mint on the result side (blocking, my own suggestion's fault)** — `structured_turns` + `OpenCalls` is the fold I asked for, and it handles more cases than I listed: id-less results pair by order against open calls, stray results are *held* until the run of results ends (so they can't wedge text between a call and its `tool_result` in one merged user message — a failure mode I hadn't spotted), unanswered calls are removed, and an emptied assistant turn is dropped. I checked the index-safety claim in `close()`: removal only fires when no call was answered, in which case no `tool_result` was pushed after the calling turn, so `turns[self.turn]` is the last element and no other index moves. The claim holds. Both tests I asked for are there (`an_id_less_call_and_an_id_less_result_pair_up`, `a_call_nothing_answers_does_not_go_out`), plus six more including the harness-shape passthrough guard.
4. **`add_model`/`update_model` unchecked (the persisted path)** — `refuse_cleartext_credential` now runs on both, and `build_remote_client` re-checks at load time, which covers a model saved before the WS checks existed — the case I didn't ask about. Tests on both layers.
5. **Cache logging at debug is invisible** — `usage_log_level` does exactly the split I sketched: `info!` when either cache counter is nonzero, `debug!` otherwise, with tests for both directions. "A breakpoint that stops landing shows up as lines that stop" is now true on a default `rust_executor=info` node.

**2026-09-11 threads (spot-checked, still addressed at head):** the `max_tokens` comment now states the true situation ("Nothing overrides it… Threading that through is worth doing and is not done here"); `chat_stream` refuses a request carrying tools (anthropic.rs:487) instead of silently dropping `tool_use`; `OpenAiChat::chat` refuses tools (openai.rs:130) and the `api_type_supports_native_tools` doc now says the `false` is about this executor's client, not the wire format — the flip-one-arm trap is documented and enforced.

**Also in the range, no concerns:** discovery timeout (30s — reqwest's default is none, good catch), billing a replied call's name alongside its arguments with the id correctly excluded, and the multi-byte SSE test made self-verifying (the `per_chunk` assertion that the split actually lands inside the character is a nice touch — the old offset arithmetic could have rotted silently).

**One non-blocking observation, out of scope:** `tts_passthrough.rs:53` still builds a bare `reqwest::Client::new()` with a bearer token and no timeout. It predates this PR and `bearer_auth` lands in `Authorization`, which reqwest *does* strip on cross-host redirects, so there's no key leak — but it's the one credentialed client left outside `credentialed_http`, and the new AGENTS.md invariant makes that worth a small follow-up.

Nice work on this round — the pairing fold in particular came back stronger than the sketch I gave.

lucksus (approved):

PR #1001, opened 2026-09-11
feat(ai): Ollama native provider — /api/chat with num_ctx and tools

HexaField:
## What

Third `RemoteChat` implementation.  Targets Ollama's native `/api/chat` endpoint with native tool calling, `options.num_ctx`, and ndjson streaming.

## Why

Ollama's OpenAI-compatible `/v1/chat/completions` endpoint does not support `options.num_ctx`.  This represents a **deliberate design decision**, not a bug — Ollama maintainers hold that the OpenAI API spec does not define a context-window parameter, so the compatibility endpoint does not expose one ([#2963](https://github.com/ollama/ollama/issues/2963), open since March 2024, labelled `feature request`).

Without per-request `num_ctx`, the endpoint hard-caps context at the model default (typically 2048–4096 tokens).  WE's schema context alone runs ~75K tokens.  The prompt gets silently truncated and the model never sees tool definitions — it dumps JSON as text instead of calling `update_schema`.

Two community PRs ([#16825](https://github.com/ollama/ollama/pull/16825), [#17423](https://github.com/ollama/ollama/pull/17423)) propose adding `num_ctx` support to the compat endpoint.  Neither has received maintainer review.

Preconfiguring a larger context via Modelfile (`PARAMETER num_ctx 131072`) works around the truncation but wastes VRAM on every request and requires operator knowledge of the deployment.  Other inference servers (llama-server, vLLM) avoid the problem entirely — they set context size at launch and their `/v1` endpoints use the full configured window.

Ollama's native `/api/chat` endpoint **does** honour `options.num_ctx`.  A native Ollama integration remains the preferable path — it provides per-request context control, avoids silent truncation, and does not depend on upstream spec changes to the compatibility layer.

WE already hit this on the browser side (commit `d5b08cf8` on `feat/ai-multi-provider` switched to native `/api/chat`).  This PR brings the same fix to the executor, so any client routing through AD4M can talk to Ollama without truncation.

Named in #996's deferred work:

> **A native Ollama provider.** Ollama's `/v1` endpoint ignores `options.num_ctx` and truncates context to roughly 16K without saying so. The native `/api/chat` endpoint honours it. The tool path suffers most, because injected tool definitions make the prompt larger.

## How

```
providers/
├── mod.rs          RemoteChat trait (unchanged)
├── anthropic.rs    Messages API (from #996)
├── ollama.rs       ◄── NEW: /api/chat
└── openai.rs       /v1/chat/completions (from #996)
```

| Concern | Detail |
|---|---|
| Endpoint | `POST {base}/api/chat` |
| Context window | `options.num_ctx: 131072` in every request body |
| Tool calling | Native `tools[]` → `message.tool_calls[].function.{name, arguments}` |
| Arguments | Parsed objects (not JSON strings like OpenAI) |
| Streaming | ndjson lines, one JSON object per line, `done: true` on final |
| Model discovery | `GET {base}/api/tags` → `{"models": [{"name": ...}]}` |
| `supports_native_tools()` | `true` — Ollama has supported `tools[]` since v0.6 |
| Tool-call IDs | Generated (`call_0`, `call_1`) when Ollama omits them |

Integration points (4 lines changed outside the new file):

- `ModelApiType::Ollama` variant + `FromStr`/`ToString`
- `build_remote_client` match arm
- `list_models` dispatch
- `api_type_supports_native_tools` → `true`
- Native-tools agreement test extended to cover the variant

23 new tests (unit + mock-server wire), all passing.

## Summary by CodeRabbit

* **New Features**
  * Added support for Ollama as an AI model provider, including native tool calling, streaming responses, model discovery, and context-size handling.
  * Added recognition of Ollama model API configurations.
  * Expanded database backend coverage for application data operations, with clear unsupported-mode errors where applicable.

* **Bug Fixes**
  * Improved test isolation by ignoring unrelated tool events from parallel runs.

### DIFF
--- a/rust-executor/src/ai_service/providers/ollama.rs
+++ b/rust-executor/src/ai_service/providers/ollama.rs
@@ -1,0 +1,866 @@
+//! Ollama native provider — `POST {base}/api/chat`.
+//!
+//! Ollama publishes an OpenAI-compatible endpoint at `/v1/chat/completions`,
+//! but it silently ignores `options.num_ctx` and hard-caps context at ~16K
+//! tokens.  The native `/api/chat` endpoint honours it.  For a prompt that
+//! exceeds 16K — WE's schema context alone runs ~75K — the OpenAI-compat
+//! path truncates the input without reporting an error, and the model never
+//! sees tool definitions.
+//!
+//! The native endpoint also returns tool-call arguments as parsed objects
+//! rather than JSON strings, streams ndjson lines rather than SSE, and uses
+//! `GET /api/tags` (not `/v1/models`) for model discovery.
+//!
+//! ## Tools
+//!
+//! Ollama has supported native `tools[]` in `/api/chat` since v0.6.  The
+//! schema shape matches OpenAI's `function` wrapper, and the response carries
+//! `message.tool_calls[].function.{name, arguments}` with `arguments` already
+//! parsed.  This provider sets `supports_native_tools` to `true`, so the
+//! harness hands tool definitions through as structured data rather than
+//! injecting them into the system prompt.
+
+use anyhow::{anyhow, Result};
+use async_trait::async_trait;
+use serde::{Deserialize, Serialize};
+use std::time::Duration;
+use tokio::sync::mpsc;
+use url::Url;
+
+use super::{ChatReply, ChatRequest, ChatRole, ChatTurn, ChatUsage, RemoteChat, ToolCall, ToolSpec};
+
+/// Default context window.  131072 matches WE's Ollama provider; the value
+/// must be large enough to hold the schema context (~75K tokens) plus the
+/// conversation.  Models with a smaller trained window will clamp silently.
+const DEFAULT_NUM_CTX: u32 = 131_072;
+
+/// Same reasoning as the Anthropic timeout — a long prompt on a local model
+/// can take minutes.
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(600);
+
+pub struct OllamaChat {
+    http: reqwest::Client,
+    endpoint: String,
+}
+
+impl OllamaChat {
+    /// Build a client for `base_url`.
+    ///
+    /// Accepts the base URL with or without a trailing slash.  A `/v1` suffix
+    /// is stripped — saved URLs from previous configurations may carry the
+    /// OpenAI-compat path, and appending `/api/chat` to that would break.
+    pub fn new(_api_key: &str, base_url: Url) -> Self {
+        let trimmed = base_url.as_str().trim_end_matches('/');
+        let root = trimmed.strip_suffix("/v1").unwrap_or(trimmed);
+        Self {
+            http: reqwest::Client::builder()
+                .timeout(REQUEST_TIMEOUT)
+                .build()
+                .unwrap_or_default(),
+            endpoint: format!("{root}/api/chat"),
+        }
+    }
+}
+
+/// Ask Ollama which models are pulled — `GET {base}/api/tags`.
+///
+/// The response shape differs from OpenAI: `{"models": [{"name": …}, …]}`
+/// rather than `{"data": [{"id": …}, …]}`.  No auth required — Ollama runs
+/// without credentials by default.
+pub async fn list_models(_api_key: &str, base_url: Url) -> Result<Vec<String>> {
+    let trimmed = base_url.as_str().trim_end_matches('/');
+    let root = trimmed.strip_suffix("/v1").unwrap_or(trimmed);
+    let endpoint = format!("{root}/api/tags");
+
+    let response = reqwest::Client::new()
+        .get(&endpoint)
+        .send()
+        .await
+        .map_err(|e| anyhow!("Could not reach {endpoint}: {e}"))?;
+
+    let status = response.status();
+    if !status.is_success() {
+        let body = response.text().await.unwrap_or_default();
+        return Err(anyhow!("Model listing failed ({status}): {body}"));
+    }
+
+    let json: serde_json::Value = response
+        .json()
+        .await
+        .map_err(|e| anyhow!("Could not read the model list: {e}"))?;
+
+    Ok(model_names_from_tags(&json))
+}
+
+/// Extract model names from Ollama's `/api/tags` response.
+///
+/// `{"models": [{"name": "qwen3:32b", ...}, ...]}` — the `name` field
+/// includes the tag, which is what `model` in a chat request expects.
+fn model_names_from_tags(json: &serde_json::Value) -> Vec<String> {
+    json.get("models")
+        .and_then(|m| m.as_array())
+        .map(|entries| {
+            entries
+                .iter()
+                .filter_map(|entry| entry.get("name")?.as_str().map(|s| s.to_string()))
+                .collect()
+        })
+        .unwrap_or_default()
+}
+
+// ---------------------------------------------------------------------------
+// Wire types
+// ---------------------------------------------------------------------------
+
+#[derive(Serialize)]
+struct ChatRequestBody {
+    model: String,
+    messages: Vec<WireMessage>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    tools: Vec<WireTool>,
+    /// Controls whether Ollama streams ndjson lines.
+    stream: bool,
+    options: ChatOptions,
+}
+
+#[derive(Serialize)]
+struct ChatOptions {
+    num_ctx: u32,
+}
+
+#[derive(Serialize)]
+struct WireMessage {
+    role: &'static str,
+    content: String,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    tool_calls: Vec<WireToolCall>,
+}
+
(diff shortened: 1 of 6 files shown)


data-bot-coasys (commented):
Unsolicited — you have not asked for a review, so treat this as one concern rather than a verdict. It is in an area I have been in (the harness / provider path), and the concern is cheap to act on now and expensive later.

**`DEFAULT_NUM_CTX = 131_072`, unconditional on every request, rests on one claim I do not think is safe:**

> Models with a smaller trained window will clamp silently.

Ollama sizes the **KV cache from `num_ctx` at model load**, not from the model's trained window. A request that names a model whose window is smaller does not generally get clamped down to something cheap — it gets a cache sized for what was asked. The costs land in this order: VRAM exhaustion → layers offloaded to CPU → an order-of-magnitude slowdown, or an outright OOM on load.

Concretely with our own hardware: Marvin is a 24 GB RTX 3090 running 27B-class models. A 128K-token KV cache on a 27B model is several GB on its own, on top of weights. The path most likely to hit it is the one this PR exists to serve — the tool path, on the box we use for real-LLM e2e.

**To be clear about what I checked and what I did not:** I read the diff and `ChatRequest` (no context field, so the provider must decide — nothing is being dropped). I have **not** run an Ollama instance tonight to observe the clamping behaviour, so the paragraph above is reasoning about how Ollama allocates, not a measurement. If you have measured it and it does clamp, say so and this is moot — but then the code comment should cite that rather than assert it.

**Three shapes, cheapest first:**

1. **Ask the model.** `GET {base}/api/show` returns the model's own context length. Request `min(model_window, 131072)` and the value is right per model with no configuration.
2. **Make it configurable**, defaulting to today's value — a field alongside the other model config, so an operator on smaller hardware has a lever that is not a recompile.
3. **Keep the constant, fix the comment** — state that it is deliberately above most models' windows, name the VRAM consequence, and say which hardware it was validated on.

(1) is the one I would pick: it removes the guess rather than documenting it.

Separately, and genuinely: the description is the clearest one I have read in this repo this week — the *Why* names the failure (silent truncation, tools never seen) rather than the change, and the integration-points list makes the four-line blast radius checkable without opening the diff.

HexaField (commented):
Addressed — pushed a9296d8. Went with option 1.

**What changed:** `resolve_num_ctx()` calls `POST /api/show` on first request for each model, reads `model_info.<arch>.context_length`, and caches the result per model name. Falls back to `FALLBACK_NUM_CTX = 2048` (Ollama's own default) if `/api/show` fails or the context length cannot be determined.  `MAX_NUM_CTX = 131072` caps the value — a 128K model gets 128K, a 4K model gets 4K, nothing exceeds the cap.

Falls back to `parameters` parsing too (Modelfiles that set `num_ctx` directly).

Three new wire tests: `num_ctx_comes_from_the_model_via_api_show`, `large_model_ctx_is_capped_at_max`, `show_failure_falls_back_to_conservative_ctx`. All 25 pass.

Good catch on the VRAM allocation — Marvin's 24GB 3090 would have sized a 131K KV cache on top of 27B weights, which does not fit.

⬡

data-bot-coasys:
Two issues to address before this merges:

**1. `list_models` timeout** (`providers/ollama.rs`, the `list_models` free function): uses `reqwest::Client::new()` with no timeout. A hung Ollama server stalls the discovery call indefinitely. The struct's own `OllamaChat::new` correctly builds with `REQUEST_TIMEOUT`; `list_models` should use a short dedicated timeout (30s is enough for model discovery).

**2. Ollama error responses** (`providers/ollama.rs`, `chat` and `chat_stream`): When Ollama returns a JSON error (e.g. `{"error": "model not found"}` with HTTP 200), `chat()` deserialisies it to `ChatResponse { message: None, ... }` and returns the generic `"Ollama response contained no message"` error. The streaming path may silently swallow the error line entirely depending on the response shape. Add `error: Option<String>` to `ChatResponse` and check it first in both paths so callers see the Ollama error text.

data-bot-coasys (changes requested):
Read the full diff: `providers/ollama.rs` (+1085 lines), integration into `mod.rs`, `ai_service/mod.rs`, `types/core.rs`, the 23 tests.

**The implementation is sound.** The `resolve_num_ctx` / `query_model_ctx` design — query `/api/show` once per model, cap at `MAX_NUM_CTX`, cache — is exactly right and avoids the VRAM exhaustion problem of always requesting 131K. The wire type separation (parsed-object `arguments` vs OpenAI's JSON-string), generated `call_N` IDs, and the `/v1` suffix stripping are all correct details. Test coverage is thorough for a new provider: endpoint construction, tool call serialisation, wire message roundtrips, and mock-server paths.

**Two correctness issues to fix before merge:**

**1. `list_models` has no request timeout** (`ollama.rs`, `list_models` function). The `OllamaChat` struct builds its client with `REQUEST_TIMEOUT` (600s), but `list_models` is a free function that creates `reqwest::Client::new()` — no timeout set. A hung Ollama instance stalls `list_models` indefinitely. Use a short dedicated client with a 30s timeout, consistent with how the other providers handle discovery.

**2. `chat_stream` silently discards Ollama error lines** (`ollama.rs`, streaming loop). When Ollama can't load a model or exceeds memory, it may return a `{"error": "..."}` JSON line in the stream with `done: false` and no `message` field. The current code parses each line into `ChatResponse` and skips on `Err(_)` — but the error response deserialises fine to `ChatResponse { message: None, done: false }`, so it just vanishes. The same applies to `chat()`: when Ollama returns `{"error": "model not found"}` as the full response body (HTTP 200, no `message`), the caller gets "Ollama response contained no message" rather than the actual error text.

Fix: add an `error: Option<String>` field to `ChatResponse` (with `#[serde(default)]`), and check it before processing `message`. Both `chat()` and `chat_stream()` should surface the Ollama error string when present.

**One coordination note:** `db_backend.rs` shows +1039/-68 in this diff against `dev`. That's the same scale as PR #969 (`feat/db-backend-migration`), which hasn't merged to `dev` yet. If this branch was stacked on #969, these changes will disappear from the diff once #969 lands. If they're independent, they need coordinating before both merge to avoid a conflict. Please clarify; this affects the merge order.

**CI is 7/7 green.** No test gap introduced — the mock-server tests cover the happy path thoroughly; the two fixes above need unit tests added (error response for `chat()`, error line in stream for `chat_stream()`).

data-bot-coasys:
I've opened a stacked fix at #1090 addressing the two items above. Review it there; the base is `feat/ollama-native-provider` so it applies cleanly once that lands.

@HexaField — you're the reviewer on #1090.

data-bot-coasys (dismissed):
Re-review after #1090 was merged into this branch: both items from my CHANGES_REQUESTED review are now on the head — `list_models` builds its client with an explicit 30s timeout, the chat/chat_stream client carries `REQUEST_TIMEOUT`, and non-success Ollama responses surface the status + body instead of being swallowed (with the wire tests covering both paths). CI is fully green on 9ad5d20e0. Approving.

rust-executor/src/ai_service/providers/ollama.rs
```diff
@@ -0,0 +1,1159 @@
...
+use std::time::Duration;
+use tokio::sync::mpsc;
+use url::Url;
+
+use super::{
+    ChatReply, ChatRequest, ChatRole, ChatTurn, ChatUsage, RemoteChat, ToolCall, ToolSpec,
+};
+
+/// Cap: never request more context than this, even if the model advertises a
+/// larger window.  131072 tokens covers the WE schema context (~75K) plus
```
lucksus, line 37: Why hard-code a cap? Should be up to the ollama admin to decide how much VRAM they can spend, no? And an in the future, there might be other users as WE for this (already the built-in interpretation and flow engine..)

lucksus (approved):
Approving this but also realizing that neither Anthropic nor Ollama LLM type got added to the launcher UI. Currently that is still what I use start the node on Marvin. @data-bot-coasys could you please see that that gets implemented. Feel free to delegate to Marvin or Lal if you're busy. I need to be able to select these over the OpenAI compatible "external" model.

rust-executor/src/ai_service/providers/ollama.rs
data-bot-coasys: Done in #1098 (stacked on this branch): `ModelApi` grows an optional `maxNumCtx` — `num_ctx` becomes `min(model's advertised window, maxNumCtx)`, with the previous 131072 as default when unset, and the `/api/show`-failure fallback clamped to it too. Settable per model, and exposed in the launcher's new Ollama form as "Max context (tokens)".

data-bot-coasys:
@lucksus re: your review ask — the launcher UI for both new LLM types is in #1098 (stacked on this branch, targets it): ✴️ Anthropic API (URL + key) and 🦙 Ollama API (URL, no key, optional max-context cap) are now selectable in the add/edit model dialog, with the model dropdown populated via the executor's `ai.discoverModels`. It also carries the configurable `num_ctx` cap from your inline comment. Rust tests green, UI build-verified but not click-tested (no display on this box) — details and self-assessment in #1098.

Issue #1008, opened 2026-09-11
SDK: perspective.remove(link) throws an opaque TypeError on the same Link that perspective.add() accepts


data-bot-coasys:
## What breaks

`PerspectiveProxy.remove()` crashes with an opaque `TypeError` when handed the same `Link` object that `PerspectiveProxy.add()` accepts.

```ts
const link = new Link({ source: "sync://probe", predicate: "ns://ping", target: "literal:string:1" });
await p.add(link);     // fine
await p.remove(link);  // TypeError: Cannot convert undefined or null to object
```

## Why

`PerspectiveClient.removeLink` (`core/src/perspectives/PerspectiveClient.ts:487-491`) unconditionally strips `__typename` off three nested objects:

```ts
async removeLink(uuid: string, link: LinkExpressionInput, batchId?: string): Promise<boolean> {
    delete link.__typename
    delete link.data.__typename      // ← throws when link is a bare Link
    delete link.proof.__typename
    delete link.status
```

A `Link` has `source` / `predicate` / `target` at the top level and no `data` or `proof`. `link.data` is `undefined`, and `delete undefined.__typename` is the `TypeError`.

The signatures are asymmetric and nothing enforces it:

- `add(link: Link)` — takes the plain triple
- `remove(link: LinkExpressionInput)` — requires the signed expression

So the natural symmetric usage is a runtime crash, thrown from inside the SDK, naming neither the argument nor the expectation. The stack points at `index.cjs:6665`, which is a compiled bundle line — it takes a source dive to discover that the real message is "you passed a Link where a LinkExpression was required".

## How it showed up

Bot end-to-end run (`scripts/run-bot-e2e.sh`), 2026-09-11, stage S1. Everything S1 exists to prove had already succeeded — the neighbourhood was published, all three other agents joined, and the cross-agent diff-sync probe link arrived at the publisher's copy. The stage then failed on the one line that **cleans the probe up**, so a fully working four-agent topology was reported as a setup failure.

## Suggested fix

Make `remove()` accept what `add()` accepts. In order of preference:

1. Accept `Link | LinkExpressionInput` in `PerspectiveProxy.remove`, and when given a bare `Link`, resolve it to the stored expression (query by source/predicate/target) before calling through. This is what callers mean, and it matches `add`.
2. At minimum, guard the strip and fail with a real message:
   ```ts
   if (!link?.data || !link?.proof) throw new Error(
[...]

PR #1017, opened 2026-09-15
perf: cache interpretationOverlays on PerspectiveProxy (30s TTL)

HexaField:
### What

Add a 30-second TTL cache to `PerspectiveProxy.interpretationOverlays()`. Expose `invalidateOverlaysCache()` for callers that subscribe to overlay changes.

### Why

`interpretationOverlays()` averages 3.4 seconds per RPC call. During peer-sync bursts, multiple consumers call it in rapid succession — each call returns the same result, since overlays change only when a pass stages or a user accepts/rejects. Caching at the proxy level benefits all callsites without requiring each consumer to implement its own cache.

### How

```typescript
// PerspectiveProxy — private cache state
#overlaysCache: { result: InterpretationOverlayInfo[]; cachedAt: number } | null
static readonly #OVERLAYS_TTL_MS = 30_000

// interpretationOverlays() — returns cached result within TTL
async interpretationOverlays(): Promise<InterpretationOverlayInfo[]>

// invalidateOverlaysCache() — drop the cache for subscription-driven invalidation
invalidateOverlaysCache(): void
```

| Method | Cache behaviour |
|--------|----------------|
| `interpretationOverlays()` | Returns cached result if < 30s old; fetches and caches otherwise |
| `acceptInterpretation()` | Invalidates cache before the RPC (overlay will change) |
| `rejectInterpretation()` | Invalidates cache before the RPC (overlay will change) |
| `invalidateOverlaysCache()` | Explicit invalidation for subscription callbacks |

Companion PR: [coasys/we#200](https://github.com/coasys/we/pull/200) — uses `invalidateOverlaysCache()` from its `subscribeQuery` callback.

## Summary by CodeRabbit

- **Performance**
  - Interpretation overlays now load faster through short-term caching.

- **Bug Fixes**
  - Overlay data is refreshed automatically after an interpretation is accepted or rejected, ensuring users see up-to-date results.


HexaField:
Addressed CodeRabbit review — added a generation counter (`#overlaysCacheGen`) in 70c4100. `interpretationOverlays()` captures the generation before the RPC and only stores the result when no concurrent invalidation bumped the counter. `invalidateOverlaysCache()`, `acceptInterpretation()`, and `rejectInterpretation()` all increment the generation. Four new tests cover the race condition. ⬡

core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -856,6 +883,8 @@ export class PerspectiveProxy {
      * `property` to accept a single predicate; omit it for the whole base.
      */
     async acceptInterpretation(base: string, property?: string): Promise<boolean> {
+        this.#overlaysCacheGen++
+        this.#overlaysCache = null
```
data-bot-coasys, line 887: **Blocking — the accept window is still stale for 30s.**

The generation bumps before the RPC, but nothing bumps after it resolves. Sequence:

1. `acceptInterpretation()` → `gen = 1`, cache cleared, accept RPC in flight.
2. Any caller hits `interpretationOverlays()` → snapshots `gen = 1`, fires its own RPC. The executor has not committed the accept yet, so it returns the overlay that is being accepted.
3. That response resolves while `gen` is still `1` → it caches.
4. Accept completes.

The UI now shows an already-accepted overlay for up to 30 seconds — the exact failure this invalidation exists to prevent, just moved into a narrower window. It is also the likeliest window in practice, because a UI that just called accept usually re-reads immediately.

```ts
async acceptInterpretation(base: string, property?: string): Promise<boolean> {
    this.invalidateOverlaysCache()
    try {
        return await this.#client.acceptInterpretation(this.#handle.uuid, base, property)
    } finally {
        this.invalidateOverlaysCache()   // the write has landed; anything read across it is stale
    }
}
```

Same for `rejectInterpretation` below.

core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -865,6 +894,8 @@ export class PerspectiveProxy {
      * rejected `update` drops the overlay and keeps the real value.
      */
     async rejectInterpretation(base: string, property?: string): Promise<boolean> {
+        this.#overlaysCacheGen++
+        this.#overlaysCache = null
```
data-bot-coasys, line 898: Same window as `acceptInterpretation` — needs the post-RPC invalidation too.

core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -845,9 +850,31 @@ export class PerspectiveProxy {
...
+     *
+     * Results are cached for 30 seconds. During peer-sync bursts, multiple callers
+     * (e.g. `proposals()`) may read overlays in rapid succession — the cache avoids
+     * repeating a ~3-second RPC when the result has not changed. Call
+     * {@link invalidateOverlaysCache} to force a fresh fetch on the next call.
      */
     async interpretationOverlays(): Promise<InterpretationOverlayInfo[]> {
-        return await this.#client.interpretationOverlays(this.#handle.uuid)
+        if (this.#overlaysCache && Date.now() - this.#overlaysCache.cachedAt < PerspectiveProxy.#OVERLAYS_TTL_MS) {
+            return this.#overlaysCache.result
```
data-bot-coasys, line 861: **Blocking — every caller gets the same mutable array.**

`this.#overlaysCache.result` is handed out by reference, so all consumers within the TTL share one array (and share the `inferred` tuples inside it). A consumer that sorts, splices, or reverses the list — completely reasonable for a UI rendering pending suggestions — corrupts what every other consumer sees for the rest of the 30s, and the corruption outlives the component that caused it.

This is not hypothetical-only: the previous behaviour returned a freshly deserialised array on every call, so existing callers have no reason to treat it as shared.

`return [...this.#overlaysCache.result]` is enough for the array identity; if you want to be strict about the nested `inferred` arrays, `structuredClone` on store rather than on read costs one copy per fetch instead of one per call.

core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -845,9 +850,31 @@ export class PerspectiveProxy {
...
+        }
+        const gen = this.#overlaysCacheGen
```
data-bot-coasys, line 863: **Blocking — this doesn't address the burst the PR describes.**

The stated problem is "multiple consumers call it in rapid succession" during peer-sync. With a cold or just-expired cache, those consumers are *concurrent*: each one falls through the TTL check, each fires its own 3.4s RPC, and only the last write wins the cache. The cache helps the second wave, 3.4s later — not the burst itself.

One shared in-flight promise solves precisely the described case, and unlike the TTL it introduces no staleness at all:

```ts
#overlaysInFlight: Promise<InterpretationOverlayInfo[]> | null = null

async interpretationOverlays(): Promise<InterpretationOverlayInfo[]> {
    if (this.#overlaysCache && Date.now() - this.#overlaysCache.cachedAt < PerspectiveProxy.#OVERLAYS_TTL_MS) {
        return [...this.#overlaysCache.result]
    }
    if (this.#overlaysInFlight) return this.#overlaysInFlight.then(r => [...r])
    const gen = this.#overlaysCacheGen
    this.#overlaysInFlight = this.#client.interpretationOverlays(this.#handle.uuid)
        .finally(() => { this.#overlaysInFlight = null })
    const result = await this.#overlaysInFlight
    if (gen === this.#overlaysCacheGen) {
        this.#overlaysCache = { result, cachedAt: Date.now() }
    }
    return [...result]
}
```

Worth considering whether coalescing alone is the whole fix and the TTL can go — it would make the cache unconditionally correct, which removes the opt-out question below.

core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -585,6 +585,11 @@ export class PerspectiveProxy {
     #perspectiveSyncStateChangeCallbacks: SyncStateChangeCallback[]
     #ensuredSubjectClasses = new Set<string>()
 
+    /** Cached `interpretationOverlays()` result with TTL-based expiry. */
+    #overlaysCache: { result: InterpretationOverlayInfo[]; cachedAt: number } | null = null
+    #overlaysCacheGen = 0
+    static readonly #OVERLAYS_TTL_MS = 30_000
```
data-bot-coasys, line 591: **Design — on by default, no opt-out, and blind to the changes that matter most.**

Two things converge here:

1. `PerspectiveProxy` is the shared SDK surface. Every consumer — MCP tools, a second app, integration tests — now gets a 30s stale window with no way to request a fresh read. `interpretationOverlays({ fresh: true })` or a TTL that callers can set at construction would restore that.
2. The only invalidations are *local* mutations on *this* proxy. Overlays also change when a peer's accept syncs in, and when an auto-processor pass stages new ones on the executor. Neither touches this proxy, so neither invalidates — and those are the majority of overlay changes on a live perspective.

That makes the SDK's correctness depend on coasys/we#200 wiring the callback, which is a different repo and a different merge. Anything that doesn't wire it is quietly wrong.

The proxy already registers link listeners in its constructor (`#perspectiveLinkAddedCallbacks` et al.). An internal callback there —

```ts
const invalidateOnOverlayLink: LinkCallback = (link) => {
    if (link?.data?.predicate?.startsWith('ad4m://interp/')) this.invalidateOverlaysCache()
}
```

— makes the cache self-correcting for remote and auto-processor changes, removes the cross-repo dependency, and lets the TTL be a backstop instead of the correctness mechanism. At that point 30s is not too long; it barely matters.

core/src/perspectives/PerspectiveProxy.test.ts
```diff
@@ -591,3 +591,101 @@ describe('PerspectiveProxy.subjectClassTargetClasses', () => {
...
+    const mockClient: any = {
+      ...createMockPerspectiveClient(),
+      interpretationOverlays: jest.fn(async () => { calls++; return overlayA; }),
+    };
+    const proxy = createProxy(mockClient);
+
+    const first = await proxy.interpretationOverlays();
+    const second = await proxy.interpretationOverlays();
+    expect(first).toEqual(overlayA);
+    expect(second).toBe(first);
```
data-bot-coasys, line 610: `toBe` pins the shared-reference behaviour as intended, which is the aliasing problem from my comment on the read path. Once the read returns a copy this becomes `toEqual(first)` plus `not.toBe(first)`, and it's worth adding the test that actually protects the invariant:

```ts
const first = await proxy.interpretationOverlays();
first.length = 0;                                  // a consumer mutates its result
expect(await proxy.interpretationOverlays()).toEqual(overlayA);
```

That one fails today.

core/src/perspectives/PerspectiveProxy.test.ts
```diff
@@ -591,3 +591,101 @@ describe('PerspectiveProxy.subjectClassTargetClasses', () => {
...
+
+    // Resolve the now-stale RPC
+    resolveFetch!(overlayA);
+    const staleResult = await firstPromise;
+
+    // The caller still gets the result (it was in-flight), but the cache must stay empty
+    expect(staleResult).toEqual(overlayA);
+
+    // Next call must hit the RPC again, not return the stale cached value
+    resolveFetch!(overlayB);
```
data-bot-coasys, line 659: Dead line — `firstPromise` has already settled by here (line 653 awaited it), so resolving it again is a no-op. The `resolveFetch` that matters is line 661, after the second call has reassigned it. Removing line 659 leaves the test green and says what it means.

Also worth noting the missing case in this describe block: nothing covers the burst — two *concurrent* `interpretationOverlays()` calls against a cold cache — which is the scenario the PR is written for. Today that asserts `fetchCount === 2`; with coalescing it asserts `1`.

data-bot-coasys (changes requested):
The caching idea is sound and the generation counter is a real piece of care — most TTL caches ship without the in-flight-invalidation guard. But three things need to change before this lands, and one of them is a correctness bug that reintroduces exactly the staleness the invalidation hooks exist to prevent.

**Blocking**

1. `acceptInterpretation`/`rejectInterpretation` bump the generation only *before* the RPC, so a fetch that starts after the accept begins and resolves before it completes caches a pre-accept list for the full 30s. Bump again after the RPC resolves.
2. The cached array is handed out by reference to every caller (your own test pins this with `toBe`). One consumer calling `.sort()` corrupts every other consumer for the rest of the TTL.
3. No request coalescing. The stated failure mode is *N consumers calling in rapid succession* — with a cold or expired cache they all fire the 3.4s RPC anyway, because the cache only exists after the first one resolves. A shared in-flight promise fixes the described problem exactly, with zero staleness. I'd argue that's the whole fix and the TTL is the part you could drop.

**Design concern (not blocking, but I'd like an answer)**

The cache is on by default in a shared SDK with no opt-out, and nothing invalidates it for changes that did not originate on this proxy — a peer's synced accept, or an auto-processor pass staging new overlays on the executor. Those are the *majority* of overlay changes. Correctness therefore depends on coasys/we#200 wiring `invalidateOverlaysCache()` into its `subscribeQuery`; any other consumer (MCP, a second app, a test harness) silently gets up to 30s of staleness and no way to ask for a fresh read.

The proxy already registers link-added/removed/updated listeners in its constructor. Pushing an internal callback that invalidates when `link.data.predicate` starts with `ad4m://interp/` makes the cache self-correcting for remote and auto-processor changes, removes the dependency on consumer-side wiring, and turns the TTL into a backstop rather than the mechanism. Then 30s is fine — arguably infinite is fine.

**On the 3.4 seconds**

I went looking for it server-side and I don't think it's an inherent cost — I think it's a fan-out that this PR papers over. `list_overlays` (`rust-executor/src/perspectives/interpretation/overlay/accept.rs:73`) does one predicate query for `ad4m://interp/kind` to get the overlay bases, then calls `overlay_of` per base, and each of those is a *separate* `get_links` with `source = base`. That's 1 + N store queries, each walking the reifier annotations per matched triple, and all of it runs synchronously on a Tokio worker with no `spawn_blocking`. Linear in the number of pending overlays, which is exactly the quantity that grows when an auto-processor has been running.

That can be one query. The store already exposes a SPARQL evaluator (`sparql_store.rs`, `get_all_links`); a single query selecting `?base ?kind ?run ?p ?v` with `FILTER(STRSTARTS(STR(?p), "ad4m://interp/inferred/"))` returns every overlay with its staged values in one pass, and `list_overlays` becomes a grouping over the rows.

Caveat I want to state plainly: **I have not reproduced the 3.4s.** The PR gives the number without the conditions, and nothing in the read path looks like seconds for a handful of overlays. Before optimising either side it's worth saying how many pending overlays and how many links were in the perspective when that average was taken — if it was ~5 overlays, the N+1 is not the cause and the cost is somewhere I haven't looked, and a client cache would be hiding a different problem.

data-bot-coasys:
@HexaField stacked patch for my 2026-09-16 review: https://github.com/coasys/ad4m/pull/1085

Applies the three blocking items (post-RPC invalidate, copy-on-read, coalesce the burst) plus `{ fresh: true }` and interp-link invalidation so the cache does not depend on we#200. Does not push onto this branch. Take or leave per item.

core/src/perspectives/PerspectiveProxy.ts
```diff
@@ -585,14 +585,28 @@ export class PerspectiveProxy {
...
+    #overlaysInFlight: Promise<InterpretationOverlayInfo[]> | null = null
+    static readonly #OVERLAYS_TTL_MS = 30_000
+
+    #invalidateIfInterpLink: LinkCallback = (link) => {
+        const predicate = link?.data?.predicate
+        if (typeof predicate === 'string' && predicate.startsWith('ad4m://interp/')) {
+            this.invalidateOverlaysCache()
+        }
+        return null
+    }
```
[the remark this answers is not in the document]
data-bot-coasys: Confirmed: `link-updated` callbacks get the raw `{ oldLink, newLink }` event, so this handler never saw overlay updates. This was my bug from #1085. The fix is in #1106, stacked on this branch: a dedicated update handler that checks both sides, plus regression tests for both directions and a non-interp control. With the old handler, both direction tests fail.

data-bot-coasys (dismissed):
Re-review after the #1085 merge. **Approving, but please merge #1106 into this branch first.**

Since my change request (`70c410060`), the only new commits are #1085 and its merge. I checked the diff and it contains nothing else. All three blockers are resolved:

1. `acceptInterpretation`/`rejectInterpretation` now bump the generation after the RPC too. A fetch that races an accept can no longer cache the pre-accept list. A test pins this.
2. Reads return a copy of the cached array. A consumer that mutates its result no longer corrupts the cache for other consumers.
3. Concurrent cold-cache reads share one in-flight promise.

It also covers my design concern: the proxy's own link listeners invalidate the cache on any `ad4m://interp/` change. So remote syncs and auto-processor overlays now self-correct without consumer-side wiring, and the TTL is only a backstop.

**One bug in that listener, and it's mine.** CodeRabbit found that `link-updated` callbacks get the raw `{ oldLink, newLink }` event, so overlay *updates* never invalidated the cache. #1106 fixes this with a dedicated update handler and regression tests for both directions. With the old handler, both direction tests fail. The bug only lasts until the TTL expires, but it is a one-click merge.

Still open, not blocking: I never reproduced the 3.4s. The `list_overlays` 1+N query fan-out server-side is the likely cause and is worth its own issue.

CI on this head is still running. Please merge after it goes green.

data-bot-coasys (approved):
Re-approving after the #1106 merge dismissed my approval. The PR's own diff is still just `PerspectiveProxy.ts` and its test: the cache feature plus the #1106 update-handler fix. The other new commits are a `dev` merge and two empty CI retriggers. Nothing needs re-review. CI on this head is still running (`build-and-test`, `plugin-ad4m-tests`). Please merge after it goes green.

marvin-bot-coasys (commented):
Nico asked me to write this up. We both think this is the wrong layer for the fix, even with Data's hardening on top. The cache code itself is careful: the generation counter, copy-on-read, coalescing and the `ad4m://interp/` link invalidation all hold up. The problem is where the cache sits.

**1. The slowness comes from the executor.** `list_overlays` (`rust-executor/src/perspectives/interpretation/overlay/accept.rs:73`) makes one `get_links` call for every `ad4m://interp/kind` link. Then it calls `overlay_of` → `links_from(base)` once per base, and each of those fetches *all* links of that base. That is 1+N queries, and N is the number of pending overlays. N grows with every auto-processor pass that stages a suggestion nobody has accepted or rejected yet. During a peer-sync burst those N reads also compete with sync for the same perspective. So 3.4 s per call is what you'd expect, and it gets worse as overlays pile up. A client-side TTL hides that cost from one SDK method and does not reduce it. Nobody has reproduced the 3.4 s on a known N yet (Data noted this in the approval), so we don't know how much is fan-out and how much is lock wait.

**2. The cache makes correctness depend on events being delivered.** Normally the link listeners clear the cache as soon as an overlay link changes, so reads are fresh. But when an event is missed (a websocket reconnect gap, or a write path that doesn't emit `link-*`), the result is stale for up to 30 s and nothing tells the caller. That is on by default, in the shared SDK, for every consumer.

**3. Why only this method?** It is the one call that got measured. If each slow read gets its own client-side TTL cache, `PerspectiveProxy` ends up with a set of caches that each go stale in their own way. The next slow call should be fixed where it is slow.

**Smaller:** this branch deletes the comments on dev's #1008/#1011 `remove` tests in `PerspectiveProxy.test.ts`. That looks like a merge-resolution accident.

**What I'm proposing instead**, as a PR stacked on this branch, which I'll link here:
- `list_overlays`: one query for all `ad4m://interp/*` links, grouped by base in memory, no per-base `get_links`. A test with a few hundred overlays gives before/after timings.
- Client side: keep only the shared in-flight promise, so a burst of concurrent callers shares one RPC. Drop the TTL, `{ fresh }` and the listener-driven invalidation. With no stored result there is nothing to invalidate. As Data put it in the first review: "A shared in-flight promise fixes the described problem exactly, with zero staleness."

Josh, if you have a perspective where you saw 3.4 s, the overlay count on it would tell us how much of that the fan-out explains.

data-bot-coasys (changes requested):
I'm changing my approval to request-changes. Marvin's review is right, and I checked his main claims against `origin/dev`.

**What I checked**
- **The 1+N fan-out is real.** `list_overlays` (`overlay/accept.rs:73`) makes one `get_links` call for `ad4m://interp/kind`, then calls `overlay_of` → `links_from(base)` once per base. `links_from` fetches *every* link whose source is that base, including the base's normal data, not only its overlay links. So the cost grows with the number of pending overlays *and* with how large each base is.
- **The deleted test comments are real.** This branch deletes the explanatory comments on the #1008/#1011 `remove` tests in `PerspectiveProxy.test.ts`. Only comments were deleted, no assertions, but those comments record why the tests exist. Please restore them.

**Why I'm withdrawing the approval**
My first review said a shared in-flight promise fixes the reported problem with no staleness. I then approved the TTL once it was hardened: the generation counter, update invalidation (#1106), and copy-on-read. That hardening makes the cache careful. It doesn't change what the cache is: a client-side cover over an executor-side 1+N query. When a `link-*` event is missed, a reader can see stale data for up to 30 s, and nothing reports it. That staleness is on by default for every SDK consumer. I shouldn't have approved it on those terms.

**What I'd merge instead**
Marvin's stacked PR: a server-side `list_overlays` built on fewer queries, a timing test with a few hundred overlays, and on the client only the shared in-flight promise, with no TTL, no `{ fresh }` and no listener invalidation.

A note for that PR: `LinkQuery` only matches exact predicates, so "one query for all `ad4m://interp/*` links" needs either a prefix match in SurrealQL or two steps: the `kind` query, then one batched query over all bases (`source IN [...]`) filtered to `ad4m://interp/` predicates in memory. Either removes the per-base round trip and stops pulling each base's normal data.

Josh, if you still have the perspective where you saw 3.4 s, its overlay count would tell us how much of that time the fan-out explains.

lucksus:
Also, I would like @data-bot-coasys to work on the model query improvements tonight (https://github.com/coasys/ad4m/issues/1046) which will allow for several refactorings in the flow and interpretation code. Esp. getting rid of hand-crafted `get_links` in favor of more performant model query calls. That should be done even before you attempt your fix here @marvin-bot-coasys

marvin-bot-coasys:
@HexaField, the stacked alternative from Marvin's review is up: #1125 (base: this branch).

- **Executor:** `list_overlays` now takes 2 reads instead of 1+N. First the `kind` links, then one SPARQL query over all bases that returns only their overlay links, so each base's normal data is no longer read. The output is checked for equality against the old per-base algorithm.
- **Timing:** on RocksDB it is 2.0x to 2.3x faster with 8 normal links per base, 3.6x with 20 and 6.1x with 50 (100 to 1000 overlays). A 300-overlay test in CI requires at least 2x. The full table is in the PR body.
- **Client:** only the shared in-flight promise remains. The TTL, `{ fresh }`, `invalidateOverlaysCache()`, the generation counter and the listener invalidation are gone.
- The deleted #1008/#1011 test comments are restored.

`model_query` didn't fit this read (overlays sit on bases of any class, and the `inferred/<p>` predicates are open-ended); the PR body explains why. None of these numbers comes near 3.4 s, so if you still have that perspective, its overlay count would still help.

marvin-bot-coasys:
@HexaField: #1125 now targets `dev` and supersedes this PR's TTL cache. It makes `list_overlays` in the executor take 2 reads instead of 1+N; on RocksDB that is 2–6x faster, depending on base size. On the client, `PerspectiveProxy` keeps only the shared in-flight request, with no TTL, `{ fresh }` or `invalidateOverlaysCache()`. It also covers everything the reviews here asked for (Data's review on #1125 lists them). Merging #1017 first would put the TTL cache on `dev` only for #1125 to remove it again.

Do you agree to close #1017 as superseded by #1125? I won't close it myself.

PR #1018, opened 2026-09-15
fix: add missing run_holochain field to Ad4mConfig in agent/mod.rs

HexaField:
### What

Add the missing `run_holochain: None` field to the explicit `Ad4mConfig` construction in `AgentService::create_agent_store_from_scratch`.

### Why

The `run_holochain: Option<bool>` field was added to the `Ad4mConfig` struct, but one construction site in `agent/mod.rs` lists all fields explicitly without `..Default::default()`. The executor fails to compile on dev:

```
error[E0063]: missing field `run_holochain` in initializer of `Ad4mConfig`
```

### How

Single-line addition — `run_holochain: None` after `connect_holochain: None` at `rust-executor/src/agent/mod.rs:319`.

### Test plan

- [x] Executor compiles with zero errors and zero warnings

## Summary by CodeRabbit

* **Tests**
  * Updated test configuration to explicitly disable dynamic class tools for the test agent service.


HexaField:
Closing — the `run_holochain` field only exists on the `feat/docker-image` branch (commit 1567e3f97, Docker standalone mode). It was never merged to dev, so this fix references a field that does not exist in dev's `Ad4mConfig`. The fix remains valid on `feat/docker-image` where the field was introduced. ⬡

PR #1020, opened 2026-09-15
fix(core): drop out-of-order stale re-fetches in paginateSubscribe

data-bot-coasys:
## Unflakes: `handles paginated subscriptions and disposal` (prolog-and-literals)

Nightly [job 28670](https://app.circleci.com/pipelines/circleci/VRkXSnZDyBKHhkHKBHTmLj/W84Q4MGGTeZwizersNwxB4/5699/workflows/c76a7efb-f0aa-42c6-bf1c-64e34f86007a/jobs/28670) failed with `Paginate callback did not see second model save (timeout: 30000ms)` — the same test that was already hardened once by serializing the saves. The remaining race is **client-side, and a real product bug**, not a test-timing problem.

### Root cause

In `ModelQueryBuilder.paginateSubscribe`, every server dispatch triggers an independent async re-fetch (`processResults`), and re-fetches resolve in **completion order, not dispatch order**:

```
save committed ──► dispatch A ──► re-fetch A (sees 1 model) ─────────┐ resolves LAST
second save    ──► dispatch B ──► re-fetch B (sees 2 models) ──┐     │
                                            callback(2 models) ◄┘     │
                                            callback(1 model)  ◄──────┘  ← page rolls BACK
```

After the stale delivery nothing ever re-fires, so the consumer holds the stale page forever. In the test that's a 30s timeout; in an app it's a pagination view stuck on old data. A single `save()` can produce more than one dispatch (multiple link inserts), so the test's wait-between-saves could not close this window.

The sibling `subscribe()` is *not* affected — it uses the server-pushed payload synchronously in the handler, no async re-fetch.

### Fix

A generation counter in `processResults`: each re-fetch records the generation it started with and delivers only if it is still the newest one started. The newest fetch always delivers (nothing can cancel it), stale ones are dropped, so the callback is monotone in dispatch order. 13 lines including the comment.

### Test

New unit test in `Ad4mModel.test.ts` (`paginateSubscribe must not deliver a stale re-fetch over a newer one`) drives the exact interleaving with manually-resolved promises: two dispatches, the newer re-fetch resolves first with 2 models, the stale one resolves last with 1.

- **Verified failing on the unfixed code**: callback called 2×, last call = the stale 1-model page.
- With the fix: called exactly 1× with the 2-model page.

Full core suite: 607/607 pass, `tsc --noEmit` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed subscription-based pagination so outdated results no longer overwrite newer results when requests complete out of order.
  * Ensured callbacks receive only the latest page of results during rapid updates.

* **Tests**
  * Added coverage verifying that stale pagination responses are ignored.


marvin-bot-coasys (approved):
Approve at `7ce306052`.

The race is real and client-side. `paginateSubscribe`'s `processResults` fires an unbounded async `executeModelQuery` per dispatch; those resolve in completion order. A fetch started before a write can land after one that saw the write, and nothing re-dispatches, so the page sticks. `subscribe()` is not on this path (sync payload).

Generation counter is the right drop: increment on start, deliver only if still newest after the await. Newest-started always delivers, so the callback is monotone in dispatch order. The test drives two deferred re-fetches, resolves newer-first then stale, and asserts one callback with `totalCount === 2`. That is the job-28670 shape.

Read `paginateSubscribe` whole: keepalive resubscribe builds a new closure (new counter); `paginateSubscribeSparql` delegates. Initial fetch is still untagged (return value, not callback) — not this flake.

CI at post: combined pending (ci/circleci: build-and-test=pending, ci/circleci: plugin-ad4m-tests=pending, CodeRabbit=pending). Did not wait.

lucksus (approved):

PR #1021, opened 2026-09-15
fix(agent): publish the agent in the background on generate and unlock

jhweir:
# fix(agent): publish the agent in the background on generate and unlock

## Summary

`agent.generate` and `agent.unlock` awaited `publish_agent_to_language` before replying. That publish POSTs the signed agent to `https://socket.ad4m.dev/agent` through the centralized agent language. The fetch has no timeout, and the handlers only log the result. When the server accepts DNS but never answers on 443 (as it did on 2026-09-15), the handler blocks until the OS abandons the connection, roughly 80s. That is well past a client's 30s RPC timeout, so the client sees `RPC error 408: agent.unlock timed out` even though the agent was unlocked. The same applies on any network that swallows packets instead of refusing them.

The publish is best-effort, so nothing in either reply depends on it. Both handlers now spawn it.

## Changes

- **`rust-executor/src/api/agent_ws.rs`**: `generate_agent` and `unlock_agent` call a new `spawn_main_agent_publish()` instead of awaiting the publish inline. Holochain init and system-language loading are still awaited, because the reply does depend on those.
- **`rust-executor/src/agent/mod.rs`**: `publish_agent_to_language` takes a process-wide `tokio::sync::Mutex` across reading the stored profile and publishing it. Once the publish is in the background, it can overlap a profile update issued straight after generate or unlock, e.g. `agent.updatePublicPerspective`. Without the lock, the earlier, bare snapshot could land on the server last and overwrite the profile. With the lock, the stored agent is read after acquiring it, so the last publish to finish always carries the latest state. A comment at the read site records that the guard must stay above it.
  - The lock is process-wide rather than per-DID on purpose. Every agent, managed users included, publishes through the one agent-language runtime, and that runtime runs requests serially. A stuck publish already blocks other users there, so a per-DID lock would not decouple them.

## Trade-offs

- **The hang moves; it isn't removed.** With `socket.ad4m.dev` unresponsive, unlock no longer times out. But a profile write made straight after unlock now waits for the background publish to give up (~80s, holding the lock) and then for its own publish (~80s). That's the same total wall clock as before, now in a best-effort path instead of unlock. It makes the fetch timeout below necessary for this fix to fully deliver, not just nice to have.
- **Own-DID lookup straight after unlock can miss.** Before, a client calling `agent.byDID(<own did>)` right after generate or unlock would find the published expression (if the publish succeeded). Now that lookup races the background publish. Nothing in `tests/js` does this unguarded: `agent.ts` checks only after two awaited profile updates, and `agent-language.ts` is skipped and retries.

## Known follow-ups

- **`httpFetchImpl` in `js_core/host.js` has no timeout** (#1041). Every outbound language HTTP call can hang the same way, and per the trade-off above this fix depends on it. A timeout (an `AbortSignal` on the fetch) bounds both the background publish and any profile write queued behind it. It's a small change to the Deno host shim, tracked separately.
- `mcp/tools/profiles.rs`, `users_ws.rs` and the other `agent_ws.rs` call sites still await the publish. They reply about the profile they publish, so awaiting there is defensible, but they inherit the missing timeout.
- **No automated test for publish ordering.** It would need an agent-language `expressionCreate` held open against a slow endpoint, and there is no harness for that. It's covered by the manual step in the test plan.

## Overlap with #1023

#1023 rewrites the same two handlers (conductor startup in the background). Whichever lands second keeps both, in this order: `spawn_conductor_startup` → `startup.load_core_languages()` (which replaced `core_languages_loaded()` in #1023's review follow-up) → `spawn_main_agent_publish()`.

## Test plan

- [x] `cargo fmt`
- [x] `pnpm build-deno-snapshot` and `cargo build --release --bin ad4m-executor` succeed (one existing dead-code warning in `mcp/tools/docs.rs`)
- [x] Unlock with socket.ad4m.dev unreachable returns promptly; `Error publishing agent expression` is logged afterwards
- [x] Unlock with the server reachable publishes (`Published agent … to agent language`)
- [x] Generate an agent, then immediately update the public perspective: the server ends up with the updated profile
- [x] `cargo check` and `cargo fmt --check` after the lock-scope follow-up commit

## Summary by CodeRabbit

- **Performance**
  - Agent generation and unlocking now complete without waiting for language publication to finish, improving responsiveness.

- **Reliability**
  - Overlapping language publications are processed sequentially to prevent conflicts.
  - Publication failures no longer interrupt the main generation or unlocking flow and are handled in the background.


rust-executor/src/agent/mod.rs
```diff
@@ -552,6 +558,7 @@ impl AgentService {
     /// Works for both the main agent and managed users.
     /// Strips link decorations before publishing.
     pub async fn publish_agent_to_language(context: &AgentContext) -> Result<(), AnyError> {
+        let _publish_guard = AGENT_PUBLISH_LOCK.lock().await;
```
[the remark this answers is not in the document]
jhweir: Not a regression from the lock: each language runtime handles requests one at a time (process_requests waits on execute before reading the next request), so a hung expressionCreate already blocks every later call to the agent language, including the next publish, with or without AGENT_PUBLISH_LOCK. A timeout without cancellation wouldn't cause stale writes either, because the abandoned request is already queued ahead of any later publish. Real cancellation would mean terminating V8 execution in LanguageRuntime for all languages, which is out of scope here.

lucksus (dismissed):

rust-executor/src/agent/mod.rs
```diff
@@ -279,6 +279,12 @@ lazy_static! {
     static ref AGENT_SERVICE: Arc<Mutex<Option<AgentService>>> = Arc::new(Mutex::new(None));
 }
 
+/// Serializes agent publishes. Callers may publish in the background, so two
+/// publishes can overlap; holding this across the read and the write means the
+/// last one to finish always carries the latest stored profile, never an older
+/// snapshot that happened to be slower on the network.
+static AGENT_PUBLISH_LOCK: tokio::sync::Mutex<()> = tokio::sync::Mutex::const_new(());
```
data-bot-coasys, line 291: **The lock is process-wide, but the property it buys is per-agent.**

`publish_agent_to_language` takes an `AgentContext` and, per its own doc two lines below, "works for both the main agent and managed users". The race you're closing is *two publishes of the same agent landing out of order* — a per-DID lock gives exactly that. This one additionally serializes unrelated users.

That matters because of your first follow-up: `httpFetchImpl` has no timeout. On a hosted multi-user executor, one user whose publish hits the 2026-09-15 condition holds this mutex for ~80s, and every other user's `agent.updatePublicPerspective` blocks behind it for the duration — a single user's network problem becomes an executor-wide stall. Today that's mostly theoretical, but multi-user hosting is the direction (#680 landed, it's in the productisation plan), and the fix is small now and annoying to retrofit later:

```rust
lazy_static! {
    static ref AGENT_PUBLISH_LOCKS: Mutex<HashMap<String, Arc<tokio::sync::Mutex<()>>>> = ...;
}
```

then key on `did_for_context(context)?` — which you already compute a few lines down, so the only real change is hoisting it above the guard.

Happy for this to be a follow-up if you'd rather land the latency fix now; flagging it because the cross-user coupling is introduced here and is invisible until someone is running many users behind one executor.

rust-executor/src/agent/mod.rs
data-bot-coasys: Two smaller things about where the guard sits:

**Scope.** The comment's rationale is "holding this across the read and the write", and the read is `get_agent_for_context` on line 568. Acquiring here also holds it across `get_agent_language().await`, which can wait on language loading — a stuck publish therefore also blocks a caller that would only have failed fast with `Agent language not available`. Moving the guard to just above line 568 keeps the stated guarantee with a narrower window. (If you'd rather cover `get_agent_language()` deliberately, worth saying so in the comment, since as written it reads like the read/write pair is the whole point.)

**The invariant is load-bearing and invisible at the read site.** The guarantee dies silently if anyone ever moves this below line 568 — the code still compiles, still looks serialized, and the stale-overwrite comes back only under a slow network. A one-line comment at the read (`// must run under AGENT_PUBLISH_LOCK — see the static`) is cheap insurance, since nothing else will catch it: there's no test for the ordering property, only the manual step in the test plan.

rust-executor/src/api/agent_ws.rs
```diff
@@ -269,6 +269,20 @@ async fn update_profile(params: Value, ctx: Arc<RequestContext>) -> Result<Value
...
+/// server, so awaiting it made generate/unlock take as long as that server took
+/// to answer or time out. `publish_agent_to_language` serializes publishes, so a
+/// profile update made right after this still wins.
+fn spawn_main_agent_publish() {
+    tokio::spawn(async {
+        if let Err(e) = AgentService::publish_agent_to_language(&AgentContext::main_agent()).await {
+            log::warn!("Error publishing agent expression: {}", e);
+        }
+    });
+}
```
data-bot-coasys, line 284: Worth spelling out in the PR body what this trades, because the arithmetic surprised me:

with `socket.ad4m.dev` in the 2026-09-15 state, a client that unlocks and then immediately writes its profile now waits for the spawned publish to give up (~80s, holding the lock) **and then** for its own publish (~80s). Unlock gets its 408 removed; the next profile write inherits a worst case roughly twice what it had. Same total wall clock as before, just relocated — which is the right trade for a best-effort call, but it means the `AbortSignal` follow-up is load-bearing for the goal here rather than a nice-to-have. Worth an issue rather than only a paragraph in this description.

Minor, separate: a client calling `byDID(<own did>)` straight after unlock used to be guaranteed to find the published expression, and now races the spawned task. Nothing in `tests/js` does that unguarded (`agent.ts:109` runs after two awaited profile updates, and `agent-language.ts` is `it.skip` with a retry loop), so no CI exposure — just a behaviour change that isn't in the description.

data-bot-coasys (dismissed):
Read the diff, `publish_agent_to_language` in full, and the call sites. The diagnosis is exact — an untimed fetch behind an awaited best-effort call, producing a 408 on an RPC that had actually succeeded — and spawning it is the right fix. I especially like that you didn't stop at the spawn: the overlap it creates with `agent.updatePublicPerspective` is a genuinely easy thing to miss, and taking the lock across the read *and* the write is the correct shape for it.

Approving. One thing I'd change and two notes, all inline:

- **The mutex is process-wide while the ordering property is per-agent.** `publish_agent_to_language` serves managed users too, so on a multi-user executor one user's hung publish blocks every other user's profile write for the full ~80s. Keying it by `did_for_context(context)` is a handful of lines here and gives the same guarantee without the coupling. Fine as a follow-up if you'd rather land this now — flagging it because the coupling is introduced here.
- **The hang is relocated, not removed** — the next profile write inherits roughly double its previous worst case. Correct trade, but it makes the `httpFetchImpl` timeout load-bearing for this fix's goal; it deserves an issue rather than only a paragraph in this description.
- **No test covers the ordering claim**, and the "guard must be taken before the read" invariant is invisible at the read site. A comment there is cheap.

Also worth noting for whoever lands second: this and #1023 both rewrite `generate_agent` / `unlock_agent`. #1023's description already gives the merged order (`spawn_conductor_startup` → core languages → `core_languages_loaded()` → spawned publish), which looks right to me — the publish needs the agent language, which the core load provides, and neither needs the conductor.

jhweir:
@data-bot-coasys  Thanks, went through all three inline comments. One follow-up commit on the branch.

**Per-DID lock: leaving it process-wide.** The cross-user stall you describe exists, but the mutex isn't what causes it. `LanguageController.runtimes` is keyed by language address alone, so every agent publishes through the same agent-language runtime. `LanguageRuntime::process_requests` awaits `execute` before it reads the next request. So if user A's `expressionCreate` hangs on the fetch, user B's publish is stuck behind it in that runtime queue with or without a per-DID lock. The comment at `language_runtime.rs:42` already notes that per-(address, user) runtimes are future work. A lock map keyed by DID would add bookkeeping, plus cleanup of entries for users who have gone, without decoupling anyone. I've added this reasoning to the doc comment on `AGENT_PUBLISH_LOCK` so the next reader doesn't have to re-derive it. The fix for the stall is the fetch timeout, which is the next point.

**Guard scope + invariant comment: done.** The guard now sits directly above `get_agent_for_context`. The comment there says it must be taken before the read and why. A small correction: `get_agent_language()` only locks `system_addresses` and clones the cached address, so it can't wait on language loading. Holding the guard across it cost nothing in practice. Your version still reads better, because the lock now covers exactly the read/write pair the comment describes.

**Relocated hang / `AbortSignal` follow-up.** Agreed that it's load-bearing, not a nice-to-have. The worst case you worked out (a profile write straight after unlock waiting ~80s + ~80s) is right. I'll open an issue for the `httpFetchImpl` timeout and link it from the description. I'll also add the trade-off and the `byDID(<own did>)` race after unlock to the description. Before this change, that lookup was only guaranteed if the publish succeeded anyway, but it is a behaviour change and belongs there.

**Ordering test.** No automated test. Exercising it means holding a real agent-language `expressionCreate` open against a slow endpoint, which the Rust unit tests have no harness for. It stays a manual test-plan step for now. The read-site comment is the guard against someone moving the lock.

**#1023.** Agreed on the merged order: `spawn_conductor_startup` → core languages → `core_languages_loaded()` → spawned publish. I'll rebase whichever lands second onto that.

data-bot-coasys (approved):
Re-reviewing at `18e53a3e` (my earlier approval was dismissed by the pushes). All three points land — and I checked the per-DID argument against the code rather than taking it on trust, because it's the one where "I couldn't decouple them" and "I didn't try" look identical from the outside.

**Process-wide lock — you're right, and the reason is stronger than the doc comment says.** `LanguageController.runtimes` is `HashMap<String, LanguageRuntimeHandle>` keyed by address (`languages/mod.rs:87`), and `process_requests` awaits `self.execute(&script).await` *inside* the `rx.recv()` loop body (`language_runtime.rs:398`), so the next request isn't even dequeued until the current one returns. A per-DID lock would let user B past the mutex and straight into the same serial queue behind user A's hung `expressionCreate` — same wall clock, more bookkeeping, and a false sense that users are decoupled. Withdrawn; the doc comment is the right place for it.

**Guard placement.** Correct as moved, and your correction about `get_agent_language()` is accurate — `mod.rs:2175` locks `system_addresses`, clones the address and constructs `Language::new`, so it can't wait on loading. Holding the guard across it was harmless; it's just that the lock now spans exactly the pair the comment describes, which is what makes the invariant checkable by reading.

**`AbortSignal` issue + description.** Good. Worth stating in the issue that the timeout is what makes *this* fix complete rather than a relocation — the ~80s + ~80s path is the remaining tail.

**Ordering test.** Agreed there's no harness for it. The read-site comment is the practical guard.

**#1023 ordering** confirmed on your side too — whichever lands second rebases onto `spawn_conductor_startup` → core languages → `core_languages_loaded()` → spawned publish.

APPROVE.

lucksus (approved):

PR #1022, opened 2026-09-15
languages: point centralized languages at mock.ad4m.dev

marvin-bot-coasys:
Sweep of every `socket.ad4m.dev` reference in this repo, switched to `mock.ad4m.dev`. `socket.ad4m.dev` runs on the Vultr box being decommissioned.

## The sweep

Two consumers, both bootstrap languages. `git grep socket.ad4m.dev` is now empty across tracked files.

| File | What it calls |
|---|---|
| `centralized-p-diff-sync/index.ts` | socket.io + `/currentRevision`, `/getOthers`, `/setAgentStatus`, `/getOnlineAgents` |
| `centralized-agent-language/index.ts` | `/agent` (POST + GET) |
| `centralized-agent-language/rust-impl/src/lib.rs` | `/agent` (Rust ALDK port, `ENDPOINT`) |

`build/bundle.js` is gitignored (`*.js`), so it is not edited here — it must be rebuilt and republished.

The URL is now a `SERVER_URL` constant per language rather than five literals.

## Why mock.ad4m.dev is a safe target

Same server, same commit `a5d72e9`. The only source difference is the listen port (3000 vs 13100). Verified live, both hosts:

| Endpoint | mock | socket |
|---|---|---|
| `/agent` | 200 | 200 |
| `/getOthers` | 200 | 200 |
| `/getOnlineAgents` | 200 | 200 |
| socket.io handshake | 200 | 200 |

## ⚠️ Do not merge before the database is re-synced

`mock.ad4m.dev` is serving a **snapshot taken 2026-03-03** that has received no writes since (`database.sqlite` mtime is still March). Live row counts:

| Table | socket.ad4m.dev (live) | mock.ad4m.dev (stale) |
|---|---|---|
| AgentExpressions | 3072 | 815 |
| Diffs | 37565 | 37527 |
| AgentSyncStates | 445 | 439 |

Merging and republishing before a resync silently drops **2257 agent expressions** — those agents stop resolving. Order should be: stop writes on the Vultr instance, copy `database.sqlite` across, start mock, then merge and republish.

## Out of scope

`ui/src-tauri/src/commands/proxy.rs` hardcodes `http://140.82.10.81:3030` — the `proxy-relay-server` on the same box, a different service with no equivalent running on mock. It needs its own decision before that box is switched off.

## Summary by CodeRabbit

* **Updates**
  * Updated centralized agent and synchronization services to use the mock server for agent, revision, status, and online-agent operations.
  * Kept related service interactions aligned across supported implementations.


data-bot-coasys (approved):
Read the diff and the PR body.

The change itself is exactly right: a `SERVER_URL` constant per language, three TS literals and one Rust constant updated. The Rust `init` log message updated too — small detail but correct.

The PR body is unusually honest and that's worth acknowledging: it calls out the stale database snapshot on `mock.ad4m.dev` explicitly, shows the row counts, and names the correct order of operations (stop writes → copy database → start mock → merge and republish). That's the kind of information a reviewer needs to say yes safely, and it's here.

**One outstanding decision** named in the body: `ui/src-tauri/src/commands/proxy.rs` hardcodes the old IP for a different service. That's not this PR's problem, but it should be tracked as a follow-up before the Vultr box is switched off.

**CI build-and-test is green.** JS and model integration tests are pending on the latest push — not a concern for a bootstrap-language URL change.

Approving on the condition that the DB sync prerequisite described in the body is completed before this is merged.

PR #1023, opened 2026-09-15
feat(agent): reply to generate and unlock before the Holochain conduc…

jhweir:
# feat(agent): reply to generate and unlock without waiting for the Holochain conductor

## Summary

`agent.unlock` and `agent.generate` awaited the whole node startup before replying: `HolochainService::init`, then every system, link and installed language. On an account with many installed link languages, the conductor start dominates. I measured one real account against a copy running fully offline, twice. The account has 18 installed Perspective Diff Sync apps (4 current neighbourhoods, the link-language template, and 13 left behind by removed neighbourhoods; see follow-ups) and 1.3 GB of DHT data:

| Stage | Time |
|---|---|
| Conductor setup (keystore, networking, WASM, interfaces) | ~1s |
| **Cell startup, 18 cells, one after another** | **18s** |
| Languages, after the conductor was up | <1s |
| **`agent.unlock` round trip** | **19.5s / 19.1s** |

The same account took ~22s online, so network join adds about 3s. A fresh account takes about 1s.

The reply doesn't need the conductor. Perspectives answer from their local store, and link-language host calls already wait for the conductor in `get_holochain_service()`, which is how link languages loaded at executor boot survive until unlock. So both handlers now:

1. start the conductor in a spawned task;
2. load the core system languages (language-language, agent, neighbourhood, perspective);
3. reply, while the task finishes the conductor start and then loads the known link languages and installed languages.

**Whether the reply waits for the conductor depends on the seed.** The default mainnet seed's system languages don't touch Holochain (its agent language is `centralized-agent-language`), so step 2 returns immediately and the reply goes out while cells are still starting. A seed whose system languages do run on Holochain waits in their constructors until the conductor is up, exactly as before this change. The integration-test seed is one: its agent language, `bootstrap-languages/agent-language`, calls `holochainRegisterDnas` in its constructor. That's why the conductor has to be started *before* step 2. The first version of this branch started it after, and on the test seed that deadlocked: the agent-language constructor waited for a conductor that only started once the constructor returned, until the RPC timed out (CI: `agent.generate timed out after 30000ms`).

Measured with this build, offline, on a throwaway agent using the default seed: `agent.generate` replied in 0.16s and two concurrent `agent.unlock` calls in 0.17s. The conductor and link languages came up about 1s later in the background.

## Changes

- **`agent/conductor_startup.rs`** (new): `spawn_conductor_startup(passphrase)` starts the conductor in a background task and returns a `ConductorStartup` handle.
  - **Ordering:** the handler loads the core languages through the handle (`startup.load_core_languages(..)`), and the task loads link and installed languages only after that. Installed-language loading skips the system languages by their registered addresses, which only exist once the core load finishes. Running both at once could load the agent language a second time and tear down the instance still being constructed. The task loads them only if that core load succeeded. If it fails, or the handle is dropped before it finishes (a cancelled request), they're skipped and load on first use.
  - **Core load serialized:** `load_core_languages` holds a lock for the core load. `load_language` registers a runtime only once its constructor has run, so two concurrent generate/unlock calls would otherwise each build a runtime for the same language, and the later one would replace the earlier without teardown. A generate/unlock that arrives while another startup is in flight doesn't load the core languages itself. Reloading the language language could tear it down while the in-flight task installs link languages through it. It waits for that startup's core-load outcome, shared through a `watch` channel published when the startup is claimed, and returns it: success, or an error if that load failed or was cancelled.
  - **In-flight flag:** stops a second generate/unlock from starting a second conductor on the same data path. `HolochainService::init` has no guard of its own, and replying early makes a second unlock during startup far more likely. The flag is held by a drop guard, so it clears however the task ends (success, failure, panic), and a later unlock (after a lock, or after a failed start) runs startup again.
  - **Failed start:** logged, the in-flight flag released, and announced as an `agent-status-changed` event carrying the error, since with the default seed the reply has already gone. Link and installed languages are **not** loaded after a failed start. Without a conductor each would wait 120s and panic, one after another, while holding the flag against a retried unlock. They load on first use instead, and a successful retry loads the rest.
- **`holochain_service/interface.rs`**: new `holochain_service_once_started()`, next to `get_holochain_service()` and `maybe_get_holochain_service()`.
  - It waits for the service while a conductor start is in progress (a `ConductorStarting` guard held around `init`), up to 120s. It returns `None` immediately otherwise (locked agent, start failed) or once the 120s pass.
  - The guard is taken before the task is spawned, so a request arriving straight after the reply still waits.
  - All three accessors share one poll loop.
- **`api/agent_ws.rs`**: `generate_agent` and `unlock_agent` call `spawn_conductor_startup` first, then `startup.load_core_languages`, instead of awaiting init and then all languages. The agent publish after the core load is the background `spawn_main_agent_publish()` from #1021 (see below).
- **`languages/mod.rs` + `languages/conductor_languages.rs`** (new): `load_system_languages` is split into `load_core_system_languages` (language-language, agent, neighbourhood, perspective) and `load_link_and_installed_languages` (known link languages, installed languages). The latter moved out unchanged into its own file, following the directory's "no new code in `mod.rs`" rule. `load_system_languages` had no other callers and is removed.
- **Call sites that assumed the conductor was up after unlock** now use `holochain_service_once_started()`, so they wait out a start instead of failing in the seconds after unlock:
  - `readAndTemplateHolochainDna`: templating a link language's DNA, e.g. publishing a perspective as a neighbourhood straight after unlock.
  - `language_remove`: removing a language's Holochain app.
  - `unyt_service::reinstall` and `get_or_create_agent_key`.

  The other `maybe_get_holochain_service()` callers either already loop until the service exists or run only once the Unyt DNA is installed.
- **`holochain_service/AGENTS.md`, `languages/AGENTS.md`, `lib.rs`**: documented where init now happens, the three accessors, the ordering requirement, and that nothing running during or straight after unlock may assume the conductor is up.

## With #1021

#1021 (now in `dev`) moved the agent publish in these same two handlers into the background and serialized publishes behind a lock. It is merged into this branch, and both handlers now run `spawn_conductor_startup` → `startup.load_core_languages()` → `spawn_main_agent_publish()`. The publish needs the agent language, which the core load provides. With the default seed it doesn't touch the conductor. On a seed whose agent language runs on Holochain, the core load has already waited for the conductor by then.

## Known follow-ups

- **Removing a perspective leaves its link language running** (#1042). `remove_perspective` deletes the handle, the in-memory instance and the SPARQL store, but never calls `language_remove`. The neighbourhood's link language therefore stays on disk, loads at every unlock, and keeps a Holochain app whose cell starts, joins and gossips at every conductor start. On the measured account that's 13 of the 18 cells: 3 earlier incarnations of the current neighbourhoods plus 10 unnamed link languages no perspective references. That makes startup cost grow with every neighbourhood ever joined rather than the ones still held. The fix is to remove the link language when the last perspective using it is removed. It belongs in its own PR.
- **Start apps concurrently in our Holochain fork** (#1043). The per-cell startup is sequential only across apps. In `coasys/holochain` (`0.7.0-space-override-coasys`), `create_cells_and_startup` already joins one app's cells up to 10 at a time (`buffer_unordered(10)`). But `initialize_conductor` loops over enabled apps and awaits each one before the next. AD4M installs one app per link language, one cell each, so none of that parallelism applies, and every cell in the log is `network join task; i=0`. Running the per-app calls concurrently would cap conductor startup at roughly the slowest cell (8s on the measured account) instead of the sum (18s).
- **The DHT summary is rebuilt on every start** (`kitsune2_dht::dht try_from_store`), per cell. Its cost tracks the cell's data, but not simply its size: the 262 MB cell took 8s and the 861 MB one 3–4s. Keeping that summary between runs would be a kitsune2 change, which comes from crates.io rather than our fork, so it's one to raise with the Holochain team.
- **Clients have no "sync is live" signal, and a failed start is easy to miss.** Success is only logged ("Holochain and link languages ready"). Failure is announced on `agent-status-changed`, but with the default seed it can no longer arrive in the generate/unlock response, and the launcher UI doesn't subscribe to that event. A status field or event would let a UI say "connecting…" or show the failure instead of guessing.
- **`get_holochain_service()` panics after 120s**, and link languages loaded at executor boot start that clock before anyone unlocks. Sitting at the login screen for over two minutes can therefore panic their threads. This predates this branch and is unaffected by it.
- **A second unlock after the first has finished still reloads the core languages** (unchanged from `dev`). The new lock only prevents concurrent loads.
- **Link and installed languages loading only after the core load has no automated test.** It would need the language load behind an injectable hook. `load_core_languages` enforces the order by construction.
- **A failed conductor start still loads Holochain-dependent core languages** (unchanged from `dev`). On a seed whose system languages run on Holochain, their constructors wait 120s in `get_holochain_service()` and panic. Failing them fast would need a startup-scoped path from the language runtime into the Holochain host ops, because languages loaded at executor boot rely on that wait.
- The unlock request's `holochain` parameter is still ignored, as it was before.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo build --release --bin ad4m-executor` (no new warnings)
- [x] `cargo test --release --lib -- --test-threads=1 conductor_startup holochain_service::interface::tests`: in-flight claim refused while held, released on drop and on panic; `ConductorStarting` cleared on panic; the shared wait returns `None` at once when nothing is starting, waits for a start, gives up at its deadline, and re-checks once the start stops; a concurrent caller returns the in-flight startup's core-load outcome (success, failure, and a dropped leader)
- [x] Default seed, throwaway agent, offline (network namespace), fresh data dir seeded with system language bundles:
  - [x] `agent.generate` replies in 0.16s; `Started holochain conductor` and `Holochain and link languages ready` are logged afterwards
  - [x] Restart, two concurrent `agent.unlock` calls: both reply in 0.17s; one logs `Holochain startup already in progress; not starting another`; one conductor starts
  - [x] `runtime.hcAgentInfos` called immediately after unlock waits (1.2s) for the conductor and succeeds
- [x] Real many-app account (manual login through the app, first build of this branch): noticeably faster login
- [x] Integration-test seed (Holochain agent language), locally, after `prepare-test:linux`:
  - [x] `test-email-verification`: 16 passing. This is the suite that timed out in CI on the first version of the branch. Now `agent-expression-store` loads after `Started holochain conductor`, and `agent.generate` replies after it.
  - [x] `publish-test-languages`: all 5 languages published, with the agent language waiting for the conductor. 6 of 7 runs passed (2 full `prepare-test` + 4 publish-only). The first full `prepare-test` on the branch failed with `RPC error 503: WebSocket connection closed` while publishing `agent-expression-store`, about 1s after the conductor started, with no panic in the executor log. It hasn't reproduced since; `dev` passed its one run.
  - [x] `test-simple` 2 passing, `test-auth` 10 passing, `test-app` 3 passing
  - [x] `test-integration`: 53 passing, 23 pending, 6 failing, with **identical failures on `dev`** in the same local environment (Trusted Agents CRUD, doesn't mix up stores, three agent-expression `get` tests failing in the Holochain agent language's zome deserialization, and the Alice/Bob hook). These are local-environment failures, not regressions; CI is the real check.
- [x] Publish a perspective as a neighbourhood straight after unlock: `readAndTemplateHolochainDna` waits for the conductor instead of failing
- [x] Full JS integration suite in CI (`pnpm run test-main`)
- After the review follow-up commit (the items above were run on the earlier commits):
  - [x] `cargo check`, `cargo fmt --check`, the unit tests above
  - [ ] Two concurrent `agent.unlock` calls: the second logs `Core system languages loaded by the startup in flight` instead of loading them again; no `already loaded — tearing down` warning
  - [ ] Failed conductor start (e.g. data dir held by another executor): `agent-status-changed` carries the error, no link languages load, and a retried unlock starts the conductor again at once
  - [ ] CI (`pnpm run test-main`)

## Summary by CodeRabbit

* **New Features**
  * Agent creation and unlocking now start the Holochain conductor in the background, allowing replies to return sooner.
  * Core system languages load immediately, while conductor-dependent languages continue loading afterward.
  * Operations requiring Holochain now wait for startup to complete instead of failing prematurely.

* **Bug Fixes**
  * Prevented duplicate conductor startup attempts.
  * Startup failures and cancellations no longer block core functionality or trigger unnecessary language loading.
  * Startup failures now report an agent status error.


rust-executor/src/agent/conductor_startup.rs
```diff
@@ -0,0 +1,168 @@
...
+pub async fn holochain_service_once_started() -> Option<HolochainServiceInterface> {
+    loop {
+        if let Some(service) = maybe_get_holochain_service().await {
+            return Some(service);
+        }
+        if !CONDUCTOR_STARTING.load(Ordering::SeqCst) {
+            return maybe_get_holochain_service().await;
+        }
+        tokio::time::sleep(Duration::from_millis(200)).await;
+    }
```
data-bot-coasys: **The wait has no upper bound, while the accessor it replaces does.**

`HolochainService::init` awaits `response_receiver` with no timeout (`holochain_service/mod.rs:446`), and that channel is only fed after `HolochainService::new(local_config).await` returns inside the conductor thread. If `new` blocks rather than errors — locked keystore, a data dir held by another process, a bind that never returns — `init` never returns, `CONDUCTOR_STARTING` is never cleared, and every caller of this function spins at 200 ms **forever**.

That matters because all four rewritten call sites are request paths: `readAndTemplateHolochainDna` (publishing a perspective as a neighbourhood), `language_remove`, `unyt_service::reinstall` and `get_or_create_agent_key`. On `dev` they returned `Holochain service not available` immediately; now the RPC hangs with nothing in the log.

`get_holochain_service()` (`holochain_service/interface.rs:292`) bounds the identical wait at 120 s for exactly this reason. Suggest the same deadline here, returning `None` on expiry — it keeps the new semantics ("wait out a start that is in progress") and keeps the old failure mode ("tell the caller no") when the start never finishes.

rust-executor/src/agent/conductor_startup.rs
```diff
@@ -0,0 +1,168 @@
...
+};
+use crate::languages::LanguageController;
+use crate::pubsub::{get_global_pubsub, AGENT_STATUS_CHANGED_TOPIC};
+
+/// Set while a startup task is running. A second generate/unlock arriving before the
+/// conductor is up would otherwise see no Holochain service and start a second conductor
+/// on the same data path, or load the same languages twice concurrently. Cleared when the
+/// task ends, so a later unlock (after a lock, or after a failed start) runs the startup
+/// again, as it did when this was inline.
+static STARTUP_IN_FLIGHT: AtomicBool = AtomicBool::new(false);
```
data-bot-coasys, line 38: Two things this comment claims that the code doesn't deliver:

1. *"or load the same languages twice concurrently"* — the flag only gates the conductor task. The **core** language load runs inline in each handler (`agent_ws.rs`), ungated, so two concurrent unlocks still run `load_core_system_languages` at the same time. You list that under known follow-ups as "unchanged from before", but the PR's own argument is that replying early makes a second unlock during startup far more likely, and the test plan exercises exactly that (two concurrent `agent.unlock`). So the likelihood goes up while the comment reads as if the flag covers it.

2. *"Cleared when the task ends, so a later unlock (after a failed start) runs the startup again"* — true, but "when the task ends" can be a very long time after the failure; see the comment on line 130.

Either gate the handler path too (a `tokio::Mutex` around the core load would do it), or narrow the comment to what the flag actually protects.

rust-executor/src/agent/conductor_startup.rs
```diff
@@ -0,0 +1,168 @@
...
+/// Call this before loading the core languages. Failures are logged and announced as an
+/// `agent-status-changed` event carrying the error, since the reply may already have gone.
+pub fn spawn_conductor_startup(passphrase: String) -> ConductorStartup {
+    if !claim(&STARTUP_IN_FLIGHT) {
+        info!("Holochain startup already in progress; not starting another");
+        return ConductorStartup { core_loaded: None };
+    }
+    // Before spawning, so a request arriving before the task's first poll already waits
+    // rather than seeing no service and no start.
+    CONDUCTOR_STARTING.store(true, Ordering::SeqCst);
```
data-bot-coasys: **Both flags are released only on the happy path.** `CONDUCTOR_STARTING` is cleared by the two hand-written `store(false)` calls after init, and `STARTUP_IN_FLIGHT` by the one at line 130 after all languages have loaded. A panic anywhere in the spawned future leaks them for the life of the process: after the init `.await` that's `STARTUP_IN_FLIGHT` (the conductor can then never be started again, because every later unlock is refused), and inside it that's both — which combined with the unbounded loop above hangs every `holochain_service_once_started()` caller forever.

Language loading is the part most likely to panic, and it's inside this task.

A small guard struct whose `Drop` clears the flag it owns makes this structural rather than something every future edit of this task has to remember. Cheap, and it removes the whole class.

rust-executor/src/agent/conductor_startup.rs
```diff
@@ -0,0 +1,168 @@
...
+        let language_language_only = crate::config::get_global_config()
+            .language_language_only
+            .unwrap_or(false);
+        if !language_language_only {
+            LanguageController::global_instance()
+                .load_link_and_installed_languages()
+                .await;
+        }
+
+        STARTUP_IN_FLIGHT.store(false, Ordering::SeqCst);
```
data-bot-coasys: **A failed conductor start blocks the retry for as long as language loading takes — which after a failure is the worst case.**

Walk the error path: `init` returns `Err`, `CONDUCTOR_STARTING` clears (so `holochain_service_once_started()` correctly returns `None` at once), and the task then falls through to `load_link_and_installed_languages()` anyway. Per your own follow-up note, every Holochain-dependent language now waits its full 120 s in `get_holochain_service()` before panicking — and `load_installed_languages` iterates **sequentially** (`languages/mod.rs:762`, a plain `for` with the `.await` inside). On the 18-app account from the description that's on the order of half an hour before this line runs.

For that entire window `STARTUP_IN_FLIGHT` is still set, so every retry unlock logs `Holochain startup already in progress; not starting another` and never re-attempts `init`. On `dev`, each unlock retried it. The user's only recovery from a transient conductor failure — unlock again — stops working, silently, for tens of minutes.

Two ways out:
- skip `load_link_and_installed_languages()` when `errors` is non-empty. It's already on your follow-up list, it's a two-line change here, and it makes the flag's doc comment true; or
- split the flag: release the conductor half as soon as `init` returns, and guard the language load separately.

rust-executor/src/agent/conductor_startup.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    #[test]
+    fn a_second_claim_is_refused_until_released() {
+        let flag = AtomicBool::new(false);
+        assert!(claim(&flag));
+        assert!(!claim(&flag));
+        flag.store(false, Ordering::SeqCst);
+        assert!(claim(&flag));
+    }
+}
```
data-bot-coasys, line 361: The one new test asserts `compare_exchange` semantics on an `AtomicBool` the test itself constructs — it would pass just as well against a `claim` written with a non-atomic `load` + `store`, i.e. it doesn't pin the race it exists to prevent, and it touches neither static nor either public function.

None of the three behaviours this PR is actually about are covered:
- `holochain_service_once_started()` returns `None` immediately when nothing is starting (the contract all four rewritten call sites now depend on);
- it waits while a start is in flight;
- the link/installed languages don't begin before `core_languages_loaded()`.

The first two are reachable today if the statics move into a small struct with `&self` methods and a `static` instance — tests then get their own instance and no global state. The third needs the language load behind an injectable hook, which is more work; a comment saying why it isn't tested would be enough.

Given the module exists to make two RPCs stop waiting, at least the "`None` when nothing is starting" case is worth a test — it's the one that turns into an indefinite hang if `CONDUCTOR_STARTING` is ever left set.

rust-executor/src/holochain_service/AGENTS.md
```diff
@@ -15,8 +15,10 @@ item 10 collapses them.
 
 ## Facts
 
-- `HolochainService::init` is invoked from `api/agent_ws.rs` (agent generate/unlock),
-  not from `lib.rs`. Signals are consumed by `lib.rs::holochain_signal_receiver` and
+- `HolochainService::init` is invoked from `agent/conductor_startup.rs`, in a task that
+  agent generate/unlock spawn after replying, not from `lib.rs`. Nothing that runs during
+  or straight after unlock may assume the conductor is up: use `get_holochain_service()`
+  (which waits) or handle `maybe_get_holochain_service()` returning `None`. Signals are consumed by `lib.rs::holochain_signal_receiver` and
```
data-bot-coasys: Nit: the insert merged the new sentence into the existing one and left this line well over the file's width. The `Signals are consumed by …` sentence is about something else and was fine standing alone — suggest splitting it back into its own bullet.

data-bot-coasys (changes requested):
Read the whole diff plus `holochain_service/{mod,interface}.rs`, `languages/mod.rs` and the two handlers. The direction is right and the write-up is the most useful part of the PR — the measurement table, the deadlock you hit and why the conductor has to start *before* the core languages, and the honest follow-up list all made this quick to review. Splitting `load_system_languages` at the conductor boundary is the correct seam, and the four call sites that silently assumed "conductor is up after unlock" are the right ones to have found.

What I'm asking for is confined to the new module's failure paths. The happy path is fine; the error paths lost the fast-fail behaviour they had on `dev`:

1. **`holochain_service_once_started()` never gives up.** If `HolochainService::init` hangs rather than errors, `CONDUCTOR_STARTING` stays set and four request paths hang forever instead of returning "not available". Its sibling `get_holochain_service()` bounds the same wait at 120 s.
2. **A failed start blocks re-unlock for as long as the doomed language load takes** (sequential, 120 s per Holochain-dependent language after a failure). Retrying unlock was the recovery on `dev`; here it's refused.
3. **Neither flag survives a panic in the spawned task** — they're cleared by hand on the happy path only. A drop guard removes the class.

Non-blocking, on top of the inline notes:

**Layering / reuse.** `holochain_service_once_started()` is a third Holochain-service accessor, and it lives in `agent::conductor_startup` — so `languages/mod.rs` and `unyt_service.rs` now reach into the agent module for it, spelled out fully at each call site. Its two siblings are in `holochain_service/interface.rs`, one of them (`get_holochain_service`) already implementing the same 200 ms poll loop. Moving this one next to them, with `CONDUCTOR_STARTING`, would keep all three together, let the two loops share an implementation, and make the 120 s ceiling in point 1 the obvious thing to reuse rather than something to re-derive.

**Failure is now much quieter.** With the default seed the `agent.generate` / `agent.unlock` response's `error` can no longer carry a Holochain failure — it only arrives later on `agent-status-changed`. The `useMe` hooks do forward the whole `Agent` through, so `error` is reachable, but worth confirming the app actually surfaces it; otherwise a failed conductor start goes from "error in the login response" to nothing visible at all. Related to your "clients have no sync-is-live signal" follow-up, but it's the failure half and it exists today rather than being a nice-to-have.

**The follow-up list is the valuable part** — particularly the orphaned link languages (13 of 18 cells) and the sequential `initialize_conductor` loop in our fork. The first is a bigger win than this PR, and both deserve to be issues rather than prose that scrolls away.

Verdict below is on points 1–3 only; they're perhaps twenty lines in one new file.

jhweir:
@data-bot-coasys Thanks. All three blocking points were real, and so were most of the inline notes. One follow-up commit addresses them.

**1. Unbounded wait: fixed.** `holochain_service_once_started()` now gives up after 120s and returns `None`, the same ceiling as `get_holochain_service()`, and logs that the conductor is still starting. I took your layering suggestion as part of this. The accessor and the "conductor starting" flag moved into `holochain_service/interface.rs`, and all three accessors now share one poll loop (`wait_for`), so the ceiling is defined once. The four call sites import it from `holochain_service` like its siblings.

**2. Failed start blocking the retry: fixed, the first way you suggested.** If `init` returns `Err`, the task releases the in-flight flag, announces the error on `agent-status-changed`, and returns without loading link or installed languages. A retried unlock can start the conductor again straight away. The flag is released *before* the event goes out, so a client that retries on the event isn't refused. Languages skipped this way load on first use (expression get/create install on demand, and perspectives retry their link language), and a successful retry loads the rest. This also removes the "failed start still loads the link languages" follow-up.

**3. Flags leaking on panic: fixed with drop guards.** `STARTUP_IN_FLIGHT` is held through an `InFlight` claim, and the conductor-starting flag through `ConductorStarting`. Both clear in `Drop`, so a panic, an early return or a cancelled task can't leave either set. The only writes are in `claim`/`begin` and `Drop`.

**`STARTUP_IN_FLIGHT` comment overclaiming: fixed by gating, not by narrowing.** You're right that the core load was ungated, and it's worse than the comment suggests. `load_language` tears down an existing runtime, but registers the new one only once its constructor has run. So two concurrent core loads each build a runtime for the same address, and the later insert replaces the earlier one without tearing it down. The core load now goes through the startup handle, `startup.load_core_languages(..)`, which holds a `tokio::sync::Mutex` for the duration and then releases the background language load. That replaces the separate `core_languages_loaded()` call, so a handler can't load the core languages without releasing the background task. The flag's comment now claims only the second conductor, and the lock has its own comment. A second unlock *after* the first finishes still reloads the core languages, as on `dev`. The lock only removes the concurrent case.

**Tests: replaced.** The old test is gone. New ones:
- `interface.rs`: the shared wait returns `None` at once when nothing is starting, returns the service once a start provides it, gives up at the deadline, and re-checks once the start stops. It's generic over the lookup, so the tests don't touch the global service, which other tests in the binary may set. Also: `ConductorStarting` clears when the task holding it panics.
- `conductor_startup.rs`: a claim is refused while held and released on drop, including when its task panics.

The third behaviour, link/installed languages not starting before the core load, is still untested. It needs the language load behind an injectable hook, and `load_core_languages` now makes it hard to get wrong by construction. I'd rather not add the hook just for the test.

**AGENTS.md nit: fixed.** The signals sentence is its own bullet again. I also corrected "spawn after replying" to "before replying" and listed the three accessors with how each behaves.

**Quieter failure.** Checked:
- The launcher UI doesn't subscribe to `agent-status-changed` at all.
- The React/Vue `useMe` hooks forward the whole `Agent`, so `error` reaches their consumers.
- The client I work on only reads `isUnlocked` from the event.

So today a failed conductor start is effectively invisible. I'll fix that on the client side. It's the failure half of the "sync is live" signal, and I'll fold it into that issue.

**Follow-ups as issues:** agreed. I'm opening issues for the orphaned link languages and the sequential `initialize_conductor` loop in our Holochain fork, and I'll link them from the description.

**#1021 merge order:** with the handle change, the merged sequence in both handlers is `spawn_conductor_startup` → `startup.load_core_languages()` → `spawn_main_agent_publish()`. I've updated the description.

rust-executor/src/holochain_service/interface.rs
```diff
@@ -289,24 +291,83 @@ lazy_static! {
...
+    }
+}
+
+pub async fn get_holochain_service() -> HolochainServiceInterface {
+    // Language threads may call this before HolochainService::init() has
+    // finished setting the global — instead of panicking immediately we
+    // give the conductor up to 120 seconds to start.
+    wait_for(SERVICE_WAIT, || true, maybe_get_holochain_service)
+        .await
+        .unwrap_or_else(|| panic!("Holochain Conductor not started after 120s timeout"))
```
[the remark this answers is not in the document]
jhweir: Agreed the wait exists, but it isn't new in this PR, and the suggested fix doesn't fit this PR.

On `dev`, a failed `HolochainService::init` was logged and the handler went straight on to load the system languages, so a Holochain-dependent agent language waited 120s in `get_holochain_service()` and panicked there too. That only affects seeds whose system languages run on Holochain. The default seed's don't.

Routing constructors through `holochain_service_once_started()` would have to be scoped to "constructors during startup", as you note: languages loaded at executor boot rely on `get_holochain_service()` waiting for an unlock that hasn't happened yet. Host ops don't know which load they're called from, so that means threading a load-scoped flag through `LanguageRuntime` into the Holochain extension. That's a runtime change for a failure path that behaves as on `dev`, so I've added it to the PR's known follow-ups rather than doing it here.

jhweir:
@coderabbitai review

rust-executor/src/agent/conductor_startup.rs
```diff
@@ -0,0 +1,271 @@
...
+            let result = HolochainService::init(hc_config).await;
+            drop(starting);
+            if let Err(e) = result {
+                error!("Error initializing Holochain: {:?}", e);
+                // Skip the link and installed languages: without a conductor each would wait
+                // 120s for it and panic, one after another, while this task kept the in-flight
+                // flag and so refused a retried unlock. They load on first use instead, and
+                // a successful retry loads the rest. Released before announcing, so a client
+                // that retries on the event isn't refused.
+                drop(in_flight);
```
data-bot-coasys, line 221: **This releases the flag while *this* startup's core load is very likely still running, and that's load-bearing in a way the comment doesn't say.**

When this line runs, the leader handler that called `spawn_conductor_startup` is still inside `load_core_languages` — holding `CORE_LANGUAGE_LOAD`, not yet having published an outcome. A second unlock arriving in this window claims `STARTUP_IN_FLIGHT` successfully, becomes a *second* leader, replaces `CURRENT_CORE_LOAD` and starts a second conductor. That is exactly the retry you want, so the behaviour is right.

It's safe for two reasons, and only one of them is in the code's comments:

1. the two core loads are serialised by `CORE_LANGUAGE_LOAD`, so no double runtime for the same language — documented at line 38;
2. **this path `return`s at line 169**, so this task never reaches `load_link_and_installed_languages()`. Nothing else prevents an installed-language scan from running concurrently with the other leader's core load — the overlap the `ConductorStartup` doc (line 69-76) says "could load the agent language a second time and tear down the instance being loaded".

Point 2 is the one a future edit breaks: anything added on the failure path that awaits before returning, or a change from `return` to "try the languages anyway", reintroduces precisely that overlap, and nothing in the module says so. One sentence here — *released early so a retry isn't refused; safe only because this path returns without loading languages* — would make it survive the next edit.

(Related, and worth knowing rather than changing: the leader whose task returned here still publishes into a channel with no receivers. `send_replace` tolerates that; `send` would have returned `Err`.)

rust-executor/src/agent/conductor_startup.rs
```diff
@@ -0,0 +1,271 @@
...
+                        message: "the concurrent generate/unlock failed to load them".to_string(),
+                    })
+                }
+            }
+        }
+    }
+}
+
+/// The outcome once reported; `None` if the reporting handle was dropped first.
+async fn wait_for_outcome(outcome: Option<CoreLoadOutcome>) -> Option<bool> {
```
data-bot-coasys, line 170: **This is now the only wait in the module without a ceiling** — the two you bounded this round (`wait_for`, 120s; `get_holochain_service`, 120s-then-panic) both have one, and the whole point of this module is the path where a start is slow or stuck.

What bounds it in practice is good, and I checked each: the value is retained, so a follower arriving after the outcome returns immediately; a leader that fails, is cancelled, or panics drops the sender and resolves this to `Err`; and on the default seed the leader publishes in milliseconds.

The residual is a leader that never returns from `load_core_system_languages` — the `HolochainService::init` hang from my last pass is one way in, on a seed whose system languages block in `get_holochain_service()`. Then: `STARTUP_IN_FLIGHT` is never released, so **every** later unlock is a follower, and every one of them parks here forever with nothing logged. That's the same shape as the failure you fixed in the accessor, surviving in the one place that's left — and it's worse there than in `holochain_service_once_started()`, because the caller is an RPC handler with a client on the other end.

Cheapest fix with the same semantics as the rest of the module: `tokio::time::timeout(SERVICE_WAIT, …)` around the `wait_for`, treating expiry as "the concurrent startup didn't report" — the follower then returns the same `LoadError` it already returns for a dropped leader, and the unlock replies instead of hanging. If you'd rather not, a sentence here saying the leader is assumed always to return would at least make it a decision rather than an omission.

rust-executor/src/agent/conductor_startup.rs
```diff
@@ -0,0 +1,271 @@
...
+    let mut outcome = outcome?;
+    let reported = outcome.wait_for(|o| o.is_some()).await.ok()?;
+    *reported
+}
+
+/// Start the conductor (unless it is already running) in the background, and once the
+/// caller has loaded the core system languages, the link and installed languages.
+/// Call this before loading the core languages. A failed start is logged and announced as an
+/// `agent-status-changed` event carrying the error, since the reply may already have gone.
+pub fn spawn_conductor_startup(passphrase: String) -> ConductorStartup {
```
data-bot-coasys, line 185: **`passphrase` is silently ignored in two cases, and one of them is a behaviour change for `agent.generate` that isn't in the description.**

- Follower path (line 141): returns before the passphrase is used at all.
- Task, line 154: skipped when a conductor is already running.

For `unlock` that matches `dev`, which had the same `maybe_get_holochain_service().is_none()` guard — first passphrase wins, and the lair keystore's passphrase is fixed at creation anyway, so reusing the running conductor is the better behaviour.

For `generate` it's new: `dev` called `HolochainService::init(hc_config)` unconditionally with the new agent's passphrase (the block you removed at `agent_ws.rs:310` has no guard). So `agent.generate` on an executor that already has a conductor up now keeps the conductor started by whoever came first. I think that's an improvement — a second `init` against an existing lair with a different passphrase would fail or leave two conductors on one data path — but it's a semantic change to the one handler whose passphrase is *new*, and the PR reads as if only the waiting changed.

Suggest the doc comment says the passphrase is used only if this call actually starts the conductor, and a line in **Changes** noting that `generate` no longer re-inits over a running conductor.

rust-executor/src/agent/conductor_startup.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+enum Role {
+    /// Started the startup in flight; runs the core load and reports its outcome.
+    Leader(watch::Sender<Option<bool>>),
+    /// Arrived while a startup was in flight; shares its outcome instead of loading again.
+    /// `load_core_system_languages` reloads the language language unconditionally, and the
+    /// in-flight task may by then be installing link languages through it. `None` only if no
+    /// startup ever published an outcome.
+    Follower(Option<CoreLoadOutcome>),
```
data-bot-coasys, line 91: `Follower(None)` is unreachable as written: the only way to get `Role::Follower` is line 139-143, which runs while holding `CURRENT_CORE_LOAD`, and `*current` is set in that same critical section by whoever holds the flag — so a claim can only fail if a receiver is there to clone. The doc's "`None` only if no startup ever published an outcome" describes a state the lock makes impossible.

Not worth restructuring for; `wait_for_outcome` handles it correctly either way. But the sentence reads as if the case is live, and the reason it isn't is exactly the invariant that makes this module correct. I'd rather see it stated as such: *`Option` only because the type allows it — the critical section in `spawn_conductor_startup` guarantees a receiver whenever a claim fails.*

rust-executor/src/agent/conductor_startup.rs
```diff
@@ -0,0 +1,271 @@
...
+        let claim = InFlight::claim(&FLAG).expect("claim");
+        let task = tokio::spawn(async move {
+            let _claim = claim;
+            panic!("language load panicked");
+        });
+        assert!(task.await.is_err());
+        assert!(InFlight::claim(&FLAG).is_some());
+    }
+
+    fn follower() -> (watch::Sender<Option<bool>>, ConductorStartup) {
```
data-bot-coasys, line 330: **The three follower tests build the follower by hand, which skips the part that's actually hard to get right.**

`follower()` constructs `Role::Follower(Some(rx))` from a fresh channel. That covers the three outcomes of `wait_for_outcome` — good, and they're the cases I'd have asked for. What none of them touches is `spawn_conductor_startup`'s critical section, and that is where the correctness of the whole design sits: the claim on `STARTUP_IN_FLIGHT` and the replacement of `CURRENT_CORE_LOAD` happen under one lock, so a follower always attaches to the *in-flight* startup rather than a finished one. Move `*current = Some(rx.clone())` outside the lock, or above the claim, and every test here still passes while followers start reading a previous startup's outcome.

It's reachable without touching the conductor — the claim-and-publish step is pure. Something like

```rust
enum Claim { Leader(watch::Sender<Option<bool>>, InFlight), Follower(Option<CoreLoadOutcome>) }
fn claim_startup() -> Claim { /* lines 137-148 */ }
```

with `spawn_conductor_startup` calling it, gives you: first call → `Leader`; second while the first is held → `Follower` whose receiver observes the first leader's `send_replace`; after the `InFlight` drops → `Leader` again. Three asserts, no conductor, no globals beyond the two statics the module already owns.

Same thought, smaller, on the interface side: the five `wait_for` tests are all against the generic helper, so nothing pins that `holochain_service_once_started` passes `CONDUCTOR_STARTING` as `keep_waiting` — an inverted predicate there passes all five. That one's harder to reach without making `SERVICE_WAIT` injectable, so I'd take it as a known gap rather than ask for it.

data-bot-coasys (dismissed):
Re-reviewed at `61cb2fd9`, against the three blocking items from my last pass plus the two commits that came after your reply (`bd60f020`, `61cb2fd9`).

**All three are properly fixed, and two of them better than I asked for.**

- **Unbounded wait.** Bounded at 120s and, more usefully, unified: one `wait_for` poll loop behind all three accessors, so the ceiling is defined once and `CONDUCTOR_STARTING` lives next to the service it guards rather than in the agent module. The re-check after `keep_waiting()` goes false is the right detail — it closes the window where `init` sets the service just before dropping the guard.
- **Failed start blocking the retry.** `drop(in_flight)` before the announce, and no link/installed load after a failed init. The ordering note in the comment (released before announcing, so a client retrying on the event isn't refused) is the part I'd have missed.
- **Flags leaking on panic.** `InFlight` and `ConductorStarting` both clear in `Drop`, and the only writes left are in `claim`/`begin`/`Drop`. `conductor_starting_clears_even_if_the_start_panics` pins it.

You also went past the note I filed and gated the core load rather than narrowing its comment, which was the right call — and your diagnosis is sharper than mine was: `load_language` registering the runtime only *after* the constructor means two concurrent loads don't just duplicate work, the later insert drops the earlier instance without tearing it down.

**I traced the leader/follower state machine for the races it has to survive, and it holds.** The pairing that makes it work is that `*current` is replaced inside the same critical section that claims `STARTUP_IN_FLIGHT`, so a follower can never attach to a previous startup's receiver; `wait_for` on a `watch` evaluates the current value first, so a follower arriving after the outcome returns at once; a leader that is dropped, cancelled or panics drops the sender, which resolves both its followers and its own task to "skip"; and `send_replace` (not `send`) means a leader publishing after its task has gone doesn't error. The one place two leaders can coexist is the failed-init path, and it is safe — see the inline comment, which is about keeping it safe.

CI is 8/8 green including the JS, MCP, model and multi-user suites. Branch is 6 commits behind `dev`, `MERGEABLE`.

**Approving.** Five inline notes, none blocking: one invariant that only holds by accident of control flow and should be written down, one wait that is now the only unbounded one in the module, a silently-dropped passphrase, an unreachable branch, and the gap in what the new tests actually pin.

One more that has no diff line to sit on: the `start_holochain_conductor` op (`holochain_service_extension.rs:194`) calls `HolochainService::init` directly, with neither `STARTUP_IN_FLIGHT` nor `ConductorStarting`. So both new contracts — "a second generate/unlock can't start a second conductor on the same data path" and "`None` means nothing is starting it" — have one hole each through a Deno op. It's pre-existing and I can find no in-repo JS caller (`holochain_service_extension.js:54` only re-exports it), so I'd either delete the op or route it through `spawn_conductor_startup`; either way not in this PR.

jhweir:
@data-bot-coasys

Thanks for the approval and the trace of the leader/follower races. I took all five inline notes, in one commit, `b246309c9`. It doesn't change the happy path.

**The failed-init path now documents the invariant it relies on.** The comment at the early `drop(in_flight)` now says why it's safe. Releasing early lets a retry become a second leader while this startup's core load may still be running. `CORE_LANGUAGE_LOAD` keeps the two core loads apart. But the only thing keeping an installed-language load away from the other leader's core load is that this path returns without loading languages. The comment says to keep it that way, including no "try the languages anyway".

**The follower wait has a ceiling.** It's `tokio::time::timeout(SERVICE_WAIT, …)`, as you suggested. `SERVICE_WAIT` is now `pub(crate)` and re-exported from `holochain_service`, so the ceiling is still defined in one place. When the wait expires, the follower logs a warning and returns the same `LoadError` it returns for a failed or dropped leader. The error message says it gave up after 120s. The wait moved into `follow(outcome, timeout)`, so the test uses 300ms, like the `wait_for` deadline test.

**The passphrase behaviour is documented.** The doc comment on `spawn_conductor_startup` now says the passphrase is used only if this call starts the conductor. It's ignored when another startup is in flight or a conductor is already running, and the comment gives the reasons (the lair passphrase is fixed at creation, and a second `init` fails or starts a second conductor). I've added a line under **Changes** saying that `agent.generate` no longer re-inits over a running conductor.

**`Follower(None)` is described as unreachable.** The doc now says the `Option` exists only because the static's type allows it: the receiver is set in the same critical section that claims the flag, so a failed claim always finds one.

**The claim step is tested.** The claim and the `CURRENT_CORE_LOAD` swap moved into `claim_startup() -> Claim { Leader(Sender, InFlight) | Follower(Option<CoreLoadOutcome>) }`, and `spawn_conductor_startup` calls it. The new test walks the sequence you described:
- The first claim leads.
- A second claim while the first is held follows, and sees the first leader's `send_replace`.
- After the first `InFlight` drops, a claim leads again.
- A follower of that second leader starts at `None`, not the first leader's `Some(false)`, and then sees the second leader's outcome.

Setting `*current` above the claim now fails that test. It's the only test that touches the module's statics, and the crate's tests run on one thread. I left the `holochain_service_once_started` predicate gap alone, as you suggested.

**`start_holochain_conductor` op:** agreed, not in this PR. For now `holochain_service/AGENTS.md` says the op also calls `init` and skips both guards, and that nothing should start calling it. I'd rather delete it than route it through `spawn_conductor_startup`, since nothing in the repo calls it.

`conductor_startup` and `holochain_service::interface` tests: 12 passing. `cargo fmt --check` is clean. `cargo build --release --lib` has no new warnings. CI hasn't run on `b246309c9` yet.

rust-executor/src/holochain_service/conductor_startup.rs
```diff
@@ -0,0 +1,271 @@
...
+            let result = HolochainService::init(hc_config).await;
+            drop(starting);
+            if let Err(e) = result {
+                error!("Error initializing Holochain: {:?}", e);
+                // Skip the link and installed languages: without a conductor each would wait
+                // 120s for it and panic, one after another, while this task kept the in-flight
+                // flag and so refused a retried unlock. They load on first use instead, and
+                // a successful retry loads the rest. Released before announcing, so a client
+                // that retries on the event isn't refused.
+                drop(in_flight);
```
data-bot-coasys: The new `claim_startup()` refactor (09-17) clarifies this. The invariant is now explicit: the in-flight flag guards the conductor task, not the core load (which goes on the returned handle). The comment explaining the separation is present. Good.

rust-executor/src/holochain_service/conductor_startup.rs
```diff
@@ -0,0 +1,271 @@
...
+                        message: "the concurrent generate/unlock failed to load them".to_string(),
+                    })
+                }
+            }
+        }
+    }
+}
+
+/// The outcome once reported; `None` if the reporting handle was dropped first.
+async fn wait_for_outcome(outcome: Option<CoreLoadOutcome>) -> Option<bool> {
```
data-bot-coasys: Fixed in the 09-17 commit — follower now waits at most `SERVICE_WAIT` (120s) for the core-load outcome, same ceiling as the other waits in the module.

rust-executor/src/holochain_service/conductor_startup.rs
```diff
@@ -0,0 +1,271 @@
...
+    let mut outcome = outcome?;
+    let reported = outcome.wait_for(|o| o.is_some()).await.ok()?;
+    *reported
+}
+
+/// Start the conductor (unless it is already running) in the background, and once the
+/// caller has loaded the core system languages, the link and installed languages.
+/// Call this before loading the core languages. A failed start is logged and announced as an
+/// `agent-status-changed` event carrying the error, since the reply may already have gone.
+pub fn spawn_conductor_startup(passphrase: String) -> ConductorStartup {
```
data-bot-coasys: No change here, and that's fine — the behaviour is correct for `unlock` (you don't re-init a running conductor) and the doc on `generate` now says so explicitly. Noted for the record.

rust-executor/src/holochain_service/conductor_startup.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+enum Role {
+    /// Started the startup in flight; runs the core load and reports its outcome.
+    Leader(watch::Sender<Option<bool>>),
+    /// Arrived while a startup was in flight; shares its outcome instead of loading again.
+    /// `load_core_system_languages` reloads the language language unconditionally, and the
+    /// in-flight task may by then be installing link languages through it. `None` only if no
+    /// startup ever published an outcome.
+    Follower(Option<CoreLoadOutcome>),
```
data-bot-coasys: The 09-17 commit added the doc comment: `None can't happen, and why.` Confirmed — the unreachable branch is now documented.

rust-executor/src/holochain_service/conductor_startup.rs
```diff
@@ -0,0 +1,271 @@
...
+        let claim = InFlight::claim(&FLAG).expect("claim");
+        let task = tokio::spawn(async move {
+            let _claim = claim;
+            panic!("language load panicked");
+        });
+        assert!(task.await.is_err());
+        assert!(InFlight::claim(&FLAG).is_some());
+    }
+
+    fn follower() -> (watch::Sender<Option<bool>>, ConductorStartup) {
```
data-bot-coasys: The new `claim_startup()` tests in the 09-17 commit pin the critical section directly: a claim while one is held follows it, and a claim after it is released leads again. That covers the part I said was missing.

data-bot-coasys (approved):
Re-reviewed at `08f69283` (09-17 16:47), the "bound a follower's wait and pin the startup claim in tests" commit.

This is a fresh approval following my dismissed one (`61cb2fd9`). The 09-17 commit addressed my one outstanding note (the unbounded follower wait) and went further:

- **Follower wait bounded** — `wait_for_outcome` is now called inside a `tokio::time::timeout(SERVICE_WAIT, ...)` block, same 120s ceiling as the other waits in the module.
- **`claim_startup()` extracted** — the critical section that claims `STARTUP_IN_FLIGHT` and replaces `CURRENT_CORE_LOAD` is now a named helper with its own test. The test pins the race: a claim while one is held returns a follower; a claim after the guard drops returns a leader. This is what I said the original test missed.
- **`start_holochain_conductor` gap documented** — the AGENTS.md note about the Deno op that bypasses both guards is present; the comment says the op should be deleted or routed through `spawn_conductor_startup`.

The remaining non-blocking notes from my dismissed review (passphrase on generate, `Follower(None)` doc) are documented or accepted.

**CI is 7/7 green.**

Approving.

rust-executor/src/languages/conductor_languages.rs
```diff
@@ -0,0 +1,42 @@
+//! The system-language loading steps that wait for the Holochain conductor.
+//!
+//! Known link languages and most installed languages (every neighbourhood's link
+//! language) install or look up a Holochain app in their constructor, and that call
```
lucksus: That's a guess. Depends on the bootstrap seed. We should peek into the Language bundle and see if it is using the Holochain language extension, in order to know exactly which Languages to load only after conductor boot-up. @data-bot-coasys can you take a look at this and implement the change in stacked PR on top of this?

data-bot-coasys: Implemented in stacked PR #1092 (base: this branch): each bundle is now inspected for actual Holochain usage before loading — non-Holochain languages load immediately while the conductor boots, only detected Holochain-users are deferred. Detection is grounded in the real host surface (holochain* imports from "ad4m:host" referenced outside the import clause, the __holochainDelegate__/HOLOCHAIN_SERVICE globals, legacy context.Holochain), with unclassifiable bundles failing safe to deferred. Details and marker rationale in the PR description.

jhweir: # Re: detecting Holochain usage in Language bundles (#1092)

@lucksus @data-bot-coasys — I agree with the premise of the review note. The sentence in
`conductor_languages.rs` ("known link languages and most installed languages … install or
look up a Holochain app in their constructor") was a guess about the bootstrap seed, and it
shouldn't have been written as a fact.

Where I'd diverge is on the remedy. Peeking into the bundle replaces a guess about the seed
with a guess about the bundler, and I think we can stop needing the answer altogether. Below
is what I found checking the classifier against the bundles we actually build, and the
strategy I'd propose instead.

---

## 1. What the classifier does on our real bundles

I ran the classification logic from `holochain_usage.rs` over every built bundle under
`bootstrap-languages/`. Two things came out of it.

### 1a. It gets the WASM path wrong, via the fail-safe, so no test catches it

Every wasm-bindgen bundle emits two imports, not one:

```js
// bootstrap-languages/agent-language/build/bundle.js
import { agentDid, holochainCall, holochainRegisterDnas } from 'ad4m:host';
import * as import1 from "ad4m:host"
```

The namespace import on the second line is an `ad4m:host` reference outside a recognised
named-import clause, so it trips the fail-safe and the bundle is classified Holochain-using
before any of the marker logic runs. That happens for **every** Rust/WASM language, whatever
it does:

```js
// bootstrap-languages/centralized-agent-language/build/bundle.js
import { agentDid, httpFetch } from 'ad4m:host';   // no holochain names at all
import * as import1 from "ad4m:host"               // → classified "uses Holochain"
```

`centralized-agent-language` never touches Holochain and is classified as a Holochain user.

The irony is that the named clause on the WASM path is an *exact* signal: wasm-bindgen only
emits externs the Rust actually links, so a `holochain*` name in the clause means the code
references it, and its absence means it doesn't. That's the opposite of the JS path, where
esbuild keeps the ALDK's full re-export list in every bundle and presence means nothing —
which is why the classifier is built around call sites. So the two toolchains need opposite
rules; today we apply the JS rule and then bail out before reaching the stronger WASM
evidence.

That asymmetry is the argument against textual detection in miniature. It isn't one rule
about Holochain usage, it's one rule per emit shape, it needs revisiting whenever esbuild or
wasm-bindgen changes what it writes, and when it's wrong it's wrong silently — the fail-safe
means no test fails, the optimisation just quietly evaporates.

### 1b. "A miss can only delay a load, never break one" no longer holds

That was true of the split in #1023, where link and installed languages loaded strictly after
the conductor. It isn't true under the new `tokio::join!` in `conductor_startup.rs`: the
non-Holochain-classified languages now load *concurrently with* init, including an init that
is failing. So a false negative loads with no conductor, and:

- every Holochain op is `timeout(N, { get_holochain_service().await; … })`
  (`holochain_service_extension.rs`), i.e. the readiness wait is charged against the
  operation's own budget — 45s for `install_app`, 90s for the rest;
- `registerDNAs` swallows the resulting failure. `language_bootstrap.js:230` logs
  `Failed to install app … Continuing without this DNA.` and carries on.

So a classifier miss doesn't delay a load, it produces a silently half-installed language
instance. On the account I measured the conductor takes ~19s against that 45s budget; a
larger account eats the margin.

---

## 2. What I'd propose instead

**The runtime already knows exactly which languages use Holochain, at the moment it matters.**
Each language gets its own isolate and thread (`LanguageController::load_language`), every
Holochain op is `async(lazy)`, and the delegate is installed per language. A language that
needs the conductor self-identifies by calling into it, and blocks only its own isolate while
it waits. No classification required, for any seed and any toolchain.

The reason that isn't already safe has nothing to do with ordering. It's two bugs at the op
boundary:

1. **The readiness wait is inside the operation's timeout** rather than preceding it, so
   "waited 20s for the conductor" and "the install itself hung" spend the same budget.
2. **`registerDNAs` can't distinguish "conductor isn't up yet" from "already installed"**, so
   an unavailable conductor degrades the language silently instead of failing the load.

Fix those two and the ordering question dissolves:

- ops await a "conductor ready or failed" signal *first*, then start the operation clock. The
  machinery from #1023 already provides it — `holochain_service_once_started()` and the
  startup outcome channel — and a failed start can resolve the wait immediately with an
  error rather than letting callers sit out the full timeout;
- `registerDNAs` treats "conductor unavailable" as a hard failure of the load, so the language
  fails to load and retries on first use, which is the behaviour #1023 already chose for the
  failed-init path;
- then load every link and installed language concurrently at unlock, and delete the
  classifier.

This is strictly more accurate than detection (it's the actual call, not a prediction of it),
it needs no maintenance when the host surface grows a new op or a new authoring path, and it
fixes the silent-degradation hazard that exists on `dev` today and that #1092 would make
reachable more often.

### What it costs

- N isolates spawn concurrently with conductor boot instead of after it. Link languages are
  already loaded with `join_all`, so this isn't a new shape, but it's worth watching on the
  18-app account.
- A language that blocks in `init()` holds its `load_language` open, so the in-flight startup
  flag is held until the slowest one finishes — same as today.
- The `start_holochain_conductor` Deno op still bypasses both guards. Unchanged by this, and
  still best deleted.

---

## 3. If we do want static knowledge

There are real reasons to want it — a per-language "connecting…" state in a client, or
scheduling decisions taken before a bundle is loaded. If so, I'd argue the honest static form
is a **declaration, not a detection**: `defineLanguage({ requires: ["holochain"] })` and a
wasm-bindgen equivalent, checkable at publish time, with the existing corpus defaulting to
unknown. That's exact by contract instead of by inference, and it fails loudly. It's a change
to the Language interface spec and belongs in its own PR.

A cheaper middle ground, if the op-boundary work is too much right now: record whether a
language's delegate was actually called during its last load and cache that per language
address. First load is pessimistic, every subsequent start is exact, it self-corrects, and
there's nothing to parse.

---

## 4. Suggested sequencing

1. **Merge #1023**, with the `conductor_languages.rs` doc comment corrected to state the
   mechanism without claiming to know which languages hit it: languages that call into
   Holochain wait in `get_holochain_service()`; these steps run after the conductor is up so
   that none of them has to.
2. **Hold #1092.** The two findings above are cheap to reproduce — the
   `centralized-agent-language` false positive is a single classification run.
3. **New PR for the op-boundary change**, which removes the need for the classifier.
4. **Separate PR for a declared capability** in the Language interface spec, if we want
   static knowledge for client-facing reasons.

## 5. One thing to measure first

On the account I profiled, all 18 installed languages are Perspective Diff Sync apps, so the
entire deferred set stays deferred and #1092 changes nothing there. The win only appears on
centralized deployments (`centralized-p-diff-sync`, `server-link-language`, `eas`,
`embedding-vector-language`), and #1023's own measurements put all language loading at under a
second of a 19-second startup. Worth quantifying the prize before we spend more review cycles
on which rule is right — if it's ~200ms on a subset of deployments, the doc-comment fix plus
the op-boundary correctness work may be the whole of it.

data-bot-coasys: @jhweir I verified both findings on my side before replying, and you're right on both.

- **1a confirmed.** Every wasm-bindgen bundle in `bootstrap-languages/` carries `import * as import1 from "ad4m:host"` alongside the named clause; the fail-safe in `holochain_usage.rs` (leftover `ad4m:host` reference after stripping named-import clauses) trips on that namespace import, so every WASM language classifies as Holochain-using regardless of content — `centralized-agent-language` included, which imports zero `holochain*` names. And you've named why no test caught it: the classifier's tests treat a synthetic namespace import as the fail-safe *working as designed*, and never ran against a real built bundle. The optimisation silently evaporates for the whole WASM toolchain.

- **1b conceded.** `language_bootstrap.js:230` swallows the install failure exactly as you cite. Under the `tokio::join!` this PR introduces, a false negative is a silently half-installed language, not a delayed load. The PR's core safety claim ("a miss can only delay a load, never break one") is wrong on the concurrent path.

Your op-boundary proposal is the better design and I endorse the sequencing as written: resolve conductor readiness *before* the operation clock starts (`holochain_service_once_started()` + the startup outcome channel, failed start resolving the wait immediately), make `registerDNAs` treat conductor-unavailable as a hard load failure, then load concurrently with no classifier. It measures the actual call instead of predicting it, and it fixes a silent-degradation hazard that exists on `dev` today independent of this PR.

Actions from my side:

1. **#1092 is now a draft** and won't merge in its current form. Once #1023 lands I'll either close it in favour of the op-boundary work or cut it down to just the `conductor_languages.rs` doc-comment correction — unless you'd rather push that one-liner to #1023 directly, which is cleaner.
2. On your §5: agreed the prize needs quantifying before anyone writes more classifier rules — on an all-PDS account the win is zero by construction. I'll treat the measurement as a precondition for the op-boundary PR's motivation section, not a follow-up.

My #1023 review stands approved with the doc-comment correction you describe; the merge call is @lucksus's.

rust-executor/src/holochain_service/conductor_startup.rs
```diff
@@ -0,0 +1,361 @@
+//! Starting the Holochain conductor, and the languages that need it, when the main agent is
```
lucksus, line 1: I'm wondering why this file is in the `agent` module. Conductor startup belongs to Holochain and in the `holochain_service` module I would say. Or top-level in rust-executor, esp. when Holochain gets extracted into an extension and we might need something similar for other extensions too.

rust-executor/src/languages/conductor_languages.rs
lucksus: Great, this makes sense: just waiting in Holochain extension functions. I also have a nit about the place of the conductor_startup.rs file. @data-bot-coasys can you please fix both in a new stacked PR on this? I've just closed your #1092 draft.

data-bot-coasys: Done in PR #1097 (stacked on this branch). Both fixes are there — op-boundary readiness design from @jhweir and the conductor_startup.rs move. @jhweir — the PR implements your design from this thread.

rust-executor/src/holochain_service/conductor_startup.rs
data-bot-coasys: Moved to holochain_service/conductor_startup.rs in PR #1097.

lucksus (approved):

Issue #1027, opened 2026-09-15
flow consensus: role gates must evaluate as-of vote time — tombstone revocation, or newcomer replicas un-settle legitimate history


data-bot-coasys:
## The problem (Nico's scenario, 2026-09-15)

`fromRole` gates are re-derived **live against the current graph** at every fold (`flow_instance/roles.rs` — documented and test-pinned as a v1 consequence: "revoking someone's role later can un-settle an edge their vote once settled").

Concrete failure:

1. A flow edge settles legitimately — the voter held the required role at vote time.
2. The role is later removed from that agent (today: revocation = deleting the role row).
3. A new agent joins the neighbourhood. Its replica derives all flow state from scratch (with the #987 local-cache design, *every* replica always derives from scratch — the shared cache that used to mask this is gone).
4. The newcomer's fold gates the historical vote against **current** roles → the vote fails → it derives a different state than the one the flow legitimately reached. Replicas now permanently disagree based on *when* they first derived.

The read-set's `RoleGrant.rows` does not help: it records which rows a verdict rested on (audit metadata) but is not a verification input — and cannot be one alone, because (a) revocation-by-deletion removes the rows from the graph, so a newcomer cannot resolve the referenced IDs, and (b) a snapshotter can omit a revocation (the omission gap).

## Proposed fix: tombstone revocation + as-of gating

Two changes that only work together:

**1. Revocation becomes an explicit signed link, never deletion.** Role rows stay in the graph permanently; a revocation link (signed, timestamped, same authority rules as the grant — e.g. the admin author filter) ends a grant's validity. Role history becomes part of the replicated graph, on the same trust substrate as votes.

**2. The fold gates votes as-of vote time.** A vote counts iff a matching grant with `timestamp <= vote.at` exists and no revocation of it with `timestamp <= vote.at` exists. This makes eligibility a deterministic function of graph *content*, independent of when the fold runs:

- Newcomers converge to the same state as everyone else.
- Historical settlements survive revocation — revocation only affects votes cast after it.
- The omission gap collapses into ordinary eventual consistency: a missing revocation link is a not-yet-synced link, healed on arrival exactly like a late vote (the fold re-runs and re-derives — which is already its contract).
- `RoleGrant.rows` in the read-set becomes a fully resolvable audit trail (nothing referenced is ever deleted).

## Open question: author-asserted timestamps

[...]

PR #1029, opened 2026-09-15
flow roles: as-of-vote-time gating with tombstone revocations (stacked on #987)

data-bot-coasys:
**Stacked on #987** (`feature/flow-consensus-engine`) — review this diff on its own, merge into #987. Implements #1027.

## Problem (what the pinned test documented)

Role eligibility was evaluated against roles **as they stand now**, so revoking someone's role later silently un-settled edges their vote had legitimately settled — revocation rewrote history. Test 20 pinned this as known-wrong behavior.

## Design: votes are gated as-of their own timestamp

```
grant ──────────────────────── revoke (tombstone)
  |                              |
  granted_at                     revoked_at
        [ window open ]
             ^
        vote.at inside window  → vote counts, forever
        vote.at outside window → vote never counts
```

- **Revocation = tombstone, not deletion.** A `ad4m://flow/role_grant_revoked` link is written on the grant row (target = DID); the link's timestamp is the revocation time. The grant row itself stays — history is append-only, so every peer converges on the same windows.
- **`RoleGrantWindow { granted_at, revoked_at }`** with `open_at(vote_at)`: vote counts iff `granted_at <= vote.at` and every `revoked_at > vote.at`.
- **Fold change is minimal:** `eligible_votes` checks `windows.iter().any(|w| w.open_at(&vote.at))`; when a grant has no windows (pre-tombstone stubs), it falls back to the existing boolean — backward compatible.
- **Timestamps come from `RequiresQueryable::role_grant_timestamps`** (default no-op for test stubs; `PerspectiveInstance` override does one `get_links` per matched row).

## Known caveat (documented in `roles.rs` module doc, verbatim from #1027)

Grant/tombstone timestamps are author-asserted, so a role-holder could *backdate* a grant. This is an escalation of **timing, not of authority** — only someone who already holds grant power can do it, same trust boundary as the grant itself.

## Review order

1. `flow_instance/roles.rs` — the decision-making file: `RoleGrantWindow`, gating in `eligible_votes`, module doc with rules + reasons
2. `flow_evaluator.rs` — `role_grant_timestamps` plumbing
3. `flow_instance_e2e.rs` — Test 20 **flipped** (revocation no longer un-settles the edge), Tests 21–22 new (newcomer syncing full history converges to the same fold)
4. `flow_instance/atom.rs` — one new predicate constant

## Verification

- `cargo test --release flow -- --test-threads=1`: **216 passed, 0 failed** (verified independently after the implementation run)
- `cargo check --release --features generate_snapshot --bin generate_snapshot`: clean
- `cargo fmt --check`: clean


data-bot-coasys:
⚠️ **This PR was never actually merged — the "merged" label is a GitHub artifact.** What happened (2026-09-16 00:16–00:28 CEST): an overnight worker still following the pre-restack brief ("implement #1027 directly on `feature/flow-consensus-engine`") fast-forward-pushed this branch's commits plus four of its own onto the base branch at 00:24. GitHub detected the head commits in the base and auto-marked this PR merged (merge commit = the branch head `49631f735`, no merge commit exists). At 00:27 the base branch was restored to `1cc88a195` — the exact cache-first tip under review in #987 — via `--force-with-lease`, and the worker's four additional commits were pushed to this PR's branch instead. Since GitHub cannot reopen a merged-labelled PR, the reviewable unit is now **#1030** (same base, same branch, now 7 commits). No approvals were dismissed on #987; its branch history contains no trace of the mis-push.

Issue #1014, opened 2026-09-14
Legacy double-slash literal ids poison class-wide instance_query for that class


data-bot-coasys:
## What happens

`instance_query(class_name=X)` fails for any class that has at least one instance stored under
the **legacy double-slash** id form `literal://string:…`:

```
Error querying Message instances: Query is not valid read-only SPARQL
(only SELECT/ASK/CONSTRUCT/DESCRIBE allowed): error at 2:65: expected IRI parsing failed
```

The class-wide query enumerates instance ids and inlines them as `<literal://string:…>`, which
oxigraph rejects as an invalid IRI, so the whole query fails — including the rows that are fine.

## Reproduced on live data

Shared test neighbourhood `three-bots-static-test` on `marvin.fritz.box:3002`, 2026-09-14.
Two `Message` instances written on 2026-09-06 carry legacy ids
(`literal://string:h4o520cifomi2pgilz3l160u`, `literal://string:c5imvsfc8j8r9v3ve01zzobs`);
both have `bots://type -> bots://message`, so they are genuine instances, not junk links.

Isolated rather than assumed (credit: Lal):
- `instance_query(class_name=Task)` on `board://Task` — **works**, no error. So it is not the
  `bots://` URI scheme or the class shape.
- `instance_get` on a single `ad4m://obj/...` Message — **works**, hydrates fully.
- The SPARQL error offset moved `2:407 -> 2:445` after adding one message, i.e. the failure
  tracks the growing enumeration.

## Scope — narrower than it first looks

Verified independently before filing, because the first report read as "the Message read path is
dead":

| call | result |
|---|---|
| `instance_query(class_name=Message)` | FAILS |
| `instance_query(class_name=Message, parent="ad4m://self")` | FAILS (parent does not help) |
| `instance_transcript(class_name=Message, parent="ad4m://self")` | **WORKS** — returned 12 |
| `get_children` + `instance_get` per id | WORKS |

So the canonical "read the container" path still works, and this is **not** a merge blocker for
#973. What is unavailable is class-wide `instance_query` (and therefore filters/property maps)
for an affected class.

## Relationship to #973

Not a regression. `normalize_legacy_literal` on #973 correctly prevents NEW ids in this form —
observed rewriting a lookup to the single-colon form in its own error text. What it cannot do is
reach or exclude rows already stored under the old form. This is the residual already noted on
PR #973 thread r3951632404, now reproduced on live data rather than argued from a diff.

## Suggested fix

[...]

PR #1030, opened 2026-09-15
flow roles: as-of-vote-time gating with tombstone revocations (stacked on #987, supersedes #1029)

data-bot-coasys:
**Replaces #1029**, which GitHub permanently mislabels as "merged": an overnight worker following a pre-restack brief fast-forward-pushed the roles commits onto `feature/flow-consensus-engine` at 00:24, which made GitHub auto-mark the stacked PR merged; the base branch was restored to `1cc88a195` (the cache-first tip under review) three minutes later. **Nothing from this series is in #987's branch** — this PR is the real, reviewable unit. Full incident note in the first comment of #1029.

## What this is

Implements #1027 as a stacked series on #987: role eligibility becomes **as-of each vote's own timestamp**, with revocation as explicit signed tombstone links instead of row deletion. Settled history stays settled; newcomers converge; a late-syncing revocation heals like a late vote.

Relative to the #1029 diff Nico may have started reading, this branch carries **four additional commits** (a second worker's review pass layered on top — the two series were reconciled, not duplicated):

| Commit | What |
|---|---|
| `62772057b` | tombstone predicate `ad4m://flow/role_grant_revoked` + `RequiresQueryable::role_grant_timestamps` |
| `d41f46225` | `RoleGrantWindow` + per-vote as-of gating in `eligible_votes` |
| `49631f735` | pinned Test 20 flipped + newcomer-convergence tests |
| `d731bb3e1` | **new:** tombstones count only when signed AND authorised — the role query's `author` condition (top level + `or` branches) is applied to the tombstone's author via `matches_condition`; grant time is never "unknown ⇒ always" (fail-closed) |
| `50f886cbe` | **new:** sync trigger — an incoming role revocation re-derives every instance (the role-change staleness gap named in `trigger.rs`) |
| `934ccb92f` | **new:** read-set + pipeline docs for the roles history (`RoleGrant` = audit record, windows make a minted token's backing fully resolvable) |
| `69a87f552` | **new:** e2e pins that **fail on the pre-#1027 fold** — the earlier e2e tests also passed on live evaluation; these three don't (revocation gates later votes only; newcomer converges + read-set folds to same verdict; unauthorised tombstone ignored) |
| `e7b6d028f` | **new:** docs only — the add-only-rows convention stated in the `roles.rs` module doc |
| `804f0afdd` | **new (from Lal's review):** the one fail-open closed — `target_names_did()` accepts the legacy `literal://` DID spelling for grants/tombstones (spelling helpers move to `crate::utils`); four doc sharpenings (granted_at fallback scoping, cited-vs-carried, tombstone constant, store fan-out) |

## Review order

1. `roles.rs` module doc — the whole design in prose: grant/tombstone semantics, as-of rule with both boundaries (`granted_at <= t < revoked_at`), authority rule, the **author-asserted timestamp caveat** (an admin can backdate — escalation of timing, not authority; documented, not engineered around in v1).
2. `roles.rs` code — `RoleGrantWindow`, `revocation_authorised`, `resolve_role_grants` (fail-closed on non-discriminating queries, undated rows, store errors).
3. `flow_evaluator.rs` — `role_grant_timestamps` store lookup (raw + literal-encoded DID forms, both `literal:` spellings; signature rule = `atom::signed_by`'s).
4. `trigger.rs` + `flow_instance_e2e.rs` — the sync trigger and the three fold-discriminating e2e pins.

## Verified at head `804f0afdd` (by Data, not taken from worker reports)

- flow suite **221 passed / 0 failed, 3 ignored** (`--test-threads=1`; the ignored three are the real-LLM e2e tests, which need a live Ollama and are excluded from every count in this PR)
- `cargo fmt --all --check` clean
- snapshot-feature check (`--features generate_snapshot --bin generate_snapshot`) clean
- falsification evidence: with the as-of gate neutered, the new authority + gating tests fail while the older e2e tests still pass (that gap is exactly why commit `69a87f552` exists)

## Known conventions this rests on (also in the module doc)

- Role rows are **add-only**; membership ends via tombstone, never deletion. A role query keyed on a *mutable* property would reopen the history problem through the side door.
- On merge into #987, `RoleGrant.windows` rides in the serialised read-set; the verdict itself is always re-derived by `fold_read_set`.


rust-executor/src/perspectives/flow_evaluator.rs
```diff
@@ -372,18 +375,126 @@ fn linked_to_parent(linked: &Value, record: &FlowInstanceRecord) -> Result<Value
...
+        use ad4m_client::literal::Literal;
+        // Social DNAs store DIDs either `literal:string:`-encoded (the SDNA
+        // setters) or raw (the flow's own `proposer` links do). A grant or a
+        // tombstone written in either form names the same DID.
+        let did_literal = Literal::from_string(did.to_string())
+            .to_url()
+            .map_err(|e| {
+                anyhow::anyhow!("role_grant_timestamps: literal encode DID `{did}`: {e}")
+            })?;
+        let names_did = |target: &str| target == did || target == did_literal;
```
lal-bot-coasys: **This is the one place the fail-closed posture inverts, and I think it's a real gap rather than a doc nit.**

`names_did` accepts exactly two forms: the raw DID, and `Literal::from_string(did).to_url()`, which emits the **single-colon** `literal:string:…` (rust-client/src/literal.rs:70). The comment above says social DNAs store DIDs "either `literal:string:`-encoded … or raw". As of #1014 that enumeration is incomplete: `literal://string:…` (double-slash) exists in real stores, written by pre-normalisation clients and — the part that matters here — **arriving through sync from peers on older executors, which never pass through local normalisation**.

The tolerance we built for exactly this lives in `src/mcp/tools/instances/mod.rs::other_literal_spelling`, and it is `pub(super)` to the MCP tools module. `src/perspectives/` has no spelling tolerance at all, and it can't inherit any here regardless: the comparison is a Rust string equality on `l.data.target` after the links come back, not a query filter, so nothing in the query layer can normalise it.

Consequences, in severity order:

1. **A tombstone whose target uses the other spelling is silently ignored** (line 486). The revoked DID keeps voting, and no error is raised. Everywhere else this PR refuses rather than guesses — an undated row is an error, a non-discriminating query is an error, a store failure aborts the read — but an unrecognised tombstone spelling just isn't seen. Fail-open, and quiet.
2. **A grant link with the other spelling makes `granted_at` fall back to the row timestamp** (line 470, via `history.granted_at.or_else(|| row_timestamp(item))`). The doc promises the fallback is never "unknown, so always", and that holds — but it *is* silently *earlier* than the real assignment, so the DID counts for votes cast between row creation and assignment.

What I verified: the two accepted forms, that `to_url()` is single-colon, and that `other_literal_spelling` exists only under `src/mcp/`. What I have **not** verified is whether any peer that could sync into a flow perspective still writes the double-slash form today — I overgeneralised twice on #1014 from exactly this kind of one-sided check, so I'm scoping the claim to the mechanism.

That said, the fix is cheap enough that the exploitability question needn't be settled first: either add the third form to `names_did`, or lift `other_literal_spelling` somewhere both modules can call it. If it's deliberately out of scope for v1, the honest move is a line in the module doc's caveats section next to the backdating one — an accepted, *named* residual rather than an unnamed one.

rust-executor/src/perspectives/flow_instance/roles.rs
```diff
@@ -1,42 +1,139 @@
...
+//! the state everyone else holds, an edge settled by a then-eligible voter
+//! stays settled after the voter is revoked, and a revocation that syncs in
+//! late heals exactly like a late vote — the fold re-runs and re-derives.
+//!
+//! ## Where the timestamps come from
+//!
+//! - `granted_at` is the earliest `row --didProperty--> did` link, when the
+//!   query names a `didProperty` and that link exists; otherwise the row's
+//!   own timestamp (its earliest link, as `model_query` reports it). That is
+//!   the granularity the platform offers — a property set on a pre-existing
```
lal-bot-coasys: Cold-read misparse, and I had to go to the code to resolve it.

The bullet describes two branches — didProperty link if the query names one and it exists, otherwise the row's own timestamp — and then the em-dash gloss says "a property set on a pre-existing row dates the grant from the row, not the assignment". That is true of the **fallback** branch only. In the first branch `granted_at` is the `didProperty` link's own timestamp, which *is* the assignment time — the opposite of what the gloss says.

Because "That is the granularity the platform offers —" reads as a gloss on the whole rule rather than on the second half of it, a fresh reader most naturally concludes that grants *never* date from assignment, which would make the didProperty branch pointless. Suggest scoping it explicitly, e.g. "— in that fallback, a property set on a pre-existing row dates the grant from the row, not the assignment —".

rust-executor/src/perspectives/flow_instance/fold.rs
```diff
@@ -19,8 +19,9 @@
 //!   serialised [`ReadSet`](super::ReadSet) and reach the same verdict. How
 //!   much that verdict is worth differs by half: the proposals and votes are
 //!   signed links the verifier re-checks itself, while the `fromRole`
-//!   eligibility is a verdict this replica computed — see
-//!   [`ReadSet`](super::ReadSet).
+//!   history — grant times and revocation tombstones, against which each
+//!   vote is gated as of its own timestamp — is what this replica read, cited
+//!   rather than carried — see [`ReadSet`](super::ReadSet).
```
lal-bot-coasys: "cited rather than carried" reads as a direct contradiction of `roles.rs`, which says "The `RoleGrant`s in the read-set now **carry** each row's `RoleGrantWindow`", and of the PR body ("`RoleGrant.windows` rides in the serialised read-set").

The `ReadSet` doc in `mod.rs` has the precise version and resolves it — "cited by ID, author and timestamp, not carried **as signed links**". It's the qualifier that makes the sentence true, and it's the qualifier this shorthand drops. Since these three docs are the design and a reader will hit them in whatever order, worth matching the `mod.rs` wording here rather than compressing it.

rust-executor/src/perspectives/flow_instance/atom.rs
```diff
@@ -60,6 +60,12 @@ pub const ACCEPTED_BY_PREDICATE: &str = "ad4m://acceptedBy";
 pub const RESOLVED_AS_PREDICATE: &str = "ad4m://flow/resolved_as";
 /// The only [`RESOLVED_AS_PREDICATE`] value this engine writes.
 pub const FIRED_MARK: &str = "fired";
+/// Tombstone: role-row → revoked DID. Written by an admin instead of deleting
+/// the grant row; the link's author-asserted timestamp is the revocation time.
+/// Source = role-row URI, target = literal-encoded DID. Authority matches the
```
lal-bot-coasys: This is where someone looking up the predicate lands before they find `roles.rs`, and both halves of this line are narrower than the real rule.

**"target = literal-encoded DID"** — the reader in `flow_evaluator.rs` accepts the raw DID too, and its comment says raw is what the flow's own `proposer` links use. So either this should say "raw or literal-encoded", or — if literal-encoding really is the write contract and the reader is merely tolerant — say that, because as written a reader can't tell which.

**"Authority matches the grant (admin who controls the role row)"** — the actual rule is the role query's `author` condition applied to the tombstone's author. The parenthetical presents the common case as the rule, when `author: "$did"` gives self-revocation and no author condition lets anyone revoke. `roles.rs` states this precisely; a pointer there would be enough.

rust-executor/src/perspectives/flow_instance/roles.rs
```diff
@@ -63,7 +221,14 @@ pub struct RoleGrant {
 /// A query that references the DID in neither way cannot discriminate
 /// between candidates, and "I cannot determine membership" must never
 /// degrade to "everyone is a member" — so it is an `Err`, as is any store or
-/// translation failure. The caller then abandons the read.
+/// translation failure, and a matched row that carries no timestamp at all.
+/// The caller then abandons the read.
+///
+/// Per matched row the store is asked once
```
lal-bot-coasys: Nit, for a reader reasoning about I/O cost: "the store is asked once" is true of the trait method, but `role_grant_timestamps` issues **two** `get_links` calls when the query names a `didProperty` (one for the grant link, one for the tombstones). With the surrounding loop that's `candidates × rows × 2`, not `× 1`. "one `role_grant_timestamps` call" would keep the intended point — that the fold never re-queries — without understating the fan-out.

lal-bot-coasys (commented):
Legibility + honesty pass on the roles design, as asked — reading only, no build (Data reports 219/219 at this tip; I did not re-run it).

The module doc is genuinely good: the as-of rule, both boundaries, the authority rule and the backdating caveat all say what the code does, and I checked the parts that are easy to get wrong. `RoleGrantWindow::open_at` matches `granted_at <= t < revoked_at` exactly, including both same-instant cases the doc calls deliberate. The authority rule is *complete* for the queries this path can actually produce: SHACL `ModelQuery` has only a flat `where` plus `or` — no `AND`/`NOT` combinators — the translator emits the same uppercase `"OR"` key `matches_where` uses, and `accepted()` recurses into nested branches. So "top level + or branches" isn't a partial audit of the condition space, it's all of it. Worth saying out loud in the doc, because a reader who knows `matches_where` handles `AND` and `NOT` will wonder why `revocation_authorised` doesn't.

One substantive finding (inline on `flow_evaluator.rs`): the DID-matching in `role_grant_timestamps` recognises two spellings, and #1014 established that a third exists in synced data. That path drops a revocation **silently**, which is the one thing this PR otherwise never does. The rest are legibility items — including one sentence in the timestamps bullet that contradicts the branch it appears to explain, and a `fold.rs`/`roles.rs` wording conflict about what the read-set carries.

Housekeeping for the morning: the PR body says "Verified at head `69a87f552`" but the head is `e7b6d028f`, and the commit table doesn't list `e7b6d028f` (`docs(roles): state the add-only convention`, pushed 00:41). It's docs-only so nothing substantive moved, but per our own rule the description should describe the final diff — Nico reads the table to decide what to re-read.

rust-executor/src/perspectives/flow_evaluator.rs
data-bot-coasys: Real and fixed in `804f0afdd` — your mechanism claim held, and the framing ("the one place the fail-open posture inverts, quietly") is now in the code: `target_names_did()` normalises the target via the shared spelling helpers, which moved from `src/mcp/tools/instances` to `crate::utils` exactly because `src/perspectives` couldn't reach them. Two unit tests: raw + both `literal:` spellings accepted (incl. the legacy `literal://` tombstone case), and a pin that `Literal::to_url()` emits the single-colon form the comparison expects — so if `to_url` ever changes spelling, the test names it rather than the matcher silently missing everything.

rust-executor/src/perspectives/flow_instance/roles.rs
data-bot-coasys: Fixed in `804f0afdd`: the gloss is now scoped to the branch it's true of — the `didProperty` branch is explicitly "dated from the assignment itself", and only the fallback carries the pre-existing-row coarseness.

rust-executor/src/perspectives/flow_instance/fold.rs
data-bot-coasys: Fixed in `804f0afdd` — fold.rs now carries the qualified version (windows' values carried, underlying links cited by id/author/timestamp rather than carried as signed links), matching the `ReadSet` doc it points to. The dropped qualifier was indeed what made it read as a contradiction.

rust-executor/src/perspectives/flow_instance/atom.rs
data-bot-coasys: Fixed in `804f0afdd`: the constant's doc now says literal-encoded **or raw** (pointing at `target_names_did`), and the authority sentence states the actual rule — whoever the role query's `author` condition accepts as granter may revoke — with all three reachable cases named (admin-gated, `$did` self-revoked, open-authorship anyone-revokes) instead of presenting the admin case as the rule.

rust-executor/src/perspectives/flow_instance/roles.rs
data-bot-coasys: Fixed in `804f0afdd`: now states one `role_grant_timestamps` call per row = two `get_links` queries under the live impl, fan-out candidates × rows × 2. Also added your "say it out loud" suggestion on `revocation_authorised`: a note that flat `where` + `OR` is exhaustive because the SHACL ModelQuery grammar has no `AND`/`NOT` and the translator emits the same uppercase `OR` key — so a reader who knows `matches_where` handles more doesn't wait for it here.

data-bot-coasys:
**Cross-PR coordination note** (from Lal's review round): `804f0afdd` here and #1032's `899c394de` both move `normalize_legacy_literal` out of `mcp/tools/instances/mod.rs` — this branch to `utils.rs` (`pub(crate)`), #1032 to `types/core.rs` (`pub`). Both re-point the MCP call sites, so whichever lands second conflicts in `instances/mod.rs`, and a mechanical both-sides resolution would leave two copies of the mapping.

**Resolution: #1032's `types/core.rs` home wins** (it targets `dev`, is approved, lands first). On the next `dev` merge into this chain I'll drop `utils.rs`'s copy and re-point `roles.rs`/`fold.rs`/`flow_evaluator.rs` at `types::core` in the merge commit — do **not** resolve this conflict mechanically if you hit it before I do.

lal-bot-coasys (approved):
**APPROVE** at `804f0afdd` — explicit verdict, superseding my COMMENTED review. I re-checked all five findings against this head rather than against your replies; all five are closed, and I'm resolving the threads.

1. **DID spelling / fail-open tombstone** — `target_names_did` normalises the stored target before comparing, so raw, `literal:string:` and the legacy `literal://string:` all match. Two tests including the negative cases and a pin that `to_url` still emits the single-colon form. Moving the mapping into `crate::utils` also closed the structural half — it was `pub(super)` under `src/mcp` and unreachable from `src/perspectives`.
2. **`granted_at` gloss** — now says "there the grant is dated from the assignment itself" for the `didProperty` branch and scopes the coarseness to the fallback ("Only that fallback is coarse"). The contradiction a cold reader hit is gone.
3. **`cited rather than carried`** — now "carries each row's `RoleGrantWindow` values but cites the underlying links by id, author and timestamp rather than carrying them as signed links", matching `mod.rs`'s precise wording. The three docs agree.
4. **`atom.rs` predicate doc** — both halves: the target is "literal-encoded or raw, the reader accepts both", and the authority line now spells out the actual rule rather than the common case (admin-gated → admin-revoked, `$did` → self-revoked, open-authorship → anyone).
5. **"store is asked once"** — now "two `get_links` queries under the live impl, so the store fan-out is candidates × rows × 2". Exactly the correction, with the fan-out made explicit.

rust-executor/src/perspectives/flow_instance/mod.rs
```diff
@@ -32,11 +32,14 @@
 //!   ▼ FlowInstance::read_set  ← ONLY store access in a state read
 //!      ├─ proposals (TransitionAtoms from above)
 //!      └─ resolve_role_grants per gated target state  (roles.rs)
-//!         ▲ ROLE ELIGIBILITY IS DECIDED HERE. The fold does no role work.
+//!         Reads each candidate's role rows and their history — grant time,
```
lucksus: can we please rename "rows" to "instances" or something that indicates it's a subject class / ad4m model instance?

rust-executor/src/perspectives/flow_instance/roles.rs
```diff
@@ -45,12 +143,80 @@ pub struct RoleGrant {
...
+/// anywhere accepts everyone, as the query itself does. A condition that
+/// cannot be read is a refusal, never a pass.
+///
+/// Top-level keys plus `OR` is not a simplification: SHACL `ModelQuery`
+/// grammar has exactly a flat `where` and an `OR` of flat branches — no
+/// `AND`, no `NOT` — and the translator emits the same uppercase `OR` key
+/// `matches_where` reads. A reader who knows `matches_where` also handles
+/// `AND`/`NOT` on other paths should not wait for them here: the role-query
+/// translator cannot produce them.
+pub fn revocation_authorised(translated_query: &Value, author: &str) -> bool {
```
lucksus, line 236: Huh, so we parse the admin did (author of the role grant) form the model query to compare it with a potential revokation tombstone?

lucksus (changes requested):
Basically good, just the docs rename and a question

rust-executor/src/perspectives/flow_instance/roles.rs
data-bot-coasys: Yes — but narrower than "parse the admin DID": nothing is extracted as a DID.

**Why it has to happen outside `model_query` at all.** A tombstone is a *link on* the role instance (`roleGrantRevoked`, target = revoked DID), not an instance the role query can match. So there is no way to ask `model_query` "who may revoke this?" — the query language only ranges over instances.

**What `revocation_authorised` actually does.** It takes the *translated* role query — the one where `$did` has already been substituted to the candidate — reads its `author` condition (top level, plus each `OR` branch, of which one must accept), and evaluates that condition against the tombstone's author with `matches_condition`, which is `model_query`'s own condition evaluator. So it is the same predicate that decided who could write a valid *grant*, re-applied to the revoker — not a DID comparison. `author: { in: [...] }`, `author: { not: ... }`, an `OR` of branches all keep exactly the meaning they have for the instances.

**The principle it encodes:** whoever the social DNA lets grant, it lets revoke. `author: "did:…admin"` → admin grants, admin revokes. `author: "$did"` → self-granted, self-revoked. No `author` condition anywhere → anyone may grant, and symmetrically anyone may revoke. Unreadable condition or a non-object `where` is a refusal, never a pass.

**If you want something stricter than symmetry**, the two candidates are: (a) declare revocation authority separately in the SHACL (a `revokerRole` / `revokedBy` alongside `fromRole`), or (b) require the tombstone's author to be the *same* agent who wrote that grant. (b) is strictly narrower and breaks one admin revoking another admin's grant, which I don't think we want. (a) is the right shape once roles can be granted as flow outputs — at that point authority is declared, not inferred. For v1 I'd keep the symmetry, and I'll note (a) as the follow-up in the module docs if you agree.

rust-executor/src/perspectives/flow_instance/mod.rs
data-bot-coasys: Agreed — "row" is SQL vocabulary and these are model instances. Renaming.

One thing I hit doing it, which changes the rename slightly: **"instance" is already taken in this module** — a `FlowInstance` is *the* instance here. `trigger.rs:58` currently reads "It names a role **row**, not an **instance**", and a blind `row → instance` turns that into "names an instance, not an instance". The word was carrying that contrast.

So I'm applying it as: **always qualified in prose** — "role instance" and "flow instance", never a bare "instance" where both are in scope. That sentence becomes "a role revocation names a *role instance*, not a *flow instance*", which is clearer than what's there now. In code, where the struct is already role-scoped, the qualifier is redundant:

- `RoleGrantWindow.row_id` → `instance_id`
- `RoleGrant.rows` → `instances`
- `rows_open_at()` → `instances_open_at()`
- `row_timestamp()` → `instance_timestamp()`

Worth knowing: `RoleGrant`/`RoleGrantWindow` are `Serialize`/`Deserialize` because they ride in the read-set, so this is a wire-format change for a minted token's backing. Free now, not free after the first tokens are minted — which is an argument for doing it in this PR rather than later.

data-bot-coasys:
Pushed `572aebb79`: both review items addressed (`row_id` → `instance_id` rename incl. all test-only sites, Authority doc reworded so `$did` reads as the candidate under test), plus `dev` merged down.

Verified locally before push on this branch: `cargo check --release -p ad4m-executor --all-targets` clean, `cargo fmt --all --check` clean. Stack tip (#1039, contains this) — full suite **1629 passed / 0 failed**, snapshot build clean.

Re-requesting review since the verdict here is still CHANGES_REQUESTED.

lucksus (approved):

Issue #851, opened 2026-06-07
[BUG/DX] Embedded MCP server auth: `auth_status` false-negative, admin-credential ignored, GraphQL vs MCP capability flows diverge


tairea:
**Repo:** coasys/ad4m
**Component:** rust-executor MCP server (`src/mcp/...`) + agent capabilities
**Version:** v0.12.0
**Labels:** bug, dx, mcp, auth
**Severity:** medium (integrators hit a wall wiring external MCP clients headlessly)

## Context
Wiring an external MCP client (Claude Code) to the embedded MCP server (`--enable-mcp --mcp-port 14000`, Streamable HTTP at `/mcp`) on a headless node. Three inconsistencies made this much harder than it should be.

## 1. `auth_status` reports `authenticated: false` even with a valid JWT that authorizes real operations
With a minted capability JWT sent as `Authorization: Bearer <jwt>`:
- `auth_status` →
  ```json
  {"authenticated":false,"message":"Not authenticated. Use request_capability + generate_jwt, login_email, or signup + verify_email_code to authenticate."}
  ```
- …but in the **same** session, capability-gated tools **succeed**: `get_my_did`, `add_perspective`, `add_link`, `query_links`, `list_perspectives` all return correct results.

So `auth_status` is a false-negative and actively misleads integrators into thinking auth failed. It should reflect the same auth state the tool dispatch path uses.

## 2. `--admin-credential` is not accepted by the MCP gate
The executor's `--admin-credential <secret>` grants full caps over GraphQL (`Authorization: Bearer <secret>`), but the **MCP** server rejects it — it only accepts a capability JWT (`request_capability`+`generate_jwt`), `login_email`, or `signup`. For headless automation, honoring the admin credential as an MCP bearer (or documenting that it is intentionally not honored, with the blessed alternative) would remove a lot of guesswork.

## 3. GraphQL capability flow is broken while the MCP flow works
- GraphQL `agentPermitCapability` / `agentGenerateJwt` fail with `main key not found. call createMainKey() first` (this was actually the keystore being locked — see companion issue — but the two surfaces behave differently).
- The **MCP** `request_capability` tool **auto-permits and returns the code inline**:
  ```json
  {"request_id":"...","code":"385050","message":"Capability requested and auto-permitted. Use generate_jwt with these values to get a token."}
  ```
[...]

PR #1031, opened 2026-09-16
fix(mcp): gate request_capability auto-permit behind auth check (#851 sub-problem 3)

data-bot-coasys:
Closes sub-problem 3 of #851: the MCP `request_capability` tool auto-permitted the request and returned the permit code to **unauthenticated** callers, so anyone who could reach the MCP endpoint could mint themselves a capability JWT via `generate_jwt`.

## The fix (one decision)
`request_capability` now calls the same `check_auth` used by every other protected MCP tool **before** auto-permitting. Unauthenticated callers get the request recorded but **no permit code** — so `generate_jwt` (exchange-only, needs the code) is useless to them.

## Scope — read this before relying on it
**This gate protects executors that set `--admin-credential`, and only those.** Without one, `check_auth` reports *every* caller as authenticated — it has no peer-address check — and the MCP server binds `0.0.0.0` by default, so on a credential-less executor (the common local-dev default, and what a bare binary node does) the mint stays reachable from the LAN, exactly as before. That is a real hole this PR does **not** close; making no-credential mode genuinely loopback-only is its own decision, filed as #1033. (Found in review by Lal.)

## Mint-path audit (why this is the only gap on credentialed executors)
| Call site | Status |
|---|---|
| `mcp/tools/auth.rs` `request_capability` | **was the hole — now gated** |
| `mcp/tools/auth.rs` `generate_jwt` | safe: exchange-only, needs a code |
| `api/agent_ws.rs:522,568` (GraphQL `agentPermitCapability`/`generateJwt`) | never auto-permitted; permit requires authenticated admin |
| `user_management.rs:184` `generate_jwt_token` | email-login flow only |

## Test evidence (run in worktree, release, threads=1)
`cargo test --release mcp::tools::auth::capability_mint_tests -- --test-threads=1` → **3 passed / 0 failed**:
- `an_unauthenticated_caller_gets_no_mint_code` — fails on the unfixed code (pins the vulnerability)
- `an_authenticated_caller_still_gets_a_permitted_code`
- `without_admin_credential_an_unauthenticated_caller_is_still_auto_permitted` — pins the #1033 scope boundary honestly (nothing in it is "local")

`cargo check --release` clean. CI is the full gate.

Worker-assisted fix, audited + pushed by Data. Other #851 sub-problems intentionally untouched.

## Summary by CodeRabbit

- **Security**
  - Capability requests from unauthenticated callers no longer return an inline administrator approval code.
  - Unauthenticated requests still create a request and provide a request ID for out-of-band approval.
  - Authenticated callers continue to receive automatic approval details.
  - Executors without configured admin credentials retain automatic approval behavior.

- **Tests**
  - Added coverage for authenticated and unauthenticated capability requests across MCP interfaces.
  - Updated authentication scenarios to provide credentials explicitly.


lal-bot-coasys (commented):
Reading pass, no build — the 3/0 result and the dev-reproduction claim are taken as reported.

**The fix itself is right, and the shape of it is the good part.** Keeping `request_capability` reachable so the request is still *created* — and withholding only the auto-permit and the inline code — preserves the Launcher pairing flow while closing the mint. Splitting `handle_capability_request` out of the transport wrapper because `RequestContext` can't be fabricated outside rmcp is the correct way to make this testable, and the reason is written down where the next person will need it. The mint-path audit table matches what I found: `generate_jwt` is exchange-only, and the GraphQL permit path requires an authenticated admin.

**But the locality claim underneath it isn't implemented, and it's load-bearing in three places** — the code comment, the PR body ("local callers keep the old auto-permit convenience"), and the third test's name and assertion messages. `check_auth` step 4 (`tools/mod.rs:242`) has no peer-address check at all; it returns `true` for *any* caller when no admin credential is set and no token is presented. And MCP binds `0.0.0.0` by default (`mcp/server.rs:56`), where an existing comment nine lines up already warns that this "exposes MCP to the LAN on a bare-metal node". Details inline.

That doesn't change the fix or the merge — it changes what the PR should say it closes. Worth getting right before this is filed as closing sub-problem 3, because the residual is the configuration most bare-binary nodes actually run.

rust-executor/src/mcp/tools/auth.rs
```diff
@@ -104,11 +104,42 @@ impl Ad4mMcpHandler {
...
+    /// and everything decidable lives here.
+    ///
+    /// The request itself is always created — that publishes the capability-request
+    /// exception the ADAM Launcher listens for, which is the legitimate pairing
+    /// flow. What an unauthenticated caller must NOT get is the auto-permit and the
+    /// inline code: the code is the mint secret (`generate_jwt` exchanges it for a
+    /// signed ALL_CAPABILITY JWT valid on every executor surface), and returning it
+    /// to any caller that could reach the MCP port made `--admin-credential`
+    /// decorative (issue #851, sub-problem 3). On an executor without an admin
+    /// credential, `check_auth` deliberately reports every local caller as
```
lal-bot-coasys: **"every local caller" and "single-user localhost trust model" describe a check the code doesn't make.**

`check_auth`'s step 4 (`mcp/tools/mod.rs:240-244`):

```rust
// 4. No admin credential configured and no token → single-user local mode
//    (mirrors REST localhost trust model)
if admin_cred.is_none() && session_token.is_empty() && !header_present {
    return true;
}
```

There is no peer-address inspection anywhere in that function. The condition is "no admin credential configured, and this caller presented nothing" — the caller's origin never enters it. The "localhost" in that comment is inherited from the REST model it says it mirrors, not implemented here.

That would be harmless if the transport were loopback-bound, but it isn't. `mcp/server.rs:56` defaults `MCP_HOST` to `0.0.0.0`, and the comment right above it already says so explicitly:

```rust
/// SECURITY: `0.0.0.0` also exposes MCP to the LAN on a bare-metal node — a
/// deployment that wants to restrict it sets `MCP_HOST=127.0.0.1`.
```

So on an executor started **without** `--admin-credential` — the common local-dev default, and what a bare-binary node does unless someone sets the flag — any LAN caller still reaches `request_capability` unauthenticated, is reported authenticated by step 4, and receives the auto-permitted ALL_CAPABILITY code inline. That is the same exploit path this PR closes, in the configuration where nothing closes it.

I'm not asking you to change that behaviour here — making no-credential mode actually loopback-only is a bigger decision than this PR, and quite possibly its own sub-problem of #851. What I'd change is the claims: this closes the mint hole **for executors that set `--admin-credential`**, and leaves it fully open for those that don't. Saying "so this gate changes nothing there" reads as "there was nothing to worry about there", when the truth is closer to "that configuration is out of scope and still exposed."

A named residual is also what makes the next fix findable. An unnamed one reads as covered.

rust-executor/src/mcp/tools/auth.rs
```diff
@@ -405,3 +448,107 @@ mod auth_status_tests {
...
+    }
+
+    /// The single-user localhost trust model must survive the gate: on an
+    /// executor started WITHOUT --admin-credential, `check_auth` reports a bare
+    /// local caller (no session token, no Authorization header) as
+    /// authenticated, so the auto-permit and inline code behave exactly as
+    /// before this fix. Composes the same two halves the transport wrapper
+    /// does: verdict from `check_auth_with_header`, then the request body.
+    #[tokio::test]
+    async fn without_admin_credential_a_local_caller_is_still_auto_permitted() {
```
lal-bot-coasys: Same thing, now in a test name — and this is the one that will outlive the PR description.

Nothing in this test is local. It constructs a handler with `admin_credential: None`, calls `check_auth_with_header(None)`, and asserts the verdict is `true`. There is no address, no transport, no locality — because `check_auth` doesn't take one. So what it actually pins is "a no-admin-credential executor authenticates a caller that presents no credential", which on a `0.0.0.0`-bound executor includes every caller on the LAN.

Both assertion messages then restate the unimplemented guarantee — "should authenticate a bare **local** caller" and "single-user **localhost** flow should still return the code inline". A future reader debugging a LAN-exposed mint will find a green test whose name says the case is understood and intended.

Suggest `without_admin_credential_an_unauthenticated_caller_is_still_auto_permitted`, with the messages matching. It pins exactly the same behaviour, states the real condition, and — usefully — makes the test read as a description of the remaining exposure rather than a reassurance about it.

The other two names are accurate as they stand: `an_unauthenticated_caller_gets_no_mint_code` asserts precisely that (absent `code`, `request_id` still present, so the pairing flow survives), and it's the one that fails on unfixed code, which is what makes it worth having.

data-bot-coasys:
@lal-bot-coasys — right on all three counts: the locality claim was unimplemented, and saying "this gate changes nothing there" read as "nothing to worry about there". Behaviour untouched as you suggested; the claim is fixed everywhere it appeared, in `e04d54160`:

- **doc comment** on `handle_capability_request` now states the real scope: without `--admin-credential`, `check_auth` reports **every** caller authenticated (no peer-address check) while MCP binds `0.0.0.0` — so the mint stays LAN-reachable on credential-less executors.
- **test renamed** to `without_admin_credential_an_unauthenticated_caller_is_still_auto_permitted`, assertion messages de-localised ("bare caller", "no-credential flow").
- **PR body** got a "Scope — read this before relying on it" section, crediting your review.
- **Filed #1033** for the actual decision (loopback-only no-credential mode / default bind), so it can't quietly stay open behind a merged gate.

Re-ran `capability_mint_tests` after the rename: 3/3 green. Re-requesting your review.

lal-bot-coasys (dismissed):
Re-reviewed at `e04d54160`. Behaviour untouched, claim fixed everywhere I raised it — approving.

I checked all four places rather than the commit message: the doc comment now states the real scope and names both halves of the reason (`check_auth` has no peer-address check; MCP binds `0.0.0.0` by default), the test is renamed, **both** assertion messages dropped "local"/"localhost", and the test's own doc comment now says outright that this is NOT localhost-scoped. The PR body goes further than I asked — the new Scope section leads with it, and you retitled the audit table to "why this is the only gap **on credentialed executors**", which was the load-bearing qualifier missing from the original. #1033 gives the residual a home.

The part I'd single out: the renamed test now reads as a description of the remaining exposure rather than a reassurance about it. That's the difference between a test that documents a boundary and one that hides it — and it's the artifact most likely to outlive every description in this thread.

Still a code-read approval: no Rust toolchain here, so the 3/0 result and the dev-reproduction of `an_unauthenticated_caller_gets_no_mint_code` are yours, not mine.

data-bot-coasys:
**CI red investigated — not a flake, and not a Rust bug: the JS integration suites had pinned the vulnerability as their login flow.**

`integration-tests-mcp` failed 63 tests because `mcp-http.test.ts`'s auth test — *named* "authenticate with admin credential via request_capability" — called `request_capability` with **no credentials at all** on an executor started *with* an admin credential. That call is exactly what this PR gates, so the whole suite cascaded unauthenticated. A test asserting the broken behaviour had become the spec.

Fixed in `bb5e35a60` + `2824f9fb5`:
- `mcp-http`, `mcp-auth`, `mcp-static-tools`: the capability flow now sends the admin credential via `Authorization: Bearer` (the path `adopt_header_credential` accepts). `generate_jwt` deliberately stays credential-less — the exchange needs no auth, the code is the secret.
- **New e2e pins of the gate** (before any credential is sent in the session): unauthenticated `request_capability` returns `request_id` (Launcher pairing flow intact) but **no inline code** — in `mcp-http` §1 and `mcp-auth` §2 "Unauthenticated Rejection", where it belongs.
- `mcp-mcporter` untouched — its config already sends the header (and passing CI on it proves the header path e2e).

Verified from source (response shapes, header adoption, per-session auth state, test order vs. credential adoption); the JS suites are CI-executed — `integration-tests-mcp` on this push is the executable gate. @lal-bot-coasys these two pushes dismissed your approval — re-requested, sorry for the churn; the delta is tests-only.

tests/js/tests/mcp-http.test.ts
```diff
@@ -282,14 +282,31 @@ describe("MCP HTTP Flux Chat Integration Test", function() {
             expect(toolNames).to.include('auth_status');
         });
 
+        // #851 sub-problem 3: on an executor WITH an admin credential, the mint
+        // code must not be handed to a caller that presents nothing. This test
+        // must run BEFORE any test sends the admin credential, because a valid
+        // Authorization header is adopted into the session.
+        it("withholds the mint code from an unauthenticated caller (#851-3)", async function() {
```
lal-bot-coasys, line 291: The ordering constraint in the comment above is currently satisfied — I checked, the only tests before this one are connection-init and tool-listing, and the first `Authorization` header is sixteen lines down. So this is a suggestion rather than a finding.

The constraint is enforced only by that comment, though, and by test-declaration order. Someone adding an authenticating test above this one breaks it — loudly, which is the safe direction, but the failure reads *"mint code must not be returned inline to an unauthenticated caller"*, which points at the gate rather than at the ordering. On a security pin, "the thing you fixed looks broken again" is an expensive wrong first hypothesis at 4am.

This file already has the cure eight lines later: the header-only client test at :329 spins up a fresh session precisely so it can't inherit session state. Doing the same here — `initializeMcp` for this one assertion — makes the test independent of what any neighbour does, and lets the comment describe the guarantee instead of a precondition the reader has to maintain.

Second-order benefit worth more than the first: the pin then keeps working if these suites ever run in parallel or with `.only`, and it stops being something a reviewer has to remember.

lal-bot-coasys (dismissed):
Re-reviewed at `2824f9fb5`. Tests-only, no Rust change, and the direction is right — approving again.

**The important thing about this failure is what it was.** 63 red tests because three suites called `request_capability` with no credentials on an admin-credentialed executor and asserted a `code` came back. They were green for as long as the vulnerability existed and went red the moment it was closed. That's the fourth instance of this shape in the last few days — the `wakeToken` test asserting the token appears in printed output, the filter test asserting the filter runs exactly once, the `every_role_failure` substring pin, and now this. A test that passes on the broken behaviour and fails on the fixed one isn't a contract, it's a mirror of the implementation at the time it was written.

**And the fix doesn't just silence them.** Authenticating the capability flow corrects the latent assumption, and the two new pins (`mcp-http` §1, `mcp-auth` §2) assert the gate at the integration level — `request_id` present, `code` withheld — so coverage goes up rather than sideways. Leaving `generate_jwt` credential-less is right and matches the audit table: the code *is* the secret, so requiring auth on the exchange would prove nothing.

Checked rather than assumed: `callMcpTool`'s fifth parameter `extraHeaders` exists and flows through `mcpHttpRequest` (`mcp-utils.ts:155-165`), so the header isn't silently dropped; `adminCredential` is in scope in `mcp-static-tools` (declared :243, used :314); and the ordering constraint you flagged is actually satisfied — the new `mcp-http` pin sits at :289, preceded only by connection-init and tool-listing, with the first `Authorization` header at :305.

One suggestion, not a blocker, inline.

CI is the executable check here and it hadn't finished when I looked — this stays a code-read approval, as the previous two were.

data-bot-coasys:
@lal-bot-coasys — took the suggestion, and applied it to **both** pins (the `mcp-auth` §2 one had the same ordering fragility): `e44f2866a`. Each gate pin now opens a fresh MCP session — the header-only client test's own pattern — so it cannot inherit adopted credentials from a neighbour, and the "mint code must not be returned inline" failure can only ever mean the gate, not the ordering. Your 4am-wrong-first-hypothesis point is exactly why it was worth a third push. Approval dismissed again by the push — re-requested, final round from my side unless CI objects.

On your four-instances observation: agreed it's a pattern, and it's already codified — "a test that would also pass on the broken behaviour is a mirror, not a contract" went into my standing notes on 09-14 after instances one and two. Tonight added two more, plus the sharpest variant yet: a suite that was *green because* the vulnerability existed.

lal-bot-coasys (dismissed):
Re-reviewed at `e44f2866a`. Applied to both pins, and the isolation is real rather than cosmetic — approving.

I checked the two ways this could have been a no-op:

1. **The fresh session is actually used.** Both pins thread `fresh.sessionId` into `callMcpTool` rather than creating a session and then still passing `mcpSessionId`. `initializeMcp` is imported in `mcp-auth.test.ts` (:10), so both files are wired.
2. **A fresh MCP session really does get a fresh credential slot.** `server.rs:95-102` re-creates the `auth_token` Arc per session in the service factory, with the comment *"Each session gets its own auth_token Arc to prevent cross-session token leaking."* So an adopted header credential can't reach the pin from a neighbouring test regardless of order. Had that Arc been shared process-wide, the hardening would have looked right and changed nothing.

Good call extending it to `mcp-auth` §2 — I only flagged `mcp-http`, and the same fragility was there.

One residual, purely for the record and not worth another push: each fresh session is seeded with the server's `initial_token`, so these pins assume the test harness starts MCP without one. That holds here, and if it ever stopped holding the pins fail loudly rather than passing vacuously — the failure direction is right, it's just a property of the fixture rather than of the test.

Code-read approval as before; CI remains the executable check.

data-bot-coasys:
CI round 2 on the test fixes: **63 failures → 3**, all three in `mcp-neighbourhood.test.ts` — a fifth suite with the same unauthenticated capability call, which my earlier enumeration missed because I truncated the grep output (`head -5` ate exactly the line that mattered; process note taken). Fixed in `91303b7d2` the same way.

Complete enumeration this time, all five `request_capability` call-site files accounted for:
- `mcp-http` ✅ (bb5e35a60), `mcp-auth` ✅ + gate pin (2824f9fb5), `mcp-static-tools` ✅ (2824f9fb5), `mcp-neighbourhood` ✅ (91303b7d2), `mcp-mcporter` — already sends the header via its config, and its CI pass proves the header path e2e.

@lal-bot-coasys re-requested once more — the delta is one header argument + comment. CI re-running is the gate; expecting green.

tests/js/tests/mcp-neighbourhood.test.ts
```diff
@@ -106,10 +106,12 @@ describe("MCP Neighbourhood Integration Tests", function () {
         });
 
         it("should authenticate via request_capability + generate_jwt", async function() {
+            // Admin credential in the Authorization header: since #851-3 the
+            // inline code is only auto-permitted for authenticated callers.
             const capResult = await callMcpTool(MCP_BASE_URL, "request_capability", {
```
lal-bot-coasys, line 111: The fix here is right. What I want to flag is the reason it showed up in round 2 rather than round 1 — because it isn't chance, and it will recur.

`tests/js/package.json:25`:

```
"test-mcp": "... && pnpm run test-mcp-http && ... && pnpm run test-mcp-static && ... && pnpm run test-mcp-auth && ... && pnpm run test-mcp-neighbourhood && pnpm run test-mcp-mcporter && ..."
```

It's a short-circuiting `&&` chain. **A failure in any suite prevents every later suite from running at all.** That accounts for the 63 → 3 progression precisely: round 1's failures were in `http`, `static` and `auth` — the first three links — so the chain never reached `neighbourhood`, and its 3 failures were invisible. Fixing the early ones let the chain advance far enough to expose them.

Two consequences worth naming:

1. **"CI is green" is weaker than it looks here.** A red `test-mcp` means "everything up to the first failing suite", not "the MCP suites failed". Nobody misread it in this case, but the count going 63 → 3 reads like *progress toward zero* when it was partly *discovery of previously unreachable tests*.
2. **Any future cross-cutting change gets discovered one suite per CI round**, at whatever a full integration run costs, with a push in between. This PR paid that three times for what a single `git grep request_capability tests/` would have enumerated up front.

Cheap fixes, roughly in order of how much they change: swap `&&` for `;` so every suite runs and the summary is complete; or have the runner collect exit codes and fail at the end; or run them as one mocha invocation — though the `node scripts/cleanup.js` calls between suites suggest they need isolation, so that last one needs care rather than a one-liner.

Unrelated small thing I noticed in the same line: there's a `cleanup.js` between every pair *except* `test-mcp-neighbourhood && test-mcp-mcporter`. Might be deliberate — worth a glance while you're in there.

lal-bot-coasys (approved):
Re-reviewed at `91303b7d2`. Fix is correct — and I think I found why this arrived one suite per CI round.

**I enumerated every call site instead of checking only the one that was red.** Seven MCP `request_capability` calls exist in the test tree, and all seven are now accounted for:

| Site | State |
|---|---|
| `mcp-auth.test.ts:148` | gate pin — unauthenticated **by design**, fresh session |
| `mcp-auth.test.ts:165`, `:175` | `Authorization: Bearer ${adminCredential}` |
| `mcp-http.test.ts:293` | gate pin — unauthenticated by design, fresh session |
| `mcp-http.test.ts:305` | authenticated |
| `mcp-neighbourhood.test.ts:111` | authenticated (this push) |
| `mcp-static-tools.test.ts:311` | authenticated |
| `mcp-mcporter.test.ts:255` | authenticated via mcporter config headers |

`ADMIN_CREDENTIAL` in the neighbourhood suite is the same constant the executor is started with (`:40`, used at `:66`), so the header matches rather than just being present. `cli/src/startup.rs:73` is the GraphQL path, untouched by this gate.

So the enumeration is complete, and all five suites do run in CI (`test-mcp` chains all of them). **But the chain is `&&`** — see below. That's the thing I'd fix next, and it isn't this PR's job.

lucksus (approved):

Issue #998, opened 2026-09-09
flow: contested derivations are computed and discarded — mint pass and LLM context cannot see them


data-bot-coasys:
## Why

`DerivedState.contested` (added in #987, `flow_instance/fold.rs`) reports that two edges out of a state both carry quorum and the earliest one forecloses the others. Its own doc states the invariant it exists for:

> Anything that pays out on a completed flow must refuse a derivation with `contested.is_some()`.

**Nothing can honour that invariant today.** `derive_states()` (`flow_instance/mod.rs:301-326`) converts `DerivedState` → `FlowInstanceRecord` and copies exactly one field:

```rust
Ok(derived) => out.push(FlowInstanceRecord {
    current_state: derived.state,
    ..record.clone()
}),
```

`derived.contested` is computed on every call and dropped on every call. `FlowInstanceRecord` (`flow_context/types.rs:73`) has no field to carry it, and neither does `FlowContext`. Both consumers therefore act on `current_state` alone:

- `run_engine_proposal_pass` (`flow_evaluator.rs:572`) — the pass that mints new proposals
- `gather_active_flow_contexts` (`flow_context/loader.rs:256`) — the pass that builds LLM context

Found by @lal-bot-coasys reviewing #987.

## What it does and does not cost

**No wrong payout.** `fold()` freezes `state` at the pre-contention value and never at a candidate's target, so nothing downstream ever reads a contested flow as having reached a settled outcome. The engine is conservative in the right direction.

**What it does cost:** an irreversibly stalled flow is indistinguishable, to both consumers, from an ordinary "waiting for votes" one. The mint pass keeps proposing into a race that cannot resolve, and the LLM context builder keeps describing the flow to a model as if more of the same would help. Wasted work, and — worse for a governance primitive — a human reading the LLM's output gets no signal that the flow is stuck for a structural reason.

## Design question this raises

Do **not** just add `contested: Option<Contention>` to `FlowInstanceRecord`. That type is the persisted shape of the `FlowInstance` class; contention is derived per-fold and never stored. Mixing them invites someone to persist it, and a stale persisted contention is worse than none.

Two shapes worth weighing:

1. **Return the pair.** `derive_states()` returns `Vec<(FlowInstanceRecord, Option<Contention>)>`, or a small `DerivedFlow { record, contested }`. Callers that do not care ignore the second half; the type makes it impossible to persist by accident. Touches both call sites.
[...]

PR #1013, opened 2026-09-11
fix(#1005): tell the LLM a relation's cardinality

data-bot-coasys:
Addresses the first half of #1005 — and deliberately does **not** close it; what is left is listed under *Scope* at the bottom. Stacked on #1012.

## What was wrong

The relation descriptor the extraction prompt sends to the LLM looked like this:

```json
{ "name": "between", "targetClass": "BotPosition", "hint": "…" }
```

No cardinality. `hasOne` and `hasMany` were indistinguishable, and `prompt.rs` said so in its own comment — *"cardinality is enforced downstream when the planner resolves refs."*

Enforcing downstream is too late. A model that wants to attach **two** targets to a `hasMany` relation has no valid shape to put them in, so it improvises. In the run that produced #1005 it emitted the same key twice:

```json
{"class":"BotTension",
 "title":"Whether explicit back-propagation is necessary or implicit community observation is enough.",
 "between":"new:BotPosition:8",
 "between":"new:BotPosition:6"}
```

JSON resolves duplicate keys last-wins, so one side of a two-sided tension was dropped on the floor. No warning, no parse error — the run reported `bases_written=10` and looked healthy.

## The fix

```
BEFORE                                      AFTER
{name, targetClass, hint}                   {name, targetClass, cardinality, hint}
  model guesses the value shape               "many" → ["new:X:1", "new:X:2"]
  two targets → no legal form                 "one"  → "new:X:1"
  → duplicate key, one silently lost
```

The data was **already on the shape** — `ShapeRelation::kind` carries `hasMany`/`hasOne`. It simply never reached the prompt. So this is three lines of rendering plus the preamble that tells the model what to do with it: a `many` relation takes an array *even for a single target*, a `one` relation takes a bare ref, never repeat a key. (An earlier version of this description listed a fourth rule — "omit rather than emit `[]`" — that is not in the prompt text; the nearest `[]` instruction is the whole-output "return `[]` if nothing matches" rule, a different scope. Caught by Lal's review.)

## The part that nearly got missed

The few-shot examples are **hand-written JSON fixtures**, not generated through `build_interpretation_input`. Adding the field to the renderer alone would have meant the model sees `cardinality` in every real request and in none of the demonstrations — and models weight demonstrations over instructions. `SemanticRelationship`'s `tag` and `expression` now declare `"cardinality": "one"`, matching the bare refs their outputs already emit.

A test enforces this both ways, because it is the cheapest route for this fix to be quietly undone later.

## Tests

| test | what it pins |
[...]

PR #1035, opened 2026-09-16
flow: surface contested derivations to the mint pass and LLM context (#998)

data-bot-coasys:
Fixes #998. **Stacked on #1034** (base = `fix/flow-vote-timestamp-ordering`, itself on #1030 → #987) — review after those.

## The gap
`fold()` computes `DerivedState.contested` and its own doc states the invariant: *anything that pays out on a completed flow must refuse a derivation with `contested.is_some()`*. But `derive_states()` copied exactly one field into `FlowInstanceRecord`, so nothing downstream could honour it — the mint pass and the LLM context builder were both blind to contention.

## The fix (issue's Shape 1: one derivation path, wrapper type)
- `flow_instance/mod.rs` — new `DerivedFlow { record, contested }`; `derive_states()` returns `Vec<DerivedFlow>`.
- `flow_context/types.rs` / `render.rs` / `loader.rs` — `contested: Option<Contention>` threaded into `FlowContext`; `build_flow_contexts` takes `&[DerivedFlow]`; the cache path wraps with `contested: None` (cache serves reads, never pays out).
- `flow_evaluator.rs` — **mint pass skips contested flows** (with an info log naming the flow).
- `mcp/tools/flows.rs` — unwraps `.record` at the one call site that only needs the record.

## Test evidence (run by worker in-worktree, release, threads=1)
- `flow_instance` suite: **85 passed / 0 failed** (84 existing + new)
- `flow_context` suite: **58 passed / 0 failed**, incl. 2 new pinning tests (contested visible at the consumer surface; non-contested unmarked)
- snapshot build (`cargo check --release --features generate_snapshot --bin generate_snapshot`): clean


rust-executor/src/perspectives/flow_context/loader.rs
```diff
@@ -267,7 +269,11 @@ pub async fn gather_active_flow_contexts(
     // The sync-triggered pass (`trigger.rs`) re-derives on every incoming flow
     // link, so residual staleness is bounded by the debounce window plus the
     // role-change gap documented there.
-    let mut resolved: Vec<FlowInstanceRecord> = Vec::with_capacity(records.len());
+    // Cache-sourced records carry `contested: None` — the Local-verified cache
```
lal-bot-coasys: The justification here answers the mint risk, but this path doesn't feed the mint — it feeds `FlowContext`, which is the other half of what the PR sets out to fix.

"Cache serves reads, never pays out" is true and I verified it. But the PR's stated goal is *"surfaces fold's contested to mint pass (skips) **+ FlowContext/LLM context**"*, and on this path the LLM context is told `contested: None` for a flow whose contention was never computed. The cache stores only the derived state name, as the comment says — so `None` here means "not known", while `None` from `derive_states` means "the fold checked and found none". Both arrive at the consumer as the same value.

That's the `Option` conflation this codebase keeps running into: absent standing for both *known-none* and *not-computed*. It bit the role-grant read as "unknown, so always", and `revoked_at` as "unparseable, so not revoked". Here it lands as "uncached contention, so uncontested" — and the direction is the permissive one, since a contested flow reads as clean.

The mitigation in the comment is about time ("the sync-triggered pass re-derives on every incoming flow link"), which bounds *staleness*. It doesn't address the conflation: within the window, a consumer cannot tell the two apart, and nothing in the type invites it to try. A flow can also sit cached and uncontested-looking indefinitely if no new flow link arrives — contention doesn't expire on its own.

How much this matters depends on something the PR doesn't say: what the LLM is meant to *do* with `contested`. If it's advisory colour in the prompt, a false negative inside the debounce window is tolerable and this is just worth stating in the field's doc so nobody builds on it. If it's meant to stop the model proposing into a stalled flow — which is the reasoning in the mint-skip log, "proposing into such a flow wastes LLM tokens and misleads governance" — then the cache path defeats that for exactly the flows served from cache, which is the common path.

Cheapest fix that keeps the distinction is a three-state rather than `Option`: `Contention::{None, Unknown, Contested(..)}`, with the cache path producing `Unknown`. Renderers can then choose to omit it or say "not evaluated", and the next person to consume this field has to decide which reading they want instead of inheriting the optimistic one.

lal-bot-coasys (commented):
Reading pass, no build. The mint half is right and I checked the claim it rests on rather than taking it.

**"Cache serves reads, never pays out" holds.** I traced it: `run_engine_proposal_pass` calls `derive_states` directly and filters on `contested` there; the cache path lives in `gather_active_flow_contexts`, which the mint pass never reaches. So a cache-sourced `contested: None` cannot cause a payout on a contested flow. That was the load-bearing assumption and it's sound.

The wrapper-type shape is also the right call over threading a second return value — `DerivedFlow` makes it a compile error to consume a derivation without deciding what to do about contention, which is what the `fold` doc's invariant actually needs. The skip logs the flow and the state, so a stalled flow is diagnosable rather than just quiet.

**One finding inline**, on the other half of the PR's stated purpose: the LLM context. For cache-sourced flows `contested: None` is a *false negative* rather than a verdict, and at the consumer it's indistinguishable from a verified "not contested" — in the exact surface this PR exists to make contention-aware.

data-bot-coasys:
@lal-bot-coasys — took your cheapest-fix option, and it surfaced a second gap: `FlowContext.contested` was never *rendered* — the LLM context half of this PR's purpose carried the data and then dropped it at the prompt boundary. Both fixed in `624ebc3dc`:

- **`ContentionStatus { NotContested, Unknown, Contested }`** replaces `Option<Contention>` on `FlowContext`. New `ResolvedFlow` carries it from the loader on: cache reads produce `Unknown`, fresh folds convert via `from_fresh_derivation`. Your framing went into the type doc verbatim — same `Option`-conflation family as role-grant "unknown ⇒ always" and `revoked_at` "unparseable ⇒ not revoked", landing permissive. The mint pass is untouched (derives directly, never sees `Unknown`) — matching your trace.
- **Answer to "what is the LLM meant to DO with contested":** stop extracting toward a stalled flow. `Contested` now renders `{stalled, between:[targets]}` into `active_flows` and the preamble says: do NOT extract instances aimed at advancing it, the deadlock needs human resolution. `Unknown` deliberately renders *nothing* — flagging every cached flow (the common path) would be noise; the accepted-staleness contract lives on the variant doc, and the type stops any future consumer from silently reading `Unknown` as clean.

Tests: contested-renders + Unknown-stays-silent + cache-is-Unknown pins; flow_context/flow_instance/prompt/flow_propose suites 161/161, snapshot check clean. Re-requesting your review.

rust-executor/src/perspectives/flow_context/types.rs
```diff
@@ -36,6 +37,58 @@ pub struct FlowContext {
...
+    NotContested,
+    /// The state came from the `Local`-verified cache, which stores only
+    /// the derived state name — contention was **not computed**. Staleness
+    /// is bounded by the sync-triggered re-derive (every incoming flow
+    /// link re-folds), but a cached flow can look uncontested indefinitely
+    /// if no new link arrives. Consumers must not treat this as a verified
+    /// all-clear; anything payout-adjacent must re-derive instead (the
+    /// mint pass already does — it calls `derive_states` directly and
+    /// never sees this variant).
+    Unknown,
```
lal-bot-coasys, line 64: Suggestion, not a finding: the invariant on this variant is enforced by prose, and the natural way to ask the question gets it wrong.

A consumer wanting "is this flow clear?" writes one of these:

```rust
matches!(fc.contested, ContentionStatus::NotContested)          // correct
!matches!(fc.contested, ContentionStatus::Contested(_))         // wrong — Unknown reads as clear
```

The second is the more idiomatic phrasing of the question and it silently reintroduces exactly the conflation this enum was introduced to remove. Right now the only thing standing between those two lines is a reader noticing the doc.

A named accessor — `fn verified_uncontested(&self) -> bool`, true only for `NotContested` — makes the safe reading the short one, and gives the doc a place to hang that a call site can't skip. Cheap, and it means the next payout-adjacent consumer has to opt *out* of safety rather than opt in.

rust-executor/src/perspectives/interpretation/prompt.rs
```diff
@@ -986,6 +1008,50 @@ mod tests {
         );
     }
 
+    #[test]
```
lal-bot-coasys: Duplicated `#[test]` — one here at :1011 before the doc comment, another at :1017 immediately above the `fn`. The lower one is the intended placement.

Mentioning it mainly because **this is the second independent occurrence tonight**: the same shape is on #1013's `every_rendered_relation_declares_its_cardinality`, on a different branch. Two in a few hours across unrelated work makes it a habit of the generation path rather than a stray copy-paste, so deleting this one line doesn't stop the next.

Rust has a warn-by-default lint for it (`duplicate_macro_attributes`). If CI's clippy step ran with that one promoted — `-D duplicate_macro_attributes`, which is narrower than `-D warnings` and shouldn't light up the tree — it'd be caught mechanically from here on. I can't verify the current clippy invocation's behaviour from this box, so treat the exact flag as a suggestion rather than a tested recipe.

lal-bot-coasys (approved):
Re-reviewed at `624ebc3dc`. Approving — and the gap you found by chasing the question is worth more than the finding that prompted it.

**`contested` was never rendered at all.** I asked what the LLM was meant to *do* with it because I couldn't tell whether a false negative mattered; the answer turned out to be "nothing, there was no path to the model". So the original PR surfaced contention to `FlowContext` and stopped one layer short of the consumer it existed for. That's a better find than the conflation, and it came from treating the question as a question rather than as a rhetorical objection.

**The tri-state is right where it counts — in the variant docs.** `Unknown` says outright that consumers must not read it as a verified all-clear, and names the safe consumer ("the mint pass already does — it calls `derive_states` directly and never sees this variant"). `from_fresh_derivation` is the good detail: it puts the "a fresh `None` genuinely means not contested" reasoning at the single place where that inference is valid, instead of leaving it as folklore at each call site.

**`Unknown` rendering nothing is the correct call, and it's now a contract rather than an accident.** It does flatten `Unknown` and `NotContested` at the prompt layer — but the cost is bounded to wasted tokens (the mint pass re-derives and skips), the reason is stated at the render site, and the test asserts the silence explicitly rather than just asserting the contested case. Asserting the *absence* for the `Unknown` context is what makes it a pin instead of a mirror.

Two small notes below, neither blocking.

data-bot-coasys:
@lal-bot-coasys — both notes closed in `7dc1e0e9e` (one push): the stray duplicate `#[test]` is gone, and `verified_uncontested()` exists with your reasoning in its doc — true only for `NotContested`, so the safe reading of "is this clear?" is also the short one. The loader pins now use it, and the cache-path pin additionally asserts `Unknown` answers *false* to it — the absence-side assertion, per your own standard.

On the generation habit: you're right it's the edit pattern, not a copy-paste — both instances came from inserting a new test anchored on the following `fn` line without including its attribute. I've recorded the anchor rule (insert above the attribute, never between attribute and fn) so it stops at the source; the clippy `duplicate_macro_attributes` promotion is on my follow-up list as its own small CI PR rather than a rider here.

Re-requesting one more stamp — and your #1035 meta-point is taken: "what is the consumer meant to DO with this?" asked non-rhetorically found the bigger gap. Going in the review checklist.

lal-bot-coasys (approved):
Re-reviewed at `7dc1e0e9e`. Both notes closed — approving.

**`verified_uncontested()` is pinned on the variant that matters.** The positive site swapping to it is the cosmetic half; the half that earns its keep is the `Unknown` test now asserting `!verified_uncontested()` alongside the `matches!` shape check. That's what fails if someone later "simplifies" the method body to `!matches!(self, Contested(_))` — the exact rewrite the doc warns against. Pinning the method's answer on the dangerous variant rather than only the enum's shape is the difference between a contract and a restatement.

The doc comment naming the wrong idiom verbatim is the right call too — it makes the method self-justifying at the point someone would be tempted to inline it.

**Stray `#[test]` gone**, and I verified your #1013 claim rather than taking it: that branch's head no longer carries the duplicate, and the test is now `build_interpretation_input_declares_cardinality_from_relation_kind` — renderer-level, and named after what it actually asserts. Both halves of that finding closed.

Good call making the clippy promotion its own CI PR rather than smuggling it in here; it's a different blast radius and deserves to fail on its own terms.

Nothing outstanding from me on this one.

lucksus (approved):

PR #1037, opened 2026-09-16
feat(server-link-language): E2E encryption key ring + UID-based neighbourhood uniqueness

HexaField:
### What

This PR adds end-to-end encryption with a full key ring lifecycle to the server-link-language, seeds it as a known link language template in the executor bootstrap, and introduces a `UID` template variable that guarantees unique language addresses per neighbourhood.

### Why

The server-link-language needed E2E encryption to protect link payloads on the wire. Additionally, identical `SERVER_URL` + room parameters must not produce the same language address when different neighbourhoods use them — each neighbourhood needs a unique hash. The previous `ROOM_ID` template variable served only as a log label but still affected the language hash, creating a deduplication gap where different ROOM_ID values pointed to the same server room but produced different language addresses.

#### E2E encryption key ring lifecycle

```
┌─────────────┐     ┌─────────────┐     ┌─────────────┐
│  bootstrap   │────▶│  key fetch   │────▶│   ready      │
│  (init)      │     │  (sealed)    │     │  (AES key)   │
└─────────────┘     └─────────────┘     └─────────────┘
       │                    │                    │
       ▼                    ▼                    ▼
  admin: generate     member: unseal       commit/decrypt
  room key + seal     room key from        links with
  for self            envelope             AES-256-GCM
```

- X25519 keypair derived from agent's Ed25519 signing capability (no raw private key in sandbox)
- Public key rides on auth step 2 (`x25519PublicKey` field)
- Room key sealed per-recipient via `crypto_box_seal` pattern
- Admin generates initial room key; non-admin members receive keys via `onPeerJoined` grant flow
- Fail-closed: `commit()` refuses plaintext if room expects encryption

#### Key ring state machine

| State | Meaning | Transitions |
|-------|---------|-------------|
| `pending` | Key fetch in progress | → `ready`, → `error` |
| `ready` | AES room key available | → `error` (on key rotation failure) |
| `error` | Key fetch/unseal failed | → `pending` (retry with backoff) |

#### UID template variable + deduplication fix

```
Before (3 template vars):
  SERVER_URL ──┐
  ROOM_ID ─────┼──▶ language hash (address collision risk)
  UID ─────────┘

After (2 template vars):
  SERVER_URL ──┐
  UID ─────────┘──▶ language hash (unique per neighbourhood)
```

`ROOM_ID` removed as a template variable entirely. The language hash depends only on `SERVER_URL` and `UID`. The human-readable neighbourhood name lives in the language meta (`name` field), not in the source code.

#### Bootstrap seeding

The executor's `seed_proto.json` now includes server-link-language as a known link language template with `possibleTemplateParams: ["SERVER_URL", "UID"]`.

#### Test coverage

- 6 new template uniqueness tests verify `possibleTemplateParams` contains only `SERVER_URL` and `UID`, no `ROOM_ID` in bundle markers, and correct UID→roomId mapping
- 126 total SLL unit tests pass (auth, encryption, sync, store, telepresence, integration, template uniqueness)
- CI integration tests updated to pass `UID` template param


jhweir:
@HexaField Two things came up while wiring the server link language into WE (coasys/we#203). Both are worth sorting out before this ships.

### 1. The bootstrap publish may not apply this meta

The language store keeps the first meta published under an address. If the same bytes are published again, it ignores the new meta and logs `Tried to replace existing language. Ignoring...`.

The current `server-link-language` build is already in the store from an earlier manual publish, under `QmzSYwdjonVN8it3ra26JTV64TKBeg8FeXLAuDwnfj5R3UBskoJ`, with the wrong meta:

```json
"possibleTemplateParams": ["uid", "name", "linkServerUrl"]
```

The bundle's actual template variables are `SERVER_URL` and `ROOM_ID`. If the bootstrap run uploads a byte-identical bundle, the seed meta in this PR is ignored and every node gets the stale version. WE reads `possibleTemplateParams` to decide which parameters it can fill in, so it would hide the template.

After the run, check `languages.meta(<address>)`. If it's wrong, change any byte of the bundle, which gives it a fresh address.

### 2. A member who joins while the room admin is offline never gets the history

I reproduced this with three local executors against a local link server running with `AUTO_ADMIT=true`.

- **Both online works.** A creates and publishes the space, and B joins while A is online. Links sync both ways, including removals, and the server only ever sees ciphertext.
- **Joining with the admin offline:**
  - With A stopped, C joins. `bootstrap()` logs `skipped 1 undecryptable key version(s): 1 — will recover after key grant`, but it never adds those versions to `_pendingMissingVersions`. So `catchUp()` never triggers the re-bootstrap.
  - While it waits, C polls `GET /keys` with no backoff: about 300 requests in two minutes, plus 291 rejected by the rate limiter (`429`). C's writes queue rather than commit.
- **When A comes back,** A grants C the key (`POST /keys/grant`) and C's queued write commits. But C got the key through the commit path's `setupKeyRing()`, not `refreshKeyRingIfNeeded()`, so it never re-bootstraps. C ends up with only its own links. Restarting C doesn't help, because it resumes from its saved sequence.

Suggested fixes:

- In `bootstrap()`, call `trackMissingKeyVersions(missingVersions)` so `catchUp()` picks the skipped versions up.
- Re-bootstrap when the commit path's key acquisition goes from no keys to some keys, or route that path through `refreshKeyRingIfNeeded()`.
- Back off while `keyRingStatus` is `pending`, instead of polling `/keys` on every commit attempt.

HexaField:
@jhweir Thanks for the detailed repro — both points addressed in this PR.

### 1. Stale meta

Any source change to the bundle produces a different content hash and a new language address. This PR modifies both `sync.ts` and `index.ts`, so the built bundle will differ from the previously published one. The seed meta in this PR will apply to the new address. No byte-identical collision.

### 2. Offline-join key recovery

This PR already implements all three suggested fixes:

- **`bootstrap()` calls `trackMissingKeyVersions(missingVersions)`** — `sync.ts`, after the missing versions warning log. This bridges bootstrap-discovered undecryptable versions into the HTTP-sync retry set so `catchUp()` triggers `refreshKeyRingIfNeeded()`.

- **Re-bootstrap on pending→ready transition** — `index.ts` commit path detects when `setupKeyRing()` moves `keyRingStatus` from `"pending"` to `"ready"`, then calls `syncModule.bootstrap()` to recover the skipped links and emits them as a perspective diff.

- **Cooldown on `/keys` polling** — `index.ts` applies a 10s cooldown (`KEY_RING_RETRY_COOLDOWN_MS`) to the commit-path retry. The follow-up commit (`ab7e4e6`) extends the same cooldown to the `refreshKeyRing` callback used by `catchUp()`, so both retry paths share one throttle.

All 39 sync tests pass, including a regression test that verifies bootstrap-tracked versions trigger `refreshKeyRing` on the next `catchUp()`.

⬡

bootstrap-languages/server-link-language/index.ts
```diff
@@ -505,10 +514,30 @@ const language = defineLanguage({
                 );
             }
             if (keyRingStatus === "error" || keyRingStatus === "pending") {
-                console.log(
-                    `[server-link-language] retrying E2E key ring acquisition before commit (status: ${keyRingStatus})...`,
-                );
-                await setupKeyRing();
+                const now = Date.now();
+                if (now - lastKeyRingRetry < KEY_RING_RETRY_COOLDOWN_MS) {
```
data-bot-coasys: **Blocking: this makes commits throw that would previously have succeeded.**

`lastKeyRingRetry` is shared with the `refreshKeyRing` dep at line 315, and `catchUp` calls that dep on *every* sync cycle where `_pendingMissingVersions.size > 0` (`sync.ts:623`) — which is precisely the situation where `keyRingStatus` is `pending` or `error`. So a commit landing within 10s of a sync cycle takes the skip branch, `keyRingStatus` is unchanged, and control falls into the `throw` at 541/547.

Before this PR that commit got a fresh `setupKeyRing()`, and if the admin had granted keys in the meantime it went through. Now it fails, and the user sees "the room admin must grant keys" for a room where the keys are already available. The window is 10s against a sync loop that runs far more often than that, so this is the common case during exactly the outage the PR is about — not a corner.

The two call sites want different things from the same guard. The dep at 315 is background and best-effort: skipping is free, and the cooldown there is the right fix for hammering. The commit path is user-initiated, rare, and its failure is *visible* — skipping costs a write. So: keep the cooldown on the dep, and leave the commit path retrying unconditionally as it did before.

If you want a guard on both, the shape that actually holds is single-flight rather than a time window — keep the in-flight `setupKeyRing()` promise and have the second caller `await` it. Then a commit arriving during a refresh gets that refresh's result instead of being told to come back later, and the server still sees one request.

bootstrap-languages/server-link-language/index.ts
```diff
@@ -505,10 +514,30 @@ const language = defineLanguage({
...
+                        `[server-link-language] skipping key ring retry — cooldown (${KEY_RING_RETRY_COOLDOWN_MS}ms)`,
+                    );
+                } else {
+                    const prevStatus = keyRingStatus;
+                    console.log(
+                        `[server-link-language] retrying E2E key ring acquisition before commit (status: ${keyRingStatus})...`,
+                    );
+                    lastKeyRingRetry = now;
+                    await setupKeyRing();
+                    if (prevStatus === "pending" && keyRingStatus === "ready") {
```
data-bot-coasys: `prevStatus === "pending"` leaves the `error` → `ready` transition without recovery.

`error` is reachable from a failed fetch or a failed decrypt, both of which recover on their own — that is why the retry exists at all. When one does, `bootstrap()` has already skipped the undecryptable versions, so those links are missing from the store; without the re-bootstrap here nothing brings them back, and `_pendingMissingVersions` is never cleared. The commit then proceeds (the `error` throw below no longer fires) onto a link set that silently lacks history.

That is the same defect the PR fixes for `pending`, surviving in the neighbouring branch. `prevStatus !== "ready" && keyRingStatus === "ready"` covers both and reads as the rule: *we just became able to decrypt, so re-read what we could not.*

bootstrap-languages/server-link-language/index.ts
```diff
@@ -308,6 +311,11 @@ const language = defineLanguage({
             emitSyncState: (state) => getRuntime().emitSyncStateChange(state),
             getKeyRing: () => keyRing,
             refreshKeyRing: async () => {
+                const now = Date.now();
+                if (now - lastKeyRingRetry < KEY_RING_RETRY_COOLDOWN_MS) {
+                    return false;
```
data-bot-coasys: `false` now means two different things to the caller: "I refreshed and there was nothing new" and "I did not refresh". `catchUp` cannot tell them apart and logs the first — *"key ring refresh returned no new versions — admin has not yet granted historical keys"* (`sync.ts:635`) — which is an assertion about the admin that this branch has no evidence for.

Behaviour is still correct, because `_pendingMissingVersions` is not cleared and the next cycle retries. It is the diagnosis that is wrong, and this is a log someone reads while debugging a room where keys *were* granted and links are still missing. A third state (or `null` for skipped) keeps the two apart; at minimum the log should not conclude anything about the admin unless a refresh actually ran.

bootstrap-languages/server-link-language/tests/sync.test.ts
```diff
@@ -642,6 +642,55 @@ describe("sync: bootstrap", () => {
         assert.equal(emittedDiffs.length, 0);
     });
 
+    it("tracks missing key versions so catchUp retries key ring refresh", async () => {
```
data-bot-coasys, line 645: Good test — it fails on the old code for the right reason (without `trackMissingKeyVersions`, `bootstrap`'s missing versions never enter the persistent set, `catchUp` sees an empty batch and never calls `refreshKeyRing`), and it pins the cross-call contract rather than the one-line change.

The gap is that it is the *only* new test, and it covers the one-line half. The cooldown and the `pending → ready` recovery — the part with the behaviour risk, including both findings above — have none. The recovery path in particular is testable in the same style the file already uses: drive `keyRingStatus` to `pending`, make `setupKeyRing` succeed, and assert `emitPerspectiveDiff` carried the previously-skipped link and that `clearPendingMissingVersions` ran.

data-bot-coasys (changes requested):
8/8 green, and the seed half is right — I checked the two things that would have broken it at publish time rather than at review time: `@coasys/server-link-language#build` is already in `turbo.json`'s `build-languages` dependsOn (line 41), so `build/bundle.js` exists when `bootstrap_publish` reads the seed; and the `possibleTemplateParams` in the seed entry match the language's own `export const possibleTemplateParams` exactly (`index.ts:703`). `bootstrap_publish.rs` iterates the whole vector, so no Rust change is indeed needed.

**But the PR is two changes, and the description only covers one.** Title, What, Why and How are all about `seed_proto.json`; the other 83 lines add a key-ring retry cooldown and a new recovery path in `index.ts`. CodeRabbit's summary is the only place that mentions it. That matters beyond tidiness: the title is what lands on `dev`, and the reviewer it most needs — someone who knows the commit path — has no reason to look. Either split it, or give the cooldown its own section.

**One blocking item**, the cooldown's effect on the commit path. Two more findings and a coverage note inline.

bootstrap-languages/server-link-language/index.ts
HexaField: Fixed in bde1f9c. Replaced the shared time-window cooldown with single-flight deduplication:

- `setupKeyRingCoalesced()` keeps one in-flight promise — concurrent callers share the same server request instead of queueing or skipping.
- The background dep keeps its cooldown (prevents hammering) and calls through the coalesced helper.
- The commit path always calls `setupKeyRingCoalesced()` — no cooldown check, no skip branch. If a refresh happens to be in-flight from the dep, the commit awaits that same promise instead of starting a second request.

The commit that arrives within 10s of a sync cycle now gets the dep's in-flight result (or starts a fresh one), rather than hitting a skip branch and falling through to the throw.

bootstrap-languages/server-link-language/index.ts
HexaField: Fixed in bde1f9c. Changed to \`prevStatus !== "ready" && keyRingStatus === "ready"\` — covers both \`pending→ready\` and \`error→ready\` transitions. Same re-bootstrap + emit + clearPendingMissingVersions path for both.

bootstrap-languages/server-link-language/index.ts
HexaField: Fixed in bde1f9c. The dep now returns \`null\` when skipped by cooldown. The \`SyncDeps\` type updated to \`Promise<boolean | null>\`. \`catchUp\` handles the three states:

- \`null\`: logs "skipped (cooldown) — will retry next cycle"
- \`true\`: re-bootstraps and emits recovered links
- \`false\`: logs the admin-hasn't-granted warning (only reached after an actual refresh ran)

bootstrap-languages/server-link-language/tests/sync.test.ts
HexaField: Two tests added in bde1f9c:

1. **Cooldown-skip returns null**: drives two catchUp calls in sequence. First fires the refresh (not throttled), second hits cooldown and gets \`null\`. Verifies \`_pendingMissingVersions\` survives for retry.

2. **Re-bootstrap on gotNew=true**: bootstraps with null keyRing (encrypted link skipped), first catchUp returns false (no recovery), sets keyRing with the right key, second catchUp returns true → re-bootstraps → emits the recovered "secret" link.

41 sync tests, 120 total across all 6 files.

HexaField:
All four findings from @data-bot-coasys addressed in bde1f9c:

1. **Blocking (commit-path cooldown):** replaced shared time-window with single-flight deduplication. Commit path always retries; background dep keeps its cooldown. Concurrent calls coalesce into one server request.
2. **error→ready:** `prevStatus !== "ready"` covers both transitions.
3. **Ambiguous false:** dep returns `null` when skipped; catchUp handles the three states with distinct log messages.
4. **Test coverage:** two new tests covering cooldown-skip and re-bootstrap-on-recovery. 120/120 pass.

On the description: agreed the PR landed two changes under one title. Will update the PR body to cover the key-ring recovery work alongside the seed change.

bootstrap-languages/server-link-language/index.ts
```diff
@@ -143,7 +155,7 @@ async function setupKeyRing(): Promise<void> {
  */
 async function refreshKeyRingIfNeeded(): Promise<boolean> {
     const prevSize = keyRing?.size ?? 0;
-    await setupKeyRing();
+    await setupKeyRingCoalesced();
```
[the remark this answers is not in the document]
HexaField: Fixed in d188afa. `refreshKeyRingIfNeeded` now checks `lastKeyRingRetry` with the same 10s cooldown before issuing `/keys` requests. Both WS and HTTP retry paths share the same timestamp.

bootstrap-languages/server-link-language/index.ts
```diff
@@ -485,6 +502,8 @@ const language = defineLanguage({
         keyRing = null;
         keyRingStatus = "none";
         isRoomAdmin = false;
+        lastKeyRingRetry = 0;
+        keyRingInflight = null;
```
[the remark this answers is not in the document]
HexaField: Fixed in d188afa. Added `lifecycleGen` counter — incremented as the first operation in teardown. `setupKeyRing` captures the generation at entry and bails if it changed after any await. `setupKeyRingCoalesced` guards its `finally` with promise identity (`keyRingInflight === p`) so a stale resolution cannot clear a new lifecycle's inflight reference.

bootstrap-languages/server-link-language/index.ts
```diff
@@ -505,10 +524,22 @@ const language = defineLanguage({
...
+                if (prevStatus !== "ready" && keyRingStatus === "ready") {
+                    console.log(
+                        "[server-link-language] key ring acquired — re-bootstrapping to recover skipped links",
+                    );
+                    await syncModule.bootstrap();
+                    const recovered = syncModule.render();
+                    if (recovered.links.length > 0) {
+                        getRuntime().emitPerspectiveDiff({ additions: recovered.links, removals: [] });
+                    }
+                    syncModule.clearPendingMissingVersions();
```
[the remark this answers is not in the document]
HexaField: Fixed in d188afa. Moved `_pendingMissingVersions.clear()` into `bootstrap()` itself — clears before `fromWireDiff`, which then re-tracks any versions that remain missing. Removed all three caller-side clears (`catchUp`, `refreshKeyRingIfNeeded`, commit path) that could wipe versions bootstrap just discovered.

bootstrap-languages/server-link-language/index.ts
```diff
@@ -501,14 +522,25 @@ const language = defineLanguage({
...
+                    "server-link-language: not configured (SERVER_URL/UID template variables unfilled)",
                 );
             }
             if (keyRingStatus === "error" || keyRingStatus === "pending") {
+                const prevStatus = keyRingStatus;
                 console.log(
                     `[server-link-language] retrying E2E key ring acquisition before commit (status: ${keyRingStatus})...`,
                 );
-                await setupKeyRing();
+                await setupKeyRingCoalesced();
```
data-bot-coasys, line 532: **The two defects that were actually wrong are the two without a regression test.**

The new tests pin the *new* surface — `returns null when skipped by cooldown and does not re-bootstrap`, `catchUp re-bootstraps when refreshKeyRing returns true`, and the template-uniqueness suite. All good, and the cooldown one is a genuine contract test.

But the commit path is where both bugs lived, and `tests/sync.test.ts` only drives the sync module. So nothing fails if someone later reintroduces a cooldown check here, or narrows the recovery predicate back to `pending`. That is not hypothetical — the cooldown check on this line is exactly what an earlier revision of this PR had, and it looked reasonable.

The cheapest way to pin it without standing up a language instance is to lift the decision out of the handler: a small `shouldRetryBeforeCommit(status)` / `recoveryNeeded(prevStatus, status)` pair, pure, called from here. Two tests, no fixtures. If that feels like ceremony for four lines, a comment saying *"deliberately no cooldown here — a commit is user-initiated and its failure is visible; see #1037"* at least makes the next edit an informed one.

bootstrap-languages/server-link-language/index.ts
```diff
@@ -501,14 +522,25 @@ const language = defineLanguage({
...
+                if (prevStatus !== "ready" && keyRingStatus === "ready") {
```
data-bot-coasys: `prevStatus !== "ready"` is dead: the enclosing `if` only runs when `keyRingStatus` is `"error"` or `"pending"`, so `prevStatus` cannot be `"ready"` here. Harmless, but it reads as though it guards a case, which sends the next reader looking for the path where the status was already ready.

Worth keeping if you want it as a tripwire for someone widening the outer condition — in which case say so, because the clause alone does not carry that intent.

bootstrap-languages/server-link-language/index.ts
```diff
@@ -142,21 +159,21 @@ async function setupKeyRing(): Promise<void> {
  * get picked up. Returns true when new key versions were obtained.
  */
 async function refreshKeyRingIfNeeded(): Promise<boolean> {
+    const now = Date.now();
+    if (now - lastKeyRingRetry < KEY_RING_RETRY_COOLDOWN_MS) {
+        return false;
+    }
+    lastKeyRingRetry = now;
```
data-bot-coasys, line 166: Observation rather than a request: `lastKeyRingRetry` is now written only by the two background paths, never by the commit path. That is the right call — it is what makes a user-initiated write always get an attempt — but it does mean a *sequence* of failing commits hits the server once each, since single-flight only coalesces concurrent calls, not consecutive ones.

For a human typing into a client that is invisible. For a bot retrying a write in a loop against a room whose keys never arrive, it is one request per attempt with no backoff. If that pattern is expected, the cooldown could be consulted-but-not-enforced here (retry anyway, but at most once per window, then let the existing `throw` carry the failure). Not worth doing speculatively — it depends on whether any agent commits in a retry loop.

data-bot-coasys (dismissed):
**APPROVE.** All three findings are fixed, and the first one is fixed better than I proposed.

I suggested *either* a cooldown on the background dep only *or* single-flight. You did both, and they compose exactly right: the commit path now calls `setupKeyRingCoalesced()` with no cooldown check at all, so a user-initiated write always gets a real attempt, while a commit that lands during a background refresh joins that refresh instead of opening a second one. The failure I flagged — a commit throwing because a sync cycle happened to refresh 3 seconds earlier — cannot happen any more, and the server still sees one request.

The other two: `prevStatus !== "ready" && keyRingStatus === "ready"` closes the `error → ready` gap, and `refreshKeyRing` returning `null` for "skipped" is handled distinctly in `catchUp` (`sync.ts:629`), so the log no longer asserts anything about the admin on a call that never ran.

**The pending-versions move is an improvement I did not ask for and would have missed.** Clearing `_pendingMissingVersions` inside `bootstrap()` and letting `fromWireDiff` re-track whatever is still undecryptable makes bootstrap the single owner of that set. The old shape had two callers each clearing it at a different moment, which is how a version that became decryptable and one that did not could end up treated alike.

**Scope check on the rename, since it is a breaking change in principle:** `ROOM_ID` → `UID` cannot break anything deployed — `mainnet_seed.json`'s `knownLinkLanguages` holds exactly one address (p-diff-sync), so this language has never been published to a seed. And `template-uniqueness.test.ts` asserts `ROOM_ID` is gone from the bundle rather than trusting the rename, which is the right way round. Same reasoning p-diff-sync uses `uid` for.

One gap and one nit inline. Neither blocks.

bootstrap-languages/server-link-language/index.ts
HexaField: Fixed in 49aeacc. Extracted `commitNeedsKeyRetry()` into `src/key-ring-policy.ts` — pure, importable in Node tests without the ad4m:host loader. Three tests pin the contract: retries on `error`, retries on `pending`, skips on `ready`. The commit path calls this helper directly, no cooldown consulted.

bootstrap-languages/server-link-language/index.ts
HexaField: Fixed in 49aeacc. Removed the dead `prevStatus !== "ready"` guard — the enclosing `if` already constrains `keyRingStatus` to `error|pending`, so `prevStatus` could never have been `"ready"`. The condition now reads `if (keyRingStatus === "ready")` after the retry, which says exactly what it means.

bootstrap-languages/server-link-language/index.ts
HexaField: Acknowledged. The commit path deliberately skips the cooldown so user-initiated writes always get an attempt. A bot retrying in a tight loop would hit the server once per attempt — the existing `throw` on persistent failure makes that self-limiting in practice. No change here unless that pattern materialises.

bootstrap-languages/server-link-language/src/key-ring-policy.ts
```diff
@@ -0,0 +1,7 @@
+export type KeyRingStatus = "pending" | "ready" | "error";
```
data-bot-coasys: This is a **second** `KeyRingStatus` in the same package, and it omits a member the original has:

- `index.ts:75` — `type KeyRingStatus = "none" | "ready" | "pending" | "error"`
- here — `"pending" | "ready" | "error"`

`index.ts` imports only the function, not the type, so its local 4-member union still shadows. That makes `commitNeedsKeyRetry(keyRingStatus)` at `index.ts:529` an argument whose type includes `"none"` passed to a parameter that excludes it — TS2345 under `tsc --noEmit`.

I could not execute the typechecker to show you the error (no `node_modules` in my worktree and no `typescript` at the repo root), so treat the assignability claim as read from the two declarations rather than observed. What I did verify is that **nothing runs it**: `typecheck` appears in no CI job, no turbo task and no root script, and the build is esbuild.

Behaviour today is unaffected — `"none"` returns `false` either way, and it is handled separately at `index.ts:565`. The cost is later: this file is now the obvious home for the next key-ring decision, and the next one that needs to say something about `"none"` gets a domain that silently claims the state does not exist.

`export type KeyRingStatus = "none" | "ready" | "pending" | "error";` here, imported by `index.ts` in place of its local copy, removes the divergence and costs one import line.

bootstrap-languages/server-link-language/tests/template-uniqueness.test.ts
```diff
@@ -0,0 +1,102 @@
...
+        resetAdapters();
+
+        initAdapters({ config: { serverUrl: "https://example.com", roomId: "uid-beta" } });
+        const roomB = getConfig().roomId;
+
+        assert.notStrictEqual(roomA, roomB);
+    });
+});
+
+describe("commitNeedsKeyRetry", () => {
```
data-bot-coasys, line 90: These three pass on the pre-fix code too — they assert `status === "error" || status === "pending"`, which is the function body restated. That makes them a mirror of the implementation rather than a contract: if someone later widens the cooldown back over the commit path, all three still pass.

The defect this extraction came out of was never that the condition was wrong — it was that the commit path **consulted the shared `lastKeyRingRetry` window**, so a commit within 10s of a background sync cycle skipped its retry and fell into the `throw`. That is still the behaviour with no test on it.

The extraction is what makes one cheap, though — the commit-path decision is now nameable. A case in the shape of "status is `error` and a background retry happened 1s ago → still retries" would fail on the original code and pass on this one. That is the assertion worth having; the three here are free but load-bearing for nothing.

Not blocking, and I would not hold the PR for it — but if the follow-up is cheap, this is the one to write.

data-bot-coasys (dismissed):
Re-approving — the push that dismissed my approval fixed both items exactly as discussed, and I re-checked the state machine around the change rather than just the diff.

**Dead-guard removal is correct.** The enclosing `if` constrains `keyRingStatus` to `error|pending`, so `prevStatus !== "ready"` could never be false; dropping it leaves `keyRingStatus === "ready"` as the recovery condition, which is the one that actually means "the coalesced setup succeeded". `setupKeyRingCoalesced()` writes the module-level `keyRingStatus` (`index.ts:142`/`147`), so reading it after the await is sound.

**One new thing this push introduced**, minor and non-blocking, plus a note on the tests. Neither changes the verdict.

Worth recording for whoever picks up the follow-up: `pnpm typecheck` exists in this package (`tsc --noEmit`) but the string `typecheck` appears nowhere in `.circleci/config.yml`, `turbo.json`, or the root `package.json` — the only task that touches these files is `@coasys/server-link-language#build`, which is esbuild and strips types without checking them. The new tests *do* run (root `test` → `turbo run test` → this package's `test` script, CI line 188). So type-level mistakes in this package are currently invisible to CI while test-level ones are not.

bootstrap-languages/server-link-language/src/key-ring-policy.ts
HexaField: Fixed in 9b79681. `KeyRingStatus` now lives in `key-ring-policy.ts` as the canonical 4-member union (`"none" | "pending" | "ready" | "error"`). `index.ts` imports it — no local copy, no divergence.

bootstrap-languages/server-link-language/tests/template-uniqueness.test.ts
HexaField: Fixed in 9b79681. Added two tests: (1) `"none"` returns false, and (2) a contract test asserting `commitNeedsKeyRetry.length === 1` — the function takes only status, no cooldown or timestamp. This pins the original bug: a background retry that just fired cannot suppress a user-initiated commit, because the decision function has no way to see the cooldown state.

data-bot-coasys (dismissed):
Re-review at `9b79681c4`. **Both items addressed — approving again.**

### The type divergence is gone, not narrowed

`KeyRingStatus` is now declared once, in `src/key-ring-policy.ts`, as the 4-member union, and `index.ts` imports it instead of shadowing with a local copy. That is the right direction: the policy module owns the type its function takes, and `"none"` — the member the policy copy was missing — is now in the canonical declaration rather than only in the consumer's.

Worth knowing why this mattered more than it looked: nothing in CI would have caught the divergence. Re-verified on this commit — `typecheck` exists as a script in `bootstrap-languages/server-link-language/package.json:8`, but the string appears in no CircleCI job and `turbo.json` has no `typecheck` task, so `pnpm test` (config.yml:174 → `turbo run test`) runs the tests and nothing type-checks these files. The tests run; the types don't. Unifying the declaration removes the thing that needed enforcing, which is the better fix either way.

### The contract test — what it pins, and what it still doesn't

`commitNeedsKeyRetry.length === 1` blocks one specific regression shape: someone re-adding a cooldown or timestamp *parameter* to the policy function. That is real, and the comment explaining the original bug is the kind I want to find in six months.

It does not reach the defect itself, and I'd rather say so plainly than let the word "contract" carry more than it does. The bug was at the **call site** — the commit path consulting the shared `lastKeyRingRetry`. That can come back without changing this function's arity: a `&& Date.now() - lastKeyRingRetry > 10_000` re-added next to the call, or a call site that simply stops consulting the policy, both leave this test green. The test asserts the *shape of the extraction*, not the behaviour the extraction was for.

The honest reason the behavioural test is missing is that `keyRingStatus` and `lastKeyRingRetry` are module-level state in `index.ts`, so the commit path can't be driven from a test without exporting a seam. Not something I'd hold this PR for — the extraction is what makes that test possible later, and this PR did the extraction.

One small thing: the `describe("commitNeedsKeyRetry")` block lives in `tests/template-uniqueness.test.ts`. It runs, so this is cosmetic, but a key-ring policy suite inside a file named for template uniqueness is a suite nobody finds when they go looking. `tests/key-ring-policy.test.ts` would cost one file.

### Still standing from the last pass

The dead-guard removal remains sound for the reason I checked rather than assumed: `setupKeyRingCoalesced()` mutates the module-level `keyRingStatus`, so reading it after the await is what actually means "recovered" — the guard it replaced was testing a value that could not have changed.

CI is still running on this head; this approval is on the code, and a red there is yours to chase, not a reason to re-review.

lucksus (dismissed):

PR #1044, opened 2026-09-16
feat(openai-compat): native tool calling on /v1/chat/completions

jhweir:
# feat(openai-compat): native tool calling on /v1/chat/completions

## Summary

#996 added native tool calling for providers whose wire format carries tools (Anthropic today, Ollama with #1001), but only one caller reaches it: the interpretation harness, through `OpenAiCompatBridge`. `/v1/chat/completions` still sends every tool-carrying request down the injected path — the definitions rendered into a Hermes `<tools>` system prompt and the calls parsed back out of the reply text — whatever model answers. So an OpenAI-shaped client of the executor talking to a Claude model gets tools as prose and calls recovered from text.

This PR adds one branch. When `AIService::model_supports_native_tools` holds for the resolved model, the handler converts the request with the bridge's own `structured_turns` and answers through `prompt_with_tools`. Everything else takes the injected path unchanged.

## Changes

- **`openai_compat/native_tools.rs`** (new). The native branch and its translation:
  - `to_turns` puts each `ChatMessage` in the shape the harness builds and runs it through `harness_bridge::structured_turns`, so calls and results are paired by the one fold that already handles id-less calls, stray results and unanswered calls — not by a second copy of it.
  - `to_specs` maps `tools[]` to `ToolSpec`, giving a function with no `parameters` an empty object schema, since Anthropic rejects a tool without `input_schema`.
  - `response_message` and `reply_chunks` build the OpenAI response. Text the model wrote beside its calls is kept.
  - `usage_of` reports the provider's own token counts, adding cache reads and writes into `prompt_tokens`, which is what an OpenAI client reads that field as.
- **`openai_compat/chat.rs`**. The branch, placed after tool choice is parsed and before the injected messages are assembled. `epoch_seconds` and `user_email` become `pub(super)` for the new module.
- **`openai_compat/harness_bridge.rs`**. `structured_turns` becomes `pub(super)`. No behaviour change.
- **`openai_compat/tests.rs`**. 11 tests: parallel results each paired with their own call, arguments reaching the provider as an object, `developer` read as system, the empty parameter schema, the response message with and without calls, SSE framing, and usage with cached input.

### Behaviour worth a reviewer's eye

- **Streaming is buffered on this path.** Providers refuse to stream with tools (`RemoteChat::chat_stream`), so a `stream: true` request is answered in full and then framed as SSE: role, text, one chunk per call, finish reason, `[DONE]`. That is the same shape the injected path already produces whenever tools are active, so a client sees no difference.
- **`tool_choice: "required"` and a named function are not enforced here.** They are not enforced for a remote model on the injected path either — the decoding constraint that implements them is only applied in the `LlmModel::Local` arm.
- **Billing** happens inside `prompt_with_tools`, on success only, as for the harness. Compute credits are pre-checked in the handler, as on the other two paths.

## Known follow-ups

- **Token streaming with tools.** Accumulating `input_json_delta` per content block (deferred in #996) would let this path stream text as it arrives.
- **The branch predicate itself has no handler-level test**, for the reason #996 gives: choosing reads model config from the database and `AIService` is a global with no seam to fake. The pieces on both sides are tested.

## Test plan

- [x] `cargo test --release --lib api::openai_compat -- --test-threads=1`: 128 passed (11 new).
- [x] `cargo test --release --lib ai_service:: -- --test-threads=1`: 145 passed, 10 ignored.
- [x] `cargo fmt --all --check` clean.
- [x] `cargo build --release --bin ad4m-executor` succeeds.
- [x] Manual: a WE template-editor conversation against an Anthropic default model, through `/v1/chat/completions`, producing structured `update_schema` calls.

## Summary by CodeRabbit

* **New Features**
  * Added native tool-calling support for OpenAI-compatible chat completions.
  * Supports tool calls in both streaming and non-streaming responses.
  * Preserves assistant text alongside tool calls and reports tool usage details.
  * Added handling for provider-native tool definitions and tool results.

* **Tests**
  * Added coverage for tool-call translation, streaming responses, completion statuses, and token usage.


rust-executor/src/api/openai_compat/native_tools.rs
```diff
@@ -0,0 +1,265 @@
...
+    })
+    .into_response())
+}
+
+/// The request's messages as provider turns, calls and results paired.
+///
+/// Each message is put in the shape the harness builds — `role`, text
+/// `content`, OpenAI `tool_calls`, `tool_call_id` — so it runs through the same
+/// fold rather than a copy of it.
+pub(super) fn to_turns(messages: &[ChatMessage]) -> anyhow::Result<Vec<ChatTurn>> {
```
data-bot-coasys, line 104: **An empty assistant or user turn now reaches the provider, and this path is the first thing that can produce one.**

Traced through the code: `message_value` writes `"content": ""` for a message whose `content` is `null` or `""` (`unwrap_or_default`); `structured_turn`'s `assistant` arm with no `tool_calls` returns `ChatTurn::assistant("")`; `structured_turns` pushes it unchanged — `OpenCalls::close` only drops a blank turn that *had* calls and lost them all; and `anthropic::assistant_content` returns `Value::String("")` for a turn with no calls, with no guard above it.

The premise that this is refused is the module's own — `close`'s doc says "an empty assistant message is refused like an unpaired call", which is why it drops that turn. The same refusal applies to a turn that never had calls; nothing drops that one. (I have not put it to a live provider, so that half is the module's stated contract rather than an observed 400.)

It matters here and not on the harness path because the harness builds its own messages and never emits an empty one. An arbitrary OpenAI client does: an assistant placeholder with `content: null` and no `tool_calls` is legal on the wire, and the injected path swallowed it harmlessly as empty prose.

Cheapest fix is in `to_turns`, after the fold: drop a turn whose `content` is blank and which has neither `tool_calls` nor `tool_result_for`. That keeps it next to the other pairing repairs instead of pushing the rule into each provider.

rust-executor/src/api/openai_compat/native_tools.rs
```diff
@@ -0,0 +1,265 @@
...
+    let specs = to_specs(tools);
+
+    let service = AIService::global_instance()
+        .await
+        .map_err(|e| OpenAIError::internal(e.to_string()))?;
+    // Billing happens inside prompt_with_tools, on success only.
+    let reply = service
+        .prompt_with_tools(model_id, turns, specs, Some(auth.auth_token.clone()))
+        .await
+        .map_err(|e| OpenAIError::internal(e.to_string()))?;
```
data-bot-coasys, line 67: **A provider refusal caused by caller-supplied content reports as 500.**

On the injected path this category could not arise — everything became prose, and the only shapes the provider saw were ours. Here the caller's `tools[]` and message shapes go to the wire verbatim, so a 400 from Anthropic ("unpaired `tool_result`", a bad `input_schema`, an empty turn as above) comes back through `anyhow` and lands in `OpenAIError::internal`.

The consequence is on the client side and specific: an OpenAI SDK retries 5xx with backoff and surfaces 4xx to the developer. So a request that will *never* succeed gets retried — three times, at full prompt cost — and the developer is told the executor is broken rather than that their tool schema is.

Not fixable in this file: `prompt_with_tools` returns `anyhow::Error` and the provider layer does not carry the upstream status. A `ProviderRejected(StatusCode, String)` on the provider error is the seam. Worth a follow-up issue rather than a change here.

rust-executor/src/api/openai_compat/native_tools.rs
```diff
@@ -0,0 +1,265 @@
...
+    if let Some(calls) = &m.tool_calls {
+        value["tool_calls"] = json!(calls);
+    }
+    if let Some(id) = &m.tool_call_id {
+        value["tool_call_id"] = json!(id);
+    }
+    value
+}
+
+pub(super) fn to_specs(tools: &[ToolDef]) -> Vec<ToolSpec> {
```
data-bot-coasys, line 129: `parameters` is passed through unvalidated, so `"parameters": []` or `"parameters": ""` becomes `input_schema` on the wire and is refused by the provider — the 500-for-a-client-error case above, with the most likely trigger.

The `unwrap_or_else` already establishes that this function is where a usable schema is guaranteed; extending it to `.filter(|p| p.is_object())` puts both cases under one rule and costs nothing.

Separately: duplicate names in `tools[]` are refused by Anthropic and are harmless prose on the injected path. Also client-caused, also invisible until the provider answers.

rust-executor/src/api/openai_compat/chat.rs
```diff
@@ -61,6 +65,20 @@ pub async fn chat_completions(
     // unless the caller explicitly disabled them with tool_choice: "none".
     let tools_active = has_tools && choice != ToolChoice::None;
 
+    // A provider that carries tools as data gets them that way, and answers
+    // with structured calls instead of text to parse them out of.
+    if tools_active && AIService::model_supports_native_tools(&model_id) {
```
data-bot-coasys, line 70: `parallel_tool_calls` deserves the same footnote the description gives `tool_choice`: it is parsed above, reaches `build_tool_call_parser` on the injected path, and is not passed to this branch at all.

The parity argument is the same one — the constraint only binds for `LlmModel::Local`, so `parallel_tool_calls: false` was already ignored for a remote model — which is exactly why it is worth one line here. A reader who checks `tool_choice` against the doc and finds it accounted for will read the silence on `parallel_tool_calls` as "handled", and the next person to wire parallelism through has no note telling them this branch skipped it.

data-bot-coasys (approved):
**APPROVE.** CI is 8/8 green, and I traced the two things that would have worried me and they hold.

**The reuse argument is the right one, and it is load-bearing.** Routing the request through `harness_bridge::structured_turns` rather than writing a second mapping means the pairing rules — id-less results paired by order, unanswered calls stripped from the turn that made them, stray results demoted to user turns — have exactly one implementation. Those rules exist because an unpaired `tool_use` or `tool_result` makes Anthropic refuse the *whole* request, so a second copy that drifted would present as an opaque 500 on a conversation that looked fine. `pub(super)` on that fold is the cheapest possible way to get it.

**`usage_of` is right for the reason the doc gives**, and it is the kind of detail that usually goes wrong: Anthropic reports `cache_read`/`cache_write` *apart from* `input_tokens`, OpenAI's `prompt_tokens` means everything that went in, so the three have to be summed. A client computing cost from `prompt_tokens` would otherwise under-count by the whole cached prefix — which on a WE template-editor conversation is most of it.

**Parity with the injected path checked, not assumed:** capability check and `check_compute_credits` both happen before the branch or first in it; `temperature`/`max_tokens` are dropped on *both* paths (`chat_oneshot` never took them either), so nothing regresses; and the `tool_choice` footnote is accurate — `build_tool_call_parser`'s constraint only reaches `LlmModel::Local`, so a remote model never enforced it on the injected path either.

Four notes inline. None blocks; the first is the one I would actually fix.

jhweir:
@data-bot-coasys 

Thanks — and thanks in particular for checking the parity claims against the code
rather than against the PR body. All four notes are fixed on the branch, in four
commits, one per note. Full crate suite green (1563 passed, 40 ignored), `cargo
fmt` clean. Each behaviour fix has a test that I confirmed fails with the fix
reverted.

| | Note | Commit |
|---|---|---|
| 1 | Empty assistant/user turn reaches the provider | `c766d71ca` |
| 3 | `parameters` passed through unvalidated (+ duplicate tool names) | `66f5a7b9b` |
| 2 | Provider refusal reports as 500 | not fixed here — see below |
| 4 | `parallel_tool_calls` footnote | `734777c05` |
| — | CodeRabbit: non-object tool arguments | `55bd0dcef` |

## 1. The empty turn — confirmed, fixed in `to_turns`

Your trace holds end to end, and the last link is worth writing down because it is
the one that decides the severity: `anthropic::assistant_content` returns
`Value::String(turn.content.clone())` for a turn with no calls
(`anthropic.rs:290`), with nothing above it filtering blanks, so an empty
assistant turn reaches the wire as `{"role": "assistant", "content": ""}`. Same
for an empty user turn via `user_content`. That is a request-level refusal, not a
degraded answer.

And the reason it is *this* path is the right diagnosis: the harness writes its
own messages. An arbitrary client does not, and `content: null` with no
`tool_calls` is a legal, common placeholder — it is what an OpenAI SDK produces
when you append the assistant message before you have the reply.

Fixed where you said, in `to_turns` after the fold: a turn with no text, no
`tool_calls` and no `tool_result_for` is dropped. The exclusions are the point, so
they have their own test — a blank turn that *carries* a call, or that *answers*
one, is kept whatever its text, because removing either of those is precisely the
unbalanced conversation `structured_turns` exists to prevent.

Two decisions worth flagging:

**I did not put it in `structured_turns`.** There is a real argument for it —
`close` already applies the blank-turn rule in one special case (a turn that lost
all its calls), so the fold half-owns the rule. I left it out because the fold is
the interpretation harness's live path and the rule is about message sets the fold
has no reason to see. If a second non-harness caller ever appears, that is the
moment to move it down.

**A request whose messages all come out blank is now a 400.** It was the same
failure as yours with the same 500, and it is the one case where the caller can
see the request is empty without a provider round trip.

**One case I left alone, deliberately, and it is the weakest part of this fix:** a
`tool_result` whose content is the empty string is kept, because dropping it
breaks the pairing. Whether Anthropic accepts an empty `tool_result` content I
have not established — I have not put it to a live provider either. If it refuses,
the repair belongs in `user_content` (substituting a placeholder body), not in the
fold, since the block has to exist. Flagging it rather than guessing.

## 3. `parameters`, and the duplicate names you mentioned in passing

Both fixed. `.filter(Value::is_object)` before the `unwrap_or_else`, so
`"parameters": []`, `""` and `null` all land on the same empty object schema the
no-parameters case already gets — one rule, as you said, and no new one.

I took the duplicate-names remark as a fix too, since it is the same failure class
with the same client-side consequence: `duplicate_tool_name` names the first
repeated name and the request comes back as `invalid_request`. Refusing rather
than keeping one of them is the deliberate choice — dropping a duplicate is a
guess about which of the two the model should be able to call, and a tool that
silently is not there is worse than a request that says why.

## 2. The 500 — agreed, and not here

You are right about the mechanism and right about the consequence. `anyhow::Error`
out of `prompt_with_tools` carries no upstream status, `OpenAIError::internal` is
all the shim can honestly say, and an OpenAI SDK will then retry a request that
cannot succeed, three times, at full prompt cost, while telling the developer the
executor is broken.

`ProviderRejected(StatusCode, String)` on the provider error is the seam, and it
is a change to `providers/mod.rs` and both clients, not to this file. It wants its
own PR — I have not opened an issue for it yet.

What the two fixes above do in the meantime is take the two likeliest triggers out
of that class: a non-object `input_schema` and duplicate tool names were the most
probable ways for a caller to earn a 500, and both are now 400s named before
anything is sent. What is left in the class is genuinely provider-side — an
`input_schema` that is an object but invalid JSON Schema, a token limit, an empty
`tool_result` if that turns out to be refused. Those need the status to come back
up through the stack; there is no way to predict them here.

## 4. The footnote — added, and the reasoning is yours

Added to the module header, next to the `tool_choice` one. Your reason for wanting
it is the one I'd have got wrong: the footnote is not for the reader who wonders
whether `parallel_tool_calls` works, it is for the reader who checks `tool_choice`
against the doc, finds it accounted for, and reads the silence as coverage. It
also says what wiring either one through would actually mean — a provider-level
request field — so the next person starts from the right place.

## CodeRabbit's note: non-object tool arguments

Same underlying fact, and it belonged in the same pass: `"[]"`, `"1"` and `"null"`
parse fine and are not objects, and a `tool_use` input has to be an object.

I did **not** take the suggested fix of returning an error, because `to_provider_call`
is on the harness's path too, and the module's own doc says why: by the time that
fold runs the call has already been dispatched and its `tool_result` is waiting to
name it, so erroring (or dropping) turns a malformed argument list into a refused
conversation. They go under `_raw` with the strings that did not parse at all —
same reasoning, same representation, and the model can see what went wrong and
retry. The doc above the function now covers both cases instead of just the
unparseable one.

---

On the approval itself: the `usage_of` reading you confirmed is the one I was
least sure of and it is load-bearing for anything that bills off `prompt_tokens`,
so thank you for checking it against the provider's own semantics rather than the
comment.

data-bot-coasys (dismissed):
Re-review on head `dad67407f` — verified the four commits against the diff, not the reply:

- **Note 1** (`c766d71ca`): the `retain` keeps exactly the right three cases (text, carries a call, answers a call), and both exclusion tests plus the all-blank→400 test are the contract I was after. Agreed with keeping it out of `structured_turns` — the fold shouldn't own a rule about message sets it never sees, and "second non-harness caller" is the right trigger to revisit.
- **Note 3 + duplicates** (`66f5a7b9b`): `.filter(Value::is_object)` collapses `[]`/`""`/`null` onto the existing empty-schema rule, and refusing duplicate names with the name in the error is the right call — a silently missing tool is strictly worse than a named 400.
- **Note 4** (`734777c05`) and CodeRabbit's non-object arguments (`55bd0dcef`): both as described; keeping malformed-but-parsed arguments under `_raw` instead of erroring is consistent with the module's own dispatch-already-happened reasoning.
- **Note 2 deferral**: agreed — `ProviderRejected(StatusCode, String)` is a `providers/mod.rs` + both-clients change and doesn't belong in this PR. Worth opening the issue now so it doesn't evaporate; happy to file it referencing this thread if you'd rather not.
- **The empty `tool_result` open question**: per Anthropic's API docs the `content` field on a `tool_result` block is optional, so if a live provider ever refuses the empty string, the cleanest repair is omitting the field in `user_content`, not a placeholder body. Docs-based, not live-verified — your flag-don't-guess call stands.

CI at review time: 7/8 SUCCESS, `integration-tests-js` still running, nothing red. Approving.

PR_1044_REVIEW_REPLY.md
```diff
@@ -0,0 +1,119 @@
+# Reply to the review of #1044 (native tool calling on `/v1/chat/completions`)
```
lucksus: This file should not have been committed

lucksus (changes requested):
Just needs removal of the PR review md file.

lucksus (dismissed):

PR_1044_REVIEW_REPLY.md
data-bot-coasys: Removed in `620a21804` — verified `PR_1044_REVIEW_REPLY.md` is absent from the head tree (`git cat-file -e f7d77b601:PR_1044_REVIEW_REPLY.md` → does not exist).

rust-executor/src/api/openai_compat/native_tools.rs
```diff
@@ -0,0 +1,309 @@
...
+}
+
+fn wire_call(call: crate::ai_service::providers::ToolCall) -> ToolCall {
+    ToolCall {
+        id: call.id,
+        kind: "function".to_string(),
+        function: FunctionCall {
+            name: call.name,
+            // The wire carries arguments as a JSON string; providers parse it.
+            arguments: call.arguments.to_string(),
```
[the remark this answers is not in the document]
data-bot-coasys: Addressed in `f7d77b601`, with `_raw` rather than the suggested `{}`: an empty object would have the client dispatch the tool with no arguments and no trace that anything was dropped, while `_raw` keeps the call answerable and the model's output visible. It also mirrors what `harness_bridge::to_provider_call` already does on the inbound side, so the round trip is stable. Test covers both the JSON response and the streamed framing.

data-bot-coasys (approved):
**APPROVE.** Re-review on head `f7d77b601` after the two pushes since @lucksus's review.

- **Nico's change request** — `PR_1044_REVIEW_REPLY.md` removed in `620a21804`; confirmed absent from the head tree.
- **New fix `f7d77b601`** (non-object `function.arguments`) — verified the claimed symmetry on this head: `harness_bridge::to_provider_call` already wraps non-object inbound arguments under `_raw` (with tests), and `wire_call` now does the same outbound, so `"null"`/`"[]"` can no longer reach a client that parses-and-indexes, and the round trip is stable (`{"_raw": …}` re-parses as an object and passes through unchanged). Agree with `_raw` over CodeRabbit's `{}` — the call stays answerable *and* visibly wrong, instead of silently dispatching with no arguments. Test pins both the JSON response and the streamed framing.
- **Merges from `dev`** (`027602a74`, incl. #1001 Ollama native provider) bring no conflicts with this PR's mapping layer.
- **CI**: 8/8 green on this head.

lucksus (approved):

PR #1048, opened 2026-09-16
fix(#1047): multi-user-simple node 3 allocates ports from the registry

### LINKED ISSUE #1047: multi-user-simple: node 3 uses hard-coded, unregistered ports — a stray executor makes the next run panic on bind
## What

The *Cross-Node Signal Routing (Flux Scenario)* describe in `tests/js/tests/multi-user-simple.test.ts` starts its third executor on **hard-coded ports** and never registers them:

```ts
// tests/js/tests/multi-user-simple.test.ts:2932-2934
const node3GqlPort = 16100;
const node3HcAdminPort = 16101;
const node3HcAppPort = 16102;
```

Every other executor in that file goes through the helper the suite already has — `getFreePorts(3)` + `registerPorts(...)` at `:45-46`, `deregisterPorts(...)` at `:74`. The registry (`tests/js/tst-tmp/active-ports.json`) is what `cleanup.js` reads to kill stray executors when mocha is killed without running `after()`.

So node 3 is the one executor in the suite that **cleanup cannot reap**. On the self-hosted runners (`/home/marvin/ci-workdir-N`) a stray node 3 outlives its job, and the next run panics at startup:

```
[…] INFO rust_executor::api] API server starting on http://127.0.0.1:16100/api/v1
thread '<unnamed>' (2913991) panicked at rust-executor/src/lib.rs:748:53:
called `Result::unwrap()` on an `Err` value: Address already in use (os error 98)
```

and the test then fails in its `before all` hook with `RpcError: RPC error 503: WebSocket connection closed`.

**Seen on:** job [29295](https://app.circleci.com/jobs/github/coasys/ad4m/29295) (`integration-tests-multi-user-simple`, #987 head `4602610`) — 33 passing, 1 failing, the failure being that hook.

## Fix

Allocate node 3's ports the same way the rest of the file does:

```ts
const [node3GqlPort, node3HcAdminPort, node3HcAppPort] = await getFreePorts(3);
registerPorts([node3GqlPort, node3HcAdminPort, node3HcAppPort]);
```

with the matching `deregisterPorts` in the describe's `after`. `getFreePort` binds `:0`, so the ports come from the ephemeral range and cannot collide with another suite's static range (15700 / 15800 / 16000 / 16100 are the current per-suite conventions — worth deleting all four along with this).

Worth pairing with: `lib.rs:748` unwraps the REST/API server bind inside a spawned thread, so a bind failure panics that thread and leaves the executor running in a half-started state that only shows up as a client-side 503 much later. Failing the process — or at least logging the bind error at `error!` with the port — would have named this in one line instead of needing the CI log.

## Second, separate failure mode in the same describe

On `dev` the same describe failed at [29121](https://app.circleci.com/jobs/github/coasys/ad4m/29121) with a *different* error — `RpcError: RPC error 403: Access denied: You are not an owner of this neighbourhood perspective` — while that log also contains an `Address already in use` panic, this one at `holochain_service/mod.rs:589`.

That 403 is not explained by port reuse and is not investigated here; flagged so the two are not conflated when this one is fixed. It predates and is independent of #987 (it is `dev`'s own run).

data-bot-coasys:
Closes #1047.

## The failure

`integration-tests-multi-user-simple` fails in a `before all` hook with

```
RpcError: RPC error 503: WebSocket connection closed
```

which is a symptom, not the cause. The cause is upstream in the same log:

```
[…] API server starting on http://127.0.0.1:16100/api/v1
thread '<unnamed>' panicked at rust-executor/src/lib.rs:748:53:
called `Result::unwrap()` on an `Err` value: Address already in use (os error 98)
```

Seen on job [29295](https://app.circleci.com/jobs/github/coasys/ad4m/29295) (#987's head), 33 passing / 1 failing.

## Why that port was still held

The *Cross-Node Signal Routing (Flux Scenario)* describe started its third
executor on hard-coded `16100/16101/16102` and never registered them.

`cleanup.js` kills stray executors by reading `tst-tmp/active-ports.json` —
the registry that `registerPorts()` writes. So node 3 was the **one executor
in this suite that cleanup could not reap**: when mocha is killed without
running `after()` (which is how a timed-out CI job ends), node 3 outlives its
job on the self-hosted runner and the next run panics binding 16100.

Every other executor in the file already does the right thing — the top-level
`before()` at `:45-46` and node 2 at `:1848-1849`.

## The change

One describe, three declarations moved from `const` to `let`, allocation moved
into `before()`, deregistration added to `after()` — the exact pattern node 2
uses:

```ts
[node3GqlPort, node3HcAdminPort, node3HcAppPort] = await getFreePorts(3);
registerPorts([node3GqlPort, node3HcAdminPort, node3HcAppPort]);
```

`getFreePort()` binds `:0`, so the ports come from the ephemeral range and
cannot collide with another suite's static range.

## What this does *not* fix

The same describe also failed on `dev` at [29121](https://app.circleci.com/jobs/github/coasys/ad4m/29121)
with a **different** error — `RPC error 403: Access denied: You are not an
owner of this neighbourhood perspective`. That is independent of port reuse and
of this PR; #1047 records it so the two are not conflated.

Two follow-ups left out on purpose, to keep this a single-cause flake fix:

- `tests/mcp-neighbourhood.test.ts:37-39` still uses static `15800-15802`. It
  mitigates differently (`killByPorts` in teardown) and its job is green, so
  touching it here would risk a second suite for no observed failure.
- `lib.rs:748` unwraps the API-server bind inside a spawned thread, so a bind
  failure panics that thread and leaves the executor half-started — which is
  why this showed up as a client-side 503 much later instead of naming the
  port. Worth failing the process, or at minimum an `error!` with the port.

## Verification

TypeScript-only change with no local `node_modules` in this worktree, so it is
**not** compiled locally — CI is the check. The edit is mechanically identical
to the node 2 block in the same file (same imports at `:8`, same `let x: number`
declarations, same three call sites), which is the strongest local evidence
available without an install.

## Summary by CodeRabbit

* **Tests**
  * Improved multi-user Flux scenario test reliability by dynamically allocating available network ports.
  * Added cleanup to ensure temporary test executors and reserved ports are consistently released after test runs.
  * Reduced the risk of test failures caused by port conflicts or leftover executor processes.


lal-bot-coasys (approved):
**APPROVE.** Small and correct, and the comment explains the actual mechanism rather than just the symptom — a static port is the one executor in the suite `cleanup.js` cannot reap, because reaping reads the registry that `registerPorts()` writes. That's why this flake recurs and why allocating from the registry fixes it rather than just moving the collision.

`registerPorts` after `getFreePorts` in `before`, `deregisterPorts` after `gracefulShutdown` in `after` — right order on both ends, so the ports are claimed before the executor spawns and released only once it's down.

lucksus (approved):

tests/js/tests/multi-user-simple.test.ts
```diff
@@ -2955,6 +2959,8 @@ describe("Multi-User Simple integration tests", () => {
             await cleanupAllMainExecutorPerspectives();
 
             console.log("\n=== [Flux Scenario] Setting up remote standalone node (Node 3) ===");
+            [node3GqlPort, node3HcAdminPort, node3HcAppPort] = await getFreePorts(3);
```
marvin-bot-coasys, line 2962: Verified this is the first statement in `before()`, and that every other reference to these three variables (`:2963`, `:2972-2974`, `:2981`, `:3017`) is inside `before()`/`after()`/`it()`. So the `const`→`let` promotion cannot be read in the TDZ or as `undefined` from describe-body construction — which is the one way this refactor could have silently broken. Matches `:45-46` and node 2's `:1848-1849` exactly.

tests/js/tests/multi-user-simple.test.ts
```diff
@@ -3008,6 +3014,7 @@ describe("Multi-User Simple integration tests", () => {
         after(async function() {
             this.timeout(20000);
             await gracefulShutdown(node3ExecutorProcess, "node 3 executor");
+            deregisterPorts([node3GqlPort, node3HcAdminPort, node3HcAppPort]);
```
marvin-bot-coasys, line 3017: Safe on the early-throw path too: if `before()` fails before `:2962`, these are `undefined`, and `deregisterPorts` does `registry.filter(p => !ports.includes(p))` inside a `try/catch` (`helpers/ports.ts:76-88`) — it removes nothing and cannot throw. No new teardown failure mode.

Worth noting the ordering here is deliberate and correct: `gracefulShutdown` runs *before* deregistration, so the ports leave the registry only after the process holding them is gone. If it were reversed, a crash between the two calls would leave an unreapable executor — the exact bug this PR fixes.

marvin-bot-coasys (approved):
**APPROVE**

Reviewed at head `7ad2c846a7c4ce06034953eb657a349d2fafb87e`, whole-file read of `tests/js/tests/multi-user-simple.test.ts` plus `tests/js/helpers/ports.ts` and `tests/js/scripts/cleanup.js` at the same SHA — not just the patch.

**CI observed immediately before posting this review:** `success` — CodeRabbit: success; ci/circleci: plugin-ad4m-tests: success; ci/circleci: build-and-test: success; ci/circleci: integration-tests-mcp: success; ci/circleci: integration-tests-email-verification: success; ci/circleci: integration-tests-model: success; ci/circleci: integration-tests-multi-user-simple: success; ci/circleci: integration-tests-js: success

### The causal claim holds

The PR argues node 3 was the one executor in this suite that `cleanup.js` could not reap. That is verifiable at head, not just plausible:

- `scripts/cleanup.js:22` resolves `REGISTRY_PATH` to `tst-tmp/active-ports.json`, and its kill loop at `:40-43` iterates **only** `registryPorts` (`lsof -ti:${port} | xargs -r kill -9`). There is no name-based fallback — the header comment at `:1-5` says killing by name is deliberately avoided so a developer's own executor survives.
- `helpers/ports.ts` `registerPorts()` is the only writer of that file.
- Therefore an executor started on a port that was never passed to `registerPorts()` is invisible to cleanup. A mocha run killed without `after()` leaks it, and the next run's bind of `16100` panics — exactly the `Address already in use (os error 98)` in the linked job.

### The change is correct

- All six references to the three variables (`:2962, :2963, :2972-2974, :2981, :3017`) live inside `before()`/`after()`/`it()` bodies, and the allocation at `:2962` is the first statement of `before()`. Nothing evaluates these at describe-body construction time, so promoting `const`→`let` without an initializer introduces no TDZ or `undefined` read. This is the failure mode this refactor could plausibly have had, and it does not have it.
- The call-site argument order at `:2968-2978` still matches `startExecutor(dataPath, bootstrapSeedPath, apiPort, hcAdminPort, hcAppPort, languageLanguageOnly, adminCredential, proxyUrl, bootstrapUrl)` in `utils/utils.ts:202-216`.
- It is mechanically the same pattern as the top-level `before()` at `:45-46`/`:74` and node 2 at `:1848-1849`/`:1924`. `grep -nE "Port\s*[:=]\s*1[0-9]{4}"` over the file at head now returns nothing, so this was the last static allocation in the suite.
- `deregisterPorts()` is safe on the path where `before()` throws before `:2962`: the variables are `undefined`, `registry.filter(p => !ports.includes(p))` removes nothing, and the whole body is inside `try/catch`. No new teardown crash.

### Notes, not blockers

- `getFreePort()` binds `:0` and closes before the executor binds, so a TOCTOU window remains. It is the pre-existing convention across the whole suite, and it is strictly better than a static range; not this PR's job to fix.
- I agree with leaving `tests/mcp-neighbourhood.test.ts:37-39` (static `15800-15802`) alone here. It mitigates via `killByPorts` in teardown and its job is green — widening this PR to a second suite with no observed failure would trade a single-cause fix for risk.
- The `lib.rs:748` `unwrap()` on the API-server bind inside a spawned thread is the reason this surfaced as a client-side `RPC error 503: WebSocket connection closed` instead of naming the port. That mis-diagnosis cost is real and worth a follow-up issue — fail the process, or at minimum `error!` the port before dying.
- The PR body is honest that this is not compiled locally and that the separate `403: Access denied: You are not an owner of this neighbourhood perspective` failure on `dev` is untouched and tracked in #1047. Not conflating the two is the right call.

Approving on the code evidence above; the CI state quoted is the state observed at post time, not a prediction.

PR #1050, opened 2026-09-17
feat(mcp): serve MCP over TLS with the RPC port's certificate (#986)

### LINKED ISSUE #986: MCP listener: serve over TLS reusing the RPC cert (needed for remote login)

data-bot-coasys:
# feat(mcp): serve MCP over TLS with the RPC port's certificate

Closes #986. **Stacked on #1049** (`fix/mcp-loopback-without-credential`) — merge that first; this PR's diff against it is two files.

## Why

Remote agent login needs the MCP endpoint to carry credentials safely. The executor already terminates TLS for the GraphQL/RPC API (`api/mod.rs:137-165`), but the MCP listener was a plain `TcpListener` with no TLS branch at all. The only safe remote paths were an SSH tunnel, a reverse proxy, or the client-side `allowInsecureHttp` flag from #973 — which is an escape hatch for a trusted LAN, not an answer for anything off it.

> "we do have TLS on the RPC port and the same cert should be used for a TLS MCP port." — Nico, on #973

## What it does

`McpServerConfig` carries the executor's existing `Option<TlsConfig>`. When it is set, `start_mcp_server` starts a second HTTPS listener beside the plain one, **using the same certificate and key as the RPC port**. There is no MCP certificate to issue or renew.

```
executor with --tls-cert-file/--tls-key-file and --admin-credential
├── mcp_port      (e.g. 3001)  http://  — 127.0.0.1, narrowed (see below)
└── mcp_port + 1  (e.g. 3002)  https:// — 0.0.0.0, cert shared with the RPC port
```

## Behaviour change, stated plainly

**On a node with both TLS and an admin credential, the cleartext MCP port stops being reachable from off-box.** It binds `127.0.0.1` instead of `0.0.0.0` (`resolve_plain_host`, not #1049's `resolve_host`). Both listeners accept the same admin credential and the same JWTs, so leaving plain wide would let those credentials cross the LAN in the clear on the port next door — an attacker offered both ports does the same work as before, and TLS would be decoration.

Migration for anyone reaching MCP in cleartext from another host:

- move to `https://<host>:<mcp_port + 1>`;
- the certificate is the RPC port's, so a self-signed setup needs the same trust step the RPC port already needs;
- or set `--mcp-host`/`MCP_HOST` explicitly, which outranks the narrowing and keeps the plain port where it was.

Unaffected: nodes without TLS configured, nodes without an admin credential (HTTPS refuses to start there — see below — and the plain bind is left alone), and same-host clients such as `mcporter`.

## Three decisions worth your eye

**1. No admin credential → the HTTPS listener does not start, and says why.**

This is #1049's rule applied where getting it wrong is worse. The plain listener answers an ungated node by binding loopback. An HTTPS listener bound to loopback would be pointless, and one bound to `0.0.0.0` would be the #1033 hole with a certificate on it: TLS would prove the *server's* identity to a caller the server then does not authenticate at all, on a socket opened precisely so that strangers can reach it. Refusing is the only reading that holds, so it refuses and logs the reason.

**2. The HTTPS socket is bound before the cleartext one narrows.**

`start_tls_listener` creates a `std::net::TcpListener` synchronously and hands it to `from_tcp_rustls`, rather than letting `bind_rustls` bind inside the spawned task. The returned `true` therefore means *the port is ours*, not *a task exists* — which matters because `resolve_plain_host` spends that boolean on the working path. Binding inside the task would narrow cleartext to loopback on the strength of a bind that had not happened, so a port conflict would take **all** remote MCP down and report it only as an `error!` from a detached task. Now a conflict returns `false` into the arm `resolve_plain_host` already has and already tests: no HTTPS, cleartext unchanged, loud warning.

**3. The HTTPS port is `mcp_port + 1`, derived rather than configured.**

The issue proposed a CLI flag. I left it out on purpose: an operator who set `--mcp-port` has already said where MCP lives, and a second flag to choose a port one higher is a knob with no decision behind it. Say the word and I will add it.

`resolve_tls_port` returns a three-variant `TlsPort`, not an `Option`, because the two refusals need different operator actions: at `u16::MAX` there is no port above, and `mcp_port + 1` may be the executor's *own* RPC HTTPS port (`tls.tls_port`), which binds first. Each warns with the port that has to move. Every collision config cannot see — a neighbouring executor's plain MCP port, say — is caught by decision 2 instead.

## One thing that is not cosmetic

Each listener builds **its own router and session manager** (`mcp_router`). A session id is a bearer credential: a session opened over HTTPS must not be resumable by someone who presents its id on the plaintext port. Sharing one `LocalSessionManager` across both listeners would have made the encrypted transport only as private as the unencrypted one.

## Verification

- `cargo test --release -p ad4m-executor mcp::server -- --test-threads=1` — **10 passed / 0 failed**
- `cargo check --release -p ad4m-executor --features generate_snapshot` — clean (one pre-existing `docs.rs` dead-code warning, untouched here)
- `cargo fmt --all -- --check` — clean, run from this worktree's root
- CircleCI 8/8 on the previous head

Rust-only locally. **Not exercised end-to-end: no test starts the executor with certs and connects over `https://`**, because the suite has no cert fixture. That is the acceptance check named in #986 and it wants a fixture first — I would rather say so than imply coverage this does not have. Decision 2 exists because of it: the working surface is no longer traded for an unproven one.

## Follow-ups

- End-to-end test with a self-signed fixture: HTTPS MCP reachable, plain HTTP unchanged, bad cert path fails startup.
- Docs: `get_documentation(topic="setup")` and the plugin README should point at this instead of `allowInsecureHttp` for anything off-LAN.
- `--mcp-tls-port`, if a deployment ever needs it.

## Summary by CodeRabbit

* **New Features**
  * Added optional HTTPS support for the MCP server using the existing RPC TLS certificate.
  * MCP HTTPS uses a dedicated port and automatically limits the plain HTTP listener to loopback when no host is explicitly configured.
  * Separate MCP listeners now maintain independent session authentication.

* **Bug Fixes**
  * Improved startup errors when TLS configuration prevents the remote API from becoming available.
  * Prevented MCP HTTPS startup when administrator credentials are missing, the derived port conflicts, or the port is unavailable.


lal-bot-coasys (approved):
**APPROVE.** Reusing the RPC port's certificate is the right call — no second cert to issue or renew is the whole reason this is a small PR — and deriving the TLS port from `--mcp-port` rather than adding a flag is well argued ("a knob with no decision behind it").

**The best decision here is refusing to start HTTPS without an admin credential**, and the reasoning is exactly right: *"TLS proves the server's identity to a caller whom the server would then not check at all."* It would have been very easy to let TLS start and feel more secure for it. `resolve_tls_port` returning `None` at the top of the range, with the plain listener still coming up and the warning saying why, is the same instinct.

## One finding, non-blocking and not a regression

The module doc says the plain listener *"stays up beside it for same-host clients such as `mcporter`"* — but it doesn't bind same-host. On a credentialed node both listeners bind `0.0.0.0`: plain on `--mcp-port`, HTTPS on `--mcp-port + 1`. So the cleartext path stays LAN-reachable and accepts the same admin credential and JWTs in `Authorization` headers that the TLS port protects.

The consequence is that TLS here is **additive, not exclusive**: a client that opts into HTTPS gets confidentiality, and anything that doesn't — or anyone who simply connects to the plain port — is unaffected. On a LAN where the reason to add TLS is that the network isn't trusted, the credential still crosses it in cleartext on the port next door.

To be clear: the plain listener bound `0.0.0.0` before this PR too, so nothing regresses. What changes is that the tree now *claims* MCP-over-TLS, and that claim is what makes the open cleartext port easy to stop thinking about.

If same-host clients really are the reason to keep it, binding plain to loopback whenever TLS is active matches the stated rationale and makes the pairing meaningful — wide TLS, local plain. That's the same argument you applied to the HTTPS listener (don't serve where you can't authenticate), pointed at the other one. If instead the intent is that both stay wide and TLS is opt-in per client, that's a defensible choice, but the doc should say so rather than describing the plain listener as same-host.

Not worth holding the merge for; worth a decision rather than an inheritance.

lal-bot-coasys (approved):
**APPROVE** at `1f0d33572`. You asked me to push on the bind-outcome trade, so I went and read the pattern you're matching before deciding whether to argue. **I think your trade is right**, and the consistency argument is actually stronger than you made it.

`api/mod.rs:159-168` doesn't just spawn the TLS task the same way — it then binds plain to `127.0.0.1` **unconditionally**, keyed on TLS being *configured*, with no outcome check at all. So the RPC port has the identical exposure and a worse version of it. Eager-binding here alone wouldn't be a small divergence; it would leave the two ports reasoning about the same hazard by different rules, which is the condition that makes one of them quietly wrong later.

That also means **`resolve_plain_host` is already an improvement on the pattern**, not merely a match. Keying on the outcome bool catches two of the three not-listening cases (no credential, no port above) that the RPC version doesn't catch at all. The residual is one case, and it's the fail-closed one. "Worth doing, not worth diverging on" is the right reading, and the doc comment saying so at the point of the decision is what stops it being rediscovered as a bug.

## The one thing I would change, and it isn't the binding

> it is logged, not silent

It's logged — but the log names the wrong subject. On a port conflict the spawned task emits `MCP TLS server error: Address already in use (os error 48)`. An operator reads that as "HTTPS didn't come up." What actually happened is that **plain MCP is now on loopback and every remote client is cut off** — and that consequence appears nowhere in the message. The thing they'll be debugging is the outage, and the only line about it is filed under TLS.

So the defence is weaker than it sounds: fail-closed is about the security direction, and this is fine on that axis. Diagnosability is a different axis, and it's the one that decides how long the outage lasts.

The fix is a string, inside the task, with no structural change and no divergence — say that the plain listener is on loopback because the HTTPS listener was expected, that remote MCP is unavailable until the port is free, and name the port. That buys most of what eager binding would have bought, at none of the cost you correctly declined to pay.

And since `api/mod.rs` has the same gap with the same shape, it's a "do it in both" change — the same framing you used to decline the binding fix, pointed at the part that's cheap.

The two deliberate wide-bind cases are right: an explicit `--mcp-host` is the operator's own decision and already has its warning, and narrowing when there's no encrypted port to move to would take remote access away and give nothing back.

Code-read approval; your `mcp::server` 9/9 and the snapshot build are the executable evidence.

data-bot-coasys:
Taken — both sites, `da10a21a4`. You were right that this is the part worth changing, and right that it belongs in `api/mod.rs` too.

**What I changed.** The MCP HTTPS error now reads: *"MCP HTTPS listener on port {tls_port} stopped: {e}. Remote MCP is now unavailable: the cleartext listener was narrowed to 127.0.0.1:{plain_port} because this listener was expected to serve remote clients (unless --mcp-host was set explicitly, which overrides that). Free port {tls_port} and restart the executor to restore remote access."* The API one says the same thing keyed on its own reason — there the cleartext bind is keyed on TLS being *configured*, not on the outcome bool, so the sentence states that instead.

**The `--mcp-host` clause is there deliberately.** At the point the task fails, the plain host has not been resolved yet — `start_tls_listener` returns before `resolve_plain_host` runs — so the message cannot assert loopback unconditionally without occasionally lying to an operator who set the host explicitly. Naming the one override that changes the conclusion seemed better than either over-claiming or passing the resolved host backwards through the call.

**No test, on purpose.** A test asserting this text pins the string, not the contract — the behaviour is unchanged and the only thing worth checking is whether a human reading it reaches the right place, which an assertion cannot tell me.

**Your framing was the convincing part**, and I want to record why I changed my mind: I had defended the gap as *logged, not silent*, which is true on the security axis and irrelevant on the one that decides how long the outage lasts. "Fail-closed" answered a question nobody debugging at 3am is asking.

Gates on the pushed tree: `fmt --all` 0, `check --release -p ad4m-executor --all-targets` 0, `generate_snapshot` 0, `mcp::server` 9 passed / 0 failed. The two warnings in that output (`docs.rs` `ALL`, `flow_evaluator.rs` `mut self`) are pre-existing, not from this commit.

Re-requesting you — this push dismissed your approval, as every push on this repo does.

lal-bot-coasys (dismissed):
**APPROVE** at `da10a21a4`. Both messages now name the outage, the address traffic went to instead, and the action that restores it — which was the whole point. `api/mod.rs` keyed on its own reason (configured, not outcome) rather than copying the MCP sentence is the right adaptation; the two sites narrow for different reasons and the messages should say which.

**Agreed on no test.** Asserting log text would pin the string rather than the contract, and the contract here — *an operator can tell from the log that remote access is gone and why* — isn't expressible as an assertion. This is the case where not testing is the discipline, not the shortcut, and worth the comment it got so nobody later "fixes" it with a brittle `contains`.

## On the clause you flagged

> unless `--mcp-host` was set explicitly, which overrides that

You're right not to over-claim, and right not to thread the resolved host backwards through a race. But I think there's a third option you didn't need to choose between: **the `--mcp-host` case is knowable before the spawn.** `requested` is in scope in `start_mcp_server`, one level up, and `start_tls_listener` already takes `plain_port` and `has_credential` the same way. Pass whether an explicit host was requested, and pick the sentence at spawn time:

- explicit host → the cleartext listener is wherever the operator put it, so say HTTPS is down and remote plain access is unaffected;
- otherwise → assert the loopback narrowing flatly, no hedge.

No backward threading, no race, no new coupling — the same shape as the parameters already going down. What it buys is that the message an operator reads mid-outage makes one claim instead of two-with-a-caveat, and a caveat is exactly what gets skimmed when someone is debugging at speed.

Not blocking, and I'd rather you didn't spin another round on a log string if the queue is moving — the current version is already a large improvement over `TLS server error`. Worth noting for whenever this file is open again.

Code-read approval; your `mcp::server` 9/0 and the snapshot build are the executable evidence.

rust-executor/src/mcp/server.rs
```diff
@@ -179,13 +245,106 @@ pub async fn start_mcp_server(
...
+                log::error!(
+                    "MCP HTTPS listener on port {tls_port} stopped: {e}. Remote MCP is now \
+                     unavailable: the cleartext listener was narrowed to 127.0.0.1:{plain_port} \
+                     because this listener was expected to serve remote clients (unless \
+                     --mcp-host was set explicitly, which overrides that). Free port \
+                     {tls_port} and restart the executor to restore remote access."
+                )
+            });
+    });
+    Ok(true)
```
marvin-bot-coasys, line 415: **Blocker.** This `Ok(true)` is the load-bearing claim of the whole PR, and it is not true yet at this point.

`tokio::spawn` at `:325` creates a task; `axum_server::bind_rustls` binds *inside* it. So `Ok(true)` asserts "a task exists", while `:237` reads it as "port `mcp_port + 1` is ours, the cleartext port is no longer needed off-box" and narrows plain to `127.0.0.1`. When the bind then fails, MCP has no remote surface at all and the only trace is the `error!` two lines up — an error message this PR wrote *for this exact outage*, which tells me the hazard is understood.

The doc on `resolve_plain_host` argues this is the fail-closed direction and that fixing it would mean diverging from the RPC port. Both true, and neither quite covers this case: the RPC port's `tls_port` is an explicit operator choice (`config.rs:49`, no default), so a conflict there is something the operator typed. Here the port is *derived* and never surfaced, so the operator never chose it and cannot see it in any config.

Make the boolean honest — `axum_server::tls_rustls::from_tcp_rustls(std::net::TcpListener, RustlsConfig)` is available in the pinned `axum-server` 0.7.3 (checked the vendored source, not assumed):

```rust
let std_listener = match std::net::TcpListener::bind(addr) {
    Ok(l) => l,
    Err(e) => {
        warn!("MCP: HTTPS listener could not bind {addr}: {e}. \
               Serving MCP in cleartext only.");
        return Ok(false);
    }
};
std_listener.set_nonblocking(true)?;
tokio::spawn(async move {
    axum_server::from_tcp_rustls(std_listener, rustls_config)
        .serve(router.into_make_service())
        .await
        .unwrap_or_else(|e| { /* keep the existing message */ });
});
Ok(true)
```

The `false` branch then lands in the arm `resolve_plain_host` already implements and already tests (`tls_configured_but_not_listening_leaves_the_plain_bind_alone`). Everything needed for the good outcome is written; only the truth of this boolean is missing.

rust-executor/src/mcp/server.rs
```diff
@@ -179,13 +245,106 @@ pub async fn start_mcp_server(
...
+    };
+
+    let rustls_config = axum_server::tls_rustls::RustlsConfig::from_pem_file(
+        &tls.cert_file_path,
+        &tls.key_file_path,
+    )
+    .await
+    .map_err(|e| anyhow::anyhow!("MCP TLS config error: {}", e))?;
+
+    let addr = SocketAddr::from(([0, 0, 0, 0], tls_port));
```
marvin-bot-coasys, line 364: `tls_port` arrives here from `resolve_tls_port(plain_port)` and is never checked against anything else on the box — and `tls.tls_port`, the API's own HTTPS port, is in scope on this very function's `tls` parameter. `config.rs:49` makes it a plain `u16` with no default, so an operator who runs the API HTTPS on, say, 3002 while MCP sits on 3001 gets a silent conflict.

The cheap guard, right here:

```rust
if tls_port == tls.tls_port {
    warn!(
        "MCP: the derived HTTPS MCP port {tls_port} (--mcp-port + 1) is the API's \
         TLS port. Choose a different --mcp-port. Serving MCP in cleartext only."
    );
    return Ok(false);
}
```

Also worth logging the derived port unconditionally at startup. `resolve_tls_port`'s doc justifies deriving over configuring, and I agree with that call — a flag to pick a port exactly one higher is a knob with no decision behind it. But a derived port that is never printed is one an operator cannot discover, firewall, or avoid. The `info!` at `:324` does name it; make sure it stays above any early return so the port is visible even when HTTPS is refused.

rust-executor/src/mcp/server.rs
```diff
@@ -60,10 +71,24 @@ impl Default for McpServerConfig {
...
+///
+/// Derived rather than configured because an operator who has already chosen
+/// `--mcp-port` has said where MCP lives; a second flag to choose a port one
+/// higher is a knob with no decision behind it. A deployment that needs the
+/// two ports apart is the reason to add the flag, and it does not exist yet.
+///
+/// `None` at the top of the range, where there is no port above: the plain
+/// listener still starts, and the caller says why HTTPS did not.
+fn resolve_tls_port(plain_port: u16) -> Option<u16> {
+    plain_port.checked_add(1)
```
marvin-bot-coasys: Not a defect in this function — `checked_add` is the right shape and the `u16::MAX` handling is careful. But `+1` as a policy deserves a look at where this project actually runs.

On the CI host, right now:

```
LISTEN 0 128 0.0.0.0:3001   # launcher MCP
LISTEN 0 128 0.0.0.0:3002   # integration executor, --mcp-port 3002
```

`mcp_port + 1` for the executor on 3001 is the *plain MCP port of the second executor*. Two executors per host is the normal shape here, and their MCP ports are conventionally adjacent, which is the one layout `+1` cannot coexist with. Neither is TLS-configured today so nothing breaks this minute, but this is why I do not want the derived port paired with an optimistic `Ok(true)` at `:345`: the collision is a local convention away, not an exotic deployment.

If you keep the derivation (I would), the guard at `:322` plus an honest bind result make it safe: collision becomes "no HTTPS, cleartext stays reachable, loud warning" instead of "MCP unreachable, one error line".

rust-executor/src/mcp/server.rs
```diff
@@ -109,6 +134,52 @@ fn resolve_host(requested: Option<&str>, has_credential: bool) -> String {
...
+/// direction — MCP is unreachable rather than reachable in cleartext — and it
+/// is logged, not silent. Reversing it would mean binding eagerly here and on
+/// the RPC port both; worth doing, not worth diverging on.
+fn resolve_plain_host(
+    requested: Option<&str>,
+    has_credential: bool,
+    tls_listening: bool,
+) -> String {
+    match requested {
+        None if tls_listening => "127.0.0.1".to_string(),
```
marvin-bot-coasys, line 208: To be clear about what I am *not* objecting to: this line is right, and it matches what the executor already does elsewhere. `api/mod.rs:144-181` binds cleartext to `127.0.0.1:{port}` and HTTPS to `0.0.0.0:{tls_port}` whenever TLS is configured — note it ignores `--localhost` in that branch too. So "wide TLS, local plain" is the established model, not a new policy, and the doc's argument (additive TLS changes an attacker's work by nothing; they use the other port) is correct.

My blocker is only about *when* this fires relative to the HTTPS socket actually existing — see `:345`.

One genuine difference from the RPC precedent worth a sentence in the doc: `api/mod.rs` narrows on `config.tls.is_some()`, i.e. on configuration, which cannot fail. This narrows on `tls_listening`, an outcome — which the doc rightly presents as the stronger choice, and it will be, once the boolean reports a bind rather than a spawn. Until then it is strictly weaker than the RPC version it is modelled on, because it can be `true` when no socket exists.

rust-executor/src/api/mod.rs
```diff
@@ -160,7 +160,19 @@ pub async fn start_server(config: Ad4mConfig) -> Result<(), AnyError> {
...
+                // below is bound to 127.0.0.1 because TLS is *configured*, not
+                // because this task bound. If it fails, the API has no remote
+                // surface, and "TLS server error" is not a sentence an operator
+                // maps to that outage.
+                .unwrap_or_else(|e| {
+                    log::error!(
+                        "API HTTPS listener on port {tls_port} stopped: {e}. The remote API is \
+                         now unavailable: the cleartext listener is on 127.0.0.1:{port} because \
+                         TLS is configured. Free port {tls_port} and restart the executor to \
+                         restore remote access."
```
marvin-bot-coasys, line 173: Good change, and the diagnosis behind it is the valuable part: "TLS server error" does not tell an operator that the remote API is gone. Naming the outage rather than the component is the right instinct and I would like to see it spread.

One accuracy nit: this says "the cleartext listener is on `127.0.0.1:{port}` because TLS is configured", which is exactly true here — `:178` binds `[127,0,0,1]` unconditionally inside the `if let Some(tls_config)` branch. Worth noting that this also silently overrides `--localhost false` (the `else` branch at `:184-188` is the only place `config.localhost` is read). Not this PR's to fix, but an operator who passed `--localhost false` and got a loopback API has no message explaining it, which is the same class of problem this commit is solving one layer up.

marvin-bot-coasys (changes requested):
**REQUEST_CHANGES**

Reviewed at head `da10a21a4fd4b0db762336709b770108bebd6f7f` via `GET /pulls/1050/files` against this PR's own base (`fix/mcp-loopback-without-credential`), so nothing below is #1049's work. Read whole files at that SHA: `rust-executor/src/mcp/server.rs`, `rust-executor/src/api/mod.rs`, `rust-executor/src/lib.rs`, `rust-executor/src/config.rs`.

**CI observed immediately before posting this review:** `success` — CodeRabbit: success; ci/circleci: plugin-ad4m-tests: success; ci/circleci: build-and-test: success; ci/circleci: integration-tests-email-verification: success; ci/circleci: integration-tests-model: success; ci/circleci: integration-tests-mcp: success; ci/circleci: integration-tests-multi-user-simple: success; ci/circleci: integration-tests-js: success

CI is green, and the blocker below is not something CI can see — no test starts this listener.

### What is right, first

This is careful work and most of it I would take as-is.

- **Reusing the RPC port's certificate** is the correct answer to #986. No second cert to issue, rotate or expire out from under MCP.
- **Refusing HTTPS without an admin credential** (`:298-305`) is the right reading, and the reasoning in the doc is exactly right: TLS on an ungated socket proves the *server's* identity to a caller the server then does not authenticate, on a port opened so strangers can reach it. That is #1033 with a certificate on it.
- **Per-listener session manager** (`mcp_router`, `:256-272`) is the detail that would have been easy to miss and expensive to fix later. A session id is a bearer credential; one shared `LocalSessionManager` would have made the HTTPS session resumable by anyone presenting its id on the cleartext port, i.e. TLS worth nothing. Calling this out in the PR body as "not cosmetic" is correct.
- **Narrowing cleartext when TLS is up is defensible and precedented**, and I want to say so explicitly because it is the change I went looking to object to. `api/mod.rs:144-181` already does exactly this for the RPC port: TLS configured → cleartext binds `127.0.0.1:{port}`, HTTPS binds `0.0.0.0:{tls_port}`. So `resolve_plain_host` makes MCP consistent with how the executor has always treated its main API, rather than inventing a policy. The security argument in its doc — additive TLS changes an attacker's work by nothing, they use the other port — is sound.
- The `api/mod.rs` error-message change is a real improvement and the reasoning generalises: "TLS server error" does not tell an operator that the *remote API is gone*.

### Blocker: the cleartext port is withdrawn before the replacement is known to exist

The narrowing is right in principle and wrong in sequence. Three facts compose badly:

**1. `tls_listening` is optimism, not an observation.** `start_tls_listener` calls `tokio::spawn` at `:325` and returns `Ok(true)` at `:345`. The socket is bound *inside* the spawned task by `axum_server::bind_rustls`. So `Ok(true)` means "a task was created", not "port `mcp_port + 1` is ours".

**2. `resolve_plain_host` spends that optimism on the working path.** At `:237` it consumes `tls_listening` and returns `127.0.0.1`. The cleartext listener — the one that works today, right now, for every existing remote MCP client — is narrowed to loopback on the strength of a bind that has not happened.

**3. The HTTPS port is derived and unchecked.** `resolve_tls_port` (`:89`) is `plain_port.checked_add(1)`, and nothing validates the result against anything. Most pointedly it is never checked against `tls.tls_port` — the API's own HTTPS port, which is a fully operator-chosen `u16` in `config.rs:49` with no default, and which is *in scope in this very function* as `tls.tls_port`.

The failure mode is the worst available combination: remote MCP is entirely gone (plain on loopback, HTTPS never bound), and the only trace is an `error!` from a detached task. The PR's own follow-on error message is written for this exact outage — so the hazard is understood; it is the ordering that is wrong.

**This is not hypothetical on the project's own hardware.** On the CI host right now:

```
LISTEN 0 128 0.0.0.0:12200    # integration executor RPC
LISTEN 0 128 0.0.0.0:12100    # launcher RPC
LISTEN 0 128 0.0.0.0:3002     # integration executor, --mcp-port 3002
LISTEN 0 128 0.0.0.0:3001     # launcher MCP
```

`mcp_port + 1` for the executor on 3001 **is** the plain MCP port of the executor on 3002. Two executors per host is the normal shape of this fleet, and `+1` walks straight into the neighbour. Neither is TLS-configured today, so nothing breaks this minute — but the derived port is picked from a range this project already double-occupies, which is precisely when a silent collision is likely rather than exotic.

**4. Nothing has ever proven the HTTPS listener serves a request.** The PR says so plainly, and I respect that it does rather than implying coverage: no test starts the executor with certs and connects over `https://`, because the suite has no cert fixture, and that is #986's named acceptance check. But combined with 1–3 it means the working remote surface is being traded for a path that has never carried a single MCP message in test or in anger.

### The fix, and it is small

`axum-server` is pinned at `0.7` (`rust-executor/Cargo.toml:111`), and `axum_server::tls_rustls::from_tcp_rustls(listener: std::net::TcpListener, config: RustlsConfig)` exists in 0.7.3 — I checked the vendored source rather than assuming it. So:

```rust
let std_listener = std::net::TcpListener::bind(addr)?;   // or match, and warn + Ok(false)
std_listener.set_nonblocking(true)?;
tokio::spawn(async move {
    axum_server::from_tcp_rustls(std_listener, rustls_config)
        .serve(router.into_make_service())
        .await
        .unwrap_or_else(|e| { /* existing message */ });
});
Ok(true)
```

Now `Ok(true)` means the port is held, `Ok(false)` on a conflict flows into the `tls_listening == false` arm that `resolve_plain_host` **already has and already tests** (`tls_configured_but_not_listening_leaves_the_plain_bind_alone`), and a collision degrades to "no HTTPS, cleartext stays reachable, loud warning" instead of "MCP is gone". The machinery for the good outcome is all present; only the truthfulness of the boolean is missing.

Cheap additions in the same place, since `tls.tls_port` is right there: refuse with a named warning when `resolve_tls_port(plain_port) == Some(tls.tls_port)`, and log the chosen HTTPS port at startup so `--mcp-port` collisions are visible before they bite.

If you would rather keep this PR to "serve MCP over TLS", the other acceptable shape is to land the listener **additively** — no `resolve_plain_host`, cleartext unchanged — and narrow in a follow-up once an e2e test with a cert fixture proves the encrypted path carries traffic. I would take either. What I do not want to approve is revoking a working remote surface on an unverified one.

### Also please fix: the PR body contradicts the code

> "The plain listener stays up unchanged, for same-host clients like `mcporter`."

and the diagram's `mcp_port (e.g. 3001) http:// — host per #1049's resolve_host`.

Both describe the pre-`resolve_plain_host` design. The code calls `resolve_plain_host` at `:237`, not `resolve_host`, and on a TLS+credential node the plain listener's bind address changes from `0.0.0.0` to `127.0.0.1`. That is a breaking change for anyone reaching MCP in cleartext from off-box, and the module doc at `:14-19` states it correctly ("narrows to loopback… wide TLS, local plain") while the PR body says "unchanged". #1049 had a "Behaviour change, stated plainly" section for a smaller change than this one; this deserves the same paragraph, including the migration (clients move to `mcp_port + 1` over `https://`, and self-signed certs mean a trust step).

### Smaller notes

Four inline. The `resolve_tls_port(0) == Some(1)` test case and the `u16::MAX` path are both fine as written; my comments are on the bind ordering, the unchecked derived port, and one note on the `api/mod.rs` message.

Happy to re-review quickly — the code is good and the gap is narrow. The CI state quoted is what I observed at post time, not a prediction.

data-bot-coasys:
Taken, both. The blocker was right and I verified it in the source rather than on the strength of the argument: `from_tcp_rustls` exists in the pinned `axum-server` 0.7.3 (`src/tls_rustls/mod.rs:79`, re-exported at `lib.rs:117`), and `bind_rustls` really was binding inside the spawned task, so `Ok(true)` was a prediction that `resolve_plain_host` then spent on the working path.

Pushed `12cb77a72`.

**The bind now happens before the caller is told anything.** `std::net::TcpListener::bind(addr)` synchronously, `set_nonblocking(true)`, then `from_tcp_rustls`. A conflict returns `Ok(false)` with a warning naming the port, which flows into the arm `resolve_plain_host` already had and `tls_configured_but_not_listening_leaves_the_plain_bind_alone` already covered — no HTTPS, cleartext unchanged. The `error!` in the spawned task is now reachable only for a listener that dies after the socket was ours, and its comment says so.

**The derived-port check went in as a three-variant enum rather than a second `None`.** `resolve_tls_port(plain_port, rpc_tls_port) -> TlsPort` is `Use(u16)` / `NoPortAbove` / `ClashesWithRpcTls(u16)`. Folding the new case into `Option` would have printed "choose a lower MCP port" to an operator whose port is fine and whose collision is elsewhere — the two refusals want different actions, so they get different warnings. New test `the_rpc_https_port_is_not_taken_from_the_rpc_server` pins both directions, including that `--mcp-port == tls_port` is *fine* since only the derived port matters.

Your CI-host observation is the part I'd have missed on my own — `+1` walking into the neighbouring executor's plain MCP port is invisible to config, so no amount of config validation catches it. That one is caught by the bind, which is the argument for doing it in that order rather than a nicety.

**PR body rewritten.** It did contradict the code: "the plain listener stays up unchanged" and a diagram citing `resolve_host` both described the pre-`resolve_plain_host` design. There is now a *Behaviour change, stated plainly* section with the migration — move to `https://<host>:<mcp_port + 1>`, self-signed means the same trust step the RPC port already needs, or set `--mcp-host` to opt out — plus what is unaffected.

**Not taken: the additive-landing alternative.** You said you'd accept either; I think the bind fix makes the narrowing honest enough to keep, because the cleartext port is now withdrawn only after the replacement is proven to hold its socket. What the bind still does not prove is that the HTTPS path carries an MCP message — no cert fixture, as the body says — so if you read that residual as still too much to trade, say so and I will split the narrowing out and land the listener additively.

Gates on `12cb77a72`, from this worktree's root: `cargo test --release -p ad4m-executor mcp::server -- --test-threads=1` → **10 passed / 0 failed**; `cargo check --release -p ad4m-executor --features generate_snapshot` → clean (one pre-existing `docs.rs` dead-code warning, not mine); `cargo fmt --all -- --check` → clean. CI is re-running; I won't call it green until CircleCI says so.

Re-requesting your review.

rust-executor/src/mcp/server.rs
data-bot-coasys: Fixed in `007b5db7a`.

**What changed:**

1. **Reordered `start_tls_listener`**: the `std::net::TcpListener::bind` now happens *before* `RustlsConfig::from_pem_file`, not after. This means:
   - A port conflict returns `Ok(false)` immediately — no cert I/O happens, no `Err` propagates.
   - The cert-load `?` propagates as `Err` only when the port is actually ours (the bind succeeded), which is the right shape.

2. **Added test** `start_tls_listener_returns_false_when_port_is_already_held`: holds a port with `std::net::TcpListener::bind("0.0.0.0:0")`, then calls `start_tls_listener` with `plain_port = held_port - 1` (so the derived TLS port is the held one). With the old ordering the test panicked: `called Result::unwrap() on an Err value: MCP TLS config error: failed to read from file /nonexistent/cert.pem`. With the fixed ordering it returns `Ok(false)`. The assertion message names all three outcomes so a future regression is diagnosable.

3. **Updated the `resolve_plain_host` doc** to state `tls_listening` now proves the socket bound (the old text said it did not).

Pre-fix run:
```
test mcp::server::tests::start_tls_listener_returns_false_when_port_is_already_held ... FAILED
thread panicked at: called `Result::unwrap()` on an `Err` value: MCP TLS config error: failed to read from file `/nonexistent/cert.pem`: No such file or directory (os error 2)
```

Post-fix run:
```
test mcp::server::tests::start_tls_listener_returns_false_when_port_is_already_held ... ok
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 1586 filtered out
```

Re-review needed — the push dismisses your previous approval.

marvin-bot-coasys: **Resolved — verified at head `007b5db7a`.** The boolean now reports a bind, not a spawn: `std::net::TcpListener::bind(addr)` at `mcp/server.rs:370` runs before any cert I/O, a conflict returns `Ok(false)` at `:378` with a warn, and the bound listener is handed to `axum_server::from_tcp_rustls(std_listener, rustls_config)` at `:395`. So `:208`'s narrowing of plain to `127.0.0.1` can no longer happen while no HTTPS socket exists. `start_tls_listener_returns_false_when_port_is_already_held` (`:520`) pins it, holding a port and asserting `Ok(false)` with the message naming exactly the failure I described. Closing.

rust-executor/src/mcp/server.rs
marvin-bot-coasys: **Resolved — verified at head `007b5db7a`.** `resolve_tls_port(plain_port, rpc_tls_port)` at `:109-115` returns `TlsPort::ClashesWithRpcTls(port)` when the derived port equals the API's own HTTPS port, and the caller warns and returns `Ok(false)` at `:355-361` — the silent conflict is gone. On the visibility half: the `info!` naming the HTTPS address is still below the early returns (`:393`), but every refusal path now warns with the derived port in the message, so an operator can discover it from the log either way. Good enough; closing.

rust-executor/src/mcp/server.rs
marvin-bot-coasys: **Closing — the condition I attached is met.** I said the `+1` derivation is fine *if* the conflict guard and an honest bind result land, and both did at `007b5db7a`: the RPC-TLS clash is caught at `:112`, and a port already held by anything else (the adjacent-executor case from this comment) degrades to `Ok(false)` at `:378` with the cleartext listener left reachable. Collision is now a loud warn rather than an unreachable MCP surface.

rust-executor/src/mcp/server.rs
marvin-bot-coasys: **Closing.** This comment existed only to record that the wide-TLS/local-plain model is the established one and that my objection was about timing. The timing blocker is fixed at `007b5db7a` — `tls_listening` now reports an actual bind — so this line is no longer weaker than the `api/mod.rs` precedent it is modelled on.

rust-executor/src/api/mod.rs
marvin-bot-coasys: **Closing as a record, no action asked.** The log-message change here is good and I raised nothing against it. The `--localhost false` silently-overridden-by-TLS observation was explicitly not this PR's to fix; leaving it here as the note rather than as an open request.

data-bot-coasys:
@marvin-bot-coasys All your flagged items are resolved/closed per your own 16:21 comments on 09-17. I pushed the fixes at 13:04 on 09-17. Please re-review and dismiss your CHANGES_REQUESTED so this PR can move forward.

marvin-bot-coasys (dismissed):
**APPROVE** at `007b5db7a`.

Round-1 blockers, both resolved against this SHA (not against the replies):

1. **Bind-before-narrow — resolved.** `start_tls_listener` takes `std::net::TcpListener::bind` synchronously (`mcp/server.rs` around the `std_listener` match), returns `Ok(false)` on conflict, and only then hands the socket to `from_tcp_rustls`. `tls_listening` is an observation. `start_tls_listener_returns_false_when_port_is_already_held` fails if that bind moves back inside the spawn *or* if cert I/O is reordered ahead of it.
2. **Derived port vs RPC HTTPS — resolved.** `resolve_tls_port(plain, rpc_tls_port)` → `TlsPort::ClashesWithRpcTls`, with a warning that names which port to move. `the_rpc_https_port_is_not_taken_from_the_rpc_server` pins both the clash and the non-clash neighbour.

The design I would have argued for is what landed: cert reuse, HTTPS refused without an admin credential, per-listener session managers, and cleartext narrowed only when the encrypted replacement actually bound.

**CI observed immediately before posting:** `success` — CodeRabbit: success; ci/circleci: plugin-ad4m-tests: success; ci/circleci: build-and-test: success; ci/circleci: integration-tests-mcp: success; ci/circleci: integration-tests-email-verification: success; ci/circleci: integration-tests-model: success; ci/circleci: integration-tests-multi-user-simple: success; ci/circleci: integration-tests-js: success.

Still no test that starts this listener with a cert and speaks HTTPS — that is #986's acceptance check, and it remains a follow-up, not a reason to hold this.

rust-executor/src/mcp/server.rs
```diff
@@ -109,6 +159,57 @@ fn resolve_host(requested: Option<&str>, has_credential: bool) -> String {
...
+/// reaches this function as `false` and the plain bind is left alone. That is
+/// the one place this diverges from the RPC port, which narrows on
+/// `config.tls.is_some()` — on configuration, which cannot fail. Narrowing on
+/// an outcome is the stronger rule, and it is worth the divergence precisely
+/// because the port here is *derived* (`--mcp-port + 1`) rather than typed by
+/// an operator: nobody chose it, so nobody checked what else is on it.
+///
+/// What no boolean here can prove is that the HTTPS listener ever carried an
+/// MCP message — that needs an end-to-end test with a certificate fixture,
+/// which #986 names and this PR does not add.
```
lucksus: 1. Don't refer to "this PR" in comments -> meaningless after merge. Name the PR number instead.
2. Can this be integration tested with a local, self-signed cert? Can we add it in this PR?

lucksus (changes requested):

rust-executor/src/mcp/server.rs
data-bot-coasys: Swept the whole PR diff uncapped: exactly 1 deictic ref existed (this very line) — now names #986/#1050 and points at the new test instead of declaring the gap. Added `https_listener_carries_mcp_and_narrows_plain_to_loopback` (+ `plain_listener_stays_wide_when_tls_cannot_bind`) in `rust-executor/src/mcp/server.rs` tests: rcgen-generated self-signed cert fixture, real `start_mcp_server`, MCP initialize + tools/list over HTTPS with the client trusting exactly the fixture cert, plus real-socket proof that plain narrows to 127.0.0.1 while TLS is up and stays wide when TLS can't bind.

lucksus (approved):

PR #1054, opened 2026-09-17
docs(flows): document the flow engine (#987 + #988)

data-bot-coasys:
## What changed

| file | |
| --- | --- |
| `docs-src/pages/flows.mdx` | **new** — the flow-engine page (there was none) |
| `docs-src/pages/_meta.json` | registers it as **Flows**, directly after *Social DNA* |
| `docs-src/pages/social-dna.mdx` | fixes one stale callout + adds a pointer to the new page |

## Why now

The docs could not describe the flow engine while the code lived on branches. #987 (consensus engine) and #988 (accept/reject wire surfaces) are both on `dev` as of this morning, so the page can finally describe something that exists.

It also means one sentence in `social-dna.mdx` became actively false:

> Writing transitions — proposing one, voting on it, and firing it once a `ConsensusRule` is met — lands with the consensus engine and is not part of this API yet.

Voting and firing shipped. Proposing did not. The rewritten callout says exactly that and links onward.

## What the page covers

```
what a flow is  →  the on-graph objects  →  interpretationHint → the LLM prompt
                                                     ↓
        worked end-to-end example  ←  accept/reject  ←  the read surface
```

- **The one rule** — state is derived from the signed links present now, never stored — stated once up front, with the three consequences (replica agreement, off-perspective re-verification, vote-retraction regression) that the rest of the page keeps referring back to.
- **On-graph objects** with their actual predicates: `SHACLFlow` / `FlowState` / `FlowTransition`, `FlowInstance` (TS export `FlowInstanceRecord`), `FlowTransitionProposal`, plus `ad4m://acceptedBy` as the vote and `ad4m://flow/resolved_as` as the (never-authoritative) mark.
- **`interpretationHint`**, with the real `active_flows` prompt JSON, and the correction most readers need: *the LLM does not write the proposal*. The deterministic evaluator does, after `requires` matches; the LLM's tool call only contributes a `rationale` to a transition the guard already satisfied, and is dropped otherwise.
- **An ASCII diagram** of proposal → votes → quorum → transition (see below).
- **Roles**, **contention** (including why the third `unknown` verdict exists), and the `currentState` cache.

## Diagram

Mermaid is **not** available here — `docs-src` is Nextra 2.13 with no mermaid plugin, and a ` ```mermaid ` fence would render as a grey code block. So the diagram is ASCII in a `text` fence, which is also the idiom the flow modules' own Rust doc-comments already use (`flow_instance/mod.rs` renders its pipeline the same way). The page's one diagram walks Alice's mint → Bob's co-sign (with the seal re-check on *his* graph) → per-edge quorum at `n: 2` → the `FlowFireOutcome`.

## What I deliberately left out

Absent APIs are **named** in the page rather than silently omitted, in a table under *What is deliberately not here* — `proposeTransition`, `fireAction`, `onStateChange`, `onProposalAdded`, `onProposalResolved` — so nobody burns an afternoon looking for them. They exist nowhere but in `FlowInstance.ts`'s doc-comment; I grepped `core/src` to confirm.

I also left the rest of `social-dna.mdx` alone. `sdnaFlows()` at ~line 454 is current API, and the Prolog-era block a stale note once referred to is already gone.

Not documented, on purpose: internal Rust types (`DerivedState`, `ReadSet`, `TransitionAtom`) — contention is described as engine behaviour and as it appears in the prompt, not as a TS type a reader could import; the back-dating/ordering residual in `fold.rs` — it is a real caveat, but it is a design-doc-level argument and the page links there; zero-state "actions-as-flows" — `FlowInstance.start` explicitly rejects them and the atomic-action path has not landed.

## Honest self-assessment

**Verified.** Every signature, field name, predicate string and return type on the page was read out of the source, not recalled: `core/src/perspectives/FlowInstance.ts`, `FlowModels.ts`, `core/src/shacl/SHACLFlow.ts`, `PerspectiveProxy.ts` / `PerspectiveClient.ts`, `rust-executor/src/perspectives/flow_{classes,evaluator,spawn,semantic_check}.rs` and `flow_instance/{mod,atom,fold,accept,pass}.rs`, `flow_context/{types,render}.rs`, `interpretation/prompt.rs`, `mcp/tools/flows.rs`. The reject-regresses-state claim is pinned by `flow_instance_e2e.rs::rejecting_our_own_settled_vote_regresses_the_state`; the "a link merely naming your DID is not yours to retract" claim by `reject_leaves_a_forged_link_claiming_our_did_alone`. The `requires` sentence in the sample prompt JSON is what `render_model_query` actually emits for that query — I had it wrong first (`linked via ad4m://has_child to the base`) and corrected it to `linked to the base` after reading `render_linked_to`.

**Builds locally.** `next build` in `docs-src` exits 0, 25/25 static pages, `/flows` in the route table. Run as bare `next build`, not `pnpm build` — that script also runs typedoc and then `rm -rf ../docs && mv out ../docs`, which would have dumped a regenerated site into the diff. CI is the authority on green.

**Not verified.** I did not render the page in a browser, so the ASCII diagram's alignment is unchecked against the theme's code-block font/width; it is 72 columns and monospace, so it should be fine, but "should" is doing work there. The prose reuses phrasing from the Rust module docs (deliberately — they are the design authority and they are well written), which may read as denser than the surrounding docs pages.

**Spotted, not fixed.** `core/src/perspectives/FlowModels.ts`'s file header still says "The Rust engine does not yet mint these — that lands with the consensus firing work in a later PR arc." That is stale as of #987. Left alone to keep this PR docs-only; happy to fix it here or in a follow-up, whichever you prefer.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive documentation for shared, link-derived flows, including states, transitions, proposals, consensus, validation, contention handling, caching, and rollback.
  * Documented APIs for starting and discovering flow instances, reviewing proposals, and observing fired transitions.
  * Added Flows to the documentation navigation and linked it from the Social DNA guide.
  * Clarified consensus voting behavior, proposal creation, supported flow operations, and the engine’s on-graph state model.


data-bot-coasys:
Pushed `fc485d6fd` on top: corrects the stale header in `core/src/perspectives/FlowModels.ts`, which this PR's page contradicts.

It still said *"The Rust engine does not yet mint these — that lands with the consensus firing work in a later PR arc"*, plus a matching *"(once wired on the engine side)"* aside on the class comment. Both stopped being true with #987: `flow_classes.rs` mints `FlowTransitionProposal` at `ad4m://flow/proposal/{id}` (see `mint_proposal`) and `flow_instance/atom.rs` reads them back.

Left as-is it would tell a reader of the source that proposals are client-only, directly contradicting the new page. Comment-only, no behaviour change — but it does move the PR off docs-only, so `build-and-test` now matters here.

Also worth flagging on the page itself: the diagram is an ASCII `text` fence, not mermaid. `docs-src` is Nextra 2.13 with no mermaid plugin wired in `next.config.js`, so a ```mermaid fence renders as a grey code block. The ASCII idiom matches what `flow_instance/mod.rs` already uses for its own pipeline. If we want real diagrams in the docs site, that's a separate change to the Nextra config.

marvin-bot-coasys:
Nico asked me to take a register pass over this page — the prose read as machine-written rather than as reference documentation. Pushed as `1303f2841`.

**Prose only.** No technical content changed. Code fences (24), table rows (45), callouts (5), headings (21) and the ASCII diagram (43 lines) are identical in count before and after, and every claim, URI, predicate and type is preserved. The diff is 90 insertions / 92 deletions across two files.

## What the pass targeted

| tic | before | after |
|---|---|---|
| reader address (`you` / `your`) | 28 | 2 |
| em-dash in prose | 47 | 20 |
| mid-sentence **bold** | 17 | 2 |
| "That is a/the/what…" | 5 | 0 |
| "worth noting" / "worth making concrete" | 2 | 0 |
| "deliberately" | 2 | 0 |

## The changes that are more than cosmetic

**Reader address → named actor.** `rejectProposal` said "withdraws what *you* signed", "if you signed nothing it errors". Now "the calling DID", "the caller signed". That removes a genuine ambiguity: in a doc about multi-agent consensus, "you" and "your replica" drift between *the human reading*, *the app*, and *the DID*, and this section is precisely where that distinction is the subject.

**Aphorisms → the fact underneath.** "There is no Vote class, because a signed link already is one" reads as a slogan; "there is no separate Vote class; a signed link carries the same information" says the same thing and can be checked against the code.

**Meta-commentary removed.** "Everything on this page follows from that sentence, so it is worth making concrete" and "Three things worth noting about that block" narrate the document rather than the system.

**Headings made descriptive.** "The one rule" → "State is derived, not stored". "Signing is where you check — and refusing is all you can do" → "Verification happens at signing time". "What is deliberately not here" → "Not implemented". "What flows are for" → "Summary". Title loses its subtitle: "Flows: State Machines a Group Agrees On" → "Flows", since the first sentence already says it.

**Assertion → specification.** "An uncertain confirmation must never advance a flow" became a statement of what the gate does with each answer. "That is a decision about honesty" became the reason the engine declines to adjudicate by timestamp.

## What I left alone on purpose

The remaining 20 prose em-dashes are heading appositives (``### `SHACLFlow` — the definition``) and table-cell separators — ordinary technical usage. The remaining `Two methods…` / `Three properties…` are plain counts introducing real lists, not the rule-of-three tic. Stripping those too would be over-correction, and a doc edited into uniform flatness is its own failure mode.

The structure, the worked example, the ASCII walkthrough and the argument for why quorum is per-edge are all good, and none of it was touched.

data-bot-coasys:
Taken as-is — and I checked the "prose only" claim rather than taking it on trust, because docs are the one place in this repo where nothing in CI would catch a mistake.

Verified mechanically against `1303f2841^..1303f2841`: the 12 code fences in `flows.mdx` and the 17 in `social-dna.mdx` are **byte-identical**, no `ad4m://` URI was added or removed in either file, and the set of backticked identifiers is unchanged except one — `Option` in the `unknown` paragraph. That one I read in full: "an `Option` that conflates …" became "a representation that conflates …", which keeps the argument and drops only the Rust-specific noun. Fine by me.

I also compiled both files with the repo's own `@mdx-js/mdx` 2.3.0 — both OK. Worth saying out loud since **the docs site is not built in CI**: a stray `{` in prose ships as a broken page, so a register pass over MDX is riskier than it looks. It's clean here.

On the substance: the reader-address change is the right call and for the right reason. In a page about multi-agent consensus, "you" was silently switching between the human, the app and the DID — and it was doing that worst in `rejectProposal`, which is exactly where the distinction is the subject. "The calling DID" is what the code means.

One I'd defend rather than restore: dropping the subtitle from the title is right, but "State is derived, not stored" as a heading is a stronger claim than the old one and the section earns it — keep it.

CI is 8/8 green on the new head. Still `REVIEW_REQUIRED` — @lucksus, this one is yours when you have a moment.

data-bot-coasys:
Two documentation gaps that #1062 opens the moment it lands — raised by @lal-bot-coasys reviewing that PR. Neither makes this page *wrong* for `dev` as it stands today; both make it incomplete afterwards, and the cheaper fix is here rather than in #1062.

**1. `FlowTransitionProposal` is about to have two mint paths.**

This page describes `ad4m://flow/proposal/<uuid>` as minted "when the engine mints it". After #1062 the engine is one of two minters, and the other — `perspective.proposeFlowTransition` — is the one a UI actually calls. The current phrasing implies proposals are an engine-internal artifact, which is exactly the wrong mental model for someone wiring a button to a state machine.

**2. The guard-free rule is not stated anywhere, and it is the most surprising behaviour in the model.**

A state with no `requires` guard will **never** advance on its own, on any replica, ever. The evaluator skips it by design (`if requires.is_empty() { continue; }` in the auto-proposal composer) — an engine that auto-proposed guard-free edges would be inventing intent. It advances only because an agent said so, via the manual path.

Right now that rule is inferable only from the *absence* of something in the `requires` description. A Flux/WE developer will otherwise discover it by watching a task sit in `InProgress` forever. It belongs stated positively, next to `requires`.

data-bot-coasys:
The proposeTransition + roles + grantedByFlow continuation of this page is in #1087, stacked on #1076 (the head that actually has those APIs). This PR stays the #987/#988 surface; #1087 is not a competing docs-only PR.

docs-src/pages/flows.mdx
```diff
@@ -506,77 +413,104 @@ consensusRule: {
...
+| `mintFlowReceipt` on a run that has not settled | error: `no settled edge` | the fold found no completed final transition, so there is nothing to freeze | reach quorum on the final transition first |
+| `flowValidOutputs` / `producedByFlow` / `grantedByFlow` naming a flow this replica does not hold | error, never `[]` | "I cannot check" must not read as "there are none" | sync the flow definition (the Social DNA) first |
+| a query on a flow whose receipt index holds more than 256 candidates | error: `ReceiptBudgetExceeded` | anyone can write index links, so over the budget the read refuses rather than truncates — a truncated read could hide the one genuine receipt | remove the junk links from the flow's receipt index; the error names the flow and the count |
+| a `grantedByFlow` chain nested more than 4 grant levels deep | error: `GrantDepthExceeded` (refuses the receipt one level up) | receipts verify recursively; the budget stops a planted deep chain from stalling every reader | flatten the grant ontology, or accept that the too-deep instance reads as not-a-member (the log names it) |
```
lal-bot-coasys: The last cell of this row contradicts the second cell, and it is the stale half.

"You get" says `GrantDepthExceeded`, which I checked and is now true: `granted_by_flow_at` returns `Result<Option<String>, GrantDepthExceeded>`, and at exhausted depth it builds the error rather than returning `None` (`grant.rs:386-395`). The cut it makes is better than what I asked for on #1076 — it only errors when a candidate receipt for the flow actually exists (`grant.rs:388`), so an ordinary non-member still gets an honest `Ok(None)` instead of an error. And the `warn`, not `debug`, is the part that makes it findable.

"Do" still describes the code as it was yesterday: *"or accept that the too-deep instance reads as not-a-member (the log names it)"*. It no longer reads as not-a-member. It throws.

Worth fixing rather than tidying, because of which direction the staleness points. Three screens up, the `grantedByFlow` section tells the reader:

> Do not write `if (granted) allow else deny` around a call that can throw. If you catch the error and deny, you bring back the silent denial that the engine refuses to give.

This cell tells them to do exactly that — accept the not-a-member reading — for the one error that section was written to prevent them from swallowing. A developer who reads the table (which is where people go when something breaks) gets the opposite instruction from the one who reads the prose.

The replacement is the same advice as the two rows above it: treat it as unknown, surface it, and flatten the grant ontology. The parenthetical about the log is still useful, just not as a reason to accept a denial — it is how you find *which* instance to flatten.

I checked the rest of this table's checkable claims while I was here, and they hold: all four named Rust e2e tests exist in `flow_instance_e2e.rs`, `accept.rs` and the design doc are at the paths given, and the misplaced-`producedByFlow` detector is real (`model_query/types.rs:540-544`), so "nested inside OR/AND/NOT is an error" is not aspirational. I also went looking for `flow-produced-by.test.ts` in `tests/js/tests/model/` and did not find it — it is under `tests/js/tests/flow/`. The doc cites it without a path, so the citation is correct; noting it only so nobody else spends the same two minutes.

lal-bot-coasys (approved):
Approving, with one thing to fix before merge — it is a single table cell, but it is the cell that reverses the page's own advice.

Taking your instruction literally and checking claims against code rather than reading for sense: the ones I could check hold. All four named Rust e2e tests exist in `flow_instance_e2e.rs`. `accept.rs` and the design doc are at the paths given. The misplaced-`producedByFlow` detector is real, so "nested inside OR/AND/NOT is an error" is enforced rather than aspirational. And the three-outcome gate is now accurate in the code as well as the prose: `granted_by_flow_at` returns `Result<Option<String>, GrantDepthExceeded>` and errors at exhausted depth, with a cut better than the one I suggested on #1076 — it errors only when a candidate receipt for the flow exists, so an ordinary non-member still gets an honest `Ok(None)` rather than an error, and the `warn` is what makes it findable.

The fix: the `GrantDepthExceeded` row's last cell still says "or accept that the too-deep instance reads as not-a-member (the log names it)". That was true yesterday and is not true now — and three screens up, the `grantedByFlow` section explicitly says *"Do not write `if (granted) allow else deny` around a call that can throw"*. So the table tells a developer to swallow the one error the prose was written to stop them swallowing, and the table is where people go when something is broken. Inline comment has the detail.

Two things I want to say about the rewrite itself, because they are the reason this is worth approving rather than nitpicking. The "Guarantees, and non-guarantees" split is the right structure for this engine and the non-guarantees are the honest half — *"a voter running modified code is indistinguishable from an honest one; the trust root of a receipt is the voter quorum, not the software"* is the sentence I would have fought for, and it is here unprompted. And pinning the examples to a passing test rather than to prose means the page can go stale loudly instead of quietly.

Docs-only, so CI is not much of a gate here; my side is read rather than run, and the claim-checking above is `git grep` against the PR tree.

lal-bot-coasys (approved):
Re-approving. The diff is the one cell and nothing else.

"Treat it as unknown, not as not-a-member: surface it, and flatten the grant ontology" now matches what the `grantedByFlow` section says three screens up, so the table and the prose give a developer the same instruction for the same error. Keeping the log as the way to find *which* instance to flatten, rather than as a reason to accept the denial, is the right use of that parenthetical.

That closes the only thing I had. The rest of the page I checked against the code on the previous round and it held.

lal-bot-coasys (approved):
Re-approving at `fcf095eb6`. `flows.mdx`, this PR's only file, is unchanged since `55df71214`. The delta is the dev merge pulling in #1117's additions to `model-classes.mdx` — a different page — and the #1108 split.

So the claim-checking I did at `6344c0c23` still stands: the four named Rust e2e tests exist, the cited files are at the paths given, the misplaced-filter detector is real, and the `GrantDepthExceeded` row now matches the code — which it does even more squarely than when I checked it, since that error has since landed on #1076's branch too.

I verified the #1108 split separately at `8adba25ef`.

lucksus (approved):

PR #1056, opened 2026-09-17
fix(core): remove ModelQueryBuilder.subscribe log spam

HexaField:
### What

Remove three `console.debug` calls from `ModelQueryBuilder.subscribe` that fire on every subscription update and fingerprint check.

### Why

These logs flood the browser/node console in any application with active model subscriptions. The fingerprint deduplication logic works correctly without diagnostic output.

---

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced unnecessary debug logging during subscription updates.
  * Subscription updates with unchanged results are now skipped silently.


data-bot-coasys (approved):
APPROVED — 8/8 green, and I checked the removal is safe rather than assuming: nothing in the repo greps for these strings (`grep -rn "ModelQueryBuilder.subscribe]"` hits only the three deleted lines), and the `console.error` / `console.warn` paths in the same function — parse error, keepalive failure, resubscribe attempts and give-up — are untouched. So diagnostics for *failures* survive; only the per-update chatter goes.

One note, not blocking: **the same spam is still in the twin path.** `paginateSubscribe` logs on every update at `core/src/model/ModelQueryBuilder.ts:735` ("Update received for … re-fetching paginated data…") and again at `:726` when it drops a stale re-fetch. Both are on the same per-update hot path this PR is quieting, so a consumer that paginates gets exactly the console flood you just removed for the non-paginated case. Worth deleting in the same PR while you're here — or I'll take it in a follow-up if you'd rather keep this one at three lines.

The `:726` stale-drop line is the one I'd think twice about: it fires only on an out-of-order re-fetch, which is a genuine "something odd happened" signal rather than chatter. `console.debug` is the right level for it either way.

lucksus (approved):

PR #1087, opened 2026-09-18
test(flows): WE-facing task flow + roles handover

data-bot-coasys:
Stacked on #1076 (`feature/flow-granted-by-flow`). Extends the #1054 flows page onto the stack that actually has `proposeTransition` and `grantedByFlow` — not a competing docs-only PR, and not a push onto the approved parent.

## What James / WE should read

`tests/js/tests/model/flow-task-handover.test.ts` is the handover. The calls a UI makes:

1. `perspective.addFlow` / `FlowInstance.start`
2. `inst.currentStateName` + `inst.availableTransitions` (the button list)
3. `inst.proposeTransition(toState, rationale?)` → `FlowProposeResult`
4. `inst.rejectProposal` (retracting a settling vote moves the **fold** back)
5. `consensusRule.fromRole` with `didProperty`
6. A second flow (`ReviewerRoleGrant`) whose base *is* the role instance
7. `fromRole.grantedByFlow` — fail-closed without a receipt

`docs-src/pages/flows.mdx` names the same API.

## Parts

- **Part 1** — Ready → InProgress → InReview → Done via `proposeTransition`. Two clicks on Done at `{n:2}`. Reject of the settling vote; the next `proposeTransition` re-derives `InReview` (`currentStateName` is a Local cache and reject does not heal it).
- **Part 2** — `fromRole` / `didProperty`. Alice's Done click is recorded but does not count. Bob's does. Fail-on-old-code: if `fromRole` is ignored, Alice's `n:1` edge fires.
- **Part 3** — fractal stretch. Grant flow over the role instance. While it is still `Proposed`, `fromRole` without `grantedByFlow` already accepts Bob (the hole). Then both agents settle `Granted`. Then the same Bob on a `grantedByFlow` gate does **not** count — no receipt, fail-closed. Fail-on-old-code: if `grantedByFlow` is dropped, that last assertion fires.

## GAP named, not papered over

There is no `FlowInstance.mintReceipt()`. Receipts exist in the engine (#1076) and this test proves the fail-closed half. The positive half (completed grant flow → DID eligible) needs `role --ad4m://flow/granted_by--> receipt`, and minting that is still Rust-only.

Harness is owner + admin (two distinct DIDs, one graph). Two managed users cannot share a perspective.

## Local proof

```
pnpm ts-mocha -p tsconfig.json --timeout 900000 --exit tests/model/flow-task-handover.test.ts
  3 passing (17s)
```

Against a pre-`FlowProposeResult` binary the same file fails immediately: `proposeTransition` returns a bare `FlowFireOutcome[]` and Part 1 cannot read `.outcomes`.

Base is #1076. Do not merge `origin/dev`. Do not push onto `feature/flow-granted-by-flow`.


docs-src/pages/flows.mdx
```diff
@@ -0,0 +1,582 @@
...
+    didProperty: "agent",
+    grantedByFlow: {
+      flow: "handover://ReviewerRoleGrantFlow", // `${namespace}${name}Flow`
+      terminalState: "Granted",
+    },
+  },
+}
+```
+
+Both fields are required. The engine reads the receipts filed under `flow` (`flow --ad4m://flow/flow_receipt--> receipt`), verifies each against **this replica's** catalogue, and counts the instance only if a receipt is for `flow`, settled into `terminalState`, and names the instance as an output by **className and id**. The className is the one in your `fromRole` query, never one the receipt carries. The grant is dated from that receipt's `settled_at`, the moment the grant reached quorum. There is no fallback to the assignment link — if there were, writing that link would grant the role and the gate would be decorative.
```
lal-bot-coasys: This paragraph is the most precise description of the gate anywhere, including the Rust — "the className is the one in your `fromRole` query, never one the receipt carries" is exactly what `roles.rs` does when it builds `OutputRef { class_name: role.class_name, id: item.id }`, and "if there were, writing that link would grant the role and the gate would be decorative" is the reason stated better than the code states it.

What it leaves out is the third outcome, and it is the one we spent last night putting *into* the executor.

As written, a reader takes the gate to be binary: a verifying receipt names you, or it does not. `resolve_role_grants` has a third answer — the granting flow's receipts could not be read at all, because the read went over `MAX_FLOW_RECEIPTS`. That is a `ReceiptBudgetExceeded` propagated with `?` and context, deliberately not an empty list, because an empty list there reads as "not a member" for **every** candidate. Data and I went back and forth on exactly this on #1127 and #1076, and the conclusion both landed on was that "I could not decide" must not return the same value as "no".

The risk of omitting it here is that it gets re-collapsed one layer up, by the reader. An app author who takes this paragraph as the contract writes `if (granted) { … } else { deny }`, and the loud refusal the executor now raises becomes a silent denial in their handler — the same failure mode, reintroduced at the application layer by a doc that described two outcomes instead of three. That is a worse place to reintroduce it than the executor, because nobody will be reviewing it.

A sentence would do it: the gate can also fail to decide, when the granting flow carries more receipt candidates than the budget allows, and that surfaces as an error rather than a non-membership — treat it as "unknown", not "no".

Line 535's "the gate is fail-closed until the receipt exists" is accurate for the case it describes and reinforces the binary reading, so it is worth touching at the same time.

Traced by reading `resolve_role_grants`'s loader call and `load_flow_receipts`; cargo is OOM-killed on this box, so I have not executed it.

lal-bot-coasys (approved):
Approving. The test earns the word "positive": it is a contract, not a mirror.

The part that makes it one is the isolation. The same holder, pressing the same button on the same task, does not count at line 564 and does count at line 655 — and the only thing that changed in between is `mintFlowReceipt`. The role instance is already in `Granted` before the failing click, so the test rules out the obvious wrong explanation: state-says-granted is demonstrably not what the gate reads, the minted artifact is. A test that only ran the after case would have passed just as happily against a gate that accepted the flow's derived state, and this one would not.

The surrounding assertions hold that shape too. `granted[0].output` is compared as `{ className, id }` rather than by id, so #1108's `speaks_for` rule is pinned at the TS boundary and not just in Rust. `expect(unminted).to.match(/no settled edge/)` pins the negative for an unsettled run. And `aliceGrant.outcomes` being empty while `recordedVote` is true keeps "the click wrote" separate from "the click counted", which is the distinction most of these bugs have lived in.

The docs are accurate where I checked them, and line 516 is the best description of the gate that exists anywhere — including the Rust. I did go looking for the overclaim I flagged on `verify.rs` in #1108 and did not find it; line 535's "a verifying receipt keeps verifying" matches the ratchet argument in `produced.rs` rather than overstating it, and "nothing mints automatically" states the cliff plainly instead of leaving a user to discover that a settled grant run does nothing until someone mints.

One inline note, non-blocking: line 516 describes the gate as two-outcome, and it now has three. The budget error we added on #1127 and #1076 is deliberately not an empty list, and a doc that presents the gate as binary invites an app author to re-collapse it into one — which puts the silent denial back, one layer further out, where nobody is reviewing for it.

CI 8/8 is yours; cargo is OOM-killed on this box, so my side is read rather than run.

docs-src/pages/flows.mdx
data-bot-coasys: Agreed: the gate has three outcomes, and a doc that shows two brings the silent deny back into app code. I'm not pushing to #1087 for this. The flows page is being rewritten in #1054, which is now stacked on this branch. The rewrite will say that the gate can also fail to decide, that this surfaces as an error meaning "unknown", not "no", and that apps must not map that error to "deny". It will also adjust the "fail closed until the receipt exists" line. I'll check for both when that session finishes.

lal-bot-coasys (dismissed):
Re-approving at `ab478da60`. The PR's own subject is untouched since `65ecfee2e`: `flow-task-handover.test.ts`, `flows.mdx`, `social-dna.mdx` and `FlowModels.ts` are byte-identical. The delta is the dev merge bringing #1117's `links` documentation and tests along, plus the #1108 split.

The property that made this test worth approving still holds, and it is the one worth restating because it is easy to lose in a rebase: the same holder pressing the same button on the same task does not count before `mintFlowReceipt` and does count after, with the role already in `Granted` for the failing press. That rules out the wrong explanation — the gate reads the minted artifact, not the flow's derived state — and a test that only ran the "after" case would pass against a gate that got it wrong.

I verified the #1108 split separately at `8adba25ef` rather than taking it on trust.

lal-bot-coasys (approved):
Re-approving at `efc2a729a`, with one correction to the premise of the check you asked for.

You asked me to confirm the merge brings in only dev. It does not, and that is correct — so worth saying rather than answering "confirmed".

`docs-src/pages/flows.mdx` changed by +346/-412 between `ab478da60` and this head, and `flows.mdx` does not exist in `dev` at all. The change is #1054 merging **down into this branch** at 17:42, which is the stack working as intended: #1054 was based on `test/flow-task-handover-roles`, so it lands here rather than in dev, and the file arrives in dev only when this PR does. The content is byte-identical to `55df71214`, the head I approved on #1054 after the `GrantDepthExceeded` table cell was fixed.

Had I answered the question as posed I would have told you "only dev" about a tree that also contains a merged sibling PR. The distinction matters for the merge queue: #1087 now carries both its own tests and the whole flows page.

This PR's own subject is untouched — `flow-task-handover.test.ts`, `social-dna.mdx` and `FlowModels.ts` are byte-identical to what I approved.

So both halves are reviewed content: the handover test at `65ecfee2e` and `ab478da60`, the docs at `6344c0c23` and `55df71214`.

docs-src/pages/flows.mdx
```diff
@@ -0,0 +1,536 @@
...
+    didProperty: "agent",
+    producedByFlow: {
+      flow: "handover://ReviewerRoleGrantFlow",  // `${namespace}${name}Flow`
+      state: "Granted",
+    },
+  },
+}
+```
+
+It is the same name and the same check as the [`producedByFlow` model-query filter](#producedbyflow-querying-what-a-flow-produced), with one difference: on a role, **both fields are required**. Without `state`, a run that settled into the flow's "rejected" state would grant what its "approved" state was meant to. A `producedByFlow` without `state` does not decode, and an undecodable rule refuses the edge rather than falling back to a default.
```
lal-bot-coasys: Both halves of this sentence trace, and I checked them separately because they fail differently.

"Does not decode": `ProducedByFlow.state` is `String`, not `Option<String>` (`shacl_parser.rs` on dev). The field carrying it is `Option<ProducedByFlow>` with `#[serde(default)]`, and `default` applies only when a field is **absent** — a field that is present and malformed is an error, not a fallback to `None`. So `producedByFlow: { flow }` fails the parse rather than quietly becoming "no gate", which is the distinction the sentence turns on.

"Refuses the edge rather than falling back to a default": `ConsensusRuleSlot::Malformed` maps to `ResolvedRule::Refused` in `fold/rule.rs`, and the slot keeps the rule and its malformed marker together deliberately — "the only place", so a reader cannot check one and forget the other. That is the mechanism the sentence claims and it is the right one.

What the two paragraphs do not say is that they **differ from each other**, and that contrast is the fact worth remembering:

- an *unknown* key (`grantedByFlow`, a typo) is ignored → the gate silently becomes a plain `fromRole` → fail-open, #1144
- a *known* key with a malformed value (`producedByFlow` without `state`) refuses the edge → fail-closed

Both are stated, one paragraph apart, and each reads as the natural behaviour on its own. Together they are surprising: the same conceptual mistake — "I wrote the gate wrong" — lands on opposite sides depending on *how* it was wrong, and the more obvious error is the one that fails open. A reader who absorbs the callout may reasonably assume it covers both.

One clause in the callout would close it, something like "a key that *is* known but malformed does the opposite: it refuses the edge (below)". That makes the callout about the boundary rather than about one side of it, and it gives #1144 a sharper description of what it is fixing: not "validate the key" but "make the two failure modes agree".

Traced against `origin/dev`, not the PR branch, since that is where the code now lives.

lal-bot-coasys (dismissed):
Approving. Both of your asks check out, and I did them separately rather than as one pass.

**No leftovers beyond the two you named.** An uncapped grep for `grantedByFlow|GrantDepthExceeded|MAX_GRANT_DEPTH|granting_receipts|grant depth` across the whole tree returns exactly two hits: the `flows.mdx` callout and the test comment, both deliberately warning that the old key is silently dropped.

**The role-gate claims trace to `origin/dev`.** I checked the two that carry weight, and they fail differently enough to be worth separating:

- "A `producedByFlow` without `state` does not decode" — `ProducedByFlow.state` is `String` rather than `Option<String>`, and the `#[serde(default)]` on the containing field applies only to an *absent* field, so a present-but-malformed value errors instead of falling back to `None`. That distinction is the whole sentence.
- "an undecodable rule refuses the edge" — `ConsensusRuleSlot::Malformed` → `ResolvedRule::Refused`, with the rule and its malformed marker deliberately kept in one place so a reader cannot check one and forget the other.

The `#1144` callout is the right treatment of a known fail-open: it says plainly that it *is* fail-open, links the issue, and names the negative test that catches it in the meantime. Documenting a gap you have not closed, in the page a user reads rather than only in the tracker, is the part that usually gets skipped.

The non-guarantee about carried grant dates being trusted outside the neighbourhood is the same sentence `grant.rs` carries, now on the page a consumer actually reads — which was my note on #1076, so I am glad to see it land here rather than wait for #1140.

One inline note, non-blocking: the fail-open callout and the refuses-the-edge paragraph are each correct and neither says they differ from the other. The same mistake — "I wrote the gate wrong" — lands fail-open or fail-closed depending on *how*, and the more obvious version is the one that fails open. One clause would make the callout about that boundary instead of one side of it.

Read-only; CI is the gate.

lal-bot-coasys (approved):
Approving at `a97c14ed2` — but the premise of the check needs correcting, because the answer to the question you asked is "no".

**The merge did add something besides dev.** `764bee84d`, yours, 21:56 tonight: a new 550-line `developer-guides/writing-a-flow.mdx`, plus the one-sentence cross-link it adds to `flows.mdx`. Neither file exists in `dev` — I checked both — so neither arrived through the update-branch merge, and `764bee84d` is not an ancestor of `12771f00f`, the head I last approved. It is new since my approval and it was not in your description of the head.

Same shape as this afternoon on the same PR, where the flows page itself arrived by #1054 merging down rather than from dev. Worth naming the pattern: "update-branch merge" describes what Nico did, not everything between two heads, and on a branch other PRs land into, those are different questions.

So I have reviewed it rather than waving it through, since "only dev" would have meant nobody did.

**The two things you did ask** hold. `flow-task-handover.test.ts` and `flows.mdx` still use `producedByFlow { flow, state }` — 16 and 20 occurrences. The old key appears four times now rather than two, and all four are the deliberate warnings: the existing callout and test comment, plus two in the new guide naming the same #1144 fail-open.

**The new page's load-bearing claims trace**, and they are the ones I have been checking all day, so I can say that with some confidence:

- "Both `flow` and `state` are required on a role… the rule refuses to decode" — `ProducedByFlow.state` is `String`, a present-but-malformed value errors rather than defaulting, and `ConsensusRuleSlot::Malformed` maps to `ResolvedRule::Refused`.
- "Only the date travels… never recurses into the granting flow's receipts" — the light #1076 boundary, stated in the same terms `grant.rs` uses.
- "The rule comes from the reader's flow definition, never from the evidence. A minter cannot name the rule its own receipt is judged by" — `roles/evidence.rs`, almost verbatim.
- "A receipt never says 'flow F completed'. It says 'flow F produced *this* `(className, id)`'" — the `speaks_for` rule as #1108 landed it.
- "You left `outputs` empty… No error. The run settles, the receipt mints, and it names nothing" — matches the terminal-proposal path, where an empty output list still gets a commitment, so the run settles and the receipt binds nothing.

CI is your gate; `integration-tests-js` was still running when you pinged.

docs-src/pages/flows.mdx
```diff
@@ -0,0 +1,536 @@
...
+} catch (e) {
+  // error message names the unmet guard
+}
+
+// Write the evidence, then the same press succeeds:
+await WorkLog.create(perspective, { note: "Picked up." },
+  { parent: { id: task.id, predicate: "ad4m://has_child" } });
+const result = await instance.proposeTransition("InProgress", "I've started — see the log");
+result.outcomes;      // [{ fromState: "Ready", toState: "InProgress", ... }] — n:1 fires
+result.derivedState;  // "InProgress"
```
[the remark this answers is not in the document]
data-bot-coasys: Verified. The block defines `started`, and `instance` was undefined. Fixed in #1147 (f3721c436): the call now uses `started.proposeTransition(...)`.

docs-src/pages/flows.mdx
```diff
@@ -0,0 +1,536 @@
...
+```
+
+*From `flow-task-handover.test.ts`, Part 3.*
+
+The gate is **fail-closed**, and Part 3 pins each direction:
+
+- A settled grant run **without a minted receipt does not grant**. Bob's vote on the gated task did not count until someone called `mintFlowReceipt`. "The run completed" is not evidence. A receipt is.
+- Once the receipt exists, votes cast before the mint count on the next derivation — the vote was always valid, the gate just could not verify the grant yet. Nothing new needs to be written.
+- The receipt is bound to its instance by `(className, id)`. Alice's own role instance — same class, same `domain` — did not pass the gate, because no receipt names *her* instance. A receipt is never "flow F completed"; it is "flow F produced *this*".
+**Where the check runs, and what travels.** Each replica decides eligibility against its **own** graph, while it collects the evidence for a fold. The read-set then carries only the result: for each matched role instance, the grant date (`produced_at`, the granting run's `settled_at`). The granting receipt itself does not travel, and verifying a receipt never recurses into the granting flow's receipts.
```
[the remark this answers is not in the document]
data-bot-coasys: Verified. The paragraph rendered inside the third bullet. Fixed in #1147 (f3721c436) with a blank line before "**Where the check runs, and what travels.**". I ran the same scan (a list line followed directly by a non-list, non-indented line, outside code fences) over `flows.mdx` and `developer-guides/writing-a-flow.mdx`. This was the only hit.

tests/js/tests/model/flow-task-handover.test.ts
```diff
@@ -0,0 +1,666 @@
...
+    // Unreachable edge is an error, not a silent mint. Fail-on-old-code:
+    // a client that wrote raw proposal links used to be able to name Done
+    // from Ready; proposeTransition must refuse.
+    let unreachable = "";
+    try {
+      await started.proposeTransition("Done");
+    } catch (e: any) {
+      unreachable = String(e?.message ?? e);
+    }
+    expect(unreachable).to.match(/not reachable|Done/i);
```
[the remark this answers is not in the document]
data-bot-coasys: Verified. Both loose patterns matched the other refusal, because both messages name the target state. Fixed in #1147 (fb3b15e85) with the exact wording from `rust-executor/src/perspectives/flow_instance/propose.rs`:

- L172 `` "`{to_state}` is not reachable from `{}` — no proposal written" `` → ``/`Done` is not reachable from `Ready`/``
- L226 `` "guard for `{to_state}` is not satisfied on this replica — proposal not written" `` → ``/guard for `InProgress` is not satisfied on this replica/``

The reachability check (L166-175) runs before the guard seal (L221-228), so each call can only produce its own message. I didn't add the `.not.match(/not reachable/)` from your suggestion: the exact guard phrase already excludes that refusal. L609 was already exact, and so are the guide's assertions.

docs-src/pages/flows.mdx
data-bot-coasys: This is in the page now. `flows.mdx` on dev, L467-473: "The gate has **three** outcomes, not two" lists **Cannot decide** as an **error** meaning "unknown", not "no". The next paragraph says not to wrap the call in `if (granted) allow else deny`, and it narrows the earlier "fail-closed" to "an unproven grant never counts … does not mean that an undecidable gate reads as 'not a member'". The guide has the same line in its fail-closed list ("Cannot decide is an error, not 'no'"). Resolving.

docs-src/pages/flows.mdx
data-bot-coasys: Thanks for tracing both halves separately. I took the contrast you pointed at: the spelling callout now ends with "A key that *is* known but malformed does the opposite: `producedByFlow` without `state` refuses the edge (above). The more obvious mistake is the one that fails open." That's in #1147 (f3721c436), and it frames #1144 as making the two failure modes agree.

